PR #1543, opened 2017-06-21
[EarlGrey] Specified minimum version for EarlGrey

randallli:
so that `pod install` will result in a buildable project

If we dont do this then `pod update` is required.

### DIFF
--- a/catalog/Podfile
+++ b/catalog/Podfile
@@ -40,7 +40,7 @@ target TEST_TARGET do
   platform :ios, '8.0'
   project PROJECT_NAME
   inherit! :search_paths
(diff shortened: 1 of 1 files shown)


randallli:
My feature branch still has a catalog/Podfile.lock. Since we deleted the catalog/Podfile.lock on develop the Earl Grey version on my system could be different from yours. Resulting in a tests failing for you and not me and vice versa.

ajsecord (approved):
Waiting on CI

willlarche (approved):

willlarche:
Is this still relevant?

PR #1545, opened 2017-06-21
[TextField] Adding text field themer

jgunaratne:
Adds a themer to change color of MDCTextField and its respective controller.

### DIFF
--- a/components/TextFields/src/ColorThemer/MDCTextFieldColorThemer.m
+++ b/components/TextFields/src/ColorThemer/MDCTextFieldColorThemer.m
@@ -21,7 +21,12 @@
 + (void)applyColorScheme:(NSObject<MDCColorScheme> *)colorScheme
              toTextField:(MDCTextField *)textField {
   textField.underline.color = colorScheme.primaryColor;
-  textField.placeholderLabel.textColor = colorScheme.primaryColor;
+}
+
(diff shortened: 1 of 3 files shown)


components/TextFields/src/ColorThemer/MDCTextFieldColorThemer.m
```diff
@@ -0,0 +1,32 @@
...
+@implementation MDCTextFieldColorThemer
+
++ (void)applyColorScheme:(NSObject<MDCColorScheme> *)colorScheme
+             toTextField:(MDCTextField *)textField {
+  textField.underline.color = colorScheme.primaryColor;
+}
+
++ (void)applyColorScheme:(NSObject<MDCColorScheme> *)colorScheme
+    toTextInputController:(MDCTextInputController *)textInputController {
+  textInputController.textInput.underline.color = colorScheme.primaryColor;
```
willlarche, line 28: I think you want `underlineColorActive` on the controller. It's newish.

willlarche (changes requested):

ianegordon:
MDCTextInputController is not a UIView and does not implement UIAppearanceContainer.
The UI_APPEARANCE_SELECTOR annotations won't behave as expected.
If MDCTextInputController implements UIAppearanceContainer, how will it be triggered?  (Usually when a view is added to a view hierarchy)
Discussed offline, @willlarche will investigate.

PR #1549, opened 2017-06-22
[TextFields] Presentation styles to controller classes

willlarche:
Converting presentation styles into discrete classes conforming to a common protocol.

### DIFF
--- a/components/TextFields/src/MDCTextInputControllerDefault.h
+++ b/components/TextFields/src/MDCTextInputControllerDefault.h
@@ -22,7 +22,7 @@
  https://www.google.com/design/spec/components/text-fields.html#text-fields-single-line-text-field
 
  The placeholder text is laid out inline. If floating is enabled, it will float above the field when
- there is content or the field is being edited. The character count is below text. The Material 
+ there is content or the field is being edited. The character count is below text. The Material
  Design guidelines call this 'Floating inline labels.'
  https://material.io/guidelines/components/text-fields.html#text-fields-labels
 */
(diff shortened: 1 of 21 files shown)


components/TextFields/src/MDCTextInputControllerDefault.h
```diff
@@ -0,0 +1,56 @@
...
+ Material Design compliant text field. The logic for 'automagic' error states changes:
+ underline color, underline text color.
+ https://www.google.com/design/spec/components/text-fields.html#text-fields-single-line-text-field
+
+ The placeholder text is laid out inline. If floating is enabled, it will float above the field when
+ there is content or the field is being edited. The character count is below text. The Material
+ Design guidelines call this 'Floating inline labels.'
+ https://material.io/guidelines/components/text-fields.html#text-fields-labels
+*/
+@interface MDCTextInputControllerDefault : NSObject <MDCTextInputController>
```
randallli, line 29: I think we need to call out the other styles in the comments because "Default" looks weird all alone.

Is there a better name than "default"?

randallli: wait is this a combo of Floating and Default styles?

components/TextFields/src/MDCTextInputControllerFullWidth.h
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+#import "MDCTextInputController.h"
+
+/**
+ Material Design compliant text field. The logic for 'automagic' error states changes:
+ underline color, underline text color.
+ https://www.google.com/design/spec/components/text-fields.html#text-fields-single-line-text-field
+
+ The placeholder is laid out inline and the character count is also inline to the right.
```
randallli, line 24: so the place holder text cannot float above in this style?

randallli (approved):

components/TextFields/src/MDCTextInputControllerFullWidth.h
willlarche: Correct.

components/TextFields/src/MDCTextInputControllerDefault.h
willlarche: Yes. The layout of floating and default were identical. Floating just had that animation. So it made sense to merge them and make floating and option you can turn off. The spec has floating as the default.

PR #1551, opened 2017-06-22
[Snackbar] Resuming messages when tokens are deallocated

### LINKED ISSUE #710: [SnackBar] Allowing a suspension token to deallocate does not resume messages.
### Overview of the issue

`+[MDCSnackbarManager suspendAllMessages]` and related methods return a `MDCSnackbarSuspensionToken` with the following promise: "A token suitable for use in `+[MDCSnackbarManager resumeMessagesWithToken:]`. Letting this object deallocate is equivalent to calling `+[MDCSnackbarManager resumeMessagesWithToken:]`."

As far as I can tell from the implementation, this is not implemented.
- The `suspendAllMessages` and related methods create a properly-initialized `MDCSnackbarSuspensionToken` returns it to the client without holding a reference to the token, so if the client drops the reference then the token will deallocate.
- MDCSnackbarManagerInternal stores a NSMutableSet of NSUUID's taken from the token's `identifier` property, but NSMutableSet stores things strongly, so even if the token goes away the NSUUID in the set will remain.
- There is no mechanism in the HUD manager or the token itself to tell the manager when the token deallocates.
### Motivation and context

Found this while reviewing a [related pull request](https://github.com/google/material-components-ios/pull/700).
### Version number

14.0.0

romoore:
Updating MDCSnackbarSuspensionToken's dealloc behavior to resume any related message categories.

The SnackbarManager documentation states that when the last token is deallocated, any suspended
messages will be resumed.  This behavior was broken and the messages would remain suspended
indefinitely.

Closes #710

### DIFF
--- a/components/Snackbar/tests/unit/SnackbarManagerTests.m
+++ b/components/Snackbar/tests/unit/SnackbarManagerTests.m
@@ -42,7 +42,7 @@
   }
 
   // Then
-  [self waitForExpectations:@[expectation] timeout:1];
+  [self waitForExpectationsWithTimeout:1.0 handler:nil];
 }
 
 @end
--- a/components/Snackbar/tests/unit/SnackbarManagerTests.m
+++ b/components/Snackbar/tests/unit/SnackbarManagerTests.m
@@ -1,0 +1,43 @@
+/*
+ Copyright 2017-present the Material Components for iOS authors. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


samnm (approved):

components/Snackbar/tests/unit/SnackbarManagerTests.m
```diff
@@ -0,0 +1,43 @@
...
+@import XCTest;
+#import "MaterialSnackBar.h"
+
+@interface SnackbarManagerTests : XCTestCase
+
+@end
+
+@implementation SnackbarManagerTests
+
+- (void)testMessagesResumedWhenTokenIsDeallocated {
```
jgunaratne, line 26: Many of our component tests use the format: given/when/then (see link)

https://github.com/material-components/material-components-ios/blob/develop/components/Slider/tests/unit/SliderTests.m

jgunaratne (changes requested):

jgunaratne (approved):

PR #1553, opened 2017-06-22
[Buttons] Call designated initializer in init methods

ianegordon:
"If an object declares one or more initialization methods, you should decide which method is the designated initializer. This is often the method that offers the most options for initialization (such as the method with the most arguments), and is called by other methods you write for convenience. You should also typically override init to call your designated initializer with suitable default values."

https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/EncapsulatingData/EncapsulatingData.html


randallli (approved):
Sorry didn't realize that init was not a designated initializer of UIButton.

PR #1554, opened 2017-06-22
[Tabs] Reposition the Ink layer in ItemBar cells

### LINKED ISSUE #1480: InkView in MDCTabBar items is above title/image
### Overview
The InkView for an MDCItemBarCell is on top of everything in the item. This means if you use a dark inkView, you will cover the contents of your tab which is undesirable.

Gif of the issue: http://d.pr/i/CyJ31z

### Reproduction steps

Implement an MDCTabBar, set the inkColor to something dark (.black) and the tintColor to something light (.white). When touching a tab item, the inkColor will cover the icon.

### Version number
23.4.1

iOS 9/10

romoore:
Moving the ink layer below the image layer in ItemBar cells. The Ink
layer was incorrectly inserted above the label and image layers of
ItemBar cells. The ink would obscure the label/image when tapped.

Closes #1480

### DIFF
--- a/components/Tabs/src/private/MDCItemBarCell.m
+++ b/components/Tabs/src/private/MDCItemBarCell.m
@@ -53,6 +53,10 @@ const CGFloat kSelectedNavigationImageYOffset = -2;
 /// Duration of selection animations in applicable content styles.
 static const NSTimeInterval kSelectionAnimationDuration = 0.3f;
 
+@interface MDCItemBarCell () <MDCInkTouchControllerDelegate>
+
+@end
+
 @implementation MDCItemBarCell {
   UILabel *_titleLabel;
   UIImageView *_imageView;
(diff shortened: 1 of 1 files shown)


willlarche (approved):

randallli:
i think that the ink is intentionally above everything. Let me check with @jgunaratne

randallli:
Looks like button puts it [below the image](https://github.com/material-components/material-components-ios/blob/develop/components/Buttons/src/MDCButton.m#L258).

randallli (approved):

components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -81,6 +85,7 @@ - (instancetype)initWithFrame:(CGRect)frame {
 
     // Set up ink controller to splash ink on taps.
     _inkTouchController = [[MDCInkTouchController alloc] initWithView:self];
+    _inkTouchController.delegate = self;
     [_inkTouchController addInkView];
```
ianegordon: Did we consider:
[self sendSubviewToBack:_inkTouchController.defaultInkView];

This may be a bit simpler than adding a delegate and implementing a method.

romoore: Can-do

ianegordon:
Fix demonstrated.
Tested locally.

PR #1564, opened 2017-06-23
[CollectionCells] Move ink view behind cell contents

### LINKED ISSUE #1560: MDCCollectionViewCell inkView is above other subviews
### Overview
When performing a touch-and-hold on an MDCCollectionViewCell, the ink completely covers the contents of the cell.  This is most readily seen when the ink color is opaque (e.g., `UIColor.redColor`).

### Reproduction steps
1. Set ink color to an opaque color
1. Touch and hold on an MDCCollectionViewCell

#### Expected Result
Ink fills the area behind the cell contents

#### Actual Result
Ink obscures/covers the entire cell

### Version number
25.0.1

iOS 10.2 (Simulator) and iPhone 5s (Simulator)

### Related issues
#1480

romoore:
Move the ink view to be above the background views but below the
contents so that it doesn't hide the cell contents when ink is present.

Closes #1560

### DIFF
--- a/components/CollectionCells/src/MDCCollectionViewCell.m
+++ b/components/CollectionCells/src/MDCCollectionViewCell.m
@@ -274,7 +274,7 @@ NSString *const kDeselectedCellAccessibilityHintKey =
 - (MDCInkView *)inkView {
   if (!_inkView) {
     _inkView = [[MDCInkView alloc] initWithFrame:self.bounds];
-    [self addSubview:_inkView];
+    [self insertSubview:_inkView aboveSubview:self.backgroundView];
   }
   return _inkView;
 }
@@ -287,7 +287,7 @@ NSString *const kDeselectedCellAccessibilityHintKey =
     [_inkView removeFromSuperview];
   }
   if (inkView) {
-    [self addSubview:inkView];
+    [self insertSubview:inkView aboveSubview:self.backgroundView];
   }
   _inkView = inkView;
 }


ajsecord (approved):

ianegordon (changes requested):
Do we want to be the bottom view in the contentView hierarchy?
[self.contentView addSubview:_inkView];
[self.contentVIew sendSubviewToBack:_inkView];

Alternatively,
What happens if self.backgroundView is nil?
If self.selectedBackgroundView, don't we want to be just in front of that?

romoore:
What about placing the InkView just behind the contentView? That allows the selectedBackgroundView and backgroundView to be behind the ink, but lets the content remain visible during ink effects.

```objectivec
if (self.contentView) {
  [self insertSubview:_inkView belowSubview:self.contentView];
} else {
  [self addSubview:_inkView];
}
```

ianegordon:
contentView is nonnull, so no need to check for existence.
It is also readonly so we *should* be decently safe.

ianegordon:
Discussed offline.
Consider adding Ink as the lowest view in contentView.
@jgunaratne may have insight.

ianegordon:
@romoore - we had discussed some possible options for where to add the ink layer.  Let me know when you've made your choice and we'll merge.

romoore:
Not merging this, ink will remain above other layers.  

Related: #1584.

PR #1565, opened 2017-06-23
[Scripts] Updated scripts/build_all to take a verbose flag.

ajsecord:
The script used to poop out logs if the build failed, but it often wouldn't poop out enough.

Now `build_all` never poops out logs, while `build_all --verbose` always poops out all logs.

Also removed the unused ability to specify a different `xcodebuild` command apart from `build`.

Tested manually with `[verbose, no verbose] x [build with error, build without error]`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ script:
   - pod --version
   - swiftlint
   - scripts/prep_all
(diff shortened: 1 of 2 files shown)


ianegordon (approved):
Tested locally.

PR #1569, opened 2017-06-26
[Themers] Support alert dialog color themer in iOS 8

jgunaratne:
Fixes issue in iOS 8 where new iOS 9 method is used. Use deprecated method when iOS 8 is used.


willlarche (approved):

ianegordon:
Catalog no longer crashes on startup on iOS 8 simulator.

ianegordon (approved):
Verified locally.

Issue #1568, opened 2017-06-26
[CI] Travis CI is failing because our logs are greater than 4MB


ajsecord:
We changed our build scripts to output everything, but now our builds are failing (incorrectly) because the log output exceeds 4MB.

ajsecord:
I'm seeing if xcpretty will help...

willlarche:
The logs that we don't have access to.

romoore:
I haven't seen our builds fail for this reason in several weeks, can this be closed?

willlarche:
Yes. I fixed it with several changes.

PR #1571, opened 2017-06-26
[Scripts] Optionally use `xcpretty` in `scripts/build_all`.

ajsecord:
If `--verbose` is used in `scripts/build_all` and [`xcpretty`](https://github.com/supermarin/xcpretty) is installed, then pipe the `xcodebuild` output through that tool. The resulting output is _much_ more condensed and should help with #1568.

Errors still seem relatively clear, they look like this:
```shell
...
▸ Compiling MDCInkTouchController.m
▸ Compiling MDCInkView.m

❌  /Users/ajsecord/Source/Git/mdc-fork/components/Ink/src/MDCInkView.m:17:1: unknown type name 'THIS'

THIS IS A DELIBERATE ERROR
^
```

### DIFF
--- a/scripts/build_all
+++ b/scripts/build_all
@@ -25,12 +25,23 @@ function is_expected_failure() {
   grep --quiet "is not configured for Running" "$1" 
 }
 
+# Test if the xcpretty command is available.
+function xcpretty_available() {
+  xcpretty > /dev/null 2>&1
+  # Exit code 127 is "command not found" and standard.
+  if [ $? -eq 127 ]; then
+    return 1
+  else
+    return 0
+  fi
+}
+
 # Parse command-line arguments.
-verbose=0
+verbose=1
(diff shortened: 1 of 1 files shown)


scripts/build_all
```diff
@@ -25,12 +32,28 @@ function is_expected_failure() {
...
+  else
+    return 0
+  fi
+}
+
 # Parse command-line arguments.
-verbose=0
+#
+# Note that we're following the command-line exit status convention of zero
+# to mean "success".
```
ianegordon, line 51: 0 == success for a method is strange to me, but I understand. :/

ianegordon (approved):
Tested locally.

scripts/build_all
ajsecord: Yeah, Unix. 🤷‍♂️

PR #1572, opened 2017-06-26
[Buttons] Remove NS_UNAVAILABLE from setBackgroundColor

samnm:


components/Buttons/src/MDCButton.m
```diff
@@ -503,7 +503,6 @@ - (void)animateButtonToHeightForState:(UIControlState)state {
 #pragma mark - BackgroundColor
 
 - (void)setBackgroundColor:(nullable UIColor *)backgroundColor {
-  // Turns out swift ignores NS_UNAVAILABLE
```
randallli, line 506: I feel like we still need an explanation here.
"We recommend using setBackgroundColor:forState: This method is just a connivence method for the normal state.

Or we can keep it permanently deprecated.

randallli: After more consideration: I think we are LGTM by just putting a  comment explaining whats going on. This is mostly to make it clear that if the button is not in a normal state it wont be calling super.backgroundcolor = backgroundColor with the passed in value.

randallli (approved):

components/Buttons/src/MDCButton.m
samnm: Redefined it and added a comment.

PR #1555, opened 2017-06-22
[Buttons] Refactored Button and removed some technical debt

randallli:
This PR will not be merged as is. @samnm will be breaking it up in to smaller PRs.

Done in PR #1574:
- Deleting resetElevationForState:
- Corrected highlight vs selected: we were incorrectly use the `UIControlStateSelected` as our pressed state.
- Using UIAppearance for setting up backgroundColors and elevations of our buttons (Flat, Floating, Raised)
- Watching for state changes instead of changing elevation on UITouch events.

In progress PR #1609:
- Deprecating CustomTitleColor
- Using setTitleColor:forState: instead of custom title color.

Remaining changes to be merged:
- Deprecating DisabledAlpha

PR #1574, opened 2017-06-26
[Buttons] Elevation clean up

samnm:
This PR pulls in the elevation-related changes from #1555.

Changes:
* Removes [MDCButton resetElevationForState]
* Changes MDCButton to correctly treat UIControlStateHighlighted as the pressed state
* Moves MDCButton (and subclasses) elevation initialization to UIAppearance calls in +initialize
* Changes elevation assignment from touchesBegan/Moved/Ended to state observation
* Removes defaultElevationForState from MDCButton+Subclassing.h
* MDCFlatButton no longer depends on shouldRaiseOnTouch

### DIFF
--- a/components/Buttons/src/MDCButton.m
+++ b/components/Buttons/src/MDCButton.m
@@ -109,17 +109,6 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
 
 @dynamic layer;
 
-+ (void)initialize {
-  // Default background colors.
-  [[MDCButton appearance] setBackgroundColor:MDCColorFromRGB(MDCButtonDefaultBackgroundColor)
-                                    forState:UIControlStateNormal];
-
-  [[MDCButton appearance] setElevation:MDCShadowElevationRaisedButtonResting
-                              forState:UIControlStateNormal];
-  [[MDCButton appearance] setElevation:MDCShadowElevationRaisedButtonPressed
-                              forState:UIControlStateHighlighted];
-}
-
 + (Class)layerClass {
   return [MDCShadowLayer class];
 }
@@ -240,6 +229,10 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
   _userElevations = [NSMutableDictionary dictionary];
   _backgroundColors = [NSMutableDictionary dictionary];
   _accessibilityLabelForState = [NSMutableDictionary dictionary];
+
+  _backgroundColors[@(UIControlStateNormal)] = MDCColorFromRGB(MDCButtonDefaultBackgroundColor);
+  _userElevations[@(UIControlStateNormal)] = @(MDCShadowElevationRaisedButtonResting);
+  _userElevations[@(UIControlStateHighlighted)] = @(MDCShadowElevationRaisedButtonPressed);
 
   // Disable default highlight state.
   self.adjustsImageWhenHighlighted = NO;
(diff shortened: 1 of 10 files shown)


samnm:
I will add some tests to this before merging.

components/Buttons/src/MDCButton.m
```diff
@@ -520,12 +539,18 @@ - (void)setBackgroundColor:(UIColor *)backgroundColor forState:(UIControlState)s
 
 - (CGFloat)elevationForState:(UIControlState)state {
```
randallli, line 521: There were some tests checking elevation was getting set. can we bring those into this PR?

components/Buttons/src/MDCButton.m
```diff
@@ -338,11 +346,6 @@ - (void)touchesEnded:(NSSet *)touches withEvent:(UIEvent *)event {
 
   CGPoint location = [self locationFromTouches:touches];
   [_inkView startTouchEndedAnimationAtPoint:location completion:nil];
-
-  BOOL inside = CGRectContainsPoint(self.bounds, location);
-  if (inside && _shouldRaiseOnTouch) {
-    [self animateButtonToHeightForState:UIControlStateNormal];
```
randallli, line 334: does this need to be called to initialize the button to the correct starting height? or will the proxying do it.

randallli (approved):
lets pull in some of those tests.

components/Buttons/src/MDCButton.m
samnm: Done.

components/Buttons/src/MDCButton.m
samnm: We call shadowLayer.elevation = [self elevationForState:self.state]; in the commonInit which correctly configures the shadow layer.

components/Buttons/tests/unit/ButtonsTests.m
```diff
@@ -423,4 +532,23 @@ - (void)testAdjustsFontProperty {
                              @"Font size should be equal to MDCFontTextStyleButton's.");
 }
 
+#pragma mark utilities
+
+- (NSString*)controlState:(UIControlState)controlState {
+  NSMutableString *string = [NSMutableString string];
+  if (UIControlStateNormal) {
```
randallli: lets copy the function from the mutator tests:
https://github.com/material-components/material-components-ios/blob/develop/components/Buttons/tests/unit/ButtonTitleColorAccessibilityMutatorTests.m#L116

components/Buttons/src/MDCFlatButton.m
```diff
@@ -16,12 +16,23 @@
...
 #import "MDCFlatButton.h"
 
+#import "MaterialShadowElevations.h"
 #import "private/MDCButton+Subclassing.h"
 
 static NSString *const MDCFlatButtonHasOpaqueBackground = @"MDCFlatButtonHasOpaqueBackground";
 
 @implementation MDCFlatButton
 
++ (void)initialize {
```
randallli, line 26: lets move the logic into the common init.

If we need to override the setters in the subclass as you showed me in your experiment lets do that too.

we should also confirm the default values with a unit test.

randallli (changes requested):

components/Buttons/src/MDCFlatButton.m
randallli: I think this is ok now.
We have now done an experiment with calling super setters and that seems to clobber appearance set values.

randallli (approved):

randallli:
We need to make sure that our clients know that they need to set the highlighted state for "pressed" state colors.

PR #1581, opened 2017-06-27
[TextFields] Fix text fields placeholder Y preset

willlarche:
Fix for placeholder not raising when there is preset text.

### DIFF
--- a/components/TextFields/src/MDCTextInputControllerDefault.m
+++ b/components/TextFields/src/MDCTextInputControllerDefault.m
@@ -663,6 +663,13 @@ static inline UIColor *MDCTextInputDefaultTextErrorColor() {
   return _floatingPlaceholderColor ?: self.textInput.tintColor;
 }
 
+- (void)setFloatingEnabled:(BOOL)floatingEnabled {
+  if (_floatingEnabled != floatingEnabled) {
+    _floatingEnabled = floatingEnabled;
+    [self updatePlaceholderY];
+  }
+}
+
 - (void)setFloatingPlaceholderScale:(NSNumber *)floatingPlaceholderScale {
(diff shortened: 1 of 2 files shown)


randallli (approved):

components/TextFields/src/MDCTextInputControllerDefault.m
```diff
@@ -407,6 +409,25 @@ - (void)updatePlaceholder {
   }
 }
 
+// Sometimes the text field is showing the correct layout for its values (like when it's created
+// with .text already entered) so we make sure it's in the right place always.
```
romoore, line 413: Did you mean that sometimes we are not showing the correct layout?

components/TextFields/src/MDCTextInputControllerDefault.m
```diff
@@ -407,6 +409,25 @@ - (void)updatePlaceholder {
...
+- (void)updatePlaceholderY {
```
romoore, line 414: Since this method is called in `-init`, is it safe to use all of these `self` references and property accessors/methods?

components/TextFields/src/MDCTextInputControllerDefault.m
```diff
@@ -909,7 +918,12 @@ - (void)textInputDidBeginEditing:(NSNotification *)note {
 }
 
 - (void)textInputDidChange:(NSNotification *)note {
-  [self updateLayout];
```
romoore, line 912: Do we no longer need to call `[self updateLayout]` here? Was it overkill previously?

romoore: Nevermind! I see updatePlaceholderY calls updateLayout internally.

PR #1597, opened 2017-07-03
[Collections] Delegate method for cards' style margins

### LINKED ISSUE #1507: [Collections] Have a way to change Cards style margins
### Overview
The Cards style has a fixed right/left margin of 8pt. On large device, if the collection is full width, the cards' width is too big.
It should have a way to adjust the margin or the width of the cells.

gambard:
When the card style is used, the margins for the cards are set to 8pt.
On large devices, it would be better if the margin can be dynamically set.
This commit adds a delegate method to the MDCCollectionViewEditingDelegate
to provide this value.

Closes #1507


romoore (changes requested):
Thanks for working to improve Collections! 

As we discussed offline, a cleaner (and probably clearer) approach will be to have the `MDCCollectionViewController` use `-collectionView:layout:section:` for width computations so that subclasses can easily override the values.  I've pushed that approach up as PR #1614 .

gambard:
Thanks! Closing this.

PR #1605, opened 2017-07-06
[Tabs] Restore initial tab selection behavior

brianjmoore:
Recent changes to the tab bar unintentionally changed the initial selection behavior of MDCTabBar such that creating a tab bar and setting its items array no longer automatically selected the first item. This broke all of our examples and made the API more onerous to use for the 99% case.

This change preserves the ability to set an empty selection while fixing the common case and keeping the API simple. To help solidify this behavior it adds some unit tests and new capabilities to the example.

### DIFF
--- a/components/Tabs/src/private/MDCItemBar.m
+++ b/components/Tabs/src/private/MDCItemBar.m
@@ -198,35 +198,20 @@ static void *kItemPropertyContext = &kItemPropertyContext;
       }
     }
 
-    // Update selected item.
-    if ((newSelectedItem != _selectedItem) && ![newSelectedItem isEqual:_selectedItem]) {
-      // Update _selectedItem directly to avoid clearing _shouldPreserveEmptySelectionOnItemsChange.
-      _selectedItem = newSelectedItem;
-      [self selectedItemDidChange];
-    }
+    // Update _selectedItem directly to avoid clearing _shouldPreserveEmptySelectionOnItemsChange.
+    _selectedItem = newSelectedItem;
+    [self updateLastSelectedIndexPath];
 
     // Update collection with new items
     [self reload];
 
+    // Select tab for current item.
+    [self selectItemAtIndex:(_lastSelectedIndexPath ? _lastSelectedIndexPath.item : NSNotFound)
+                   animated:NO];
+
     // Start observing new items for changes.
     [self startObservingItems];
   }
-}
-
-- (void)selectedItemDidChange {
-  // Update the index path for the selected item.
-  NSIndexPath *selectedItemIndexPath = nil;
-  if (_selectedItem) {
-    NSUInteger index = [_items indexOfObject:_selectedItem];
-    if (NSNotFound != index) {
-      // Valid index
-      selectedItemIndexPath = [NSIndexPath indexPathForItem:index inSection:0];
-    }
-  }
-  _lastSelectedIndexPath = selectedItemIndexPath;
-
-  // Update UI to select the item.
-  [self selectItemAtIndex:NSNotFound animated:NO];
 }
 
 - (void)setAlignment:(MDCItemBarAlignment)alignment {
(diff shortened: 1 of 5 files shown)


components/Tabs/src/private/MDCItemBar.m
```diff
@@ -170,26 +170,32 @@ - (void)setItems:(NSArray<UITabBarItem *> *)items {
...
-      if (_selectedItem) {
-        NSUInteger index = [_items indexOfObject:_selectedItem];
-        if (NSNotFound != index) {
-          _lastSelectedIndexPath = [NSIndexPath indexPathForItem:index inSection:0];
-        } else {
-          _lastSelectedIndexPath = nil;
-        }
+
+    // Determine new selected item.
+    UITabBarItem *newSelectedItem = nil;
```
ianegordon: nit: nil is redundant.

components/Tabs/src/private/MDCItemBar.m
```diff
@@ -170,26 +170,32 @@ - (void)setItems:(NSArray<UITabBarItem *> *)items {
...
+    if (_selectedItem) {
```
ianegordon: I think these nested ifs could be simplified to:
if (_selectedItem && [_items containsObject:_selectedItem]) {
  newSelectedItem = _selectedItem;
} else {
  newSelectedItem = _items.firstObject;
}

components/Tabs/src/private/MDCItemBar.m
```diff
@@ -224,7 +230,9 @@ - (void)setSelectedItem:(nullable UITabBarItem *)selectedItem animated:(BOOL)ani
       }
     }
     _selectedItem = selectedItem;
-    [self selectItemAtIndex:itemIndex animated:animated];
+    [self updateLastSelectedIndexPath];
+    [self selectItemAtIndex:(_lastSelectedIndexPath ? _lastSelectedIndexPath.item : NSNotFound)
+                   animated:NO];
```
ianegordon: Do you want animated:animated instead of animated:NO?

components/Tabs/src/private/MDCItemBar.m
```diff
@@ -457,6 +465,19 @@ - (void)stopObservingItems {
   }
 }
 
+- (void)updateLastSelectedIndexPath {
+  // Update the index path for the selected item.
+  NSIndexPath *selectedItemIndexPath = nil;
```
ianegordon: nit: nil is redundant.

ianegordon (changes requested):
Perhaps I'm missing some requirement, but can we simplify _lastSelectedIndexPath to an NSInteger and use NSNotFound to indicate an unselected state?
(As we broke our initial selection behavior, I'm looking for ways to simplify.  It may be out of scope for this change)

components/Tabs/src/private/MDCItemBar.m
brianjmoore: Yep, good catch

components/Tabs/src/private/MDCItemBar.m
brianjmoore: It is redundant under ARC, but I would argue that being explicit with default values improves readability by removing a mental step for readers. Regardless, there was a better way to write this that avoids the issue.

components/Tabs/src/private/MDCItemBar.m
brianjmoore: Done, even simpler too

components/Tabs/src/private/MDCItemBar.m
brianjmoore: Removed

brianjmoore:
PTAL. I removed _lastSelectedIndexPath entirely so _selectedItem is now the sole source of truth. That variable was there from a much older implementation where we didn't have an explicit array of items to work with or concrete objects representing each tab. Good call on rethinking it.

ianegordon (approved):
LGTM.

PR #1612, opened 2017-07-10
[TextField] Update demo alert title to be variable.

### LINKED ISSUE #1611: [TextFields] Demo alert title doesn't change based on floating enabled/disabled.
### Overview
Using the catalog app viewing the Demo of TextFields, switching the style to use floating or not doesn't change the title of the alert which always says Floating Enabled (regardless or floating being enabled or not).

### Reproduction steps
Launch catalog and navigate to TextFields > Demo > STYLE. Notice in the alert title is Floating Enabled. Select No. Select STYLE to show alert again, the title still says Floating Enabled.

### Version number
 - 26.0.0

- iPhone 7 Plus - iOS 10.3 Simulator.

DonnaLea:
Update demo alert title to display based on whether the text fields are floating or not.

Closes #1611

### DIFF
--- a/components/TextFields/examples/TextFieldExample.swift
+++ b/components/TextFields/examples/TextFieldExample.swift
@@ -214,16 +214,20 @@ final class TextFieldSwiftExample: UIViewController {
   }
 
   func buttonDidTouch(sender: Any) {
-    let alert = UIAlertController(title: "Floating Enabled",
+    let isFloatingEnabled = allTextFieldControllers.first?.isFloatingEnabled ?? false
+    let alert = UIAlertController(title: "Floating Labels",
                                   message: nil,
                                   preferredStyle: .actionSheet)
-    let defaultAction = UIAlertAction(title: "Default (Yes)", style: .default) { _ in
+
+    let defaultAction = UIAlertAction(title: "Default (Yes)" + (isFloatingEnabled ? " ✓" : ""),
+                                      style: .default) { _ in
       self.allTextFieldControllers.forEach({ (controller) in
         controller.isFloatingEnabled = true
       })
     }
     alert.addAction(defaultAction)
-    let floatingAction = UIAlertAction(title: "No", style: .default) { _ in
+    let floatingAction = UIAlertAction(title: "No" + (isFloatingEnabled ? "" : " ✓"),
+                                       style: .default) { _ in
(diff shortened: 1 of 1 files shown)


willlarche:
Thanks @DonnaLea for contributing! What was intended here was to explain what you are toggling which is "Whether floating is enabled" but it seems like that wasn't clear. Could you make this that the title is "Floating Labels" and there is a ✓ next to the one that is currently selected?

willlarche (changes requested):
What was intended here was to explain what you are toggling which is "Whether floating is enabled" but it seems like that wasn't clear. Could you make this that the title is "Floating Labels" and there is a ✓ next to the one that is currently selected?

DonnaLea:
Yep, that makes more sense! And yes, the problem was that it wasn't clear what the current state that was being represented was. Thanks for the suggestion, update coming.

willlarche:
Thank you!!

DonnaLea:
Updated with an amended commit. Ready for review. (can't see how to re-assign back)

components/TextFields/examples/TextFieldExample.swift
```diff
@@ -214,16 +214,20 @@ final class TextFieldSwiftExample: UIViewController {
   }
 
   func buttonDidTouch(sender: Any) {
+    let isFloatingEnabled = allTextFieldControllers.first?.isFloatingEnabled ?? false
     let alert = UIAlertController(title: "Floating Enabled",
```
willlarche: Let's make the title "Enable Placeholder Floating" and I'd say it's done!

willlarche (changes requested):
Little change to title and it's ready.

components/TextFields/examples/TextFieldExample.swift
willlarche: Or "Floating Labels". Whichever you think is clearer.

willlarche (approved):

DonnaLea:
My bad, that got lost in my edits. Fixed!

willlarche:
Awaiting Travis CI, approval (which is a formality. There is no work in the framework.)

willlarche:
Between when this was started and today, conflicts emerged. Correct them at your convenience and we'll try submitting again.

ajsecord:
Will, GitHub claims this is ok to merge. Merge now?

PR #1614, opened 2017-07-10
[Collections] Allow custom UIEdgeInsets

### LINKED ISSUE #1507: [Collections] Have a way to change Cards style margins

romoore:
MDCCollectionViewController implements custom logic for computing the
`UIEdgeInsets` depending on the `MDCCollectionViewCellStyle` and the
`MDCCollectionViewCellLayoutType`. Subclasses need to override this
functionality to customize their presentation.

* MDCCollectionViewController will use the public
  `-collectionView:layout:section` to compute the width of its
   supplementary view instead of the internal implementation directly.
* Update the Collections layout example to have different horizontal
  edge insets depending on the view's horizontal size class.

Closes #1507


components/CollectionCells/examples/CollectionCellsLayoutExample.m
```diff
@@ -166,6 +166,29 @@ - (UICollectionViewCell *)collectionView:(UICollectionView *)collectionView
   return cell;
 }
 
+#pragma mark - <UICollectionViewDelegateFlowLayout>
+
+- (UIEdgeInsets)collectionView:(UICollectionView *)collectionView
```
gambard: Don't you prefer to add the example in the Collections category?

romoore: Yep, good callpout. I'm working on writing a new "Collections" example and an Earl Grey test to verify it.

components/Collections/examples/CollectionsSectionInsetsExample.m
```diff
@@ -0,0 +1,114 @@
...
+ limitations under the License.
+ */
+#import "CollectionsSectionInsetsExample.h"
+
+static const NSInteger kSectionCount = 5;
+static const NSInteger kSectionItemCount = 3;
+static NSString *const kReusableIdentifierItem = @"itemCellIdentifier";
+
+@implementation CollectionsSectionInsetsExample {
+  NSMutableArray *_content;
```
willlarche: Usually we use @propertys in an objc extension and let the ivars be free. If you're overriding both the setter and the getter you'll _also_ have to declare an ivar but that's a super exception.

components/Collections/examples/CollectionsSectionInsetsExample.m
```diff
@@ -0,0 +1,114 @@
...
+  NSMutableDictionary *_sectionUsesCustomInsets;
+}
+
++ (NSArray *)catalogBreadcrumbs {
```
willlarche: Move the CBC methods to the bottom of the file (or into a -supplemental) in an extension "(CatalogByConvention)"

components/Collections/examples/CollectionsSectionInsetsExample.m
```diff
@@ -0,0 +1,114 @@
...
+  return @[ @"Collections", @"Custom Section Insets" ];
+}
+
++ (NSString *)catalogDescription {
```
willlarche: catalogDescription is a description of the component itself. So, you can delete this since there's one in CollectionsSimpleDemo.m.

components/Collections/examples/CollectionsSectionInsetsExample.m
```diff
@@ -0,0 +1,114 @@
...
+  if ([_sectionUsesCustomInsets[@(section)] boolValue]) {
+    insets.left = MIN(40, 8 * section);
+    insets.right = insets.left;
+  }
+  return insets;
+}
+
+#pragma mark - UIControlEventHandlers
+
+- (void)didSet:(id)sender {
```
willlarche: Instead of creating a local variable, cast the sender in the method signature by changing 'id' to 'UISwitch*'.

components/Collections/examples/CollectionsSectionInsetsExample.m
```diff
@@ -0,0 +1,114 @@
...
+- (void)viewDidLoad {
+  [super viewDidLoad];
+  self.title = @"Custom Section Insets";
+  self.collectionView.accessibilityIdentifier = @"collectionsCustomSectionInsetsCollectionView";
+
+  // Register cell class.
+  [self.collectionView registerClass:[MDCCollectionViewTextCell class]
+          forCellWithReuseIdentifier:kReusableIdentifierItem];
+
+  // Populate content.
```
willlarche: This would be best in a -Supplemental file.

components/Collections/examples/CollectionsSectionInsetsExample.m
```diff
@@ -0,0 +1,114 @@
...
+}
+
++ (NSString *)catalogDescription {
+  return @"Demonstration of customizing section insets in Material Collections views.";
+}
+
+- (void)viewDidLoad {
+  [super viewDidLoad];
+  self.title = @"Custom Section Insets";
+  self.collectionView.accessibilityIdentifier = @"collectionsCustomSectionInsetsCollectionView";
```
willlarche: This would be best in a -Supplemental file.

components/Collections/examples/CollectionsSectionInsetsExample.m
```diff
@@ -0,0 +1,114 @@
...
+      [items addObject:itemString];
+    }
+    [_content addObject:items];
+  }
+
+  // Customize collection view settings.
+  self.styler.cellStyle = MDCCollectionViewCellStyleCard;
+}
+
+#pragma mark - <UICollectionViewDataSource>
```
willlarche: This whole section would be best in a -Supplemental file since the 'meat' of the instruction here is the UICollectionViewDelegateFlowLayout changes.

components/Collections/examples/CollectionsSectionInsetsExample.m
willlarche: Usually we use @propertys where we get the ivar for free. The exception being when we override the setter and the getter of the property when we need to have both the @property declaration and the explicit ivar.

components/Collections/examples/CollectionsSectionInsetsExample.m
willlarche: catalogDescription is a description of the component itself which has already been done in CollectionsSimpleDemo.m. So, you can delete this.

components/Collections/examples/CollectionsSectionInsetsExample.m
willlarche: Move the catalog by convention methods to an extension called (CatalogByConvention) and put it either at the end of the file or in a -Supplemental.

components/Collections/examples/CollectionsSectionInsetsExample.m
willlarche: This would be a good candidate to move to a -Supplemental file.

components/Collections/examples/CollectionsSectionInsetsExample.m
willlarche: This section also would be a good candidate to move to the -Supplemental file.

components/Collections/examples/CollectionsSectionInsetsExample.m
willlarche: This entire section would be a good candidate to move to the -Supplemental file since the 'meat' of the instruction is the UICollectionViewDelegateFlowLayout section.

components/Collections/examples/CollectionsSectionInsetsExample.m
willlarche: You can get rid of the local variable by casting sender in the method signature by changing id to UISwitch *

components/Collections/tests/earlgrey/MDCCollectionViewControllerTests.m
```diff
@@ -0,0 +1,201 @@
...
+ Navigates to a specific Example screen given an array of "breadcrumb" identifiers.
+ 
+ @param breadcrumbs an array of identifiers for the example to open.
+
+ @note Only the first two elements are currently handled.
+ */
++ (void)jumpToExampleWithBreadcrumbs:(NSArray <NSString *>*)breadcrumbs {
+  NSAssert(breadcrumbs.count >= 2, @"Breadcrumb paths must be at least 2 elements.");
+  [MDCCollectionViewControllerTests scrollToElementWithIdentifier:breadcrumbs[0]
+                                          inElementWithIdentifier:@"collectionView"
```
willlarche: I think @"collectionView" is good candidate for a const string now.

components/Collections/tests/earlgrey/MDCCollectionViewControllerTests.m
```diff
@@ -0,0 +1,201 @@
...
+                           withStep:-1
+                                tap:YES];
+}
+
+/**
+ Navigates back to the main screen from the Custom Section Insets example screen.
+ */
++ (void)returnToMainScreen {
+  // Leave the Custom Section Insets view
+  [[EarlGrey selectElementWithMatcher:grey_accessibilityID(@"back_bar_button")] performAction:grey_tap()];
```
willlarche: @"back_bar_button" would also be a good candidate for a const string.

components/Collections/tests/earlgrey/MDCCollectionViewControllerTests.m
```diff
@@ -0,0 +1,201 @@
...
+/**
+ Returns a GREYActionBlock that retrieves the @c numberOfSections property value from a 
+ UICollectionView.
+ 
+ @param numberOfSections inout parameter for storing the number of sections value.
+ */
++ (GREYActionBlock *)getNumberOfSections:(NSInteger *)numberOfSections {
+  return [GREYActionBlock actionWithName:@"getNumberOfSectionsFromUICollectionView"
+                             constraints:grey_respondsToSelector(@selector(numberOfSections))
+                            performBlock:^BOOL(id element, NSError *__strong *errorOrNil) {
```
willlarche: If element will always be a UICollectionView *, then cast it right in the block declaration.

components/Collections/tests/earlgrey/MDCCollectionViewControllerTests.m
```diff
@@ -0,0 +1,201 @@
...
+}
+
+/**
+ Returns a GREYActionBlock that retrieves the @c bounds property value from a UIView.
+ 
+ @param bounds inout parameter for storing the bounds value.
+ */
++ (GREYActionBlock *)getBounds:(CGRect *)bounds {
+  return [GREYActionBlock actionWithName:@"getBoundsFromUIView" constraints:grey_respondsToSelector(@selector(bounds)) performBlock:^BOOL(id element, NSError *__strong *errorOrNil) {
+    UIView *asView = (UIView *)element;
```
willlarche: Same note about casting in the block declaration instead of using a local variable.

components/Collections/tests/earlgrey/MDCCollectionViewControllerTests.m
```diff
@@ -0,0 +1,201 @@
...
+ evaluates as true.
+
+ @param matcher a matcher for target element
+ @param scrollableMatcher a matcher for the scrollable view
+ @param direction the scroll direction
+ @param step the amount to scroll in each repeated step until the target element is found. If @c 
+             step is non-positive, a default value is used.
+ @param action a GREYAction to perform after scrolling
+ */
++ (void)scrollToElementWithMatcher:(nonnull id<GREYMatcher>)matcher
```
willlarche: Since we can use this for all the tests, move it to a more general file.

willlarche (changes requested):
All style notes. Content is solid!

components/Collections/examples/CollectionsSectionInsetsExample.m
romoore: Done

components/Collections/tests/earlgrey/MDCCollectionViewControllerTests.m
romoore: It doesn't seem to accept that as a signature, and the explicit cast makes maintainers recognize just how brittle it is.

components/Collections/examples/CollectionsSectionInsetsExample.m
romoore: Done

components/Collections/examples/CollectionsSectionInsetsExample.m
romoore: Done

components/Collections/examples/CollectionsSectionInsetsExample.m
romoore: Done

components/Collections/examples/CollectionsSectionInsetsExample.m
romoore: Done

components/Collections/examples/CollectionsSectionInsetsExample.m
romoore: Done

components/Collections/examples/CollectionsSectionInsetsExample.m
romoore: Done

components/Collections/tests/earlgrey/MDCCollectionViewControllerTests.m
romoore: Doesn't seem to play well with the signature

willlarche: Ack!

willlarche (approved):

components/Collections/tests/earlgrey/MDCCollectionViewControllerTests.m
romoore: I was wrong, and when I tried again it was fine. Switched.

catalog/MDCCatalog/MDCNodeListViewController.swift
```diff
@@ -107,6 +107,12 @@ class MDCNodeListViewController: CBCNodeListViewController {
     self.tableView.backgroundColor = UIColor.white
     self.tableView.separatorStyle = .none
 
+    if (node.title.characters.count > 0) {
```
romoore: Remove parentheses

components/Collections/tests/earlgrey/MDCCollectionViewControllerTests.m
```diff
@@ -0,0 +1,206 @@
...
+  [[EarlGrey selectElementWithMatcher:grey_accessibilityID(@"back_bar_button")]
+      performAction:grey_tap()];
+  // Leave the Collections view
+  [[EarlGrey selectElementWithMatcher:grey_accessibilityID(@"back_bar_button")]
+      performAction:grey_tap()];
+  // Scroll to the top of the main screen
+  [[EarlGrey selectElementWithMatcher:grey_accessibilityID(@"collectionView")]
+      performAction:grey_scrollToContentEdge(kGREYContentEdgeTop)];
+}
+
```
romoore, line 116: Add pragma mark

ajsecord (approved):

ianegordon:
I've asked Felix to take a look as he recently dealt with EdgeInsets.

ianegordon (changes requested):
Waiting for a LGTM from Felix.

ianegordon:
@felixaemiliano will be reviewing / testing this on his project tomorrow.

components/Collections/src/MDCCollectionViewFlowLayout.m
```diff
@@ -391,13 +391,28 @@ - (MDCCollectionViewLayoutAttributes *)updateSupplementaryViewAttribute:
   // view rows.
   CGRect insetFrame = attr.frame;
   if (!CGRectIsEmpty(insetFrame)) {
-    UIEdgeInsets insets = [self insetsAtSectionIndex:attr.indexPath.section];
+    UIEdgeInsets insets = self.sectionInset;
```
felixaemiliano, line 394: Is the intention that the default insets be derived from -insetsAtSectionIndex? Currently this is not the case: the insets are only set to that value if the collectionView's delegate is nil (which is unlikely).

Perhaps the correct implementation would be to set the insets here to [self insetsAtSectionIndex] and conditionally override below if the flowLayoutDelegate responds to the selector which allows subclasses to customize the insets.

felixaemiliano (commented):
This is definitely a welcome change that allows subclasses more control, thanks for adding it!

However, Classroom overrides the -cellWidthAtSectionIndex: method, which means we will not be affected by this change.

components/Collections/src/MDCCollectionViewFlowLayout.m
romoore: Good catch, I can switch the implementation to start with the internals and then call the delegate.

ianegordon (approved):
LGTM.

PR #1616, opened 2017-07-10
[Docs] Replaced internal site references with public equivalents.

ajsecord:


components/Dialogs/src/MDCDialogPresentationController.m
```diff
@@ -98,7 +98,7 @@ - (CGRect)frameOfPresentedViewInContainerView {
 - (void)presentationTransitionWillBegin {
   // TODO: Follow the Material spec description of Autonomous surface creation for both
   // presentation and dismissal of the dialog.
-  // https://spec.googleplex.com/quantum/motion/choreography.html#choreography-creation
+  // https://material.googleplex.com/motion/choreography.html#choreography-creation
```
ianegordon: Isn't googleplex internal only?

s/material.googleplex.com/material.io/

ianegordon (approved):

components/Dialogs/src/MDCDialogPresentationController.m
ajsecord: Yes! Of course. Updated to correct link.

PR #1627, opened 2017-07-12
[Collections] Add StylingDelegate methods to control separator display

### LINKED ISSUE #1406: [Collections]
### Overview
When the background is hidden on the header or a cell in a collection view, a thin bar is displayed in the item.
For example if you remove the background of the header in a collection, there is a line above the item like in this screenshot (Header/Footer example in the catalog with header background hidden).
simulator screen shot may 9 2017 11 55 21 am

### Reproduction steps
 - Hide the background for the headers in CollectionsHeaderFooterExample.m in the catalog
 - Open the example
 - Look at the separation of the header and the section
 - An unexpected line is displayed

### Version number
 - 23.0.2

gambard:
Adds methods to MDCCollectionViewStylingDelegate, allowing the delegate to control the display of the separator on cells, header and footer.

Closes #1406

### DIFF
--- a/components/Collections/examples/CollectionsCellSeparatorExample.m
+++ b/components/Collections/examples/CollectionsCellSeparatorExample.m
@@ -35,6 +35,11 @@ static NSString *const kReusableIdentifierItem = @"itemCellIdentifier";
   [self.collectionView registerClass:[MDCCollectionViewTextCell class]
           forCellWithReuseIdentifier:kReusableIdentifierItem];
 
+  // Register header.
+  [self.collectionView registerClass:[MDCCollectionViewTextCell class]
+          forSupplementaryViewOfKind:UICollectionElementKindSectionHeader
+                 withReuseIdentifier:UICollectionElementKindSectionHeader];
+
   // Populate content.
   _content = [NSMutableArray array];
   for (NSInteger i = 0; i < kSectionCount; i++) {
@@ -78,6 +83,46 @@ static NSString *const kReusableIdentifierItem = @"itemCellIdentifier";
   }
 
   return cell;
+}
+
+- (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView
+           viewForSupplementaryElementOfKind:(NSString *)kind
+                                 atIndexPath:(NSIndexPath *)indexPath {
+  MDCCollectionViewTextCell *supplementaryView =
+      [collectionView dequeueReusableSupplementaryViewOfKind:kind
+                                         withReuseIdentifier:kind
+                                                forIndexPath:indexPath];
+
+  if ([kind isEqualToString:UICollectionElementKindSectionHeader]) {
+    if (indexPath.section == 3) {
+      supplementaryView.textLabel.text = @"Header with separator";
+    } else if (indexPath.section == 4) {
+      supplementaryView.textLabel.text = @"Header without separator";
+    }
+  }
+
+  return supplementaryView;
+}
+
+- (CGSize)collectionView:(UICollectionView *)collectionView
+                             layout:(UICollectionViewLayout *)collectionViewLayout
+    referenceSizeForHeaderInSection:(NSInteger)section {
+  if (section == 3 || section == 4) {
+    return CGSizeMake(collectionView.bounds.size.width, MDCCellDefaultOneLineHeight);
+  }
+  return CGSizeZero;
+}
+
+#pragma mark - <MDCCollectionViewStylingDelegate>
+
+- (BOOL)collectionView:(UICollectionView *)collectionView
+    shouldHideItemSeparatorAtIndexPath:(NSIndexPath *)indexPath {
+  return indexPath.section == 3;
+}
+
+- (BOOL)collectionView:(UICollectionView *)collectionView
+    shouldHideHeaderSeparatorForSection:(NSInteger)section {
+  return section == 4;
 }
 
 @end
--- a/components/Collections/src/MDCCollectionViewFlowLayout.m
+++ b/components/Collections/src/MDCCollectionViewFlowLayout.m
@@ -374,7 +374,7 @@ static const NSInteger kSupplementaryViewZIndex = 99;
   attr.separatorColor = self.styler.separatorColor;
   attr.separatorInset = self.styler.separatorInset;
   attr.separatorLineHeight = self.styler.separatorLineHeight;
-  attr.shouldHideSeparators = self.styler.shouldHideSeparators;
+  attr.shouldHideSeparators = [self.styler shouldHideSeparatorForCellLayoutAttributes:attr];
 
   // Set inlay and hidden state if necessary.
   [self inlayAttributeIfNecessary:attr];
@@ -397,7 +397,7 @@ static const NSInteger kSupplementaryViewZIndex = 99;
       if ([attr.representedElementKind isEqualToString:UICollectionElementKindSectionHeader]) {
         insetFrame.origin.y += insets.top;
       } else if ([attr.representedElementKind
-                  isEqualToString:UICollectionElementKindSectionFooter]) {
+                     isEqualToString:UICollectionElementKindSectionFooter]) {
         insetFrame.origin.y -= insets.bottom;
       }
       attr.frame = insetFrame;
@@ -531,8 +531,8 @@ static const NSInteger kSupplementaryViewZIndex = 99;
           (id<MDCCollectionViewEditingDelegate>)self.collectionView.dataSource;
 
       // Check if delegate can select during editing.
-      if ([editingDelegate respondsToSelector:@selector(collectionView:
-                                                  canSelectItemDuringEditingAtIndexPath:)]) {
+      if ([editingDelegate respondsToSelector:@selector
+                           (collectionView:canSelectItemDuringEditingAtIndexPath:)]) {
         attr.shouldShowSelectorStateMask = [editingDelegate collectionView:self.collectionView
                                      canSelectItemDuringEditingAtIndexPath:attr.indexPath];
       }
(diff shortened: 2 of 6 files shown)


components/Collections/src/MDCCollectionViewStylingDelegate.h
```diff
@@ -90,6 +90,38 @@
...
+
+/**
+ Asks the delegate whether the specified footer should hide its separator.
+
+ @param collectionView The collection view.
+ @param section The collection view section.
+ @return If the footer separator should be hidden at the specified section.
+ */
+- (BOOL)collectionView:(nonnull UICollectionView *)collectionView
+    shouldHideFooterSeparatorForSection:(NSInteger)section;
```
gambard, line 123: I am not sure a footer can have a separator but in case, I think it is better to have this method.

components/Collections/src/MDCCollectionViewFlowLayout.m
```diff
@@ -531,8 +531,8 @@ - (void)updateCellStateMaskWithAttribute:(MDCCollectionViewLayoutAttributes *)at
           (id<MDCCollectionViewEditingDelegate>)self.collectionView.dataSource;
 
       // Check if delegate can select during editing.
-      if ([editingDelegate respondsToSelector:@selector(collectionView:
-                                                  canSelectItemDuringEditingAtIndexPath:)]) {
+      if ([editingDelegate respondsToSelector:@selector
+                           (collectionView:canSelectItemDuringEditingAtIndexPath:)]) {
```
romoore, line 535: This looks like the automatic formatter, do we normally break-up the @selector from the method name?

components/Collections/src/private/MDCCollectionViewStyler.m
```diff
@@ -246,6 +246,35 @@ - (void)setCellStyle:(MDCCollectionViewCellStyle)cellStyle {
...
+      [attr.representedElementKind isEqualToString:UICollectionElementKindSectionHeader];
+  BOOL isSectionFooter =
+      [attr.representedElementKind isEqualToString:UICollectionElementKindSectionFooter];
+  BOOL isCell = attr.representedElementCategory == UICollectionElementCategoryCell;
+  if (isSectionHeader && [self.delegate respondsToSelector:@selector
+                                        (collectionView:shouldHideHeaderSeparatorForSection:)]) {
+    shouldHideSeparator = [self.delegate collectionView:_collectionView
+                    shouldHideHeaderSeparatorForSection:indexPath.section];
+  }
+  if (isSectionFooter && [self.delegate respondsToSelector:@selector
```
romoore: Can the same indexPath be a header/footer/cell simultaneously?  If not, it may be better to if/else-if on the cell type first, then have nested conditionals for `respondsToSelector:` inside each.  Placing the `isCell` condition first seems like the primary case and so might be incrementally faster.

components/Collections/examples/CollectionsCellSeparatorExample.m
```diff
@@ -80,4 +85,44 @@ - (UICollectionViewCell *)collectionView:(UICollectionView *)collectionView
...
+  if (section == 3 || section == 4) {
+    return CGSizeMake(collectionView.bounds.size.width, MDCCellDefaultOneLineHeight);
+  }
+  return CGSizeZero;
+}
+
+#pragma mark - <MDCCollectionViewStylingDelegate>
+
+- (BOOL)collectionView:(UICollectionView *)collectionView
+    shouldHideItemSeparatorAtIndexPath:(NSIndexPath *)indexPath {
```
romoore, line 119: Please indent the second line 4 spaces (align with the "L" in `BOOL`) since we cannot align the colons.

components/Collections/examples/CollectionsCellSeparatorExample.m
```diff
@@ -80,4 +85,44 @@ - (UICollectionViewCell *)collectionView:(UICollectionView *)collectionView
...
+  return indexPath.section == 3;
+}
+
+- (BOOL)collectionView:(UICollectionView *)collectionView
+    shouldHideHeaderSeparatorForSection:(NSInteger)section {
```
romoore, line 124: Please indent 4 spaces here as well

components/Collections/src/MDCCollectionViewStyling.h
```diff
@@ -142,6 +142,16 @@ typedef NS_ENUM(NSUInteger, MDCCollectionViewCellLayoutType) {
...
 
+/**
+ Returns whether the separator should be hidden for the cell with this layout attributes, determined
+ by the cell position and type.
+
+ @param attr The cell's layout attributes.
+ @return Whether the separtor should be hidden.
+ */
+- (BOOL)shouldHideSeparatorForCellLayoutAttributes:
+        (nonnull MDCCollectionViewLayoutAttributes *)attr;
```
romoore, line 155: This looks like it was indented a bit too much (more than 4 spaces).

components/Collections/src/private/MDCCollectionViewStyler.m
```diff
@@ -246,6 +246,35 @@ - (void)setCellStyle:(MDCCollectionViewCellStyle)cellStyle {
   _cellStyle = cellStyle;
 }
 
+- (BOOL)shouldHideSeparatorForCellLayoutAttributes:(MDCCollectionViewLayoutAttributes *)attr {
+  BOOL shouldHideSeparator = self.shouldHideSeparators;
+  if (!self.delegate)
+    return shouldHideSeparator;
```
romoore: Please add curly braces  (`{}`) to surround the expression following the conditional (`if`)

romoore (changes requested):
A few small comments, and I would like to see Unit Tests added to the styler to verify the behavior with and without a styling delegate.

romoore:
Keeping myself assigned so I can track this issue.

components/Collections/examples/CollectionsCellSeparatorExample.m
gambard: Same.

components/Collections/examples/CollectionsCellSeparatorExample.m
gambard: I am already using a 4 space indentation, and this is what clang-format is giving. It also seems consistent with the rest of the methods.

components/Collections/src/MDCCollectionViewFlowLayout.m
gambard: I don't know. This is indeed the clang-format result.

components/Collections/src/MDCCollectionViewStyling.h
gambard: Auto-formatter.

components/Collections/src/private/MDCCollectionViewStyler.m
gambard: Done.

components/Collections/src/private/MDCCollectionViewStyler.m
gambard: The same indexPath can be a header/footer/cell, but an attribute is only one of them.
Done.

gambard (commented):
Thanks, PTAL.

components/Collections/examples/CollectionsCellSeparatorExample.m
romoore: You're right. I misremembered the indentation rules from the style guide. Sorry!

romoore:
Source looks good, could you please add unit tests?   I think `shouldHideSeparatorForCellLayoutAttributes:` in `MDCCollectionViewStyler.h` looks fairly testable.  Testing `updateAttribute:` in `MDCCollectionViewFlowLayout` would be great if you could as well.

gambard:
Tests added for shouldHideSeparatorForCellLayoutAttributes.

gambard:
I think testing updateAttribute would be too complex for this CL.

components/Collections/tests/unit/CollectionsStylerTests.m
```diff
@@ -0,0 +1,136 @@
...
+                         collectionViewLayout:[[UICollectionViewFlowLayout alloc] init]];
+  MDCCollectionViewStyler* styler =
+      [[MDCCollectionViewStyler alloc] initWithCollectionView:collectionView];
+  styler.delegate = nil;
+
+  // When
+  styler.shouldHideSeparators = YES;
+
+  // Then
+  XCTAssertTrue([styler shouldHideSeparatorForCellLayoutAttributes:cell01()]);
```
romoore: Please verify all 6 cases in each test, I think it makes it clearer that all cells are configured by the property.

components/Collections/src/MDCCollectionViewStyling.h
```diff
@@ -142,6 +142,16 @@ typedef NS_ENUM(NSUInteger, MDCCollectionViewCellLayoutType) {
 /* Whether to hide the cell separators. Defaults to NO. */
 @property(nonatomic) BOOL shouldHideSeparators;
 
+/**
+ Returns whether the separator should be hidden for the cell with this layout attributes, determined
+ by the cell position and type.
```
romoore, line 147: ~~Please add a comment to mention that setting `shouldHideSeparators` to `NO` should cause this method to always return `NO`.~~

Please add a comment that this method may override the `shouldHideSeparators` property depending on which delegate methods are implemented.

romoore (changes requested):
Very minor feedback on tests.

components/Collections/tests/unit/CollectionsStylerTests.m
gambard: Done.

components/Collections/src/MDCCollectionViewStyling.h
gambard: Done.

romoore (approved):

PR #1630, opened 2017-07-12
[Collections] Fallback to default cell background color on nil

### LINKED ISSUE #1629: [Collections] Styler should use default cell background color if delegate returns nil
Styling delegate method `MDCCollectionViewStylingDelegate#collectionView:cellBackgroundColorAtIndexPath:` returns a nullable color value. However returning `nil` causes an unexpected termination.
`MDCCollectionViewStyler` should fallback to default background color if delegate returns nil.

### Version number
26.0.0

sigito:
Closes #1629.


ajsecord (approved):
Thanks!

PR #1652, opened 2017-07-17
[Ink] Ink reset passes animation value to ripples

### LINKED ISSUE #1651: [Ink] MDCInkLayer does not pass animated parameter when resetting ink
### Overview
When `[MDCInkLayer -resetAllInk:]` is called, the MDCInkLayerRipple receivers do not get the correct value of the `animated` parameter.  This is because `makeObjectsPerformSelector` is being called without the argument.

### Version number
- Found in 26.0.0

iPad Mini 4, iOS 10.3.2

romoore:
When the MDCInkLayer resets ink, it will now pass the `animated` parameter to the ripple sublayers. This fixes a bug where resetting the ink without animation would cause animations to be executed.

Closes #1651

### DIFF
--- a/components/Ink/src/private/MDCInkLayer+Testing.h
+++ b/components/Ink/src/private/MDCInkLayer+Testing.h
@@ -1,0 +1,28 @@
+/*
+ Copyright 2017-present the Material Components for iOS authors. All Rights Reserved.
+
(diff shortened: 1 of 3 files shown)


jgunaratne (approved):

romoore:
Why is the coverage going down so much?

PR #1656, opened 2017-07-18
[Buttons] Reset Ink when moved to new superview

### LINKED ISSUE #1655: [Buttons] MDCButton should reset Ink when moved between superviews (b/31320843)
### Overview
When MDCButton is moved between superviews (removed from one, added to another), it continues to animate Ink effects.  Worse yet, because it uses UITouch events to start/end Ink animations, the underlying Ink view can end up in a state where it never terminates an ink effect.  This has resulted in permanently "stacked" effects in upcoming features like [FAB Action](https://github.com/material-components/material-components-ios/tree/experiment-fabaction) when buttons are moved between views.

### Reproduction steps
1. Place a button in a view 
1. After tapping and holding on the button, remove it from the view
1. Release the tap
1. Re-add the button to the view
1. Repeat the steps above

### Expected Result
Ink should be dissipated/reset after being re-added to a view

### Actual Result
Ink effect never dissipates because neither the `touchesEnded`, `touchesCancelled`, nor `touchesMoved` delegate methods are called after the view is removed from the superview.

### Version number
27.0.0

iPad MIni 4/iOS 10.3.2

### Related issues
#1105

romoore:
MDCButton will now reset all Ink effects before it is added to a new
superview. This prevents retaining Ripple objects indefinitely if the
button is moved between superviews while a touch is held down.

Closes #1655

### DIFF
--- a/components/Buttons/src/MDCButton.m
+++ b/components/Buttons/src/MDCButton.m
@@ -283,6 +283,13 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
   return CGRectContainsPoint(UIEdgeInsetsInsetRect(self.bounds, _hitAreaInsets), point);
 }
 
+- (void)willMoveToSuperview:(UIView *)newSuperview {
+  [super willMoveToSuperview:newSuperview];
+  if (newSuperview) {
+    [self.inkView cancelAllAnimationsAnimated:NO];
+  }
+}
+
 #pragma mark - UIResponder
 
 - (void)touchesBegan:(NSSet *)touches withEvent:(UIEvent *)event {
--- a/components/Buttons/src/MDCButton.m
+++ b/components/Buttons/src/MDCButton.m
@@ -285,11 +285,7 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
 
 - (void)willMoveToSuperview:(UIView *)newSuperview {
   [super willMoveToSuperview:newSuperview];
-  if (newSuperview) {
-    [self.inkView cancelAllAnimationsAnimated:YES];
-  } else {
-    [self.inkView cancelAllAnimationsAnimated:NO];
-  }
+  [self.inkView cancelAllAnimationsAnimated:NO];
 }
 
 #pragma mark - UIResponder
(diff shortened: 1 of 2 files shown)


jgunaratne (approved):

components/Buttons/src/MDCButton.m
```diff
@@ -283,6 +283,13 @@ - (BOOL)pointInside:(CGPoint)point withEvent:(UIEvent *)event {
   return CGRectContainsPoint(UIEdgeInsetsInsetRect(self.bounds, _hitAreaInsets), point);
 }
 
+- (void)willMoveToSuperview:(UIView *)newSuperview {
+  [super willMoveToSuperview:newSuperview];
+  if (newSuperview) {
```
ianegordon: Is there a reason we aren't checking for newSuperview == nil?

ianegordon (changes requested):
Let's wait on Chat for an update on Speeddial since this affects rendering.

romoore:
Marking this blocked pending discussion with @chatura-atapattu  regarding the FAB Action Menu.

chatura-atapattu:
Taking aside the impact on the Speeddial, what would be the expected behavior for MDCButton if it was removed from one view and added to another mid-touch and ripple effect.  I think that instead of the ripple effect being immediately ended, we'd expect it to continue and end naturally, right? Would triggering the end animation achieve that behavior?

romoore:
I like where you're going with this, and I wonder if we might "clear" the ink either way when the view moves out of its parent, but differently depending on where it's going:
* Animated ink dissipation when a view is directly moved between parent views (`newSuperview != nil`)
* Unanimated ink removal when a view is removed from its parent view (`newSuperview == nil`)

What do you think of this approach?

chatura-atapattu:
Yeah, I think thats the way to go.

ianegordon (approved):

romoore:
Pulling in the latest changes and a dirty test with ActionMenu, it seems performing the animation when the view is moved will cause the ink to ripple twice when the menu is dismissed.

fab-dismiss-animatedink

It looks a little better if there is no animated dismissal of the ink, the result is basically the fade-out being cut short.

fab-dismiss-inanimateink

romoore:
A quick offline conversation with @ianegordon and I think we'll just have to dismiss the Ink in an unanimated way whenever it's removed from its superview (regardless of whether it's immediately added to another superview).  The visual artifact that results (see gif in my previous comment) is an acceptably-small side effect in exchange for removing the memory leak.  

I will trawl the remaining components in the MDC Catalog app to look for any other surprises, but I think the Action Menu/Speed Dial will need to be refactored slightly if we want to eliminate the ink dissipation artifact when the menu is dismissed.

PR #1659, opened 2017-07-18
[Snackbar] Notify MDCOverlayObservers of bottom offset changes 

### LINKED ISSUE #1643: MDCSnackbarManager.setBottomOffset doesn't update overlay observers
Internal issue 63642943.

>We have the usual setup of a FAB getting out of the way of the snackbars via MDCOverlayObserver.
We also to adjust the snackbar bottom offset depending on whether our Tab bar is shown.
When calling MDCSnackbarManager.setBottomOffset, we don't get notified via the observers if there is a snackbar displayed at the time of the method call.
In short, MDCSnackbarManager.setBottomOffset should immediately notify the overlay observers as it moves the existing overlay.

samnm:
Notify MDCOverlayObservers of `MDCSnackbarManager.bottomOffset` changes and update an example to showcase this.

Closes #1643.

### DIFF
--- a/components/Snackbar/examples/SnackbarOverlayViewExample.m
+++ b/components/Snackbar/examples/SnackbarOverlayViewExample.m
@@ -22,12 +22,13 @@
 
 static const CGFloat kFABBottomOffset = 24.0f;
 static const CGFloat kFABSideOffset = 24.0f;
+static const CGFloat kBottomBarHeight = 44.0f;
 
 @implementation SnackbarOverlayViewExample
 
 - (void)viewDidLoad {
   [super viewDidLoad];
-  [self setupExampleViews:@[@"Show Snackbar"]];
+  [self setupExampleViews:@[@"Show Snackbar", @"Toggle bottom bar"]];
   self.title = @"Snackbar Overlay View";
 
   // Make sure we're listening for overlay notifications.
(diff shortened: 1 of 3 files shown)


components/Snackbar/examples/supplemental/SnackbarExampleSupplemental.h
```diff
@@ -30,6 +30,10 @@
 /** The floating action button shown in the bottom right corner. */
 @property(nonatomic) MDCFloatingButton *floatingButton;
```
ShepJGoogle, line 32: Should we include documentation?

components/Snackbar/examples/supplemental/SnackbarExampleSupplemental.h
```diff
@@ -30,6 +30,10 @@
...
 
+@property(nonatomic) UIView *bottomBar;
+
```
ShepJGoogle, line 34: Same as above

ShepJGoogle (dismissed):

components/Snackbar/src/private/MDCSnackbarOverlayView.m
```diff
@@ -542,6 +542,20 @@ - (void)setBottomOffset:(CGFloat)bottomOffset {
...
     self.bottomConstraint.constant = -[self dynamicBottomMargin];
     [self triggerSnackbarLayoutChange];
+
+    // If there is no snackbar the following method returns CGRectNull, but we still need to notify
+    // observers of bottom offset changes.
+    CGRect frame = [self snackbarRectInScreenCoordinates];
+    if (CGRectIsNull(frame)) {
+      frame = CGRectMake(0,
+                         CGRectGetMaxY(self.frame) - self.bottomOffset,
+                         CGRectGetMaxX(self.frame),
```
ianegordon: I would expect this to be CGRectGetWidth(self.frame).
If this isn't GetWidth, consider documenting.

ianegordon (changes requested):

components/Snackbar/src/private/MDCSnackbarOverlayView.m
samnm: Good catch. Fixed.

components/Snackbar/examples/supplemental/SnackbarExampleSupplemental.h
samnm: It isn't really necessary here... I just removed the comment from the floatingButton property.

ajsecord (approved):

components/Snackbar/examples/SnackbarOverlayViewExample.m
```diff
@@ -51,17 +52,50 @@ - (void)viewDidLoad {
   self.floatingButton.frame = fabFrame;
   self.floatingButton.autoresizingMask =
       (UIViewAutoresizingFlexibleLeftMargin | UIViewAutoresizingFlexibleTopMargin);
+
+  self.bottomBar = [[UIView alloc] initWithFrame:CGRectMake(0,
+                                                            CGRectGetMaxY(self.view.bounds),
+                                                            CGRectGetMaxX(self.view.bounds),
```
ianegordon: Do we want CGRectGetWidth() here?
If self.bottomBar is a subview of self.view I would expect to use the Height / Width of self.view.bounds.
I wouldn't typically expect to use anything other than GetHeight and GetWidth with bounds.
GetMaxX and GetMaxY would be more common with self.frame.

components/Snackbar/src/private/MDCSnackbarOverlayView.m
```diff
@@ -542,6 +542,20 @@ - (void)setBottomOffset:(CGFloat)bottomOffset {
 
     self.bottomConstraint.constant = -[self dynamicBottomMargin];
     [self triggerSnackbarLayoutChange];
+
+    // If there is no snackbar the following method returns CGRectNull, but we still need to notify
+    // observers of bottom offset changes.
+    CGRect frame = [self snackbarRectInScreenCoordinates];
+    if (CGRectIsNull(frame)) {
+      frame = CGRectMake(0,
+                         CGRectGetMaxY(self.frame) - self.bottomOffset,
```
ianegordon: Consider adding a comment along the lines of:
// We use the MaxY(self.frame) instead of the Height(self.frame) since we are dealing with screen coordinates.

ajsecord:
@ianegordon is out sick, I can confirm that @samnm made the change requested.

@ShepJGoogle Can you please review your requested changes?

ianegordon (changes requested):
This appears to break on rotation.

From Snackbar Overlay View in portrait.
Toggle bottom bar
Show Snackbar
Rotate to landscape (button disappears behind snackbar and bottombar)
Rotate back to portrait (button jump to the upper right corner)

screen shot 2017-07-26 at 13 28 10

components/Snackbar/examples/SnackbarOverlayViewExample.m
samnm: Fixed.

components/Snackbar/src/private/MDCSnackbarOverlayView.m
samnm: Fixed.

samnm:
That rotation bug appears to have existed before this change. I can't find a fix for it right now so I'd rather not block this PR on that fix.

ianegordon (approved):

ianegordon:
Will solve rotation issue in another PR.
#1701 created.

PR #1666, opened 2017-07-20
[TextField] Add color themer for text fields

jgunaratne:
Adds a color themer for text fields applied through its controller.


willlarche (approved):
Looks good. Can you add a test and an example?

jgunaratne:
Added text field to themer demo:

simulator screen shot jul 20 2017 2 36 31 pm

willlarche:
I do think that’s just beautiful. 😼

PR #1672, opened 2017-07-21
Change TravisCI Destination

ianegordon:
Experiment.
DO NOT MERGE.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ branches:
     - stable
     - /^feature-.*$/
 
-xcode_workspace: catalog/MDCCatalog.xcworkspace
+xcode_workspace: ./catalog/MDCCatalog.xcworkspace
 xcode_scheme: MDCCatalog
 
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,9 @@ branches:
     - stable
     - /^feature-.*$/
 
+xcode_workspace: catalog/MDCCatalog.xcworkspace
+xcode_scheme: MDCCatalog
+
 env:
   global:
   - LC_CTYPE=en_US.UTF-8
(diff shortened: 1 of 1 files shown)


willlarche (approved):

.travis.yml
```diff
@@ -41,8 +41,9 @@ script:
   - xcodebuild -showsdks
   - pod --version
   - scripts/prep_all
-  - travis_wait scripts/build_all --verbose
-  - travis_wait scripts/test_all catalog/MDCCatalog.xcworkspace:MDCUnitTests
+  - xcodebuild build test
+#  - travis_wait scripts/build_all --verbose
+#  - travis_wait scripts/test_all catalog/MDCCatalog.xcworkspace:MDCUnitTests
```
ajsecord, line 46: Err.. why aren't we building everything? And no more unit tests?

ajsecord (changes requested):

.travis.yml
```diff
@@ -14,18 +14,18 @@ branches:
...
+
 env:
   global:
   - LC_CTYPE=en_US.UTF-8
   - LANG=en_US.UTF-8
   matrix:
-    - DESTINATION="platform=iOS Simulator,name=iPhone 7,OS=10.3.1" SDK="iphonesimulator10.3.1"
-    # The iPhone 5 was chosen because it is a 32-bit device.
-    # 8.1 was chosen because it is the earliest iOS offered here.
-    - DESTINATION="platform=iOS Simulator,name=iPhone 5,OS=8.1" SDK="iphonesimulator8.1"
```
willlarche, line 25: We really need this one still. It's 32-bit and iOS 8. Removing it lowers our coverage by .3%.

willlarche (changes requested):

.travis.yml
romoore: If we have 32-bit specific code, that's probably more important than the percentage.

.travis.yml
ianegordon: "xcodebuild build test" does build MDCCatalog, and runs the unit and UI tests.
I'm trying to determine if we can put all of our workspaces in a matrix.
This PR may not turn out to be something we merge.  (Description updated above)

ianegordon:
As a point of reference, I was able to get MDFInternationalization building and testing on all three devices.

See:
https://github.com/material-foundation/material-internationalization-ios

Config file:
https://github.com/material-foundation/material-internationalization-ios/blob/develop/.travis.yml

ianegordon:
To be continued,.....

PR #1676, opened 2017-07-24
[TextFields] Overlay views.

willlarche:
Maps leftView and rightView of text field map to leading and trailing.

Adds a mandatory trailing view to the text input protocol.

### DIFF
--- a/components/TextFields/src/MDCTextField.h
+++ b/components/TextFields/src/MDCTextField.h
@@ -30,4 +30,18 @@ extern NSString *_Nonnull const MDCTextFieldTextDidSetTextNotification;
 /** MDCTextField does not implement borders that conform to UITextBorderStyle. */
 @property(nonatomic, assign) UITextBorderStyle borderStyle NS_UNAVAILABLE;
 
+/**
+ An overlay view on the side of the input where reading and writing lines begin. In LTR this is
+ the Left side. In RTL, the Right side.
+ */
+@property(nonatomic, nullable, strong) UIView *leadingView;
+
+/**
+ Controls when the leading view will display.
+ */
+@property(nonatomic) UITextFieldViewMode leadingViewMode;
+
+/** Insets used to calculate the spacing of subviews. */
+@property(nonatomic, assign, readonly) UIEdgeInsets textInsets;
+
 @end
--- a/components/TextFields/src/MDCTextField.h
+++ b/components/TextFields/src/MDCTextField.h
@@ -39,7 +39,7 @@ extern NSString *_Nonnull const MDCTextFieldTextDidSetTextNotification;
 /**
  Controls when the leading view will display.
  */
-@property(nonatomic) UITextFieldViewMode leadingViewMode;
+@property(nonatomic, assign) UITextFieldViewMode leadingViewMode;
 
 /** Insets used to calculate the spacing of subviews. */
 @property(nonatomic, assign, readonly) UIEdgeInsets textInsets;
--- a/components/TextFields/src/MDCTextField.h
+++ b/components/TextFields/src/MDCTextField.h
@@ -31,8 +31,9 @@ extern NSString *_Nonnull const MDCTextFieldTextDidSetTextNotification;
 @property(nonatomic, assign) UITextBorderStyle borderStyle NS_UNAVAILABLE;
 
 /**
- An overlay view on the side of the input where reading and writing lines begin. In LTR this is
- the Left side. In RTL, the Right side.
+ An overlay view on the leading side.
+ 
+ Note: if RTL is engaged, this will return the .rightView and if LTR, it will return the .leftView.
  */
 @property(nonatomic, nullable, strong) UIView *leadingView;
 
--- a/components/TextFields/src/MDCTextField.m
+++ b/components/TextFields/src/MDCTextField.m
@@ -92,9 +92,15 @@ static const CGFloat MDCTextInputEditingRectRightViewPaddingCorrection = -2.f;
 
   copy.fundament = [self.fundament copy];
   copy.enabled = self.isEnabled;
+  if ([self.leadingView conformsToProtocol:@protocol(NSCopying)]) {
+    copy.leadingView = [self.leadingView copy];
+  }
   copy.leadingViewMode = self.leadingViewMode;
   copy.placeholder = [self.placeholder copy];
   copy.text = [self.text copy];
+  if ([self.trailingView conformsToProtocol:@protocol(NSCopying)]) {
+    copy.trailingView = [self.trailingView copy];
+  }
   copy.trailingViewMode = self.trailingViewMode;
 
   return copy;
--- a/components/TextFields/src/MDCTextField.m
+++ b/components/TextFields/src/MDCTextField.m
@@ -197,6 +197,22 @@ static const CGFloat MDCTextInputEditingRectRightViewPaddingCorrection = -2.f;
   _fundament.enabled = enabled;
 }
 
+- (UIView *)leadingView {
+  return self.leftView;
+}
+
+- (void)setLeadingView:(UIView *)leadingView {
+  self.leftView = leadingView;
+}
+
+- (UITextFieldViewMode)leadingViewMode {
+  return self.leadingViewMode;
+}
+
+- (void)setLeadingViewMode:(UITextFieldViewMode)leadingViewMode {
+  self.leftViewMode = leadingViewMode;
+}
+
 - (NSString *)placeholder {
   return self.fundament.placeholder;
 }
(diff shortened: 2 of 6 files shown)


romoore:
Why do we need the "mandatory trailing view"? Could you explain a bit more in the commit message for future archaeologists?

components/TextFields/src/MDCTextField.h
```diff
@@ -30,4 +30,18 @@ extern NSString *_Nonnull const MDCTextFieldTextDidSetTextNotification;
...
+/**
+ An overlay view on the side of the input where reading and writing lines begin. In LTR this is
+ the Left side. In RTL, the Right side.
+ */
+@property(nonatomic, nullable, strong) UIView *leadingView;
+
+/**
+ Controls when the leading view will display.
+ */
+@property(nonatomic) UITextFieldViewMode leadingViewMode;
```
romoore: (nonatomic, assign) please.

components/TextFields/src/MDCTextInput.h
```diff
@@ -130,6 +130,17 @@
...
+/**
+ An overlay view on the side of the input opposite from where reading and writing lines begin. In
+ LTR this means it will show on the Right side. In LTR, the Left side.
+ */
+@property(nonatomic, nullable, strong) UIView *trailingView;
+
+/**
+ Controls when the trailing view will display.
+ */
+ @property(nonatomic) UITextFieldViewMode trailingViewMode;
```
romoore: (nonatomic, assign) please

components/TextFields/tests/unit/TextFieldTests.swift
```diff
@@ -52,11 +64,28 @@ class TextFieldTests: XCTestCase {
       XCTAssertEqual(textField.font, textFieldCopy.font)
       XCTAssertEqual(textField.hidesPlaceholderOnInput, textFieldCopy.hidesPlaceholderOnInput)
       XCTAssertEqual(textField.isEnabled, textFieldCopy.isEnabled)
+
+      if let leadingViewCopy = textField.leadingView as? UILabel {
```
romoore: This isn't a copy, is it?  Isn't this just a pointer to the actual object?

components/TextFields/tests/unit/TextFieldTests.swift
```diff
@@ -52,11 +64,28 @@ class TextFieldTests: XCTestCase {
...
+        XCTAssertEqual(leadingViewCopy, leadingView)
+        XCTAssertEqual(leadingViewCopy.text, leadingView.text)
+      } else {
+        XCTFail("No leading view or it isn't a UILabel")
+      }
+
+      if let trailingViewCopy = textField.trailingView as? UILabel {
```
romoore: Same - maybe I'm misreading this code, but this doesn't look like a copy.

components/TextFields/tests/unit/TextFieldTests.swift
```diff
@@ -100,10 +129,29 @@ class TextFieldTests: XCTestCase {
 
     XCTAssertTrue(textField.subviews.contains(leftView))
     XCTAssertTrue(textField.subviews.contains(rightView))
+
+    XCTAssertEqual(textField.leadingView, leftView)
```
romoore: Should we check the orientation first? This test will fail if the Simulator is in an RTL locale, no?

romoore (changes requested):
Minor property nits, some concern about RTL/LTR unit test, and nits about object copies vs pointers.

willlarche:
@romoore We need trailing views because the material guidelines have trailing views on all text fields but only require leading views on single line text fields. So the spec is the paper trail!

components/TextFields/src/MDCTextInput.h
willlarche: Ugh. I don't know how I missed those. Done

components/TextFields/src/MDCTextField.h
willlarche: Done

components/TextFields/tests/unit/TextFieldTests.swift
willlarche: This test makes sure the copy's propertys match the original. The copy happens on the line 
```    if let textFieldCopy = textField.copy() as? MDCTextField {```

File: components/TextFields/tests/unit/TextFieldTests.swift (hunk 5 above)
Comment (willlarche) [reply components/TextFields/tests/unit/TextFieldTests.swift]: I would LOVE to have actual RTL specific test code but we don't right now. I could add it in but I'd add a bunch more if I knew we had it.

File: components/TextFields/tests/unit/TextFieldTests.swift (hunk 3 above)
Comment (romoore) [reply components/TextFields/tests/unit/TextFieldTests.swift]: So shouldn't it be textFieldCopy.leadingView ?

File: components/TextFields/tests/unit/TextFieldTests.swift (hunk 5 above)
Comment (romoore) [reply components/TextFields/tests/unit/TextFieldTests.swift]: At a minimum, can we skip this test if we're in RTL?

File: components/TextFields/src/MDCTextInput.h [hunk 6]
```diff
@@ -130,6 +130,17 @@
  */
 @property(nonatomic, nonnull, strong, readonly) UILabel *trailingUnderlineLabel;
 
+/**
+ An overlay view on the side of the input opposite from where reading and writing lines begin. In
+ LTR this means it will show on the Right side. In LTR, the Left side.
+ */
```
ajsecord, line 145: I'm not sure re-explaining what "trailing" means is useful here. I think you can omit all but the first sentence.

ajsecord (approved):
Small nit

components/TextFields/tests/unit/TextFieldTests.swift
willlarche: Yup! That's a typo. Which explains why it was passing. UILabel doesn't conform to copy. Pulling this out.

components/TextFields/tests/unit/TextFieldTests.swift
willlarche: Same issue as above. Done.

components/TextFields/src/MDCTextInput.h
willlarche: I took another crack at it. Made it more specific.

components/TextFields/tests/unit/TextFieldTests.swift
willlarche: I put in a check. It requires iOS 9 but seeing as we don't have an RTL device testing, much less one that's on iOS 8, I think it'll be fine.

components/TextFields/src/MDCTextField.m
```diff
@@ -76,15 +83,25 @@ - (void)dealloc {
...
+  [aCoder encodeInteger:self.leftViewMode forKey:MDCTextFieldLeftViewModeKey];
+  [aCoder encodeInteger:self.rightViewMode forKey:MDCTextFieldRightViewModeKey];
 }
 
 - (instancetype)copyWithZone:(NSZone *)zone {
   MDCTextField *copy = [[[self class] alloc] init];
 
   copy.fundament = [self.fundament copy];
   copy.enabled = self.isEnabled;
+  if ([self.leadingView conformsToProtocol:@protocol(NSCopying)]) {
```
romoore: It seems the call to `-init` does not call `-commonMDCTextFieldInitialization`.  Is this intentional? Does a copy need to have all of its views, observers, etc?

components/TextFields/tests/unit/TextFieldTests.swift
```diff
@@ -29,18 +29,20 @@ class TextFieldTests: XCTestCase {
     XCTAssertEqual(textField.attributedText?.string, string)
   }
 
-  func testCopyingTextField() {
+  func testCopying() {
     let textField = MDCTextField()
 
     textField.clearButtonColor = .red
```
romoore, line 45: Test `super.borderStyle` is assigned .none?

romoore (commented):
Not sure if `copy` should call commonInit or not.

components/TextFields/tests/unit/TextFieldTests.swift
willlarche: borderStyle is not supported.

components/TextFields/src/MDCTextField.m
willlarche: The init is not overridden. When you call init on a UIView, it calls the designated initializer which is initWithFrame which calls the common init.

But to be even clearer, I've made it the initWithFrame explicitly instead of implicitly.

romoore: Awesome.

components/TextFields/tests/unit/TextFieldTests.swift
romoore: Yes, so should we confirm that we are explicitly setting it to .none in the commonInit?

romoore:
Might just wanna rebase and force push?

willlarche:
@romoore I am working on the rebase. My local develop was pointing at my fork on my home computer.

romoore (approved):

PR #1677, opened 2017-07-24
[TextFields] Reorganizing and internal clean up

willlarche:
In preparation for multiple text input classes, non-shared logic has been moved into text field and out of the fundament and text insets have been renamed.

### DIFF
--- a/components/TextFields/src/MDCTextField.m
+++ b/components/TextFields/src/MDCTextField.m
@@ -123,6 +123,20 @@ static const CGFloat MDCTextInputEditingRectRightViewPaddingCorrection = -2.f;
                       object:self];
 }
 
+#pragma mark - Layout
+
+- (UIEdgeInsets)textInsets {
+  UIEdgeInsets textInsets = UIEdgeInsetsZero;
+
+  textInsets.top = MDCTextInputFullPadding;
+  textInsets.bottom = MDCTextInputFullPadding;
+
+  if ([self.positioningDelegate respondsToSelector:@selector(textInsets:)]) {
+    return [self.positioningDelegate textInsets:textInsets];
+  }
+  return textInsets;
+}
+
 #pragma mark - Properties Implementation
 
 - (UIButton *)clearButton {
@@ -298,14 +312,14 @@ static const CGFloat MDCTextInputEditingRectRightViewPaddingCorrection = -2.f;
 #pragma mark - UITextField Overrides
 
 // This method doesn't have a positioning delegate mirror per se. But it uses the
-// textContainerInsets' value that the positioning delegate can return to inset this text rect.
+// textInsets' value that the positioning delegate can return to inset this text rect.
 - (CGRect)textRectForBounds:(CGRect)bounds {
   CGRect textRect = bounds;
 
   // Standard textRect calculation
-  UIEdgeInsets textContainerInset = [_fundament textContainerInset];
-  textRect.origin.x += textContainerInset.left;
-  textRect.size.width -= textContainerInset.left + textContainerInset.right;
+  UIEdgeInsets textInsets = self.textInsets;
+  textRect.origin.x += textInsets.left;
+  textRect.size.width -= textInsets.left + textInsets.right;
 
   // Adjustments for .leftView, .rightView
   // When in RTL mode, the .rightView is presented using the leftViewRectForBounds frame and the
@@ -435,7 +449,7 @@ static const CGFloat MDCTextInputEditingRectRightViewPaddingCorrection = -2.f;
 }
 
 - (CGFloat)centerYForOverlayViews:(CGFloat)heightOfView {
-  CGFloat centerY = [_fundament textContainerInset].top +
+  CGFloat centerY = self.textInsets.top +
                     (self.placeholderLabel.font.lineHeight / 2.f) - (heightOfView / 2.f);
   return centerY;
 }
--- a/components/TextFields/src/MDCTextField.m
+++ b/components/TextFields/src/MDCTextField.m
@@ -174,10 +174,7 @@ static const CGFloat MDCTextInputEditingRectRightViewPaddingCorrection = -2.f;
 }
 
 - (BOOL)needsUpdateUnderlinePosition {
-  NSInteger currentConstant = (NSInteger)(self.underlineY.constant * 1000);
-  NSInteger idealConstant = (NSInteger)([self underlineYConstant] * 1000);
-
-  return currentConstant != idealConstant;
+  return MDCCGFloatEqual(self.underlineY.constant, [self underlineYConstant]);
 }
 
 - (void)updateUnderlinePosition {
--- a/components/TextFields/src/MDCTextField.m
+++ b/components/TextFields/src/MDCTextField.m
@@ -37,12 +37,11 @@ static const CGFloat MDCTextInputEditingRectRightViewPaddingCorrection = -2.f;
 
 @property(nonatomic, strong) MDCTextInputCommonFundament *fundament;
 
-/** 
- 
- This constraint controls the default center Y of the underline view. 
+/**
+ Constraint for center Y of the underline view.
 
- Default is from the top of the text field with constant of height of input's font line-height +
- half padding.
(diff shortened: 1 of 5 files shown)


components/TextFields/src/MDCTextField.m
```diff
@@ -34,6 +34,8 @@ @interface MDCTextField ()
 
 @property(nonatomic, strong) MDCTextInputCommonFundament *fundament;
 
+@property(nonatomic, strong) NSLayoutConstraint *underlineY;
```
romoore: Comment to document this property?

components/TextFields/src/MDCTextField.m
```diff
@@ -106,6 +111,72 @@ - (void)commonMDCTextFieldInitialization {
...
+                               attribute:NSLayoutAttributeTrailing
+                               relatedBy:NSLayoutRelationEqual
+                                  toItem:self
+                               attribute:NSLayoutAttributeTrailing
+                              multiplier:1
+                                                                        constant:0];
+  underlineTrailing.priority = UILayoutPriorityDefaultLow;
+  underlineTrailing.active = YES;
+
+  CGFloat estimatedTextHeight = MDCCeil(self.font.lineHeight * 2.f) / 2.f;
```
romoore: Does this estimation work for CJK fonts?

components/TextFields/src/MDCTextField.m
```diff
@@ -106,6 +111,72 @@ - (void)commonMDCTextFieldInitialization {
...
+                               attribute:NSLayoutAttributeTop
+                              multiplier:1
+                                constant:[self textInsets].top + estimatedTextHeight +
+   MDCTextInputHalfPadding];
+  _underlineY.priority = UILayoutPriorityDefaultLow;
+  _underlineY.active = YES;
+}
+
+- (CGFloat)underlineYConstant {
+  CGFloat estimatedTextHeight = MDCCeil(self.font.lineHeight * 2.f) / 2.f;
```
romoore: This same estimation is in intrinsicContentSize and above in setupUnderlineConstraints. Consider turning it into a method.

components/TextFields/src/MDCTextField.m
```diff
@@ -106,6 +111,72 @@ - (void)commonMDCTextFieldInitialization {
...
+
+  return [self textInsets].top + estimatedTextHeight + MDCTextInputHalfPadding;
+}
+
+- (BOOL)needsUpdateUnderlinePosition {
+  return self.underlineY.constant != [self underlineYConstant];
```
romoore: Better to compare these floats to within some delta rather than exactly the same?

romoore (changes requested):
I think the height estimation can be turned into a method/function. I'm also concerned about direct comparison of floating point values.

components/TextFields/src/MDCTextField.m
willlarche: CJK support will be added later. Right now it's guaranteed on latin, arabic and hebrew.

components/TextFields/src/MDCTextField.m
willlarche: Done

components/TextFields/src/MDCTextField.m
willlarche: Good idea. Done.

components/TextFields/src/MDCTextField.m
willlarche: So true. Done.

romoore (approved):

components/TextFields/src/MDCTextField.m
```diff
@@ -34,6 +34,15 @@ @interface MDCTextField ()
 
 @property(nonatomic, strong) MDCTextInputCommonFundament *fundament;
 
+/** 
+ 
+ This constraint controls the default center Y of the underline view.
```
ajsecord: Delete the leading empty line. Probably just "Constraint for the default center y-coordinate of the underline view."

components/TextFields/src/MDCTextField.m
```diff
@@ -34,6 +34,15 @@ @interface MDCTextField ()
 
 @property(nonatomic, strong) MDCTextInputCommonFundament *fundament;
 
+/** 
+ 
+ This constraint controls the default center Y of the underline view. 
+
+ Default is from the top of the text field with constant of height of input's font line-height +
+ half padding.
```
ajsecord: Too complicated to understand in a single go. Maybe add ", e.g. below the text field at a distance of one-half of the padding"?

components/TextFields/src/MDCTextField.m
```diff
@@ -106,6 +118,78 @@ - (void)commonMDCTextFieldInitialization {
...
+                               relatedBy:NSLayoutRelationEqual
+                                  toItem:self
+                               attribute:NSLayoutAttributeTrailing
+                              multiplier:1
+                                                                        constant:0];
+  underlineTrailing.priority = UILayoutPriorityDefaultLow;
+  underlineTrailing.active = YES;
+
+  _underlineY =
+  [NSLayoutConstraint constraintWithItem:self.underline
```
ajsecord: +4

components/TextFields/src/MDCTextField.m
```diff
@@ -106,6 +118,78 @@ - (void)commonMDCTextFieldInitialization {
...
+- (CGFloat)underlineYConstant {
+  return [self textInsets].top + [self estimatedTextHeight] + MDCTextInputHalfPadding;
+}
+
+- (BOOL)needsUpdateUnderlinePosition {
+  NSInteger currentConstant = (NSInteger)(self.underlineY.constant * 1000);
+  NSInteger idealConstant = (NSInteger)([self underlineYConstant] * 1000);
+
+  return currentConstant != idealConstant;
+}
```
ajsecord: This logic is weird. Was 1000 just chosen at random?

components/TextFields/src/MDCTextInputControllerFullWidth.m
```diff
@@ -731,21 +731,21 @@ - (void)updateFontsForDynamicType {
  MAX(underlineLabelsOffset,MDCTextInputVerticalHalfPadding)           // Padding and/or labels
  */
 // clang-format on
-- (UIEdgeInsets)textContainerInset:(UIEdgeInsets)defaultInsets {
+- (UIEdgeInsets)textInsets:(UIEdgeInsets)defaultInsets {
   // NOTE: UITextFields have a centerY based layout. But you can change EITHER the height or the Y.
   // Not both. Don't know why. So, we have to leave the text rect as big as the bounds and move it
   // to a Y that works. In other words, no bottom inset will make a difference here for UITextFields
-  UIEdgeInsets textContainerInset = UIEdgeInsetsZero;
```
ajsecord, line 738: In the future, this kind of search-and-replace is a nice candidate for a single, trivially-reviewed, PR.

ajsecord (changes requested):

components/TextFields/src/MDCTextField.m
willlarche: Yeah this one. @romoore correctly pointed out that I was doing floating point comparison which is bad. So I need to convert these. I chose 1000 to be conservative. I believe the most decimal places It'll ever be is actually 1 which would be 10. Should I change to 10?

components/TextFields/src/MDCTextField.m
willlarche: Took another stab at clarity.

components/TextFields/src/MDCTextField.m
willlarche: Done.

components/TextFields/src/MDCTextField.m
willlarche: Done.

components/TextFields/src/MDCTextField.m
romoore: How about MDCCGFloatEqual ?

willlarche: Perfect! Done.

ajsecord (approved):

PR #1678, opened 2017-07-24
[TextFields] Tests for default priorities of constraints.

willlarche:
All the constraints created internally by the element need to have a rather low priority so they can be overridden by a controller if necessary.

### DIFF
--- a/components/TextFields/tests/unit/TextFieldTests.swift
+++ b/components/TextFields/tests/unit/TextFieldTests.swift
@@ -29,6 +29,8 @@ class TextFieldTests: XCTestCase {
     XCTAssertEqual(textField.attributedText?.string, string)
   }
 
+  // All the constraints created internally by the MDCTextField need to have a rather low priority 
+  // so they can be overridden by a controller from the outside.
   func testConstraintPriorities() {
     let textField = MDCTextField()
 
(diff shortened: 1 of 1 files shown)


components/TextFields/tests/unit/TextFieldTests.swift
```diff
@@ -29,6 +29,14 @@ class TextFieldTests: XCTestCase {
     XCTAssertEqual(textField.attributedText?.string, string)
   }
 
+  func testConstraintPriorities() {
+    let textField = MDCTextField()
+
+    for constraint in textField.constraints {
+      XCTAssertLessThanOrEqual(constraint.priority, UILayoutPriorityDefaultLow + 10, String(describing: constraint))
```
romoore, line 38: Maybe a more descriptive comment mentioning why they should all be so low?

romoore (approved):
1 comment but not worth blocking

components/TextFields/tests/unit/TextFieldTests.swift
willlarche: Done

romoore (approved):
Great, I meant in the XCTest call, but this works. 😸

willlarche:
@romoore , I did put it in the test:
```
 // All the constraints created internally by the MDCTextField need to have a rather low priority 
 // so they can be overridden by a controller from the outside.
 func testConstraintPriorities() {
```

ajsecord (approved):

PR #1681, opened 2017-07-24
[Buttons] Undo MDCFlatButton title color changes

samnm:
It seems defaulting MDCFlatButton to a white title color is a bad idea.


ajsecord (approved):
Shouldn't the default be whatever we've been shipping?

samnm:
Yes it should. Changing it to white was a bad idea.

PR #1687, opened 2017-07-25
[Ink] Don't assign compositeRipple frame twice

### LINKED ISSUE #1685: MDCInkLayer reassigns compositeRipple frame twice in layoutSublayers
### Overview
In MDCInkLayer `-layoutSublayers`, it assigns the `.frame` of `_compositeRipple` twice - the first is its own frame and then a computed frame from the InkLayer's bounds.  We can discard the first assignment, which is actually fairly expensive because it requires the view to compute the layer's frame.

### Reproduction steps
1. Generate a large number of MDCCollectionViewCells in a collection view controller
1. Run the Time Profiler
1. Check how much time is spent in `[MDCInkLayer layoutSublayers]`

### Version number
28.0.0

iPhone 5 (Simulator) iOS 10.2 (Simulator)

romoore:
`_compositeRipple.frame` was being assigned InkLayer's frame before
being assigned a second rectangle based on InkLayer's bounds. This can
cause the InkLayer compute its frame within its parent layer/view and is
inefficient.

Closes #1685

### DIFF
--- a/components/Ink/src/private/MDCInkLayer.m
+++ b/components/Ink/src/private/MDCInkLayer.m
@@ -484,7 +484,6 @@ static NSString *const kInkLayerBackgroundOpacityAnim = @"backgroundOpacityAnim"
 
 - (void)layoutSublayers {
   [super layoutSublayers];
-  _compositeRipple.frame = self.frame;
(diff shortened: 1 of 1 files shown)


jgunaratne (approved):

willlarche (approved):
Have @jgunaratne look at it too.

PR #1692, opened 2017-07-25
[Collections] Only change selected icon if needed

### LINKED ISSUE #1708: [Collections] MDCCollectionViewCell unnecessarily assigns "(un)selected" image
### Overview
When scrolling MDCCollectionViewCells, `prepareForReuse` will call `setSelected` which will load a UIImage for the (un)selected state and assign it. If the selection imageView is nil or if the state has not changed since the last time the cell was used, these are wasted CPU time.

### Reproduction steps
1. Open the MDCCatalog in Instruments (Core Animation)
1. Scroll any of the Collections example with a reasonable number of items (at least 2 screens worth)
1. Check the time used for loading icons

### Version number
29.0.0

iPhone 5c/iOS 10.3.3
iPad Mini 4/iOS 10.3.2

romoore:
The `-setSelected:` method reassigns the selection icon image every time
a cell is dequeued/reused. Since most cells are unselected most of the
time, this results in a lot of unnecessary UIImage calls.  Only updating
the selected/unselected image if the view exists (it won't until
initially changed to Editing Mode) and if the state has changed.
    
Closes #1708

### Original performance
screen shot 2017-07-27 at 9 48 25 am

### Performance with changes
**MDCIcon no longer appears in the trace
screen shot 2017-07-27 at 9 57 10 am

### DIFF
--- a/components/CollectionCells/src/MDCCollectionViewCell.m
+++ b/components/CollectionCells/src/MDCCollectionViewCell.m
@@ -424,14 +424,19 @@ NSString *const kDeselectedCellAccessibilityHintKey =
 #pragma mark - Selecting
 
 - (void)setSelected:(BOOL)selected {
+  BOOL previousSelectedState = self.selected;
   [super setSelected:selected];
   if (selected) {
-    _editingSelectorImageView.image = [MDCIcons imageFor_ic_check_circle];
-    _editingSelectorImageView.tintColor = self.editingSelectorColor;
+    if (_editingSelectorImageView && previousSelectedState != selected) {
+      _editingSelectorImageView.image = [MDCIcons imageFor_ic_check_circle];
+      _editingSelectorImageView.tintColor = self.editingSelectorColor;
+    }
     self.accessibilityTraits |= UIAccessibilityTraitSelected;
   } else {
-    _editingSelectorImageView.image = [MDCIcons imageFor_ic_radio_button_unchecked];
-    _editingSelectorImageView.tintColor = HEXCOLOR(kCellGrayColor);
+    if (_editingSelectorImageView && previousSelectedState != selected) {
+      _editingSelectorImageView.image = [MDCIcons imageFor_ic_radio_button_unchecked];
+      _editingSelectorImageView.tintColor = HEXCOLOR(kCellGrayColor);
+    }
     self.accessibilityTraits &= ~UIAccessibilityTraitSelected;
   }
   _editingSelectorImageView.image =
--- a/components/CollectionCells/src/MDCCollectionViewCell.m
+++ b/components/CollectionCells/src/MDCCollectionViewCell.m
@@ -430,17 +430,19 @@ NSString *const kDeselectedCellAccessibilityHintKey =
     if (_editingSelectorImageView && previousSelectedState != selected) {
       _editingSelectorImageView.image = [MDCIcons imageFor_ic_check_circle];
       _editingSelectorImageView.tintColor = self.editingSelectorColor;
+      _editingSelectorImageView.image = [_editingSelectorImageView.image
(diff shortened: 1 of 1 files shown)


components/CollectionCells/src/MDCCollectionViewCell.m
```diff
@@ -58,6 +68,10 @@ @interface MDCAccessoryTypeImageView : UIImageView
 @implementation MDCAccessoryTypeImageView
 @end
 
+@interface MDCCollectionViewCell ()
+@property (nonatomic, assign) MDCCollectionViewCellSelectionState selectionState;
```
ianegordon: I 100% agree with the motivation, but I'm concerned about adding state that mirrors existing state.

As an alternative, could we keep class / static references to the images in question?
static UIImage *s_checkCircleImage = [MDCIcons imageFor_ic_check_circle]; // Through dispatch_once
Then the flow would be:
if (selected) {
  if (_editingSelectorImageView.image != s_checkCircleImage) {
    _editingSelectorImageView.image = s_checkCircleImage;
    _editingSelectorImageView.tintColor = self.editingSelectorColor;
  } 
  self.accessibilityTraits |= UIAccessibilityTraitSelected;
}

This may not be a good idea, why don't we catch up in person tomorrow?

ianegordon (changes requested):

ajsecord (changes requested):
Please add ref to performance issue/bug report.

willlarche (approved):

ianegordon (approved):

components/CollectionCells/src/MDCCollectionViewCell.m
```diff
@@ -424,14 +424,19 @@ - (void)updateInterfaceForEditing {
...
   } else {
-    _editingSelectorImageView.image = [MDCIcons imageFor_ic_radio_button_unchecked];
-    _editingSelectorImageView.tintColor = HEXCOLOR(kCellGrayColor);
+    if (_editingSelectorImageView && previousSelectedState != selected) {
+      _editingSelectorImageView.image = [MDCIcons imageFor_ic_radio_button_unchecked];
+      _editingSelectorImageView.tintColor = HEXCOLOR(kCellGrayColor);
+    }
     self.accessibilityTraits &= ~UIAccessibilityTraitSelected;
   }
   _editingSelectorImageView.image =
```
ianegordon: Should this image generation be moved up into both of the if (we want to do stuff) {  } clauses?

ianegordon (changes requested):

components/CollectionCells/src/MDCCollectionViewCell.m
romoore: Like this?

```objectivec
if (_editingSelectorImageView && previousSelectedState != selected) {
  if (selected) {
    ... image stuff 
  } else {
    ... image stuff
  }
}
if (selected) {
  ...trait stuff
} else {
  ... trait stuff
}
_editingSelectorImageView.iamge = ...
```

ianegordon: We should only call  the imageWithRenderingMode: if the image has changed.  I would expect the if blocks to look similar to the following:

```
if (_editingSelectorImageView && previousSelectedState != selected) {
     _editingSelectorImageView.image = [MDCIcons imageFor_ic_radio_button_unchecked];
     _editingSelectorImageView.tintColor = HEXCOLOR(kCellGrayColor);
     _editingSelectorImageView.image = 
         [_editingSelectorImageView.image imageWithRenderingMode:UIImageRenderingModeAlwaysTemplate]
 }
```

ianegordon (approved):

ianegordon:
Please merge when Travis gives the OK.

PR #1702, opened 2017-07-26
[TextFields] Fundament implementation of trailingViewMode.

willlarche:


components/TextFields/src/private/MDCTextInputCommonFundament.m
```diff
@@ -159,6 +161,7 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder {
         [aDecoder decodeObjectForKey:MDCTextInputFundamentCharacterRelativeSuperviewKey];
     _textColor = [aDecoder decodeObjectForKey:MDCTextInputFundamentTextColorKey];
     _trailingUnderlineLabel = [aDecoder decodeObjectForKey:MDCTextInputFundamentTrailingLabelKey];
+    _trailingViewMode = (UITextFieldViewMode)[aDecoder decodeIntegerForKey:MDCTextInputFundamentTrailingViewKey];
```
ianegordon, line 164: Do we need to check for the existence of these keys before we decode them?
```
if ([aDecoder containsValueForKey:MDCTrailingViewKey]) {
  _trailingViewMode = (UITextFieldViewMode)[aDecoder decoderIntergerForKey:MDCTrailingViewKey];
}
```
And the other ivars? (Happy to have that in another PR if we do need to do so)

ianegordon (changes requested):

components/TextFields/src/private/MDCTextInputCommonFundament.m
willlarche: Spoke with you offline. I do indeed need these checks. Let's do that in another PR for there are more than just this one.

ianegordon:
SGTM.
#1709 opened.

ianegordon (approved):

PR #1706, opened 2017-07-26
[Buttons] Fix storyboard backgroundColor assignment

### LINKED ISSUE #1686: [Buttons] Interface Builder background color is not longer working as expected
From an internal client:

We use Interface Builder for a few screens, and it looks like MDCButtons on those screens don't have the correct background color set anymore when loading the .xib files. I suspect it was a recent MDCButton change in the last few weeks, but I had a look through CS and couldn't figure out the suspect CL.

samnm:
[Before](https://user-images.githubusercontent.com/1251373/28644517-e2fda6f0-7227-11e7-8752-63fc60072abb.png)/[After](https://user-images.githubusercontent.com/1251373/28644523-e95d55f4-7227-11e7-8b0c-83618f6e410b.png)

Closes #1686.

### DIFF
--- a/components/Buttons/src/MDCButton.m
+++ b/components/Buttons/src/MDCButton.m
@@ -121,6 +121,7 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
   self = [super initWithFrame:frame];
   if (self) {
     [self commonMDCButtonInit];
+    [self updateBackgroundColor];
   }
   return self;
 }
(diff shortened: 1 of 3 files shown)


ianegordon:
I've reached out to an internal team for verification of the fix.

ianegordon:
Awaiting response from an internal team.

components/Buttons/src/MDCButton.m
```diff
@@ -207,18 +213,20 @@ - (void)commonMDCButtonInit {
...
   _uppercaseTitle = YES;
   _userElevations = [NSMutableDictionary dictionary];
-  _backgroundColors = [NSMutableDictionary dictionary];
   _accessibilityLabelForState = [NSMutableDictionary dictionary];
 
-  _backgroundColors[@(UIControlStateNormal)] = MDCColorFromRGB(MDCButtonDefaultBackgroundColor);
+  if (!_backgroundColors) {
+    // _backgroundColors may have already been initialized by setting the backgroundColor setter.
+    _backgroundColors = [NSMutableDictionary dictionary];
+    _backgroundColors[@(UIControlStateNormal)] = MDCColorFromRGB(MDCButtonDefaultBackgroundColor);
```
ianegordon, line 221: Should this be:
```
if (!_backgroundColor) {
  _backgroundColors = [NSMutableDictionary dictionary];
}
_backgroundColors[@(UIConrolStateNormal)] = MDCColorFromRGB(MDCColor);
```

Do we not need to [self updateBackgroundColor] as we are below?

ianegordon (changes requested):

components/Buttons/src/MDCButton.m
samnm: [self updateBackgroundColor] is called at the end of both initializers.

We don't want to always write to the Normal backgroundColor because we might already have a value there, so we only write to it if the dictionary is uninitialized.

ianegordon: SGTM

ianegordon (approved):

Issue #1707, opened 2017-07-27
[Collections] Computing inlays in MDCCollectionViewFlowLayout is n² operation


romoore:
### Overview
For every element in the collectionView,  the flow layout iterates over inlaid element (which is likely every single one in Editing Mode) twice.  For example, in the Catalog / Collections / Cell Editing Example, if you change the number of sections to 2 and the number of items to 10, then there are 39 calls to `-inlayAttributeIfNecessary`. Each call will iterate over all 20 items to check their cell type, and possibly section, and indexPath.

If you have an "infinite" scrolling collection view with editing enabled (photo/email picker?) then this can quickly degrade performance.  

**Note:**
* `-[MDCCollectionViewController collectionViewWillBeginEditing:]` tells the styler to apply the inlay to all items.
* As long as the styler allows multiple inlays, it will add all IndexPaths from the CV into its `_inlaidPathSet`
* `-[MDCCollectionViewFlowLayout inlayAttributeIfNecessary]` retrieves all of these IndexPaths and iterates them to find those that it must adjust the insets
* This code path is called by both `-collectionView:flowLayout:sizeForItemAtIndexPath` and `layoutAttributesForElementsInRect`, so setting the initial "Editing" state can be incredibly slow for large numbers of available indexPaths.

### Reproduction steps
1. Edit the catalog "CollectionsEditingExample.m" file constants to increase the number of sections and/or items
1. Run the catalog and open that example
1. Scroll without Editing
1. Turn on Editing mode and scroll

### Version number
29.0.0

### Operating system and device
iPhone 5c/iOS 10.3.3 (loses ~15fps)
iPad Mini 4/iOS 10.3.2 (FPS ok, CPU notably higher)

### iPhone 5c/iOS 10.3.3 Performance##
**Instruments data "groupings"**

1. App launch
1. Scrolling without Editing
2. Open "Editing" mode
3. Scrolling with editing (highlighted)
![screen shot 2017-07-27 at 9 29 30 am](https://user-images.githubusercontent.com/1753199/28672598-2aff7858-72ae-11e7-9a7c-b210e169ec9b.png)

ianegordon:
We are not intending on any significant updates to the current Collections component. 😞
We are working on a much simpler [List component](https://github.com/material-components/material-components-ios/tree/develop/components/List). 😄

PR #1710, opened 2017-07-27
[Collections] Add "editing" performance example

romoore:
New example to demonstrate scrolling performance when a collection
contains a large number of cells. "Editing" mode is currently very
inefficient.

Related to: #1707


ianegordon (changes requested):
Travis is upset about a name collision on kReusableIdentifierItem.
That name might be a little generic to declare in a header.
Move to source?
Add detail / kCollectionsSectionsInsetsIdentifier ?

ianegordon (approved):

PR #1711, opened 2017-07-27
[Collections] - Fix for possible infinite recursion

ShepJGoogle:
In the case that a UICollectionViewController's UICollectionViewDelegate methods fire before the view had been loaded, there is an infinite recursion that begins. As we call self.collectionView in our delegate methods, loadView is continuously called.

- Changes delegate method chain to use the passed in value for the collectionView rather than self.collectionView

### DIFF
--- a/components/Collections/src/MDCCollectionViewController.m
+++ b/components/Collections/src/MDCCollectionViewController.m
@@ -176,15 +176,15 @@ NSString *const MDCCollectionInfoBarKindFooter = @"MDCCollectionInfoBarKindFoote
     sizeForItemAtIndexPath:(NSIndexPath *)indexPath {
   UICollectionViewLayoutAttributes *attr =
       [collectionViewLayout layoutAttributesForItemAtIndexPath:indexPath];
-  CGSize size = [self sizeWithAttribute:attr];
-  size = [self inlaidSizeAtIndexPath:indexPath withSize:size];
+  CGSize size = [self sizeWithAttribute:attr collectionView:collectionView];
+  size = [self inlaidSizeAtIndexPath:indexPath withSize:size collectionView:collectionView];
   return size;
 }
 
 - (UIEdgeInsets)collectionView:(UICollectionView *)collectionView
                         layout:(UICollectionViewLayout *)collectionViewLayout
         insetForSectionAtIndex:(NSInteger)section {
-  return [self insetsAtSectionIndex:section];
+  return [self insetsAtSectionIndex:section collectionView:collectionView];
 }
 
 - (CGFloat)collectionView:(UICollectionView *)collectionView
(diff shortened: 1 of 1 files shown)


components/Collections/src/MDCCollectionViewController.m
```diff
@@ -199,20 +199,40 @@ - (CGFloat)collectionView:(UICollectionView *)collectionView
...
+                                 collectionView:collectionView];
   return CGSizeMake(width, height);
 }
 
+- (CGFloat)cellWidthAtSectionIndex:(NSInteger)section
+                    collectionView:(UICollectionView *)collectionView {
+  CGFloat bounds = CGRectGetWidth(
+      UIEdgeInsetsInsetRect(collectionView.bounds, collectionView.contentInset));
+  UIEdgeInsets sectionInsets = [self insetsAtSectionIndex:section collectionView:collectionView];
+  /*[self collectionView:self.collectionView
```
romoore: remove commented-out code?

components/Collections/src/MDCCollectionViewController.m
```diff
@@ -199,20 +199,40 @@ - (CGFloat)collectionView:(UICollectionView *)collectionView
...
+
+  CGFloat insets = sectionInsets.left + sectionInsets.right;
+  if (_styler.cellLayoutType == MDCCollectionViewCellLayoutTypeGrid) {
+    CGFloat cellWidth = bounds - insets - (_styler.gridPadding * (_styler.gridColumnCount - 1));
+    return cellWidth / _styler.gridColumnCount;
+  }
+  return bounds - insets;
+}
+
 - (CGFloat)cellWidthAtSectionIndex:(NSInteger)section {
```
romoore: Please comment to explain why this method cannot call the one above.

components/Collections/src/MDCCollectionViewController.m
```diff
@@ -248,9 +269,10 @@ - (UIEdgeInsets)insetsAtSectionIndex:(NSInteger)section {
...
 
-- (CGSize)inlaidSizeAtIndexPath:(NSIndexPath *)indexPath withSize:(CGSize)size {
+- (CGSize)inlaidSizeAtIndexPath:(NSIndexPath *)indexPath
+                       withSize:(CGSize)size
+                 collectionView:(UICollectionView *)collectionView {
   // If object is inlaid, return its adjusted size.
-  UICollectionView *collectionView = self.collectionView;
   if ([_styler isItemInlaidAtIndexPath:indexPath]) {
     CGFloat inset = MDCCollectionViewCellStyleCardSectionInset;
     UIEdgeInsets inlayInsets = UIEdgeInsetsZero;
```
romoore, line 259: Should `_collectionViewLayout` below be the one attached to `collectionView` or the one this controller holds?

ShepJGoogle: Done

components/Collections/src/MDCCollectionViewController.m
```diff
@@ -199,23 +199,24 @@ - (CGFloat)collectionView:(UICollectionView *)collectionView
   return 0;
 }
 
-- (CGSize)sizeWithAttribute:(UICollectionViewLayoutAttributes *)attr {
+- (CGSize)sizeWithAttribute:(UICollectionViewLayoutAttributes *)attr
+             collectionView:(UICollectionView *)collectionView {
   CGFloat height = MDCCellDefaultOneLineHeight;
   if ([_styler.delegate respondsToSelector:@selector(collectionView:cellHeightAtIndexPath:)]) {
     height =
         [_styler.delegate collectionView:self.collectionView cellHeightAtIndexPath:attr.indexPath];
```
romoore: `[_styler.delegate collectionView:self.collectionView...` should probably switch to the parameter rather than the property as well.

romoore (changes requested):

components/Collections/src/MDCCollectionViewController.m
```diff
@@ -298,6 +301,29 @@ - (CGSize)inlaidSizeAtIndexPath:(NSIndexPath *)indexPath withSize:(CGSize)size {
...
+- (CGFloat)cellWidthAtSectionIndex:(NSInteger)section {
+  CGFloat bounds = CGRectGetWidth(
+                                  UIEdgeInsetsInsetRect(self.collectionView.bounds,
+                                                        self.collectionView.contentInset));
+  UIEdgeInsets sectionInsets = [self collectionView:self.collectionView
+                                             layout:self.collectionView.collectionViewLayout
+                             insetForSectionAtIndex:section];
+
+  CGFloat insets = sectionInsets.left + sectionInsets.right;
+  if (_styler.cellLayoutType == MDCCollectionViewCellLayoutTypeGrid) {
```
ianegordon: Do we need to check that _styler is != nil?

I would add a check that _styler.gridColumnCount > 0 to this if statement unless we are 100% confident it will never be < 1.

ianegordon (changes requested):

romoore (approved):
LGTM

components/Collections/src/MDCCollectionViewController.m
romoore: MDCCollectionViewCellLayoutTypeList is the equivalent of `nil`, so it will skip this conditional and return the default.  

I would suggest that anyone setting a Grid style with no columns in the grid wouldn't be surprised that the cells have an invalid width. I would support a check (and maybe an NSAssert to warn devs) if the styler is in an inconsistent state, but I think the div-by-zero should be pretty obvious as well.

ianegordon (approved):

PR #1719, opened 2017-07-28
[FeatureHighlight] Fix voiceover regression

samnm:
See internal issue b/64137830


ianegordon:
The navigate back gesture works as expected in our catalog.  Based on the bug report, I'm not 100% sure they did not try the back gesture.  Waiting on confirmation.

samnm:
Closing since the back gesture fulfills the purpose much better than the change in this PR does.

PR #1728, opened 2017-08-03
[ButtonBar]  Alignment

Eccelor:
[ButtonBar]
Allowed Automatic Calculation of Width Based on the ButtonBar Frame to create equal width buttons
ButtonBar was ignoring the tintColor of UIBarButtonItem to set as inkColor

[PageControl]
Send UIControlValueChanged Event from setCurrentPage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 29.0.0
+
+* Reverts changes to the title color of MDCFlatButton and MDCRaisedButton.
+
 # 28.0.0
 
 ## API Diffs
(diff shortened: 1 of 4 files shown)


ianegordon:
Hi @Eccelor, thanks for signing the CLA.  I'm looking at this now.

A couple things jumped out:
1. We typically create our PRs off of develop.  We reserve stable for releases.  If you rebase off develop, you will likely see the conflicting CHANGELOG.md commit in the PR go away.
2. We may want to break this into more than one PR since we are changing multiple components, one for the ButtonBar and one for the PageControl.

In any case, thanks for the PR!  I'm digging into it now.

-Ian

Eccelor:
@ianegordon Really appreciate the pointers! Will keep those in mind when submitting new PRs. Let me know if there are any changes required from this side.

ianegordon:
Hi @Eccelor, thank you for your patience.

We decided against landing #1746 to match Apple's functionality, but thank you for looking into it.

Does #1797 fix your alignment issues?  It looks like romoore was also looking into button bar.

Finally, per the spec, we don't currently want to change the buttonbar tint color as part of the tintColor change, but you do raise an interesting point.  I've opened #1837 to track the issue.

Huge thanks for the PR!  Please do reach out in the future if you have more ideas, PRs or questions.
(And feel free to re-open if we haven't addressed your concerns)

-Ian

PR #1737, opened 2017-08-04
[TextFields] Multiline text fields element

willlarche:
This is the element and support for it in the fundament.


ajsecord (approved):

willlarche:
Submitting this feature in pieces. Monitoring Travis to see if it stays happy or if something I've withheld causes a break because of dependency.

PR #1739, opened 2017-08-04
[Collections] Allow selection of cells with hidden ink

### LINKED ISSUE #1446: [Collections] Hiding ink prevent interaction
### Overview
Having collectionView:hidesInkViewAtIndexPath: returning YES prevents interaction with the cells at the given indexPath. If the function returns NO, the interactions are fine.

### Reproduction steps
 - Implements collectionView:hidesInkViewAtIndexPath:, returning YES for some cells, NO for some others
 - collectionView:didSelectItemAtIndexPath: is never called for the cell which have the ink hidden

### Version number
23.2

romoore:
MDCCollectionViewController does not receive messages for
`-collectionView:didSelectItemAtIndexPath:` if it responds `NO` to
`-collectionView:shouldHighlightCellAtIndexPath:`. It seems by allowing
highlight, but not performing a highlight animation on the cell, the delegate
will receive *didSelect* messages for cells with hidden ink.

Closes #1446

### DIFF
--- a/components/Collections/src/MDCCollectionViewController.h
+++ b/components/Collections/src/MDCCollectionViewController.h
@@ -49,9 +49,6 @@
  as well as ink during the highlight/unhighlight states.
  */
 
-- (BOOL)collectionView:(nonnull UICollectionView *)collectionView
-    shouldHighlightItemAtIndexPath:(nonnull NSIndexPath *)indexPath NS_REQUIRES_SUPER;
-
 - (void)collectionView:(nonnull UICollectionView *)collectionView
     didHighlightItemAtIndexPath:(nonnull NSIndexPath *)indexPath NS_REQUIRES_SUPER;
 
--- a/components/Collections/src/MDCCollectionViewController.h
+++ b/components/Collections/src/MDCCollectionViewController.h
@@ -49,6 +49,9 @@
  as well as ink during the highlight/unhighlight states.
  */
 
+- (BOOL)collectionView:(nonnull UICollectionView *)collectionView
+    shouldHighlightItemAtIndexPath:(nonnull NSIndexPath *)indexPath NS_REQUIRES_SUPER;
+
 - (void)collectionView:(nonnull UICollectionView *)collectionView
     didHighlightItemAtIndexPath:(nonnull NSIndexPath *)indexPath NS_REQUIRES_SUPER;
 
--- a/components/Collections/src/MDCCollectionViewController.m
+++ b/components/Collections/src/MDCCollectionViewController.m
@@ -398,6 +398,11 @@ NSString *const MDCCollectionInfoBarKindFooter = @"MDCCollectionInfoBarKindFoote
 
 #pragma mark - <UICollectionViewDelegate>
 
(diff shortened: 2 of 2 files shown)


romoore:
Need to fix the header.

gambard:
Thanks for doing this!
Overall it looks good, however, I am not sure about calling startTouchEndedAnimationAtPoint:completion: on the inkView if you did not start the ink animation. From my understanding of the code, it should be fine.

romoore:
@gambard  Awesome, thanks for the review!  If ink is disabled for that indexPath, then the method exits early and the ink animation should not behavior.  There should not be a change in behavior for inked cells, only that those without ink will have selection actions passed to the delegate.

ianegordon:
Hi @romoore.  I believe you were investigating this a bit.  Are you ready for me to review / merge?

romoore:
Yep, whenever you have time. 👍

components/Collections/src/MDCCollectionViewController.h
```diff
@@ -49,9 +49,6 @@
  as well as ink during the highlight/unhighlight states.
  */
 
-- (BOOL)collectionView:(nonnull UICollectionView *)collectionView
```
ianegordon: This is in the public API.  If we want to remove it, we should go through the deprecation process.  I could see people subclassing MDCCollectionViewController just to overwrite methods like this.

components/Collections/src/MDCCollectionViewController.m
```diff
@@ -399,16 +399,12 @@ - (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView
 
 #pragma mark - <UICollectionViewDelegate>
 
-- (BOOL)collectionView:(UICollectionView *)collectionView
-    shouldHighlightItemAtIndexPath:(NSIndexPath *)indexPath {
-  if ([_styler.delegate respondsToSelector:@selector(collectionView:hidesInkViewAtIndexPath:)]) {
-    return ![_styler.delegate collectionView:collectionView hidesInkViewAtIndexPath:indexPath];
-  }
```
ianegordon: Can we ignore the styler and just return YES to fix the functionality?

ianegordon (changes requested):

components/Collections/src/MDCCollectionViewController.m
romoore: Are you suggesting keeping the shouldHighlight... and returning YES each time, but then in didHighlight keeping the proposed changes?

components/Collections/src/MDCCollectionViewController.h
romoore: The [default implementation is to return YES](https://developer.apple.com/documentation/uikit/uicollectionviewdelegate/1618070-collectionview?language=objc). Do you see an advantage in us explicitly returning YES as well?

ianegordon:
Discussed offline.
We will keep shouldHighlight: but the implementation will just return YES.  The new handling will allow Ink and Selection to co-exist.

ianegordon (approved):

ianegordon:
LGTM.  Please merge when Travis is happy.

PR #1746, opened 2017-08-07
[PageControl] Send UIControlEventChanges Event on setCurrentPage

ianegordon:
Call sendActions:UIControlerEventValueChanged when setCurrentPage is called.

Originally part of #1728 by @Eccelor.

### DIFF
--- a/components/PageControl/src/MDCPageControl.m
+++ b/components/PageControl/src/MDCPageControl.m
@@ -175,6 +175,7 @@ static inline CGFloat normalizeValue(CGFloat value, CGFloat minRange, CGFloat ma
                                                    // more to ensure
                                                    // no hidden indicators remain.
                                                    [self revealIndicatorsReversed:shouldReverse];
+                                                   [self sendActionsForControlEvents:UIControlEventValueChanged];
                                                  }];
                        [self revealIndicatorsReversed:shouldReverse];
                      });
@@ -193,6 +194,7 @@ static inline CGFloat normalizeValue(CGFloat value, CGFloat minRange, CGFloat ma
     [CATransaction setDisableActions:YES];
     [_indicators[previousPage] setHidden:NO];
     [CATransaction commit];
+    [self sendActionsForControlEvents:UIControlEventValueChanged];
   }
 }
 


ianegordon:
googlebot seems confused by this PR.  Possibly because it was generated from a cherry-pick from another PR.  @Eccelor has signed the cla.

components/PageControl/src/MDCPageControl.m
```diff
@@ -193,6 +194,7 @@ - (void)setCurrentPage:(NSInteger)currentPage
     [CATransaction setDisableActions:YES];
     [_indicators[previousPage] setHidden:NO];
     [CATransaction commit];
+    [self sendActionsForControlEvents:UIControlEventValueChanged];
```
romoore, line 197: I believe the CATransaction is asynchronously dispatched into the main run loop. I think we should place the `sendActionsForControlEvents:` into the completion block of the transaction to ensure that it's only invoked once the changes are actually made.

```objectivec
    dispatch_async(dispatch_get_main_queue(), ^{
      NSLog(@"Dispatched before");
    });
    [CATransaction begin];
    [CATransaction setDisableActions:YES];
    [_indicators[previousPage] setHidden:NO];
      NSLog(@"Completion block");
    }];
    [CATransaction commit];
    dispatch_async(dispatch_get_main_queue(), ^{
      NSLog(@"Dispatched after");
    });

    NSLog(@"Subsequent code");
```

```
2017-08-07 16:29:03.806 MDCCatalog[61661:7937194] Subsequent code
2017-08-07 16:29:03.807 MDCCatalog[61661:7937194] Dispatched before
2017-08-07 16:29:03.807 MDCCatalog[61661:7937194] Completion block
2017-08-07 16:29:03.807 MDCCatalog[61661:7937194] Dispatched after
```

romoore (changes requested):

components/PageControl/src/MDCPageControl.m
ajsecord: Hrm. Does UIButton dispatch when the user lifts their finger or after the animation completes?

ianegordon:
Hi @Eccelor.  After looking into it, we likely do not want to send an action on programmatic changes.  

From Apple's UIControl documentation:
"Action methods are called when the user interacts with the control in specific ways."

The specific bit is the "user interacts".  It specifically does not mention programmatic changes.

You can see this for yourself if you create a UIPageControl and wire up the valueChanged action.

I'm going to close this, but feel free to re-open if you have questions.  And even though we aren't planning on landing the PR, many thanks for sending it our direction!

👍

PR #1743, opened 2017-08-04
[TextFields] Multiline examples.

willlarche:
Dependent on #1740 and #1742

PR #1747, opened 2017-08-07
[Text fields] Trailing view support

willlarche:
Dependent on #1743

### DIFF
--- a/components/TextFields/src/MDCMultilineTextField.m
+++ b/components/TextFields/src/MDCMultilineTextField.m
@@ -46,6 +46,9 @@ static NSString *const MDCMultilineTextFieldTrailingViewModeKey =
 @property(nonatomic, strong) NSLayoutConstraint *textViewMinHeight;
 @property(nonatomic, strong) NSLayoutConstraint *textViewTop;
 @property(nonatomic, strong) NSLayoutConstraint *textViewTrailing;
+@property(nonatomic, strong) NSLayoutConstraint *textViewTrailingTrailingViewLeading;
+@property(nonatomic, strong) NSLayoutConstraint *trailingViewCenterY;
+@property(nonatomic, strong) NSLayoutConstraint *trailingViewTrailing;
 
 @end
 
@@ -318,6 +321,8 @@ static NSString *const MDCMultilineTextFieldTrailingViewModeKey =
     [self sendSubviewToBack:self.textView];
   }
 
+  [self updateTrailingViewAlpha];
+
   [self.fundament layoutSubviewsOfInput];
 }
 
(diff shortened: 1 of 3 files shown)


components/TextFields/src/MDCMultilineTextField.m
```diff
@@ -46,6 +46,9 @@ @interface MDCMultilineTextField () {
 @property(nonatomic, strong) NSLayoutConstraint *textViewMinHeight;
 @property(nonatomic, strong) NSLayoutConstraint *textViewTop;
 @property(nonatomic, strong) NSLayoutConstraint *textViewTrailing;
+@property(nonatomic, strong) NSLayoutConstraint *textViewTrailingTrailingViewLeading;
```
ianegordon: TrailingTrailingViewLeading?
Would you add a comment to explain.

components/TextFields/src/MDCMultilineTextField.m
```diff
@@ -387,6 +394,72 @@ - (CGFloat)estimatedTextViewLineHeight {
...
 }
 
+#pragma mark - Trailing View Implementation
+
+- (void)updateTrailingViewLayout {
+  if (!self.trailingView) {
+    return;
+  }
+
+  self.trailingView.translatesAutoresizingMaskIntoConstraints = NO;
```
ianegordon, line 407: I would only expect to set this once, when trailingView is created or added.
There may be a deeper reason I'm not seeing.  If so, perhaps a comment.

components/TextFields/src/MDCMultilineTextField.m
```diff
@@ -387,6 +394,72 @@ - (CGFloat)estimatedTextViewLineHeight {
...
+
+  if (![self.trailingView isDescendantOfView:self]) {
+    [self addSubview:self.trailingView];
+  }
+
+  if (!self.trailingViewTrailing) {
```
ianegordon, line 413: Once created, will self.trailingViewTrailing ever be nil again?
dispatch_once ?
(dispatch_once may be less clear)
Perhaps a comment?
// Create trailing view trailing constraint if it doesn't exist yet.
If this is only created once, will it break if we change self.trailingView?

ianegordon (changes requested):

ianegordon (approved):

ianegordon:
Discussed offline: The constraints will get destroyed if the view is changed.  This is why they are recreated in-line.

components/TextFields/src/MDCMultilineTextField.m
willlarche: Done!

Issue #1735, opened 2017-08-04
[FeatureHighlight] Tap to dismiss is no longer working


FloresRobles:
### Overview
Tapping outside the circles to dismiss isn't working.

### Reproduction steps
 - Trigger the component in the catalog, and tap outside the circle

### Version number
 - Found in 30.0.0

### Operating system and device
iOS 10 and iOS 11, all iPhones

ajsecord:
Sam, this is closed, right?

samnm:
Yes it is.

PR #1749, opened 2017-08-07
[FeatureHighlight] Fix gesture recognizers

samnm:
The most recent feature highlight fix broke part of the tap-to-dismiss #1735. This PR fixes it correctly.

Core change:
* Allow both gesture recognizers to act simultaneously.

### DIFF
--- a/components/FeatureHighlight/src/private/MDCFeatureHighlightDismissGestureRecognizer.m
+++ b/components/FeatureHighlight/src/private/MDCFeatureHighlightDismissGestureRecognizer.m
@@ -86,7 +86,11 @@
   [super touchesEnded:touches withEvent:event];
 
   if (_hasTouch) {
-    self.state = UIGestureRecognizerStateEnded;
+    if (MDCCGFloatEqual(_progress, 1.0)) {
+      self.state = UIGestureRecognizerStateCancelled;
+    } else {
+      self.state = UIGestureRecognizerStateEnded;
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


components/FeatureHighlight/src/private/MDCFeatureHighlightDismissGestureRecognizer.m
```diff
@@ -86,7 +86,11 @@ - (void)touchesEnded:(NSSet<UITouch *> *)touches withEvent:(UIEvent *)event {
   [super touchesEnded:touches withEvent:event];
 
   if (_hasTouch) {
-    self.state = UIGestureRecognizerStateEnded;
+    if (MDCCGFloatEqual(_progress, 1.0)) {
```
ianegordon: Could you add a comment explaining when _progress == 1.0, and why we want to cancel in that case?

ianegordon (approved):
Please add comment before merge.

ianegordon:
Voice Over back gesture worked as expected on an iPhone 6 @ 10.x.

components/FeatureHighlight/src/private/MDCFeatureHighlightDismissGestureRecognizer.m
samnm: I removed that change... I did a bit more research on what the cancelled state means and I think my initial assumption was incorrect.

PR #1755, opened 2017-08-09
[Buttons] Creating animations for FAB appearance

romoore:
The Floating Action Button (FAB) is supposed to "pop-in" to a view when the
view appears. Adding some basic animation code to allow the FAB to be
animated-in and -out of views.

fab-animate

### DIFF
--- a/components/Buttons/examples/ButtonsTypicalUse.m
+++ b/components/Buttons/examples/ButtonsTypicalUse.m
@@ -20,17 +20,15 @@
 #import "MaterialTypography.h"
 
 @interface ButtonsTypicalUseViewController ()
-
+@property (nonatomic, strong) MDCFloatingButton *floatingButton;
 @end
 
 @implementation ButtonsTypicalUseViewController
 
 - (MDCButton *)buildCustomStrokedButton {
   MDCButton *button = [[MDCButton alloc] init];
-  [button setBackgroundColor:[UIColor clearColor]
-                    forState:UIControlStateNormal];
-  [button setTitleColor:[UIColor colorWithWhite:0.1 alpha:1.0]
-               forState:UIControlStateNormal];
+  [button setBackgroundColor:[UIColor clearColor] forState:UIControlStateNormal];
+  [button setTitleColor:[UIColor colorWithWhite:0.1 alpha:1.0] forState:UIControlStateNormal];
   button.inkColor = [UIColor colorWithWhite:0 alpha:0.06];
   button.layer.borderWidth = 1;
   button.layer.borderColor = [UIColor blackColor].CGColor;
@@ -119,26 +117,21 @@
 
   // Floating action button
 
-  MDCFloatingButton *floatingButton = [[MDCFloatingButton alloc] init];
-  [floatingButton setTitleColor:titleColor forState:UIControlStateNormal];
-  [floatingButton sizeToFit];
-  [floatingButton addTarget:self
+  self.floatingButton = [[MDCFloatingButton alloc] init];
+  [self.floatingButton setTitleColor:titleColor forState:UIControlStateNormal];
+  [self.floatingButton sizeToFit];
+  [self.floatingButton addTarget:self
                      action:@selector(didTap:)
            forControlEvents:UIControlEventTouchUpInside];
-  floatingButton.translatesAutoresizingMaskIntoConstraints = NO;
+  self.floatingButton.translatesAutoresizingMaskIntoConstraints = NO;
 
   UIImage *plusImage = [UIImage imageNamed:@"Plus"];
-  [floatingButton setImage:plusImage forState:UIControlStateNormal];
-  [self.view addSubview:floatingButton];
+  [self.floatingButton setImage:plusImage forState:UIControlStateNormal];
+  [self.view addSubview:self.floatingButton];
 
   self.buttons = @[
-    raisedButton,
-    disabledRaisedButton,
-    flatButton,
-    disabledFlatButton,
-    strokedButton,
-    disabledStrokedButton,
-    floatingButton
+    raisedButton, disabledRaisedButton, flatButton, disabledFlatButton, strokedButton,
+    disabledStrokedButton, self.floatingButton
   ];
 
   [self setupExampleViews];
@@ -146,6 +139,30 @@
 
 - (void)didTap:(id)sender {
   NSLog(@"%@ was tapped.", NSStringFromClass([sender class]));
+  if (sender == self.floatingButton) {
+    [self.floatingButton shrink:YES
+                               completion:^{
+                                 dispatch_after(
+                                     dispatch_time(DISPATCH_TIME_NOW, (int64_t)(1 * NSEC_PER_SEC)),
+                                     dispatch_get_main_queue(), ^{
+                                       [self.floatingButton grow:YES completion:nil];
+                                     });
+                               }];
+  }
+}
+
+- (void)viewWillAppear:(BOOL)animated {
+  [super viewWillAppear:animated];
+  if (animated) {
+    [self.floatingButton shrink:NO completion:nil];
+  }
+}
+
+- (void)viewDidAppear:(BOOL)animated {
+  [super viewDidAppear:animated];
+  if (animated) {
+    [self.floatingButton grow:YES completion:nil];
+  }
 }
 
 @end
--- a/components/Buttons/examples/ButtonsTypicalUse.m
+++ b/components/Buttons/examples/ButtonsTypicalUse.m
@@ -20,7 +20,7 @@
 #import "MaterialTypography.h"
 
 @interface ButtonsTypicalUseViewController ()
-@property (nonatomic, strong) MDCFloatingButton *floatingButton;
+@property(nonatomic, strong) MDCFloatingButton *floatingButton;
 @end
 
 @implementation ButtonsTypicalUseViewController
@@ -28,18 +28,18 @@
 - (MDCButton *)buildCustomStrokedButton {
   MDCButton *button = [[MDCButton alloc] init];
   [button setBackgroundColor:[UIColor clearColor] forState:UIControlStateNormal];
-  [button setTitleColor:[UIColor colorWithWhite:0.1 alpha:1.0] forState:UIControlStateNormal];
-  button.inkColor = [UIColor colorWithWhite:0 alpha:0.06];
+  [button setTitleColor:[UIColor colorWithWhite:0.1F alpha:1] forState:UIControlStateNormal];
+  button.inkColor = [UIColor colorWithWhite:0 alpha:0.06F];
   button.layer.borderWidth = 1;
   button.layer.borderColor = [UIColor blackColor].CGColor;
-  button.disabledAlpha = 0.38;
+  button.disabledAlpha = 0.38F;
   return button;
 }
 
 - (void)viewDidLoad {
   [super viewDidLoad];
 
-  self.view.backgroundColor = [UIColor colorWithWhite:0.9 alpha:1.0];
+  self.view.backgroundColor = [UIColor colorWithWhite:0.9F alpha:1];
   UIColor *titleColor = [UIColor whiteColor];
 
   // Raised button
@@ -121,8 +121,8 @@
   [self.floatingButton setTitleColor:titleColor forState:UIControlStateNormal];
   [self.floatingButton sizeToFit];
   [self.floatingButton addTarget:self
-                     action:@selector(didTap:)
-           forControlEvents:UIControlEventTouchUpInside];
+                          action:@selector(didTap:)
+                forControlEvents:UIControlEventTouchUpInside];
   self.floatingButton.translatesAutoresizingMaskIntoConstraints = NO;
 
   UIImage *plusImage = [UIImage imageNamed:@"Plus"];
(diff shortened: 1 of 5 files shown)


components/Buttons/src/MDCFloatingButton+Animation.m
```diff
@@ -0,0 +1,92 @@
...
+@implementation MDCFloatingButton (Animation)
+
+- (void)grow:(BOOL)animated completion:(void (^_Nullable)(void))completion {
+  void (^growActions)() = ^{
+    self.transform = CGAffineTransformIdentity;
+  };
+  if (animated) {
+    [UIView animateWithDuration:kMDCFloatingButtonAnimationDuration
+                     animations:^{
+                       growActions();
```
willlarche: Since the block is a void block, I'd just pass it to animations: instead of invoking it inside.

components/Buttons/src/MDCFloatingButton+Animation.m
```diff
@@ -0,0 +1,92 @@
...
+
+@implementation MDCFloatingButton (Animation)
+
+- (void)grow:(BOOL)animated completion:(void (^_Nullable)(void))completion {
+  void (^growActions)() = ^{
+    self.transform = CGAffineTransformIdentity;
+  };
+  if (animated) {
+    [UIView animateWithDuration:kMDCFloatingButtonAnimationDuration
+                     animations:^{
```
willlarche: This probably needs a curve / motion design. Did it get vetted by a motion designer?

components/Buttons/src/MDCFloatingButton+Animation.h
```diff
@@ -0,0 +1,37 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import "MDCFloatingButton.h"
+
+@interface MDCFloatingButton (Animation)
+
+/**
+ Causes this button to return to its "normal" size.
```
willlarche: I would love to have the actual diameter in this comment. It's GTK.

components/Buttons/examples/ButtonsTypicalUse.m
```diff
@@ -20,17 +20,15 @@
...
 @implementation ButtonsTypicalUseViewController
 
 - (MDCButton *)buildCustomStrokedButton {
   MDCButton *button = [[MDCButton alloc] init];
-  [button setBackgroundColor:[UIColor clearColor]
-                    forState:UIControlStateNormal];
-  [button setTitleColor:[UIColor colorWithWhite:0.1 alpha:1.0]
-               forState:UIControlStateNormal];
+  [button setBackgroundColor:[UIColor clearColor] forState:UIControlStateNormal];
+  [button setTitleColor:[UIColor colorWithWhite:0.1 alpha:1.0] forState:UIControlStateNormal];
```
willlarche: You probably need to cast these floats to float with f. Build for 32 and see.

components/Buttons/examples/ButtonsTypicalUse.m
```diff
@@ -119,33 +117,52 @@ - (void)viewDidLoad {
...
   ];
 
   [self setupExampleViews];
 }
 
 - (void)didTap:(id)sender {
   NSLog(@"%@ was tapped.", NSStringFromClass([sender class]));
+  if (sender == self.floatingButton) {
+    [self.floatingButton shrink:YES
+                               completion:^{
```
willlarche: This alignment looks a little off. I think it should be aligned to colons.

components/Buttons/examples/supplemental/ButtonsTypicalUseSupplemental.m
```diff
@@ -89,12 +84,11 @@ - (void)viewWillLayoutSubviews {
     CGFloat colOffset = CGRectGetWidth(self.view.frame) / 4;
```
willlarche: I know this isn't your writing but I'd love to see 'col' not abbreviated.

willlarche (changes requested):

components/Buttons/src/MDCFloatingButton+Animation.m
romoore: Nope, marked as "needing research" until I can get the correct motion metrics.

components/Buttons/src/MDCFloatingButton+Animation.m
romoore: Good idea! Done

components/Buttons/src/MDCFloatingButton+Animation.h
romoore: I'm a little wary of documenting the exact value (in case we change it) - how about "scaled to less than 0.1% of the original size"?

components/Buttons/examples/ButtonsTypicalUse.m
romoore: Agreed, seems I can make it ever-so-slightly better.

components/Buttons/examples/supplemental/ButtonsTypicalUseSupplemental.m
romoore: Agreed. Done.

components/Buttons/examples/ButtonsTypicalUse.m
```diff
@@ -20,28 +20,26 @@
...
-  [button setTitleColor:[UIColor colorWithWhite:0.1 alpha:1.0]
-               forState:UIControlStateNormal];
-  button.inkColor = [UIColor colorWithWhite:0 alpha:0.06];
+  [button setBackgroundColor:[UIColor clearColor] forState:UIControlStateNormal];
+  [button setTitleColor:[UIColor colorWithWhite:0.1F alpha:1] forState:UIControlStateNormal];
+  button.inkColor = [UIColor colorWithWhite:0 alpha:0.06F];
   button.layer.borderWidth = 1;
   button.layer.borderColor = [UIColor blackColor].CGColor;
-  button.disabledAlpha = 0.38;
+  button.disabledAlpha = 0.38F;
```
ajsecord: Floats in our codebase use lowercase 'f'.

components/Buttons/src/MDCFloatingButton+Animation.h
```diff
@@ -0,0 +1,39 @@
...
+
+ @param animated YES if the size change should be animated.
+ @param completion a completion block to call after the size change is complete.
+ */
+- (void)enter:(BOOL)animated completion:(void (^_Nullable)(void))completion;
```
ajsecord: Please name this more descriptively—at the call site reading `enter` doesn't make it clear what is going on. Consider naming it for the actual animation, in case there are multiple kinds in the future.

ajsecord: (and same for `exit`, obviously)

components/Buttons/src/MDCFloatingButton+Animation.m
```diff
@@ -0,0 +1,222 @@
...
+    self.layer.transform = CATransform3DIdentity;
+    self.layer.opacity = 1;
+    self.imageView.layer.transform = CATransform3DIdentity;
+    [self.layer removeAnimationForKey:@"kMDCFloatingButtontransform"];
+    [self.layer removeAnimationForKey:@"kMDCFloatingButtonopacity"];
+    [self.imageView.layer removeAnimationForKey:@"kMDCFloatingButtontransform"];
+    if (completion) {
+      completion();
+    }
+    NSLog(@"%@", [self.layer animationKeys]);
```
ajsecord: Remove all debugging logs.

components/Buttons/src/MDCFloatingButton+Animation.m
```diff
@@ -0,0 +1,222 @@
...
+
+  [self.layer removeAllAnimations];
+  [self.imageView.layer removeAllAnimations];
+
+  if (animated) {
+    [CATransaction begin];
+    [CATransaction setDisableActions:YES];
+    [CATransaction setCompletionBlock:enterActions];
+
+    CABasicAnimation *overallScaleAnimation = [CABasicAnimation animationWithKeyPath:@"transform"];
```
ajsecord: These animation blocks are suspiciously similar. Refactor into functions/methods.

components/Buttons/src/MDCFloatingButton+Animation.m
```diff
@@ -0,0 +1,222 @@
...
+#endif
+    [self.layer addAnimation:opacityAnimation forKey:@"kMDCFloatingButtonopacityexit"];
+
+    [CATransaction commit];
+  } else {
+    exitActions();
+  }
+}
+
+#pragma mark - UIView overrides to handle non-identity transforms
```
ajsecord: This needs more documentation—why is this required?

ajsecord (changes requested):

components/Buttons/examples/ButtonsTypicalUse.m
romoore: It didn't complain when I tried building, but I changed anyway.

components/Buttons/examples/ButtonsTypicalUse.m
romoore: Done

components/Buttons/src/MDCFloatingButton+Animation.h
romoore: I was trying to align to how we named ink enter/exit.  How about "expand" and "collapse" to align with the wording on the [motion guidelines]( https://material.io/guidelines/motion/transforming-material.html)?

components/Buttons/src/MDCFloatingButton+Animation.m
romoore: Yep, this commit was a work-in-progress save.

components/Buttons/examples/ButtonsTypicalUse.m
ianegordon: This is still capitalized in the PR on origin.
git push?

components/Buttons/src/MDCFloatingButton+Animation.m
```diff
@@ -0,0 +1,222 @@
...
+ limitations under the License.
+ */
+
+#import "MDCFloatingButton+Animation.h"
+
+#if TARGET_IPHONE_SIMULATOR
+float UIAnimationDragCoefficient(void);  // Private API for simulator animation speed
+#endif
+
+static const CGFloat kMDCFloatingButtonTransformScale = 0.00001F;
```
ianegordon: s/F/f/
Here and below.

components/Buttons/src/MDCFloatingButton+Animation.m
```diff
@@ -0,0 +1,222 @@
...
+static const NSTimeInterval kMDCFloatingButtonExitIconDuration = 0.135F;
+static const NSTimeInterval kMDCFloatingButtonExitIconOffset = 0.000F;
+
+static const NSTimeInterval kMDCFloatingButtonOpacityDuration = 0.015F;
+static const NSTimeInterval kMDCFloatingButtonOpacityEnterOffset = 0.030F;
+static const NSTimeInterval kMDCFloatingButtonOpacityExitOffset = 0.150F;
+
+@implementation MDCFloatingButton (Animation)
+
+#if TARGET_IPHONE_SIMULATOR
```
ianegordon: Add a comment why we need this?  And why only on the simulator?

components/Buttons/src/MDCFloatingButton+Animation.m
ianegordon: While I understand we can't use frame for layout when our transform isn't the identity, changing .frame is very concerning.
Let's discuss some alternatives.

ianegordon (changes requested):

components/Buttons/src/MDCFloatingButton+Animation.m
romoore: Done

components/Buttons/src/MDCFloatingButton+Animation.m
romoore: Done

components/Buttons/src/MDCFloatingButton+Animation.m
romoore: For debugging animations  using the simulator's "Slow animations" option.

components/Buttons/src/MDCFloatingButton+Animation.m
romoore: I've tried adding a container view, but MDCButton uses a shadowLayer so we can't swap them out.  It might be possible to refactor MDCButton to have an MDCButtonContainerView with a shadowLayer, then have all of the subviews in there.  The button view itself would just be an empty/transparent rectangle.  I can spend a day on that approach to see how far I can get if you prefer.

components/Buttons/src/MDCFloatingButton+Animation.m
```diff
@@ -0,0 +1,242 @@
...
+  if (fabs(beginOffset) > DBL_EPSILON) {
+    animation.beginTime = CACurrentMediaTime() + (beginOffset * [self fab_dragCoefficient]);
+  }
+#endif
+
+  return animation;
+}
+
+#if TARGET_IPHONE_SIMULATOR
++ (float)fab_dragCoefficient {
```
willlarche, line 80: Tell me more about this.

willlarche (approved):

components/Buttons/src/MDCFloatingButton+Animation.m
romoore: The purpose is to slow-down our custom Core Animation animations to match those of UIViews.  In the simulator, turning on Slow Animation changes this value.  Since it's a private API we don't want it compiled into "real" targets.  I stole the code from the FlexibleHeaderView.

components/Buttons/src/MDCFloatingButton+Animation.m
```diff
@@ -0,0 +1,242 @@
...
+ the frame will be modified. Apple is very clear that if a UIView's transform is not the identity
+ transform, then the value of the frame is undefined and should be ignored.
+
+ Unfortunately, much of our layout code assumes that frames are valid and relies on them instead of
+ the view's center and bounds for positioning information. As a result, this method will determine
+ what the frame would look like without the transform by using the center and bounds.
+
+ https://developer.apple.com/documentation/uikit/uiview/1622621-frame
+ */
+- (CGRect)frame {
```
ajsecord: I think this is too much magic and will burn us at some point in the future. The `frame` property is pretty central and special. For example, the page you link to says

> Changing the frame rectangle automatically redisplays the receiver without invoking the draw(_:) method.

How would that interact with this? Are we sure that this will continue to work in the future? 

Let's chat and figure out a more maintainable solution.

ajsecord (changes requested):

romoore:
During an offline discussion with ajsecord@, we've decided to remove the frame manipulation code from the MDCFloatingButton and provide clear guidance and reference to Apple's docs about how the frame is affected by transforms.

components/Buttons/src/MDCFloatingButton+Animation.h
```diff
@@ -0,0 +1,37 @@
...
+@interface MDCFloatingButton (Animation)
+
+/**
+ Expand this button to its unscaled (normal) size.
+
+ @param animated YES if the size change should be animated.
+ @param completion a completion block to call after the size change is complete.
+ */
+- (void)expand:(BOOL)animated completion:(void (^_Nullable)(void))completion;
+
```
ajsecord: As discussed, please add docs about how this manipulates the transform and include a pointer to Apple's relevant docs.

ajsecord (changes requested):

components/Buttons/src/MDCFloatingButton+Animation.h
```diff
@@ -0,0 +1,42 @@
...
+/**
+ Collapses this button so that it becomes smaller than 0.1% of its normal size.
+
+ @param animated YES if the size change should be animated.
+ @param completion a completion block to call after the size change is complete.
+
+ @note This method will modify the transform property of the button. Apple's documentation about
```
ajsecord, line 40: Copy this to `expand:completion:`.

ajsecord (approved):

samnm (approved):

PR #1759, opened 2017-08-09
Fab animate alt

romoore:
Different way to do FAB animation by adding a wrapper view to MDCButton


romoore:
This would break background color.

PR #1764, opened 2017-08-10
[Collections] Rewrite color helpers are functions

romoore:
One convenience macro was being used to transform integer red/green/blue
values into CGFloat for UIColor.  Another was transforming an integer
value into 8-bit red/green/blue components. Both are being changed to
static inline functions.

References #1682


romoore:
Nevermind, I'm going to try and switch to palettes anyway.  Plus I'm pretty sure these functions will have a naming conflict.

PR #1765, opened 2017-08-10
[ActivityIndicator] - Added voice accessibility support for MDCActivityIndicator

ShepJGoogle:
Added accessibility to MDCActivityIndicator

Accessibility Label:
- Returns "In progress" for indeterminate mode isSpinning
- Returns "{value} Percent Complete" for determinate mode isSpinning
- Returns "Progress Halted" for both modes determinate and indeterminate

Accessibility Value:
- Returns "1" for in progress
- Returns "0" for not in progress

Values From internal issue: b/27796425


ShepJGoogle:
Open to suggestions to change values or labels while waiting for accessibility advice

components/ActivityIndicator/src/MDCActivityIndicator.m
```diff
@@ -809,4 +814,73 @@ - (void)applyPropertiesWithoutAnimation:(void (^)(void))setPropBlock {
...
+  return YES;
+}
+
+- (NSString *)accessibilityLabel {
+
+  if (self.isAnimating) {
+    if (self.indicatorMode == MDCActivityIndicatorModeIndeterminate) {
+      NSString *key =
+      kMaterialActivityIndicatorStringTable[kStr_MaterialActivityIndicatorInProgressAccessibilityLabel];
+      return NSLocalizedStringFromTableInBundle(key,
```
romoore, line 850: If we don't localize the strings outside of English, and we're retrieving them from our bundle, can clients still provide translations?

components/ActivityIndicator/src/MDCActivityIndicator.m
```diff
@@ -809,4 +814,73 @@ - (void)applyPropertiesWithoutAnimation:(void (^)(void))setPropBlock {
...
+  }
+}
+
+- (UIAccessibilityTraits)accessibilityTraits {
+  return UIAccessibilityTraitUpdatesFrequently;
+}
+
+- (NSString *)accessibilityValue {
+  if (self.isAnimating) {
+    return @"1";
```
romoore: Should this be the localized value of the number 1 or should it always be an Arabic numeral "1" as a string?

components/ActivityIndicator/src/MDCActivityIndicator.m
```diff
@@ -809,4 +814,73 @@ - (void)applyPropertiesWithoutAnimation:(void (^)(void))setPropBlock {
...
+  } else {
+    return @"0";
```
romoore: Same for "0"

components/ActivityIndicator/src/MDCActivityIndicator.m
ShepJGoogle: Per @ianegordon request we wanted to get an initial english version out

romoore: What happens if the client wants to set an accessibility label? On a UIProgressView the default label is "Progress" and setting the accessibilityLabel works.

components/ActivityIndicator/src/MDCActivityIndicator.m
```diff
@@ -809,4 +814,73 @@ - (void)applyPropertiesWithoutAnimation:(void (^)(void))setPropBlock {
...
+      NSString *key =
+      kMaterialActivityIndicatorStringTable[kStr_MaterialActivityIndicatorInProgressAccessibilityLabel];
+      return NSLocalizedStringFromTableInBundle(key,
+                                                kMaterialActivityIndicatorStringsTableName,
+                                                [[self class] bundle],
+                                                @"In Progress");
+    } else {
+      NSUInteger percentage = (int)(self.progress * 100);
+      NSString *key =
+      kMaterialActivityIndicatorStringTable[kStr_MaterialActivityIndicatorProgressCompletedAccessibilityLabel];
```
willlarche, line 857: +4

components/ActivityIndicator/src/MDCActivityIndicator.m
```diff
@@ -809,4 +814,73 @@ - (void)applyPropertiesWithoutAnimation:(void (^)(void))setPropBlock {
...
+      NSString *localizedString = NSLocalizedStringFromTableInBundle(key,
+                                                kMaterialActivityIndicatorStringsTableName,
```
willlarche, line 859: Double check spacing with format_all.

components/ActivityIndicator/src/MaterialActivityIndicator.bundle/Resources/en.lproj/MaterialActivityIndicator.strings
```diff
@@ -0,0 +1,8 @@
+/* Accessibility label for indeterminate activity indicator in progress */
+"MaterialActivityIndicatorInProgressAccessibilityLabel" = "In Progress";
```
willlarche, line 2: I think the words that aren't the start of each quote should be lowercase. Sentence case.

willlarche (approved):

components/ActivityIndicator/src/MDCActivityIndicator.m
ajsecord: Rob, we're not expecting clients to provide translations, that's our responsibility.

Also, how would you suggest we return a _dynamic_ label without this code?

romoore: I think the dynamic label is great and we may want to mark the setter as not available, or something else (NSAssert) to let downstream dependencies know that setting the accessibilityLabel will have no effect.  Maybe we want to provide these as default values that clients can override.

ianegordon:
@romoore has raised some concerns that we should investigate which I've captured in #1841.

I'm will merge as this implementation is an improvement over our existing behavior and any improvements we make will depend on the localization capabilities added here.

PR #1776, opened 2017-08-10
[TextFields] Making a disabled color property on controllers

willlarche:


ajsecord (approved):

components/TextFields/src/MDCTextInputControllerFullWidth.m
```diff
@@ -424,6 +424,22 @@ - (void)setCharacterCountViewMode:(UITextFieldViewMode)characterCountViewMode {
   }
 }
 
+- (void)setDisabledColor:(UIColor *)disabledColor {
```
romoore, line 427: mark the parameter as __unused?

components/TextFields/src/MDCTextInputControllerFullWidth.m
```diff
@@ -424,6 +424,22 @@ - (void)setCharacterCountViewMode:(UITextFieldViewMode)characterCountViewMode {
...
+  [self updateUnderline];
+}
+
+- (UIColor *)disabledColor {
+  return [UIColor clearColor];
+}
+
++ (void)setDisabledColorDefault:(UIColor *)disabledColorDefault {
```
romoore, line 435: mark the parameter as __unused?

components/TextFields/src/MDCTextInputControllerFullWidth.m
willlarche: My understanding is you need __unused for suppressing 'unused variable' warnings. This doesn't trigger one in my Xcode 9. Are you seeing it? It's defined in a protocol so it has to be implemented. This particular class doesn't _yet_ need it.

romoore: I know that __unused can be used to suppress compiler warnings (not sure why it wouldn't be in this case), but I think it's also a healthy call-out to creators and maintainers that there is an argument being discarded.  Unused parameters are often completely innocuous, I just think it's a nice-to-have for source code bookkeeping.

romoore: I added some unused warnings to the project and I did get a pile of unused parameter complaint from the compiler.

```
-Wunused-comparison -Wunused-const-variable -Wunused-exception-parameter -Wunused-function -Wunused-label -Wunused-member-function -Wunused-parameter -Wunused-private-field -Wunused-property-ivar -Wunused-result -Wunused-value -Wunused-variable -Wunused-volatile-lvalue -Wused-but-marked-unused
```

willlarche: Please propose in weekly meeting. If it's accepted as a team best practice, then let's do it.

PR #1789, opened 2017-08-11
[TextFields] Removing unused constraints in default.

willlarche:


willlarche:
codecov is mistaken. This does not affect code coverage at all.

romoore (approved):
Hooray for less code!

PR #1804, opened 2017-08-15
[Scripts] Stop installing contributor tools for CI builds

romoore:
The Travis-CI .yaml script currently installs all of our "contributor
tools" before building.  These tools are not needed for continuous
integration builds, which target unit tests and code coverage.


ajsecord (approved):
Hm, I think they were required at some point, 🤷‍♂️

PR #1811, opened 2017-08-16
[Collections] Limit the drag needed to dismiss an item

### LINKED ISSUE #1780: [Collections] Threshold for swipe to dismiss is too high for large screens
### Overview
When using the swipe-to-dismiss gesture on collection, the threshold to compute if the swipe was successful (i.e. the item should be dismissed) if half of the width of the collection.
On large collection (e.g. iPad pro in landscape), the threshold is too big.

### Reproduction steps
On iPad landscape:
 - Open one of the swipe-to-dismiss examples in Catalog
 - Try to swipe-to-dismiss

### Version number
 - 31

gambard:
When using the swipe-to-dimiss gesture on large screens, the distance on which the user has to move the item can be big.
This commit limits the distance by using the width of the item instead of the width of the collection.

Closes #1780

### DIFF
--- a/components/Collections/src/private/MDCCollectionViewEditor.m
+++ b/components/Collections/src/private/MDCCollectionViewEditor.m
@@ -768,7 +768,7 @@ typedef NS_ENUM(NSInteger, MDCAutoscrollPanningDirection) {
 // The distance an item must be panned before it is dismissed. Currently half of the bounds width.
 - (CGFloat)distanceThresholdForDismissal {
   if (_cellSnapshot) {
-    return _cellSnapshot.bounds.size.width / 2;
+    return CGRectGetWidth(_cellSnapshot.bounds) / 2;
   } else {
     return _collectionView.bounds.size.width / 2;
   }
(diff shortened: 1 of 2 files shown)


components/Collections/src/private/MDCCollectionViewEditor.m
```diff
@@ -767,7 +767,11 @@ - (void)cleanupDismissingInformation {
 
 // The distance an item must be panned before it is dismissed. Currently half of the bounds width.
 - (CGFloat)distanceThresholdForDismissal {
-  return _collectionView.bounds.size.width / 2;
+  if (_cellSnapshot) {
+    return _cellSnapshot.bounds.size.width / 2;
```
romoore: Please use CGRectGetWidth(_cellSnapshot.bounds)

romoore (changes requested):

components/Collections/src/private/MDCCollectionViewEditor.m
gambard: Done.

romoore (approved):

PR #1813, opened 2017-08-16
[TextFields] Adding border view.

willlarche:


components/TextFields/src/MDCTextInputBorderView.m
```diff
@@ -0,0 +1,171 @@
...
+#import "MDCTextInputBorderView.h"
+
+static NSString *const MDCTextInputBorderViewBorderFillColorKey =
+    @"MDCTextInputBorderViewBorderFillColorKey";
+static NSString *const MDCTextInputBorderViewBorderPathKey =
+    @"MDCTextInputBorderViewBorderPathKey";
+static NSString *const MDCTextInputBorderViewBorderStrokeColorKey =
+    @"MDCTextInputBorderViewBorderStrokeColorKey";
+
+static inline NSString * _Nullable MDCNSStringFromCGLineCap(CGLineCap lineCap) {
```
ianegordon, line 26: naming-nit: You typically don't include the object prefixes in the function name.
Consider [UIView initWithCoder:].  It is initWithCoder, not initWithNSCoder.
Or NSClassFromString, not NSClassFromNSString.
s/MDCNSStringFromCGLineCap/MDCStringFromCGLineCap/
However, CGLineCap is in CoreGraphics, which is still dealing with CGRefs instead of object pointers, so it still has the prefix.
Maybe,...  naming is hard.

components/TextFields/src/MDCTextInputBorderView.m
```diff
@@ -0,0 +1,171 @@
...
+    case kCGLineCapButt:
+      lineCapString = kCALineCapButt;
+      break;
+    case kCGLineCapRound:
+      lineCapString = kCALineCapRound;
+      break;
+    case kCGLineCapSquare:
+      lineCapString = kCALineCapSquare;
+      break;
+    default:
```
ianegordon: Consider removing the default clause.

If we change this switch statement to:
switch(lineCap) {
  case kCGLineCapButt:
    return kCALineCapButt;
  case kCGLineCapRound:
    return kCALineCapROund;
  case kCGLineCapSquare:
    return kCALineCapSquare;
}
return nil;
}

If Apple adds another entry to CGLineCap in iOS 12.3, the compiler will automagically throw a warning along the lines of 'You aren't handling every case in your switch'.  Otherwise, with our current implementation someone will upgrade to iOS 12.3 and wonder why we aren't supporting kCGLineCapTriangle and we will have no warning.

Here and below.

components/TextFields/src/MDCTextInputBorderView.m
```diff
@@ -0,0 +1,171 @@
...
+}
+
+- (instancetype)initWithCoder:(NSCoder *)coder
+{
+  self = [super initWithCoder:coder];
+  if (self) {
+    _borderFillColor = [coder decodeObjectForKey:MDCTextInputBorderViewBorderFillColorKey];
+    _borderPath = [coder decodeObjectForKey:MDCTextInputBorderViewBorderPathKey];
+    _borderStrokeColor = [coder decodeObjectForKey:MDCTextInputBorderViewBorderStrokeColorKey];
+    [self commonMDCTextInputBorderViewInit];
```
ianegordon, line 82: commonMDCXYZInit before decoded values or decoded values will get overwritten by defaults set in commonInit.
Looking at commonInit, this will work as intended.
We should determine the 'one, true way' we want to initWithCoder.
Happy to do so in another Issue / meeting / PR.

components/TextFields/src/MDCTextInputBorderView.m
```diff
@@ -0,0 +1,171 @@
...
+  [super encodeWithCoder:aCoder];
+  [aCoder encodeObject:self.borderFillColor forKey:MDCTextInputBorderViewBorderFillColorKey];
+  [aCoder encodeObject:self.borderPath forKey:MDCTextInputBorderViewBorderPathKey];
+  [aCoder encodeObject:self.borderStrokeColor forKey:MDCTextInputBorderViewBorderStrokeColorKey];
+}
+
+- (nonnull id)copyWithZone:(nullable NSZone *)zone {
+  MDCTextInputBorderView *copy = [[[self class] alloc] initWithFrame:self.frame];
+  copy.borderFillColor = self.borderFillColor;
+  copy.borderPath = self.borderPath.copy;
```
ianegordon: Nit: /self.borderPath.copy/[self.borderPath copy]/

components/TextFields/src/MDCTextInputBorderView.m
```diff
@@ -0,0 +1,171 @@
...
+- (void)layoutSubviews {
+  [super layoutSubviews];
+
+  [self updateBorder];
+}
+
+#pragma mark - Properties
+
+- (void)setBorderFillColor:(UIColor *)borderFillColor {
+  if (_borderFillColor != borderFillColor) {
```
ianegordon: Consider:
if (![_borderFillColor isEqual:borderFillColor]) 

I usually prefer isEqual: for NSObjects.
Here and below.

ianegordon (changes requested):

components/TextFields/src/MDCTextInputBorderView.m
willlarche: Great point!

components/TextFields/src/MDCTextInputBorderView.m
willlarche: Oops. Thank you.

components/TextFields/src/MDCTextInputBorderView.m
willlarche: Good idea!

components/TextFields/src/MDCTextInputBorderView.m
willlarche: Spoke in person. Decided to leave as is.

components/TextFields/src/MDCTextInputBorderView.m
willlarche: I'm actually a fan of the way I'm doing it here. 

1. Read from file.
2. Handle nils in commonInit.
3. Profit.

willlarche:
@ianegordon , this is ready for re-review.

components/TextFields/src/MDCTextInputBorderView.m
ianegordon: My concern is that we should be consistent.  Either commonMDCInit before decoding -or- decoding before commonMDcInit with nil handling.  I'll open an issue.
LGTM.

willlarche: You got it.

ianegordon (approved):

PR #1814, opened 2017-08-16
[TextFields] Input changes for new styles.

willlarche:
Dependent on #1813


components/TextFields/src/MDCMultilineTextField.m
```diff
@@ -74,7 +76,7 @@ - (instancetype)initWithFrame:(CGRect)frame {
 
 - (instancetype)initWithCoder:(NSCoder *)aDecoder {
   self = [super initWithCoder:aDecoder];
-  if (self) {
+  if (self) {\
```
ajsecord: Danger: '\'

ajsecord (approved):
One typo.

components/TextFields/src/MDCMultilineTextField.m
willlarche: So, total weirdness. There's no \ in the file. I've seen this twice now where GH says there is.

PR #1815, opened 2017-08-16
[TextFields] Renaming old default to legacy default.

willlarche:
Rename of files and opting out of new functionality.

No new functionality.

### DIFF
--- a/components/TextFields/src/MDCTextInputControllerLegacyDefault.m
+++ b/components/TextFields/src/MDCTextInputControllerLegacyDefault.m
@@ -212,10 +212,10 @@ static UITextFieldViewMode _underlineViewModeDefault = UITextFieldViewModeWhileE
   self = [self init];
   if (self) {
     _textInput = textInput;
-  }
 
-  // This should happen last because it relies on the state of a ton of properties.
-  [self setupInput];
+    // This should happen last because it relies on the state of a ton of properties.
+    [self setupInput];
+  }
 
   return self;
 }
--- a/components/TextFields/src/MDCTextInputControllerLegacyDefault.m
+++ b/components/TextFields/src/MDCTextInputControllerLegacyDefault.m
@@ -285,7 +285,7 @@ static UITextFieldViewMode _underlineViewModeDefault = UITextFieldViewModeWhileE
   _characterCountViewMode = UITextFieldViewModeAlways;
   _disabledColor = [[self class] disabledColorDefault];
   _floatingEnabled = [[self class] isFloatingEnabledDefault];
-  _internalCharacterCounter = [MDCTextInputAllCharactersCounter new];
+  _internalCharacterCounter = [[MDCTextInputAllCharactersCounter alloc] init];
   _underlineViewMode = [[self class] underlineViewModeDefault];
   _textInput.hidesPlaceholderOnInput = NO;
 
--- a/components/TextFields/src/MDCTextInputControllerLegacyDefault.m
+++ b/components/TextFields/src/MDCTextInputControllerLegacyDefault.m
@@ -19,6 +19,7 @@
 #import "MDCMultilineTextField.h"
 #import "MDCTextField.h"
 #import "MDCTextInput.h"
+#import "private/MDCTextInputArt.h"
 #import "MDCTextInputCharacterCounter.h"
 #import "MDCTextInputUnderlineView.h"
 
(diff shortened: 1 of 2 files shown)


components/TextFields/src/MDCTextInputControllerLegacyDefault.m
```diff
@@ -0,0 +1,1261 @@
...
+
+  return self;
+}
+
+- (instancetype)initWithCoder:(NSCoder *)aDecoder {
+  self = [super init];
+  if (self) {
+    [self commonMDCTextInputControllerDefaultInitialization];
+
+    _activeColor =
```
ianegordon, line 187: If the decoder doesn't have a value for ActiveColorKey, won't this overwrite the default value we just set for _activeColor to nil?
Check for existence of keys?

components/TextFields/src/MDCTextInputControllerLegacyDefault.m
```diff
@@ -0,0 +1,1261 @@
...
+}
+
+- (instancetype)initWithTextInput:(UIView<MDCTextInput> *)textInput {
+  self = [self init];
+  if (self) {
+    _textInput = textInput;
+  }
+
+  // This should happen last because it relies on the state of a ton of properties.
+  [self setupInput];
```
ianegordon: Consider moving inside if (self) block?
Usually I don't expect to see anything expect return self; after the if block.

components/TextFields/src/MDCTextInputControllerLegacyDefault.m
```diff
@@ -0,0 +1,1261 @@
...
+@property(nonatomic, strong) UIFont *customTrailingFont;
+
+@end
+
+@implementation MDCTextInputControllerLegacyDefault
+
+@synthesize characterCounter = _characterCounter;
+@synthesize characterCountMax = _characterCountMax;
+@synthesize characterCountViewMode = _characterCountViewMode;
+@synthesize errorColor = _errorColor;
```
ianegordon: If you explicitly define setErrorColor and errorColor below do you need to @synthesize?
(Here and above / below)

components/TextFields/src/MDCTextInputControllerLegacyDefault.m
```diff
@@ -0,0 +1,1261 @@
...
+- (void)dealloc {
+  [self unsubscribeFromNotifications];
+  [self unsubscribeFromKVO];
+}
+
+- (void)commonMDCTextInputControllerDefaultInitialization {
+  _characterCountViewMode = UITextFieldViewModeAlways;
+  _disabledColor = [[self class] disabledColorDefault];
+  _floatingEnabled = [[self class] isFloatingEnabledDefault];
+  _internalCharacterCounter = [MDCTextInputAllCharactersCounter new];
```
ianegordon: Internally, we prefer [[MDCClass alloc] init] over new.

components/TextFields/src/MDCTextInputControllerLegacyDefault.m
```diff
@@ -0,0 +1,1261 @@
...
+    [self updatePlaceholderY];
+    [CATransaction commit];
+  } else {
+    [self updateLayout];
+  }
+
+  // Accessibility
+  if (self.textInput.isEditing && self.characterCountMax > 0) {
+    NSString *announcementString;
+    if (!announcementString.length) {
```
ianegordon, line 1125: announcementString was just created / set to nil on the line above.
Will announcementString.length ever not be 0?
Do we mean to test another string here?

ianegordon (changes requested):
This is the first pass.  I'll take another when we've landed the PR this depends on so I can step through.

components/TextFields/src/MDCTextInputControllerLegacyDefault.m
willlarche: So true! There were a few others too.

components/TextFields/src/MDCTextInputControllerLegacyDefault.m
willlarche: .activeColor actually handles nil in its setter and getting. It uses the default!

components/TextFields/src/MDCTextInputControllerLegacyDefault.m
willlarche: Done!

components/TextFields/src/MDCTextInputControllerLegacyDefault.m
willlarche: Done!

components/TextFields/src/MDCTextInputControllerLegacyDefault.m
willlarche: Good catch. This is from copying code where the state is different. Done.

willlarche:
Ready for re-review.

ianegordon (approved):

PR #1829, opened 2017-08-17
[Scripts] Retry unit tests on Travis

romoore:
Trying to deflake Travis builds by retrying unit tests - our largest
source of failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ script:
   - pod --version
   - scripts/prep_all
   - travis_wait scripts/build_all --verbose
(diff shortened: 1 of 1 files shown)


romoore:
[Travis retried the unit tests where they would have failed. Now they succeed!](https://travis-ci.org/material-components/material-components-ios/jobs/265739511#L2965)

romoore:
Attempted multiple restarts of the job, always succeed even when unit test would have failed on one of the builds.

willlarche (approved):
So wonderful.

PR #1851, opened 2017-08-19
[Collections, TextFields] Fix static analyzer warnings

romoore:
Two instances of values being assigned without being read (they were
always being reassigned), and one example of confusing property
attributes (a copied NSMutableArray). There should be no functional
change in the code, but now the static analyzer has no warnings.

### DIFF
--- a/components/Collections/src/MDCCollectionViewFlowLayout.m
+++ b/components/Collections/src/MDCCollectionViewFlowLayout.m
@@ -394,7 +394,7 @@ static const NSInteger kSupplementaryViewZIndex = 99;
   // view rows.
   CGRect insetFrame = attr.frame;
   if (!CGRectIsEmpty(insetFrame)) {
-    UIEdgeInsets insets = self.sectionInset;
+    UIEdgeInsets insets;
 
     // Retrieve the insets from the Flow Layout delegate to maintain consistency with the CVC
     if ([self.collectionView.delegate
--- a/components/TextFields/src/MDCMultilineTextField.m
+++ b/components/TextFields/src/MDCMultilineTextField.m
@@ -434,9 +434,16 @@ static NSString *const MDCMultilineTextFieldTrailingViewModeKey =
                                                              constant:0];
   }
   self.trailingViewCenterY.active = YES;
-  
+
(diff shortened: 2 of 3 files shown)


components/TextFields/src/MDCMultilineTextField.m
```diff
@@ -448,7 +455,7 @@ - (void)updateTrailingViewAlpha {
 - (CGFloat)trailingViewAlpha {
   // The trailing view has the same behavior as .rightView in UITextField: It has visual precedence
   // over the clear button.
-  CGFloat trailingViewAlpha = self.trailingView.alpha;
```
ajsecord, line 451: The warning is wrong here. If self.trailingViewMode is garbage (not one of the enums), then trailingViewAlpha is an arbitrary value. The original code is safer.

components/Collections/src/MDCCollectionViewFlowLayout.m
```diff
@@ -394,7 +394,7 @@ - (MDCCollectionViewLayoutAttributes *)updateSupplementaryViewAttribute:
   // view rows.
   CGRect insetFrame = attr.frame;
   if (!CGRectIsEmpty(insetFrame)) {
-    UIEdgeInsets insets = self.sectionInset;
```
ajsecord, line 397: I think the original code is better here—seeing an uninitialized local variable makes me read the rest of the code just to figure out if it got assigned in all cases. Plus future edits to the following logic might change things and miss that the variable wasn't assigned.

ajsecord (changes requested):

components/TextFields/src/MDCMultilineTextField.m
willlarche: Also, I had that line as a subliminal reminder to the next dev what we're talking about: self.trailingView's alpha.

components/Collections/src/MDCCollectionViewFlowLayout.m
willlarche: Same comment.

romoore:
Alright. aside from an ugly (void) cast to disable the analyzer, there isn't much to do here.

PR #1856, opened 2017-08-21
[Buttons] Respect minimum touch target size

### LINKED ISSUE #223: Buttons sizeToFit/sizeThatFits should enforce minimum touch sizes
Buttons should respect the minimum touch size of Material: https://www.google.com/design/spec/layout/metrics-keylines.html#metrics-keylines-touch-target-size

romoore:
All Buttons should support a minimum touch target size of 48 points for
both width and height.  Unless custom |hitAreaInsets| are set on the
button, the touch target will be expanded to 48x48 points and centered
around the button's center.

Closes #223

### DIFF
--- a/components/Buttons/tests/unit/ButtonsTests.m
+++ b/components/Buttons/tests/unit/ButtonsTests.m
@@ -265,8 +265,7 @@ static NSString *controlStateDescription(UIControlState controlState) {
   [button setElevation:normalElevation forState:UIControlStateNormal];
 
   // Then
-  XCTAssertEqualWithAccuracy([button elevationForState:button.state],
-                             normalElevation,
+  XCTAssertEqualWithAccuracy([button elevationForState:button.state], normalElevation,
                              kEpsilonAccuracy);
 }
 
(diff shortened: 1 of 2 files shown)


romoore:
Marking as "author will merge" because this is a significant proposed change in behavior, particularly for "mini" Floating Action Buttons.

components/Buttons/tests/unit/ButtonsTests.m
```diff
@@ -276,15 +275,14 @@ - (void)testCurrentElevationHighlighted {
   CGFloat normalElevation = 10;
   CGFloat elevation = 40;
   [button setElevation:normalElevation forState:UIControlStateNormal];
-  [button setElevation:elevation forState:UIControlStateHighlighted];;
+  [button setElevation:elevation forState:UIControlStateHighlighted];
+  ;
```
samnm: ;

There are two more below also.

samnm (approved):
Some tiny cleanup changes requested, otherwise LGTM.

components/Buttons/tests/unit/ButtonsTests.m
romoore: Done

ianegordon (changes requested):
Would you please add a test and/or example testing what happens when buttons smaller than the minimum size are adjacent? 
I immediately thought of a keypad-like interface with a 3 x 3 grid of number buttons claiming touches intended for their neighbor.

romoore:
This approach is too surprising. I will push up another CR that takes a gentler approach.

PR #1857, opened 2017-08-21
Release candidate 32.0.0

ajsecord:


components/Buttons/src/MDCFloatingButton+Animation.m
```diff
@@ -123,20 +123,22 @@ - (void)expand:(BOOL)animated completion:(void (^_Nullable)(void))completion {
...
-                 beginOffset:kMDCFloatingButtonEnterIconOffset];
-    [self.imageView.layer addAnimation:iconScaleAnimation forKey:kMDCFloatingButtonTransformKey];
+    CALayer *iconPresentationLayer = self.imageView.layer.presentationLayer;
+    if (iconPresentationLayer) {
+      CABasicAnimation *iconScaleAnimation = [MDCFloatingButton
+          animationWithKeypath:@"transform"
+                       toValue:[NSValue valueWithCATransform3D:iconPresentationLayer.transform]
+                     fromValue:[NSValue
+                                   valueWithCATransform3D:CATransform3DConcat(
+                                                              iconPresentationLayer.transform,
```
romoore, line 133: This line (fromValue) should be self.layer.presentationLayer.transform, not the iconLayer's presentationLayer.

romoore (approved):
Would prefer a minor fix-up if possible for the FAB animation, but since it's new and would only be a visual bug I think it's low-priority.

PR #1866, opened 2017-08-21
[Buttons] Default hitAreaInsets for FAB

### LINKED ISSUE #223: Buttons sizeToFit/sizeThatFits should enforce minimum touch sizes

romoore:
The Floating Action Button should have default hitAreaInsets so that it
adheres the minimum touch target in the Material Design guidelines.
This commit introduces a set of default hitAreaInsets for the different
FAB styles as well as a warning if a button's touch target is smaller
than the 48-point minimum.

Closes #223


romoore:
Marking "author will merge" to ensure I get feedback from senior team members.

ianegordon (changes requested):
There is concern about unexpected behavior on adjacent buttons.  I believe the current plan is to expand the hitAreaInsets only for the small FAB?

romoore:
Added examples:

### ButtonBar
simulator screen shot - iphone 4s - 2017-08-22 at 11 00 52

### NavigationBar
simulator screen shot - iphone 4s - 2017-08-22 at 11 01 07

simulator screen shot - iphone 4s - 2017-08-22 at 11 00 41

ianegordon (approved):

components/Buttons/src/MDCButton.m
```diff
@@ -305,7 +310,22 @@ - (void)layoutSubviews {
...
+    return CGRectContainsPoint(UIEdgeInsetsInsetRect(self.bounds, self.hitAreaInsets), point);
+  }
+
+  // If the bounds are smaller than the minimum touch target, produce a warning once
+  CGFloat width = CGRectGetWidth(self.bounds);
+  CGFloat height = CGRectGetHeight(self.bounds);
+  if (width < MDCButtonMinimumTouchTargetWidth || height < MDCButtonMinimumTouchTargetHeight) {
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+      NSLog(@"Button touch target is too small. Button: %@, Touch Target: %@", [self description],
```
ianegordon: Indicate target size.
Button: [self description] touch target (24 x 34) is smaller than the recommended (48 x 48).

components/Buttons/src/MDCButton.m
```diff
@@ -305,7 +310,22 @@ - (void)layoutSubviews {
 }
 
 - (BOOL)pointInside:(CGPoint)point withEvent:(UIEvent *)event {
-  return CGRectContainsPoint(UIEdgeInsetsInsetRect(self.bounds, _hitAreaInsets), point);
+  // If there are custom hitAreaInsets, use those
+  if (!UIEdgeInsetsEqualToEdgeInsets(self.hitAreaInsets, UIEdgeInsetsZero)) {
+    return CGRectContainsPoint(UIEdgeInsetsInsetRect(self.bounds, self.hitAreaInsets), point);
```
ianegordon: Do you need to CGRectStandardized(self.bounds)?

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -125,6 +126,17 @@ - (UIEdgeInsets)defaultContentEdgeInsets {
   }
 }
 
+- (UIEdgeInsets)defaultHitAreaInsets {
+  switch (_shape) {
+    case MDCFloatingButtonShapeDefault:
+      return UIEdgeInsetsZero;
+    case MDCFloatingButtonShapeMini:
+      return UIEdgeInsetsMake(-4, -4, -4, -4);
```
ianegordon: Consider adding a comment to lend some context to why -4s were chosen:
// Set EdgeInsets to ensure that the hit target of the Mini meets the minimum recommendation of 48 x 48.

components/Buttons/tests/unit/ButtonsTests.m
```diff
@@ -265,8 +265,7 @@ - (void)testCurrentElevationNormal {
   [button setElevation:normalElevation forState:UIControlStateNormal];
 
   // Then
-  XCTAssertEqualWithAccuracy([button elevationForState:button.state],
-                             normalElevation,
+  XCTAssertEqualWithAccuracy([button elevationForState:button.state], normalElevation,
```
ianegordon: Per our style guide, if we line-break on one parameter, each parameter should get its own line.

ianegordon (changes requested):

components/Buttons/src/MDCButton.m
romoore: Yes, it appears that UIEdgeInsetsInsetRect performs a literal manipulation as defined in the docs and does not standardize the rect first.

romoore: After writing the unit test, it seems that setting either the frame or the bounds property of a UIView standardizes the rect in the setter.  I will add the call to CGRectStandardize, but I'm having difficulty coming up with a unit test.

components/Buttons/tests/unit/ButtonsTests.m
romoore: Done

components/Buttons/src/MDCFloatingButton.m

components/Buttons/src/MDCButton.m
romoore: Done

components/Buttons/src/MDCButton.m
```diff
@@ -305,7 +310,26 @@ - (void)layoutSubviews {
...
+        point);
+  }
+
+  // If the bounds are smaller than the minimum touch target, produce a warning once
+  CGFloat width = CGRectGetWidth(self.bounds);
+  CGFloat height = CGRectGetHeight(self.bounds);
+  if (width < MDCButtonMinimumTouchTargetWidth || height < MDCButtonMinimumTouchTargetHeight) {
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+      NSLog(@"Button touch target does not meet minimum size guidlines of (48, 48). Button: %@, "
```
ianegordon: Super-nit: consider using MDCButtonMinimumTouchTargetWidth / Height instead of hard-coding 48 x 48.

components/Buttons/src/MDCButton.m
```diff
@@ -305,7 +310,26 @@ - (void)layoutSubviews {
...
+            @"Touch Target: %@",
+            [self description],
+            NSStringFromCGSize(CGSizeMake(width, height)));
+    });
+  }
+  return [super pointInside:point withEvent:event];
```
ianegordon: If we are switching to [super pointInside:withEvent:] here, do we also want to in our implementation if we have custom hitAreaInsets above?

Alternatively, should this return CGRectContainsPoint()?

If there is cause for these two cases to be different, we should document why.

components/NavigationBar/examples/NavigationBarLayoutExample.m
```diff
@@ -1,24 +1,19 @@
...
  */
 
-//
-//  NavigationBarLayoutExample.m
-//  Pods
-//
-//  Created by Rob Moore on 7/28/17.
-//
-//
-
```
ianegordon: I appreciate this cleanup, but changes to license clauses can draw extra scrutiny.
Consider pulling the changes to this file into a separate PR.

ianegordon (changes requested):

components/Buttons/src/MDCButton.m
romoore: If we have hitAreaInsets, then we should use those along with our bounds to compute the touch target. If we don't, then we leave it up to our superclass to determine.  I think that's a bit better in terms of class hierarchies since here we won't assume how the default touch target should be calculated.

components/NavigationBar/examples/NavigationBarLayoutExample.m
romoore: ok

romoore: Done

components/Buttons/src/MDCButton.m
romoore: Done

ianegordon (approved):

ianegordon:
LGTM. Please merge when Travis is happy.

PR #1876, opened 2017-08-22
[TextFields] Renaming text field controllers to have 'Legacy'

willlarche:
No functional changes. Renames and title changes only.


ajsecord:
I tested locally with `scripts/test_all` because Travis CI is having a hard time today.

ajsecord (approved):

Issue #1805, opened 2017-08-15
[Typography] mdc_preferredFont: is returning the incorrect version of .SFUI Font Family


sigito:
`[MDCTypography titleFont]` returns a _20pt_ font with font family `.SFUIDisplay-Medium`, when 
`[UIFont mdc_preferredFontForMaterialTextStyle:MDCFontTextStyleTitle]` returns `.SFUIText-Medium`. While I cannot find this anywhere officially, but according to [San Francisco display vs text / compact vs normal a brief review](http://protosketch.io/san-francisco-display-vs-text-compact-vs-normal-a-brief-review/):

> The OS will change the two fonts automagically **text is for sizes up to 19pt and display is from sizes above 19pt**. This will not be the case in design software though.

Indeed the `UIFont` family returned for _19pt_ is `.SFUIText`, and for _20pt_ - `.SFUIDisplay`.

It seems that the fix would be to cache values for _Text_ and _Display_ fonts in [UIFontDescriptor+MaterialTypography.m](https://github.com/material-components/material-components-ios/blob/develop/components/Typography/src/UIFontDescriptor%2BMaterialTypography.m#L39) and select a proper one depending on `materialTraits.pointSize`.

ianegordon:
Hi @sigito. Good catch! Fixing this may get pulled into removing some of the font legacy code once we drop support for iOS 8 (Pre-SF), but your suggestion sounds like a good solution.

sigito:
Does this mean we would have to wait until iOS 8 support is dropped? Is there a tracking issue or an ETA for it?

ianegordon:
Currently there is no firm schedule on dropping iOS 8 support.
That conversation will likely not begin until iOS 11 reaches general availability.

If you are interested in getting a fix in sooner rather than later, PRs are always welcome and I would be happy to work with you to get a PR landed. If you aren't comfortable putting together a PR, let me know and I'll try and get this issue priority higher in the backlog.

Your proposed solution SGTM.  I would replace systemFontFamilyName with smallSystemFontFamilyName and largeSystemFontFamilyName, setting the UIFontDescriptorFamilyAttribute as appropriate.  I would want to verify things work as expected on iOS 8, and perhaps add unit tests to demonstrate we are returning the right family for each size.

PR #1880, opened 2017-08-23
[Typography] Match large font family on iOS 9+

sigito:
This change only affects existing large fonts. Visually display font family has smaller tracking values. However they are still different than default system font would have [1].
Some images to compare for `UIContentSizeCategory`: 
- [UIContentSizeCategoryLarge](https://image.ibb.co/ea1KP5/UIContent_Size_Category_Large.png);
- [UIContentSizeCategoryExtraExtraExtraLarge](https://image.ibb.co/m9NR45/UIContent_Size_Category_Extra_Extra_Extra_Large.png).
Where:
_System_ - `[UIFont systemFontForSize:weight:]`;
_Old_ - state before this change;
_New_ - change results;
_Typography_ - `[MDCTypography *Font]`.

This has no effect on iOS 8, since there is only a single font used everywhere. Checked in _iPhone 5 (8.4)_ simulator.

Partially addresses #1805.

[1] With iOS 11 it should be also possible to load with https://developer.apple.com/documentation/uikit/uifontdescriptor.traitkey and related API.

### DIFF
--- a/components/Typography/src/UIFontDescriptor+MaterialTypography.m
+++ b/components/Typography/src/UIFontDescriptor+MaterialTypography.m
@@ -35,9 +35,9 @@
 
   // Store the system font family name to ensure that we load the system font.
   // If we do not explicitly include this UIFontDescriptorFamilyAttribute in the
-  // FontDescriptor the OS will default to Helvetica.
+  // FontDescriptor the OS will default to Helvetica. On iOS 9+, the Font Family
+  // changes from San Francisco to San Francisco Display at point size 20.
   static NSString *smallSystemFontFamilyName;
-  // For font size 20+, iOS returns a different font family. Query it separately.
   static NSString *largeSystemFontFamilyName;
   static dispatch_once_t onceToken;
   dispatch_once(&onceToken, ^{
--- a/components/Typography/src/UIFontDescriptor+MaterialTypography.m
+++ b/components/Typography/src/UIFontDescriptor+MaterialTypography.m
@@ -36,24 +36,32 @@
   // Store the system font family name to ensure that we load the system font.
   // If we do not explicitly include this UIFontDescriptorFamilyAttribute in the
   // FontDescriptor the OS will default to Helvetica.
-  static NSString *systemFontFamilyName;
+  static NSString *smallSystemFontFamilyName;
+  // For font size 20+, iOS returns a different font family. Query it separately.
+  static NSString *largeSystemFontFamilyName;
   static dispatch_once_t onceToken;
   dispatch_once(&onceToken, ^{
-    UIFont *systemFont;
+    UIFont *smallSystemFont;
+    UIFont *largeSystemFont;
     if ([UIFont respondsToSelector:@selector(systemFontOfSize:weight:)]) {
-      systemFont = [UIFont systemFontOfSize:12 weight:UIFontWeightRegular];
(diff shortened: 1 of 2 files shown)


romoore:
Apple HIG discussion of selecting font families by point size.  https://developer.apple.com/ios/human-interface-guidelines/visual-design/typography/

components/Typography/src/UIFontDescriptor+MaterialTypography.m
```diff
@@ -36,24 +36,32 @@ + (nonnull UIFontDescriptor *)mdc_preferredFontDescriptorForMaterialTextStyle:
   // Store the system font family name to ensure that we load the system font.
   // If we do not explicitly include this UIFontDescriptorFamilyAttribute in the
   // FontDescriptor the OS will default to Helvetica.
```
romoore: Nit: Could you please update this to include iOS 9+ with San Francisco?  Your PR discussion was really good, so something like, "On iOS 9+, the Font Family changes from San Francisco to San Francisco Display at point size 20".  You could probably put the entire comment together and remove the one-liner just below about font size 20+.

components/Typography/src/UIFontDescriptor+MaterialTypography.m
```diff
@@ -36,24 +36,32 @@ + (nonnull UIFontDescriptor *)mdc_preferredFontDescriptorForMaterialTextStyle:
...
+      largeSystemFont = [UIFont systemFontOfSize:20];
     }
-    systemFontFamilyName = [systemFont.familyName copy];
+    smallSystemFontFamilyName = [smallSystemFont.familyName copy];
+    largeSystemFontFamilyName = [largeSystemFont.familyName copy];
   });
 
   NSDictionary *traits = @{ UIFontWeightTrait : @(materialTraits.weight) };
+  NSString *fontFamily =
+      materialTraits.pointSize < 19.5 ? smallSystemFontFamilyName : largeSystemFontFamilyName;
```
romoore, line 60: Nit/Suggestion: Should we make this 20 instead of 19.5? Apple's documentation says that "20 and above" should use SF Pro Display. Does the 19.5 look better visually as a breaking point?

romoore (commented):
LGTM with some minor suggestions, would like @ianegordon  to review.

components/Typography/src/UIFontDescriptor+MaterialTypography.m
sigito: My assumption is probably wrong here. 

I thought of a good pivot point, and didn't think of a case that people might use a half-point font. _20_ is concerning me a bit, since it is a floating point value (_double_ to be exact). So there is no guarantee that `pointSize` is not represented as 19.99999997, or is there?

romoore: That sounds good enough to me. 👍  I thought you might have "split the difference" between 19 and 20, but I wasn't sure if there were other advantages as well.  

In general, it seems unlikely that the float would have representational errors like that, but you make a good point that it's not worth the risk. 😸

romoore:
Trimming commit message summary to fit in 50 columns

components/Typography/src/UIFontDescriptor+MaterialTypography.m
sigito: Done.

ianegordon (approved):
Excellent, thank you @sigito !

PR #1887, opened 2017-08-23
[Buttons] Make uppercaseTitle nondestructive

samnm:
Since uppercaseTitle defaults to YES, it was impossible to turn it off with UIAppearance and have titles behave as expected. All existing uppercaseTitle tests still pass and a new test was added to ensure this fix will not regress.

### DIFF
--- a/components/Buttons/src/MDCButton.m
+++ b/components/Buttons/src/MDCButton.m
@@ -259,7 +259,7 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
 
   // Uppercase all titles
   if (_uppercaseTitle) {
-    [self uppercaseAllTitles];
+    [self updateTitleCase];
   }
 }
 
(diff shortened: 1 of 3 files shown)


components/Buttons/src/MDCButton.m
```diff
@@ -182,8 +182,8 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder {
     }
     [self updateBackgroundColor];
 
-    if ([aDecoder containsValueForKey:MDCButtonAccessibilityLabelsKey]) {
```
ajsecord, line 185: If someone restores from a previous version, this code will drop the title because you've changed the key. You should check for both keys.

ajsecord (approved):
Nice!

ajsecord (changes requested):
Whoops, still need that key decode change.

components/Buttons/src/MDCButton.m
samnm: Fixed.

ajsecord (approved):

PR #1889, opened 2017-08-24
[FeatureHighlight] Shown views can receive taps

### LINKED ISSUE #1888: [Feature Highlight] Taps ignored in shown view when highlight is visible
### Overview
"Showing" a view along with Feature Highlight that has a gesture recognizer (like a button with an action) prevents actions from triggering on the presented view.  It looks like the pan gesture recognizer is cancelling taps that it does not accept.

### Reproduction steps
Create a new MDCButton with an action and have Feature Highlight show that view.  For example, in FeatureHighlightTypicalUseViewController.m

```objectivec
- (void)didTapButton:(id)sender {
  MDCFloatingButton *fab = [MDCFloatingButton floatingButtonWithShape:MDCFloatingButtonShapeDefault];
  [fab setImage:[UIImage imageNamed:@"Plus"] forState:UIControlStateNormal];
  [fab sizeToFit];
  fab.backgroundColor = UIColor.orangeColor;
  [fab addTarget:self action:@selector(fabDidTap:) forControlEvents:UIControlEventTouchUpInside];
  fab.center = _button.center;

  MDCFeatureHighlightViewController *vc = [[MDCFeatureHighlightViewController alloc] initWithHighlightedView:_button andShowView:fab completion:nil];

  vc.titleText = @"Hey this is a multiline title for the Feature Highlight";
  vc.bodyText = @"This is the description of the feature highlight view controller.";
  [self presentViewController:vc animated:YES completion:nil];
}

- (void)fabDidTap:(id)sender {
  NSLog(@"Fab tapped");
}
```

### Version number
31.0.0&ndash;33.0.0

All iPhone/iPads that were tested

romoore:
Shown views with gesture recognizers or shown buttons with actions
should be able to receive taps while Feature Highlight is active.

Closes #1888

### DIFF
--- a/components/FeatureHighlight/examples/FeatureHighlightShownViewExample.m
+++ b/components/FeatureHighlight/examples/FeatureHighlightShownViewExample.m
@@ -27,25 +27,25 @@
   [fab setImage:[UIImage imageNamed:@"Plus"] forState:UIControlStateNormal];
   [fab sizeToFit];
   fab.backgroundColor = UIColor.orangeColor;
-  [fab addTarget:self action:@selector(fabDidTap:) forControlEvents:UIControlEventTouchUpInside];
   fab.center = _button.center;
 
   MDCFeatureHighlightViewController *vc =
       [[MDCFeatureHighlightViewController alloc] initWithHighlightedView:_button
                                                              andShowView:fab
-                                                              completion:nil];
+                                                              completion:^(BOOL accepted) {
+                                                                if (accepted) {
+                                                                  [self fabDidTap:fab];
+                                                                }
+                                                              }];
 
+  [fab addTarget:vc action:@selector(acceptFeature) forControlEvents:UIControlEventTouchUpInside];
   vc.titleText = @"Shown views can be interactive";
-  vc.bodyText =
-      @"Tapping the button below will dismiss it. Tap again to dismiss Feature Highlight.";
+  vc.bodyText = @"The shown button has custom tap animations.";
   [self presentViewController:vc animated:YES completion:nil];
 }
 
 - (void)fabDidTap:(MDCFloatingButton *)sender {
-  [sender collapse:YES
-        completion:^(void) {
-          [sender removeFromSuperview];
-        }];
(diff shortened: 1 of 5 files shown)


garrettmurray (approved):

samnm:
I chatted briefly with Rob about this offline, but its worth mentioning Feature Highlight is intended to be used by providing a completion block. The completion block specifies whether the user accepted or dismissed the highlight. I think it's worth landing the functional change here so that buttons still can have a pressed state, but clients shouldn't be listening to touches on the button as a way to dismiss the feature highlight.

garrettmurray:
@samnm Just a note: Without the change to the gesture recognizer, the completion block doesn't fire. The gesture rec is stealing the taps and not passing them through. But still noted that we should be using the completion block in general, so thanks.

components/FeatureHighlight/examples/FeatureHighlightShownViewExample.m
```diff
@@ -0,0 +1,51 @@
...
+  MDCFeatureHighlightViewController *vc =
+      [[MDCFeatureHighlightViewController alloc] initWithHighlightedView:_button
+                                                             andShowView:fab
+                                                              completion:^(BOOL accepted) {
+                                                                if (accepted) {
+                                                                  [self fabDidTap:fab];
+                                                                }
+                                                              }];
+
+  [fab addTarget:vc action:@selector(acceptFeature) forControlEvents:UIControlEventTouchUpInside];
```
samnm: Since we're adding `panRecognizer.cancelsTouchesInView = NO;`, we should also be listening for this control event inside MDCFeatureHighlightView. Clients shouldn't have to do anything extra like this to get Feature Highlight to work.

samnm (changes requested):
@garrettmurray ah yeah that is definitely a bug then.

Rob, if we set `cancelsTouchesInView = NO` and bake in the shownView TouchUpInside listening then we should be good.

samnm (approved):

romoore:
OK to merge once Travis gets off its duff.

PR #1890, opened 2017-08-24
[Activity Indicator] Added swift example

codeman7:
[Activity Indicator] Added swift example

I would like to go through and add examples for all components that are currently missing examples.  This is not currently an issue so it doesn't close any issues.


codeman7:
I signed it!

ianegordon:
Hi @codeman7 and welcome!

Thank you for the pull request. Having Swift samples for all of the components would be great.  If you are so inclined, feel free to open issues along those lines:
[ActivityIndicator] Swift example needed

But, we're definitely happy to see PRs either way!

I'll take a look this afternoon.

Many thanks!

codeman7:
@ianegordon Thank you for the warm welcome.  I will open issues for them as I go along is that ok? I just wanted to do one to make sure you guys were ok with it.

components/ActivityIndicator/examples/ActivityIndicatorExample.swift
```diff
@@ -0,0 +1,85 @@
+/*
+ Copyright 2016-present the Material Components for iOS authors. All Rights Reserved.
```
ianegordon: s/2016/2017/

components/ActivityIndicator/examples/ActivityIndicatorExample.swift
```diff
@@ -0,0 +1,85 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+
+import UIKit
+import MaterialComponents
+
+struct MDPalette {
+   static let blue: UIColor = UIColor(colorLiteralRed: 0.129, green: 0.588, blue: 0.953, alpha: 1.0)
```
ianegordon: This is throwing a compiler warning.
UIColor(red:green:blue:alpha:) seems to fix the problem.

components/ActivityIndicator/examples/ActivityIndicatorExample.swift
```diff
@@ -0,0 +1,85 @@
...
+   static let red: UIColor = UIColor(colorLiteralRed: 0.957, green: 0.263, blue: 0.212, alpha: 1.0)
+   static let green: UIColor = UIColor(colorLiteralRed: 0.298, green: 0.686, blue: 0.314, alpha: 1.0)
+   static let yellow: UIColor = UIColor(colorLiteralRed: 1.0, green: 0.922, blue: 0.231, alpha: 1.0)
+}
+
+class ViewController: UIViewController {
```
ianegordon: Our naming tend to be verbose here.

Consider:
ActivityIndicatorSwiftController

components/ActivityIndicator/examples/ActivityIndicatorExample.swift
```diff
@@ -0,0 +1,85 @@
...
+   }
+   
+   
+}
+
+extension ViewController : MDCActivityIndicatorDelegate {
+   func activityIndicatorAnimationDidFinish(_ activityIndicator: MDCActivityIndicator) {
+      return
+   }
+}
```
ianegordon: We need to implement catalogBreadcrumbs so the view controller will show up in the catalog.

If you add the following, your view controller should be visible under the catalog activity indicator tile.

// MARK: Catalog by convention
extension ViewController {
  class func catalogBreadcrumbs() -> [String] {
    return ["Activity Indicator", "Activity Indicator (Swift)"]
  }
}

ianegordon (changes requested):
Some minor things, but we are pretty close.

Many thanks!

components/ActivityIndicator/examples/ActivityIndicatorExample.swift
codeman7: I am confused here as to what you would like me to change???

ianegordon: Apologies, that wasn't very explicit.
This is a new example, and I expect you copy & pasted the license from another file.
That usually works, with the exception that we must set the copyright year to the year the file was created.
So in this case it should be 2017 instead of 2016.

codeman7: Yeah sorry for the confusion. If I have questions about the roadmap or design strategies where would be a good place to discuss that, discord app?

ianegordon (changes requested):
It looks like the Copyright date is the last piece before we can merge.
👍

components/ActivityIndicator/examples/ActivityIndicatorExample.swift
ianegordon: Feel free to open an issue with any questions.  That can be useful if other people may also be wondering.
Someone is usually on Dischord as well and that may be a good spot for the easier ones.

Roadmap and design sound like bigger questions :)  But, we can always chat in Dischord and document the useful pieces.

ianegordon (approved):

ianegordon:
PR merged, thank you for becoming a contributor @codeman7 ! 😀

PR #1894, opened 2017-08-24
[TextFields] Improving hit zone for multiline

willlarche:


components/TextFields/src/MDCMultilineTextField.m
```diff
@@ -400,6 +400,19 @@ - (CGFloat)estimatedTextViewLineHeight {
...
 
+#pragma mark - Touch (UIView)
+
+- (UIView *)hitTest:(CGPoint)point withEvent:(UIEvent *)event {
+  UIView *receiver = [super hitTest:point withEvent:event];
+
+  // We do this so the hit zone to become first responder is larger than just the small text view.
+  if (receiver == self && self.isEditing == NO) {
+    return self.textView;
+  }
```
ianegordon, line 411: Does this have any impact on accessibility when VoiceOver mode is on?
Do we want to set a custom accessibility frame if so?

ianegordon (changes requested):

components/TextFields/src/MDCMultilineTextField.m
willlarche: This does not affect existing VO but VO isn't ideal on multiline. Should be improved in separate issue.

willlarche: #1895

ianegordon (approved):

ianegordon:
LGTM.  Please merge when Travis is happy.

PR #1900, opened 2017-08-25
[TextFields] Changing insets to react to new property

willlarche:


components/TextFields/src/private/MDCTextInputCommonFundament.m
```diff
@@ -665,7 +680,46 @@ - (void)setTextInput:(UIView<MDCTextInput> *)textInput {
 }
 
 - (UIEdgeInsets)textInsets {
-  return _textInput.textInsets;
+  UIEdgeInsets textInsets = UIEdgeInsetsZero;
+
+  textInsets.top = MDCTextInputFullPadding;
+
+  CGFloat leadingOffset = MDCCeil(self.leadingUnderlineLabel.font.lineHeight * 2.f) / 2.f;
```
ianegordon, line 687: Consider adding a comment indicating why the leading and trailing offsets are linked to lineHeight.
I'm not caffeinated yet, but do we need to ceil(x * 2) / 2?  Can we not just ceil(x) ?

ianegordon (approved):
Please address comments before merge.

components/TextFields/src/private/MDCTextInputCommonFundament.m
romoore: Looks like rounding code that permits half-points?

willlarche: Yes, @romoore . This is a quick estimation of uilabel height based on font line height. I measure a bunch and saw the pattern that it was ceiling to .5's. 

Would be fabulous if someone could find out for SURE what the formula is and put it in typography or something.

willlarche: @ianegordon , these are the estimated measurements of the height that will be needed for the leading and trailing underline labels. So it's not leading or trailing space. It's leading and trailing underline labels' heights.

willlarche: Improving estimation in a new PR because conflicts.

PR #1901, opened 2017-08-25
[TextFields] Example updates due to text inset changes.

willlarche:
Dependent on #1899 and #1900

### DIFF
--- a/components/TextFields/examples/TextFieldKitchenSinkExample.swift
+++ b/components/TextFields/examples/TextFieldKitchenSinkExample.swift
@@ -397,6 +397,20 @@ final class TextFieldKitchenSinkSwiftExample: UIViewController {
   }
 
   // MARK: - Multiline
+
+  func setupAreaTextFields() -> [MDCTextInputControllerTextArea] {
+    let textFieldArea = MDCMultilineTextField()
+    scrollView.addSubview(textFieldArea)
+    textFieldArea.translatesAutoresizingMaskIntoConstraints = false
+
+    textFieldArea.textView?.delegate = self
+    textFieldArea.placeholder = "This is a text area"
+
+    let textFieldControllerArea = MDCTextInputControllerTextArea(textInput: textFieldArea)
+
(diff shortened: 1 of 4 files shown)


components/TextFields/examples/TextFieldKitchenSinkExample.swift
```diff
@@ -96,16 +97,37 @@ final class TextFieldKitchenSinkSwiftExample: UIViewController {
     setupExampleViews()
   }
 
-  func setupDefaultTextFields() -> [MDCTextInputControllerLegacyDefault] {
+  func setupBoxTextFields() -> [MDCTextInputControllerTextFieldBox] {
+    let textFieldBox = MDCTextField()
+    scrollView.addSubview(textFieldBox)
+    textFieldBox.translatesAutoresizingMaskIntoConstraints = false
```
ianegordon: When we addSubview a setNeedsLayout is implicitly invoked. 
Does the state of .translatesAutoresizingMaskIntoConstraints matter at this point or does it not matter until the layout() actually happens?  To put it another way, will constraints get created, then disabled after the .translatesAutoresizingMaskIntoConstraints is set to false?
I typically set .translatesAutoresizingMaskIntoConstraints = false before addSubview just in case, but I have no evidence that I should be.  Curious if you knew either way.

ianegordon (approved):
LMK if you know the answer to the question inline.
I'll let you merge in case you want to get the earlier PR landed and rerun Travis.

components/TextFields/examples/TextFieldKitchenSinkExample.swift
willlarche: I have never checked! Would be really good to know.

willlarche: I poked around and this use does not create constraints yet. However, I think it makes sense. What if a layout pass happened when we didn't want it yet? Would be good best practice to make sure you turn tamic on or off before adding as a subview.

PR #1911, opened 2017-08-25
[Catalog] Verify all examples provide CatalogByConvention methods

### LINKED ISSUE #1897: [Catalog] All example view controllers should implement Catalog By Convention
Some of our examples do not implement:
catalogBreadcrumbs, catalogShouldHideNavigation, etc.

Every example should implement these methods.

### Required
```objectivec
+ (nonnull NSArray<NSString *> *)catalogBreadcrumbs;
+ (BOOL)catalogIsPrimaryDemo;
```

### Optional
```objectivec
- (nonnull NSString *)catalogStoryboardName;
- (nonnull NSString *)catalogDescription;
```

## Informal Protocol (MDCNodeListViewController)
```objectivec
- (BOOL)catalogShouldHideNavigation;
```

romoore:
CatalogByConvention classes require two methods to be implemented:

```objectivec
+ (nonnull NSArray<NSString *> *)catalogBreadcrumbs;
+ (BOOL)catalogIsPrimaryDemo;
```

Closes #1897

### DIFF
--- a/catalog/MDCCatalog/MDCNodeListViewController.swift
+++ b/catalog/MDCCatalog/MDCNodeListViewController.swift
@@ -355,7 +355,15 @@ extension MDCNodeListViewController {
     var vc: UIViewController
     if node.isExample() {
       let contentVC = node.createExampleViewController()
-      if contentVC.responds(to: NSSelectorFromString("catalogShouldHideNavigation")) {
+      var shouldHideNavigation = contentVC.responds(to: NSSelectorFromString("catalogShouldHideNavigation"))
+      if (shouldHideNavigation) {
+        if let _ = contentVC.perform(NSSelectorFromString("catalogShouldHideNavigation")) {
+          shouldHideNavigation = true
+        } else {
+          shouldHideNavigation = false
+        }
+      }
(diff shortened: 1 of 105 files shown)


ianegordon:
This is likely a question for @jverkoey as well, but does it make sense to make isPrimaryDemo optional?  Right now it default to 'NO', and that *seems* to be working as intended.

romoore:
I think the biggest challenge with our current approach is that it's hard to understand how to do it correctly.  Correct behavior is defined by a protocol that no class should declare implementing, it works if you do it wrong, and some of the behavior is undocumented.  My suggestion is to make the protocol more formal, and fail if you forget something required.  This makes it easier for contributors to learn how to do it correctly and avoid mistakes.

romoore:
I found a situation where NO example was declared primary and another where two examples were declared primary.  In both cases, the runtime just picked one arbitrarily and went with that.

jverkoey (approved):
This looks like a good iteration.

In a separate change it may be worth considering alternative ways to construct the catalog's metadata that resolve some of these implicit patterns that aren't necessarily obvious to a newcomer.

PR #1915, opened 2017-08-26
[TextFields] Making a test pass on 3x simulator.

willlarche:


components/TextFields/tests/unit/TextFieldControllerLegacyTests.swift
```diff
@@ -246,7 +246,7 @@ class TextFieldControllerDefaultLegacyTests: XCTestCase {
 
     controller.helperText = "Helper"
     textField.sizeToFit()
-    XCTAssertEqual(textField.frame.height, 84.5)
+    XCTAssertTrue(MDCCGFloatEqual(MDCFloor(textField.frame.height), 84.0))
```
romoore: This test is failing on iPad 2 simulator when I run the entire Unit Test suite, ~~but passes when I run it individually~~. I'm a little concerned that there is some interaction between the tests.  

Update: The test doesn't actually run when I try to run it individually in Xcode. 😕 

On iPad 2, the height is 85.0 instead of 84.0.

romoore (changes requested):

components/TextFields/tests/unit/TextFieldControllerLegacyTests.swift
willlarche: Let's try this version.

romoore (approved):
LGTM. Passed on all devices I tried.

PR #1926, opened 2017-08-28
[Catalog] Remove unused sample option

ianegordon:
Outlined textfields always have the floating title label.
This removes a selection interface that has no effect.


romoore (approved):
LGTM

PR #1928, opened 2017-08-29
[ActivityIndicator] - Reduce CPU demands of MDCActivityIndicator

ShepJGoogle:
- Adds a dispatch_after buffer in completion block addStrokeRotationCycle method before calling strokeRotationCycleFInishedFromState: to alleviate CPU intensity

See thread 1
Before:
before

After:
after

### DIFF
--- a/components/ActivityIndicator/src/MDCActivityIndicator.m
+++ b/components/ActivityIndicator/src/MDCActivityIndicator.m
@@ -432,8 +432,13 @@ static const CGFloat kSingleCycleRotation =
 
   [CATransaction begin];
   {
+    __weak typeof(self) weakSelf = self;
     [CATransaction setCompletionBlock:^{
-      [self strokeRotationCycleFinishedFromState:MDCActivityIndicatorStateIndeterminate];
+      dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.03 * NSEC_PER_SEC)),
+                     dispatch_get_main_queue(), ^{
+          MDCActivityIndicator *strongSelf = weakSelf;
+          [strongSelf strokeRotationCycleFinishedFromState:MDCActivityIndicatorStateIndeterminate];
+                     });
     }];
 
     // Outer 5-point star detent rotation.


jgunaratne (approved):

ianegordon:
@romoore I believe you might find this interesting.

romoore:
Yes, we've been unable to reproduce in MDC Catalog.  It may be a weird interaction with the client code.

components/ActivityIndicator/src/MDCActivityIndicator.m
```diff
@@ -432,8 +432,13 @@ - (void)addStrokeRotationCycle {
...
   [CATransaction begin];
   {
+    __weak typeof(self) weakSelf = self;
     [CATransaction setCompletionBlock:^{
-      [self strokeRotationCycleFinishedFromState:MDCActivityIndicatorStateIndeterminate];
+      dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.03 * NSEC_PER_SEC)),
+                     dispatch_get_main_queue(), ^{
+          MDCActivityIndicator *strongSelf = weakSelf;
+          [strongSelf strokeRotationCycleFinishedFromState:MDCActivityIndicatorStateIndeterminate];
+                     });
```
ianegordon, line 441: I'm not sure I follow this change.
Let's catch up offline.

ianegordon (dismissed):
To be discussed offline.

components/ActivityIndicator/src/MDCActivityIndicator.m
romoore: I'd also like to see if we can determine why the client app has increased CPU but not the MDC Catalog.

randallli: Please report back here after offline conversation to unblock.

romoore (commented):
Without knowing that the root cause is actually in MDCActivityIndicator, I'm having a hard time supporting dropping from 60hz (or 120hz) to 30hz.  When this was tested on MDCCatalog, I don't believe the same high CPU utilization was observed as it was in the client application.

randallli:
original issue b/64975595

ShepJGoogle:
I'm going to close this pr. Outside of photos, the activity indicator doesn't pose any similar numbers to warrant the change.

PR #1930, opened 2017-08-30
[ActivityIndicator] Use explicit class for bundle

brianjmoore:
Using [self class] to determine the bundle to load resources/strings here breaks if you subclass MDCActivityIndicator and that subclass ends up being placed in a different bundle. Writing MDCActivityIndicator here explicitly ensures that we find the intended resources: those included alongside the base definition.


romoore (approved):
LGTM

ShepJGoogle (approved):

PR #1936, opened 2017-09-01
[ButtonBar] - Update setting of insets to use size classes and device type

ShepJGoogle:
Update setting of insets to use size classes and device type


samnm (approved):

components/ButtonBar/src/private/MDCAppBarButtonBarBuilder.m
```diff
@@ -167,7 +178,7 @@ + (UIEdgeInsets)contentInsetsForButton:(MDCButton *)button
     return sum;
   };
 
-  const BOOL isPad = userInterfaceIdiom == UIUserInterfaceIdiomPad;
```
romoore, line 170: Nit: I think you can leave this as const, I don't see anywhere that you change it.

romoore (approved):
Just a small nit

PR #1937, opened 2017-09-01
[AnimationTiming | ActivityIndicator] Add swift examples

codeman7:
Added Swift examples for both AnimationTiming and ActivityIndicator. For ActivityInicator just updates some things to match styles recommended.


codeman7:
@romoore Is it looking better now?

romoore:
I still see there appear to be conflicts; getting Git out of a situation like this can be frustratingly difficult. 😞   I think the easiest might be the following:
1. Make a copy of the two files somewhere outside your normal working directory (like ~/Desktop)
1. Reset your `develop` branch locally and on GH so that it mirrors MDC. [Here's one way to do it](https://stackoverflow.com/questions/9646167/clean-up-a-fork-and-restart-it-from-the-upstream), there are many.
1. Branch from the now "clean" `develop` and then copy the files back to your local repo
1. Add the files as a new commit and push up for a PR.

I think once you get your `develop` back to a mirror of MDC things will be a lot simpler going forward.

codeman7:
Is there a way I can see on my end that I have resolved the issue or where can I see the issue?

PR #1938, opened 2017-09-01
[FeatureHighlight] Fix broken properties

samnm:
Feature Highlight previously loaded the view at view controller instantiation time. With #1883 FH now only loads the view in loadView. That change broke the property routing through the view controller to the view, as generally the view is not instantiated when people are setting those properties. This change fixes that.


romoore (approved):
Unit test to catch an accidentally-nil property?

PR #1896, opened 2017-08-24
[Buttons] Deprecate subclassing to set cornerRadius and contentEdgeInsets

samnm:
Previously cornerRadius and contentEdgeInsets were only configurable via subclassing MDCButton. This PR deprecates that behavior in favor of setting the properties explicitly.

PR #1948, opened 2017-09-05
[Buttons] Revert Deprecate subclassing to set cornerRadius and contentEdgeInsets

samnm:
Reverts #1896


ianegordon:
Alert buttons are working as expected on iPhone 4s Sim @ 8.1.

ianegordon (approved):
We can pull in the required dialog changes and roll forward.
Please wait for Travis / test locally before merge.

PR #1949, opened 2017-09-05
[Catalog] Remove local color overrides

romoore:
The MDC Catalog should demonstrate what our components look like in their default state without any local overrides.  It should also not use any default theme, but support theming via the Themes section of the catalog.

PR #1958, opened 2017-09-08
[Catalog] Remove local color overrides

romoore:
Removing example code that overrides the theme/default appearance of our
components except where values are needed as part of the example.


randallli:
this appears to have the same PR title as https://github.com/material-components/material-components-ios/pull/1949 can we rename/drop one?

romoore:
Closed #1949 - that one is a superset of this one.  @ajsecord  and @samnm have requested that the changes be made in stages.

randallli:
ok cool, once you remove the planningChanges label then we know its read for review.

PR #1963, opened 2017-09-11
[TextFields] Fix for placeholders being colored active when isEditing = NO

willlarche:
Solution is to define the floating placeholder color parameter to be for 'normal' state.


components/TextFields/src/MDCTextInputControllerDefault.m
```diff
@@ -503,14 +500,19 @@ - (void)updatePlaceholder {
...
-            ? self.errorColor
-            : self.floatingPlaceholderColor;
+            ? self.errorColor : nonErrorColor;
   } else {
-    self.textInput.placeholderLabel.textColor = self.inlinePlaceholderColor;
+    placeholderColor = self.textInput.isEditing ? self.activeColor :
+        self.inlinePlaceholderColor;
   }
+  self.textInput.placeholderLabel.textColor = placeholderColor;
+  NSLog(@"%@", placeholderColor);
```
jgunaratne: Remove logging?

jgunaratne (approved):

components/TextFields/src/MDCTextInputControllerDefault.m
willlarche: Oops! Thank you.

PR #1982, opened 2017-09-14
[Tabs] TabBarViewController should respect safeAreaInsets

### LINKED ISSUE #1977: [Tabs] MDCTabBarViewController should adjust the TabBar position based on the safeAreaInsets
### Overview
With the iPhone X affordance for the Home Screen indicator, the MDCTabBarViewController should reposition the TabBar above the Home Screen indicator.

**iPhone X**
simulator screen shot - iphone x - 2017-09-14 at 00 13 32

**iPhone SE**
simulator screen shot - iphone se - 2017-09-14 at 00 14 35

### Reproduction steps
1. Open MDC Catalog on iPhone X &rarr; Tab Bar &rarr; TabBarViewController
2. Make sure the Tab Bar items in the Bar at the bottom are fully visible and not overlapping the Home Screen indicator

### Version number
34.0.2

iPhone X/iOS 11 (Simulator)

### Related issues
#1974 

---
#### Internal data
- Associated internal bug: [b/117179383](http://b/117179383)

romoore:
Starting with iOS 11 (primarily due to iPhone X), we need to respect the safeAreaInsets values when positioning elements directly on the edges of the screen. 

* Position the embedded MDCTabBar within the safeArea of the MDCTabBarViewController's view
* Set the backgroundColor of the MDCTabBarViewController to match the barTintColor of the MDCTabBar
* Position the MDCTabBar in `-viewWillLayoutSubviews` rather than in `-viewDidLayoutSubviews` so that subviews can be positioned relative to the MDCTabBar.
* Set `-additionalSafeAreaInsets` on any embedded child View Controllers based on [Apple's documentation for positioning content](https://developer.apple.com/documentation/uikit/uiview/positioning_content_relative_to_the_safe_area)

Closes #1977

### DIFF
--- a/components/Tabs/src/MDCTabBarViewController.m
+++ b/components/Tabs/src/MDCTabBarViewController.m
@@ -47,7 +47,7 @@ const CGFloat MDCTabBarViewControllerAnimationDuration = 0.3f;
   tabBar.alignment = MDCTabBarAlignmentJustified;
   tabBar.delegate = self;
   self.tabBar = tabBar;
-  _tabBarWrapperView = [[UIView alloc] init];
+  _tabBarWrapperView = [[UIView alloc] initWithFrame:CGRectZero];
   [_tabBarWrapperView addSubview:tabBar];
   [self.view addSubview:_tabBarWrapperView];
   [self updateTabBarItems];
@@ -207,7 +207,7 @@ const CGFloat MDCTabBarViewControllerAnimationDuration = 0.3f;
     [super viewSafeAreaInsetsDidChange];
   }
 #endif
-  [self updateLayout];
+  [self.view setNeedsLayout];
 }
 
 - (void)updateLayout {
--- a/components/Tabs/src/MDCTabBarViewController.m
+++ b/components/Tabs/src/MDCTabBarViewController.m
@@ -202,7 +202,7 @@ const CGFloat MDCTabBarViewControllerAnimationDuration = 0.3f;
 
 - (void)viewSafeAreaInsetsDidChange {
   // TODO(rsmoore): Remove check when we drop Xcode 7/8 support
-#ifdef __IPHONE_11_0
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
   if (@available(iOS 11.0, *)) {
     [super viewSafeAreaInsetsDidChange];
   }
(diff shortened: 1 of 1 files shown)


brianjmoore:
Thanks for this! I wonder if we should consider extending the tab bar itself to the bottom with extra insets rather than wrapping it with another view so ink effects wouldn't be clipped. From some brief preliminary research, UITabBarController seems to take this approach with its UITabBar.

romoore:
@brianjmoore I agree it would be better if MDCTabBar supported more flexible layouts. From my initial research, MDCTabBar tries to position its cells within its entire bounds. Specifically, by making the MDCTabBar taller, it was offsetting each of the cells vertically by half of the additional space. 
 
Adding a wrapping view to the VC (or alternatively a view below the TabBar) seems to be the simplest solution to the problem right now.  If we refactor/improve MDCTabBar in the future, we can remove the wrapper from MDCTabBarViewController and there shouldn't be any impact on clients.

components/Tabs/src/MDCTabBarViewController.m
```diff
@@ -41,7 +47,9 @@ - (void)viewDidLoad {
   tabBar.alignment = MDCTabBarAlignmentJustified;
   tabBar.delegate = self;
   self.tabBar = tabBar;
-  [view addSubview:tabBar];
+  _tabBarWrapperView = [[UIView alloc] init];
```
ianegordon: When possible use designated initializers:
[[UIView alloc] initWithFrame:CGRectZero]

components/Tabs/src/MDCTabBarViewController.m
```diff
@@ -192,17 +200,43 @@ - (nullable UIViewController *)controllerWithView:(nullable UIView *)view {
...
 }
 
+- (void)viewSafeAreaInsetsDidChange {
+  // TODO(rsmoore): Remove check when we drop Xcode 7/8 support
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    [super viewSafeAreaInsetsDidChange];
+  }
+#endif
+  [self updateLayout];
```
ianegordon: Do we want to updateLayout or setNeedsLayout?
Does [UIView viewSafeAreaInsetsDidChange] not already call setNeedsLayout?

ianegordon (changes requested):
Hi @romoore thanks for investigating.
safeAreaInsets affects many of our components. 
Why don't we figure out a strategy that will work for all of them before we make one-off changes?

components/Tabs/src/MDCTabBarViewController.m
romoore: `setNeedsLayout` makes sense. This gets called just before `viewDidLayoutSubviews` when we load/rotate.

components/Tabs/src/MDCTabBarViewController.m
romoore: Done

ianegordon:
Thanks Rob!  Your work here should get pulled into #2261.

PR #1986, opened 2017-09-14
Workaround -Wunused-parameter, since it can't handle usage inside of asserts.

ajsecord:
-Wunused-parameter is enabled by -Wextra, but can't handle cases like

```
- (void)doSomething:(int)foo {
  NSAssert(foo < 42);
  ...
}
```

It will work in Debug but not Release.


ianegordon:
If we are only using the parameter as an Assert sanity check, perhaps we don't need the parameter at all.
If someone is invoking a method or changing state inside of an assert, we likely want to call that behavior out, which this warning does.
How does this fit with people implementing delegate methods or dealing with inheritance?
In that case can they use the __unused annotation to silence the warning / error?

romoore:
Marking the parameter as `__unused` would give a warning that you're using a parameter marked unused. We could instead disable that warning `-Wused-but-marked-unused`.  I would agree in general that we should not put asserts on a parameter that isn't used regardless.

romoore (changes requested):
I would prefer marking it as `__unused` and disabling the `-Wused-but-marked-unused` warning instead.  I think the `-Wunused-parameter` has merit in requiring developers to explicitly acknowledge parameters.

ianegordon:
I like Rob's proposal as we can add -Wno-used-but-marked-unused only for DEBUG builds and continue to get strict checking on RELEASE.  Why don't we discuss tomorrow in person?

ajsecord:
We don't have a mechanism for specifying different warnings in Debug/Release modes, and we shouldn't spend the time to do so at the moment. I'll add a hack into the problematic file for now, but if this fires again anywhere else we should revisit.

ajsecord:
PTAL

romoore (approved):
Do we know why there is a parameter assert if the parameter isn't used?

ajsecord:
Either future expansion or a subclassing API. In any case, can we move this forward?

ianegordon (approved):
Updated tracking issue.  
We should revisit when we have a moment.
LGTM, please merge once Travis is happy.

ianegordon:
Travis is happy, merging.

Issue #1987, opened 2017-09-14
Components should respect the device's safe area


ianegordon:
The iPhone X' screen has a cut out on the top for sensors and rounded corners.

Material components should respect a device's safe area.

The height of the status bar will also be variable.

ianegordon:
Apple's HIG - iPhone X
https://developer.apple.com/ios/human-interface-guidelines/overview/iphone-x/

ianegordon:
Consider back-porting UIView safeAreaLayoutGuide to use across our components.
https://developer.apple.com/documentation/uikit/uiview/2891102-safearealayoutguide?language=objc

ianegordon:
If we choose to backport, we will likely need to handle the entire API:
UILayoutGuide safeAreaLayoutGuide
UIEdgeInset safeAreaInsets
-[UIView safeAreaInsetsDidChange]

UILayoutGuide was introduced in iOS 9.

romoore:
Also [[UIViewController viewSafeAreaInsetsDidChange]](https://developer.apple.com/documentation/uikit/uiviewcontroller/2891116-viewsafeareainsetsdidchange)

gambard:
FYI, it seems that the applications compiled with an iOS < 11 sdk have black strips at the top/bottom of iPhone X.

chuga:
Marking this as closed as we have issues for each component.

PR #1988, opened 2017-09-15
[Collections] Update collections to support iPhone X

### LINKED ISSUE #1980: [Collections] Collections does not respect the safe area on landscape on iPhone X
### Overview
On iPhone X, the collection with MDC Card style is not respecting the safe area, hiding some content.
simulator screen shot - iphone x - 2017-09-14 at 09 42 31

### Reproduction steps
Launch MDC Catalog on iPhone X → Collections → Demo

### Version number
34.0.01

iPhone X/iOS 11 (Simulator)

gambard:
The collections use the contentInset to compute the width of the cells.
With iPhone X, the width must take into account possible inset of the side.
This patch makes sure that the collection takes into account the adjustedContentInset, which may take the safe area into consideration.

Closes #1980

### DIFF
--- a/components/Collections/examples/CollectionsALaCarteExample.m
+++ b/components/Collections/examples/CollectionsALaCarteExample.m
@@ -63,7 +63,7 @@ static NSString *const kReusableIdentifierItem = @"itemCellIdentifier";
 
 #if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
   if (@available(iOS 11.0, *)) {
-    self.collectionView.contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentAlways;
+    _customCollectionView.contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentAlways;
   }
 #endif
 
--- a/components/Collections/examples/CollectionsALaCarteExample.m
+++ b/components/Collections/examples/CollectionsALaCarteExample.m
@@ -61,6 +61,12 @@ static NSString *const kReusableIdentifierItem = @"itemCellIdentifier";
                                                  collectionViewLayout:self.collectionViewLayout];
   self.collectionView = _customCollectionView;
 
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    self.collectionView.contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentAlways;
+  }
+#endif
+
   // Register cell class.
   [self.collectionView registerClass:[MDCCollectionViewTextCell class]
           forCellWithReuseIdentifier:kReusableIdentifierItem];
--- a/components/Collections/examples/CollectionsSimpleSwiftDemo.swift
+++ b/components/Collections/examples/CollectionsSimpleSwiftDemo.swift
@@ -26,6 +26,10 @@ class CollectionsSimpleSwiftDemo: MDCCollectionViewController {
   override func viewDidLoad() {
     super.viewDidLoad()
 
+    if #available(iOS 11.0, *) {
+      self.collectionView?.contentInsetAdjustmentBehavior = .always
+    }
+
     // Register cell class.
     self.collectionView?.register(MDCCollectionViewTextCell.self,
                                        forCellWithReuseIdentifier: reusableIdentifierItem)
--- a/components/Collections/examples/CollectionsSimpleSwiftDemo.swift
+++ b/components/Collections/examples/CollectionsSimpleSwiftDemo.swift
@@ -26,12 +26,6 @@ class CollectionsSimpleSwiftDemo: MDCCollectionViewController {
   override func viewDidLoad() {
     super.viewDidLoad()
 
-#if swift(>=3.2)
-    if #available(iOS 11.0, *) {
-      self.collectionView?.contentInsetAdjustmentBehavior = .always
-    }
-#endif
-
     // Register cell class.
     self.collectionView?.register(MDCCollectionViewTextCell.self,
                                        forCellWithReuseIdentifier: reusableIdentifierItem)
--- a/components/Collections/examples/CollectionsSimpleSwiftDemo.swift
+++ b/components/Collections/examples/CollectionsSimpleSwiftDemo.swift
@@ -26,9 +26,11 @@ class CollectionsSimpleSwiftDemo: MDCCollectionViewController {
   override func viewDidLoad() {
     super.viewDidLoad()
 
+#if swift(>=3.2)
     if #available(iOS 11.0, *) {
       self.collectionView?.contentInsetAdjustmentBehavior = .always
     }
+#endif
 
     // Register cell class.
     self.collectionView?.register(MDCCollectionViewTextCell.self,
--- a/components/Collections/examples/CollectionsStoryboardExample.m
+++ b/components/Collections/examples/CollectionsStoryboardExample.m
@@ -32,12 +32,6 @@ static NSString *const kReusableIdentifierItem = @"customCell";
   [super viewDidLoad];
   self.title = @"Storyboard Demo";
 
-#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
-  if (@available(iOS 11.0, *)) {
-    self.collectionView.contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentAlways;
-  }
-#endif
-
   // Populate content.
   _content = [NSMutableArray array];
   for (NSInteger i = 0; i < kSectionCount; i++) {
--- a/components/Collections/examples/CollectionsStoryboardExample.m
+++ b/components/Collections/examples/CollectionsStoryboardExample.m
@@ -32,6 +32,12 @@ static NSString *const kReusableIdentifierItem = @"customCell";
   [super viewDidLoad];
   self.title = @"Storyboard Demo";
 
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    self.collectionView.contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentAlways;
+  }
+#endif
+
   // Populate content.
   _content = [NSMutableArray array];
   for (NSInteger i = 0; i < kSectionCount; i++) {
--- a/components/Collections/src/MDCCollectionViewController.m
+++ b/components/Collections/src/MDCCollectionViewController.m
@@ -214,17 +214,12 @@ NSString *const MDCCollectionInfoBarKindFooter = @"MDCCollectionInfoBarKindFoote
 - (CGFloat)cellWidthAtSectionIndex:(NSInteger)section
                     collectionView:(UICollectionView *)collectionView {
   UIEdgeInsets contentInset = collectionView.contentInset;
-  // On the iPhone X, we need to offset
-  if ([self.view respondsToSelector:@selector(adjustedContentInset)]) {
-    NSMethodSignature *adjustedContentInsetSignature =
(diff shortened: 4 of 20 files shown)


components/Collections/examples/CollectionsALaCarteExample.m
```diff
@@ -61,6 +61,12 @@ - (void)viewDidLoad {
                                                  collectionViewLayout:self.collectionViewLayout];
   self.collectionView = _customCollectionView;
 
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    self.collectionView.contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentAlways;
```
gambard: I think it is better to have it on every class and not override this behavior in MDCCollectionViewController, because UICollectionViewController does not.

components/Collections/examples/CollectionsSimpleSwiftDemo.swift
```diff
@@ -26,6 +26,10 @@ class CollectionsSimpleSwiftDemo: MDCCollectionViewController {
   override func viewDidLoad() {
     super.viewDidLoad()
 
+    if #available(iOS 11.0, *) {
```
gambard: I don't know if this will compile on older XCode versions.

components/Collections/examples/CollectionsALaCarteExample.m
romoore: Do we need both checks? Xcode 8 and earlier do not have `__IPHONE_11_0` defined in the SDK, and the max allowed is `__IPHONE_10_3` in Xcode 8.3.2. 

Discussed offline: yes, we do. Just in case the SDK being compiled against isn't the latest offered in the Xcode version.

components/Collections/examples/CollectionsSimpleSwiftDemo.swift
codeman7: I know it will on XCode 8.3.3, I don't know if the team has already shifted to XCode 9 or not.

romoore: This was added in swift 2, so it should be OK in Xcode 8. It is worthwhile to test on Xcode 7, but I [believe it is supported](https://developer.apple.com/library/content/documentation/Xcode/Conceptual/WhatsNewXcode-Archive/Articles/xcode_7_0.html).

components/Collections/src/MDCCollectionViewController.m
```diff
@@ -213,8 +213,20 @@ - (CGSize)sizeWithAttribute:(UICollectionViewLayoutAttributes *)attr
 
 - (CGFloat)cellWidthAtSectionIndex:(NSInteger)section
                     collectionView:(UICollectionView *)collectionView {
-  CGFloat bounds =
-      CGRectGetWidth(UIEdgeInsetsInsetRect(collectionView.bounds, collectionView.contentInset));
+  UIEdgeInsets contentInset = collectionView.contentInset;
+  // On the iPhone X, we need to offset
+  if ([self.view respondsToSelector:@selector(safeAreaInsets)]) {
+    NSMethodSignature *adjustedContentInsetSignature =
+        [[UIScrollView class] instanceMethodSignatureForSelector:@selector(safeAreaInsets)];
```
romoore: Shouldn't these two selectors be `@selector(adjustedContentInset)` and checking that `self.collectionView` responds? I believe all phones respond to this selector once they're running iOS 11.

romoore (changes requested):

components/Collections/src/MDCCollectionViewController.m
gambard: That was a mistake.
Done.

components/Collections/src/MDCCollectionViewController.m
```diff
@@ -213,8 +213,20 @@ - (CGSize)sizeWithAttribute:(UICollectionViewLayoutAttributes *)attr
 
 - (CGFloat)cellWidthAtSectionIndex:(NSInteger)section
                     collectionView:(UICollectionView *)collectionView {
-  CGFloat bounds =
-      CGRectGetWidth(UIEdgeInsetsInsetRect(collectionView.bounds, collectionView.contentInset));
+  UIEdgeInsets contentInset = collectionView.contentInset;
+  // On the iPhone X, we need to offset
+  if ([self.view respondsToSelector:@selector(adjustedContentInset)]) {
```
romoore: This should be
```objectivec
if ([collectionView respondsToSelector:@selector(adjustedContentInset)]) {
```

However, I think it may be cleaner to just wrap the whole thing in a macro (for SDK checking) and availability check (for iOS checking).  What do you think?

romoore (changes requested):

components/Collections/examples/CollectionsALaCarteExample.m
gambard: :)

components/Collections/examples/CollectionsSimpleSwiftDemo.swift
gambard: Ok, I am letting it for now.

components/Collections/src/MDCCollectionViewController.m
gambard: Yes, but you used NSInvocation in your fix. Moving to macro + @available as you did.

gambard (commented):
Thanks, PTAL.

romoore (approved):
LGTM. Please test both Xcode 8 and 9, also multiple iOS versions for safety.

ianegordon (changes requested):
Travis is failing on:
components/Collections/examples/CollectionsSimpleSwiftDemo.swift:30

gambard:
It does not compile on swift because there is no way to exclude code from compilation and the contentInsetAdjustmentBehavior property is only define on the iOS 11 sdk.
I can either use perform selector and translating the enum value to its int value (which I dislike), or calling some Objective-C code containing the #ifdef code.
Or if I missed a way to exclude code at compile time in swift, please tell me!

WDYT?

brianjmoore:
A `#if swift(>=3.2)` block could help in this situation? It would allow Xcodes < 9 to ignore the code entirely.

ianegordon:

chuga:
I'll be looking at this today, thanks for your patience!

chuga:
This looks like a step in the right direction. Let's add the `#if swift(>=3.2)` in the swift examples and merge this in.

chuga:
Actually, since the MDCCollectionViewController *is* the owner of the collection view, we should be setting contentInsetAdjustmentBehavior to UIScrollViewContentInsetAdjustmentAlways ourselves. Otherwise our component wouldn't be iOS 11 friendly out of the box, and every single instance of MDCCollectionViewController would need to make this change to support the new safe area insets. Any concerns with making this the default on iOS 11?

gambard:
We should not make it the default because with the "grouped" style you probably don't want this style ?
I would be in favor of doing the same as iOS: automatic let the collection display elements outside of the safe area by default.

chuga:
Yeah, you do have a point about automatic being the default. I'll wait for Ian to weigh in before we merge this in.

Btw, is this the grouped style you're referring to? what's wrong with it?
simulator screen shot - iphone x - 2017-09-29 at 14 37 21

gambard:
Yes it is the style I am referring to. Now, if you also toggle the other setting (groups maybe?) you have something very similar to a table view. UITableView have cells full width with the content of the cell respecting the safe area. I don't know if this is the behavior we want, but it is something worth looking at.

chuga:
Oh, I see what you're saying, and that's something we should fix in this PR. The change to MDCCollectionViewController will probably need to take into account the style before we change the cell width, for the reason you're describing here.

chuga:
I've been playing with a vanilla UICollectionViewController and it looks like `UIScrollViewContentInsetAdjustmentAlways` only adds an extra horizontal inset when the width of the cell is less than the cv's width - (safeAreaInset.left + safeAreaInset.right). Also, the contentView's bounds is always the same as the cell's. So we don't necessarily *need* to go down the route Apple took with UITableViews, but we probably should given how our collection view is expected to look and behave.

I'm leaning towards the following fix:

Instead of changing the actual size of the cells, we should fix MDCCollectionViewController's `|insetsAtSectionIndex:collectionView|` with something like this for the card style:
```
  ...
  if (isCardStyle) {
    insets.left = inset;
    insets.right = inset;
      if (!UIEdgeInsetsEqualToEdgeInsets(self.collectionView.contentInset,
                                         self.collectionView.adjustedContentInset)) {
        insets.left += self.collectionView.adjustedContentInset.left;
        insets.right += self.collectionView.adjustedContentInset.left;
      }
    }
#endif
  }
  ...
```
For the other styles we should adjust the cells `contentView` frame to account for the safe area insets (only if the collection view has been set to `UIScrollViewContentInsetAdjustmentAlways`). This would replicate the new behavior of UITableViews, where cells take the entire width and only the contentView gets adjusted to account for the safe area insets.

Thoughts?

chuga:
Bumping this to the top of your inbox

gambard:
I am currently ooo, so I might be slow to answer, feel free to land changes without my input. 
If I understand correctly your change, the adjusted content inset would not be used for non-card collection, at the collection level. The cells will adjust their content to respect the safe area.

I don't really mind this behavior, but I am concern that if a collection override the insetForSection: method, it would break the card behavior (or the override would need to use the adjusted content inset).

I would prefer to stick with the behavior provided with UIKit. If the user wants to have a table view style, the adjusted content inset behavior should be kept automatic. In the provided example with the different style, I guess we can switch the adjusted behavior at the same time as we switch the style of the cell.

chuga:
The only problem I see with that approach is that the content of the non-card style cells will go under the notch when in landscape. I don't see a way around that that doesn't involve cells adjusting its content to respect the safe area (if there's any). Right?

screen shot 2017-10-02 at 4 48 37 pm

components/Collections/examples/CollectionsStoryboardExample.m
```diff
@@ -32,6 +32,12 @@ - (void)viewDidLoad {
   [super viewDidLoad];
   self.title = @"Storyboard Demo";
 
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    self.collectionView.contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentAlways;
+  }
+#endif
```
ianegordon: Can this be set in the storyboard?

ianegordon (commented):
On iOS 11, it looks like we now require:
self.collectionView.contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentAlways

Is that the case?  Should we update the documentation?  (If so, happy do so in another PR)

Can we set that value in the storyboard?  If so, we should likely include that in our future doc PR.

ianegordon:
I can definitely see the concern about overriding MDCCollectionViewController's collectionView.contentInsetAdjustment behavior to be UIScrollViewContentAdjustmentAlways.

That said, I don't think we can rely on clients to explicitly set it in each case where they use cards.  It should work out of the box whether you are displaying cards or non-card cells.

Is this something we can push to the layout to set the appropriate insets?  MDCInsetLayout -vs- UIFlowLayout?  (In general, perhaps not specifically with our current architecture)

chuga:
Just wanted to give a quick update. I've been looking into this with Ian, and after trying out a few approaches to try to get a UITableViewController-like behavior for the non-cards cells (full width + adjusted contentView inset), we've come to the conclusion that the amount of work and complexity we would need to add to a handful of classes doesn't it feasible in the short term. So we're planning on going forward with what was originally proposed in this PR, but making contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentAlways the default. The reason for that change being that we want to minimize the amount of work needed to get our components to support the iPhone X out of the box, and that this is actually the expected behavior for the vast majority of use cases for this component.

gambard:
Ok, lgtm.
Concerning your comment from yesterday, in my understanding of the behavior of table view, the cell is supposed to go below the notch and the content view of the cell is adjusted to respect the safe area. So you can select the cell by tapping outside of the safe area.

But I think that having the content adjustment behavior set to always by default and letting the user set it to automatic if the wanted behavior is exactly the same as the table view looks reasonable to me.

ianegordon (approved):

Issue #1975, opened 2017-09-13
[Catalog] TabBar Bottom Navigation example cut-off on iPhone X


romoore:
### Overview
It looks like the Home Bar (?) is pushing up the content of the TabBar items in the Bottom Navigation example, but the height of the TabBar itself is not being adjusted to match.

**iPhone X**
![tab-bottom-ipx](https://user-images.githubusercontent.com/1753199/30381033-394fee94-98d6-11e7-820b-5b861cb808ab.png)

**iPhone 8+**
![tab-bottom-ip8](https://user-images.githubusercontent.com/1753199/30381037-3d86f4ee-98d6-11e7-9148-17d83492958f.png)

### Reproduction steps
Open MDC Catalog on an iPhone X &rarr; Tab Bar &rarr; Bottom Navigation

### Version number
34.0.01

### Operating system and device
iPhone X/iOS 11 (Simulator)

PR #1989, opened 2017-09-15
[NavigationBar, TabBar] Hide `safeAreaInsets` from unsupported SDKs

romoore:
Xcode 7 and 8 (and iOS before 11) do not support `safeAreaInsets` and
`safeAreaLayoutGuide` selectors. Using compile-time checks to hide these
from previous SDKs and new iOS 11 runtime checks to prevent them from
being used on older iOS versions.

Related to #1975, #1866

### DIFF
--- a/components/NavigationBar/examples/NavigationBarIconsExample.m
+++ b/components/NavigationBar/examples/NavigationBarIconsExample.m
@@ -77,10 +77,13 @@
   self.navigationItem.rightBarButtonItem = trailingButtonItem;
   self.navigationItem.backBarButtonItem = backButtonItem;
 
-  if ([self.view respondsToSelector:@selector(safeAreaLayoutGuide)]) {
-    UILayoutGuide *layoutGuide = [self.view performSelector:@selector(safeAreaLayoutGuide)];
+  // TODO(rsmoore): Remove this check when we drop support for Xcode 7/8
+#ifdef __IPHONE_11_0
+  if (@available(iOS 11.0, *)) {
+    UILayoutGuide *layoutGuide = self.view.safeAreaLayoutGuide;
     [layoutGuide.topAnchor constraintEqualToAnchor:self.navigationBar.topAnchor].active = YES;
   } else {
+#endif
     [NSLayoutConstraint constraintWithItem:self.topLayoutGuide
                                  attribute:NSLayoutAttributeBottom
                                  relatedBy:NSLayoutRelationEqual
@@ -89,7 +92,9 @@
                                 multiplier:1.0
                                   constant:0]
         .active = YES;
+#ifdef __IPHONE_11_0
   }
+#endif
   NSDictionary *viewsBindings = NSDictionaryOfVariableBindings(_navigationBar);
 
   [NSLayoutConstraint
--- a/components/NavigationBar/examples/NavigationBarIconsExample.m
+++ b/components/NavigationBar/examples/NavigationBarIconsExample.m
@@ -77,13 +77,19 @@
   self.navigationItem.rightBarButtonItem = trailingButtonItem;
   self.navigationItem.backBarButtonItem = backButtonItem;
 
-  // TODO(rsmoore): Remove this check when we drop support for Xcode 7/8
+  // TODO(rsmoore): Remove this check when we drop support for iOS 8 and can use anchors
+  if ([self.topLayoutGuide respondsToSelector:@selector(bottomAnchor)]) {
+    NSLayoutYAxisAnchor *topLayoutAnchor = self.topLayoutGuide.bottomAnchor;
+    // TODO(rsmoore): Remove this check when we drop support for Xcode 7/8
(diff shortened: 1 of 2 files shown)


components/NavigationBar/examples/NavigationBarIconsExample.m
```diff
@@ -77,10 +77,13 @@ - (void)viewDidLoad {
   self.navigationItem.rightBarButtonItem = trailingButtonItem;
   self.navigationItem.backBarButtonItem = backButtonItem;
 
-  if ([self.view respondsToSelector:@selector(safeAreaLayoutGuide)]) {
-    UILayoutGuide *layoutGuide = [self.view performSelector:@selector(safeAreaLayoutGuide)];
+  // TODO(rsmoore): Remove this check when we drop support for Xcode 7/8
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
```
willlarche: If this is gonna be used a lot, can we get more reusable code like a macro macro?

willlarche (approved):

components/NavigationBar/examples/NavigationBarIconsExample.m
ianegordon: Positive: We could simplify this and eliminate for the first __IPHONE_11_0 to get out of sync with the second.
Negative: Why verbose this is 'standard' macro magic that most iOS engineers can immediate parse / recognize.
If we do create a macro it should likely be standard across all of MDC-iOS and used universally.

willlarche: ^ That was my hope (the second)

components/NavigationBar/examples/NavigationBarIconsExample.m
```diff
@@ -77,10 +77,13 @@ - (void)viewDidLoad {
   self.navigationItem.rightBarButtonItem = trailingButtonItem;
   self.navigationItem.backBarButtonItem = backButtonItem;
 
-  if ([self.view respondsToSelector:@selector(safeAreaLayoutGuide)]) {
```
ianegordon, line 80: Can we identify the top constraint in an if statement?
That the execution path is the same, just with a slightly different parameter.
NSLayoutYAxisAnchor *topLayoutAnchor = self.topLayoutGuide.topAnchor;
// Determine best way to identify if we have a safeAreaLayoutGuide and add as the if clause.
if ([UIView respondsTo:safeAreaLayoutGuide]) {
  topLayoutAnchor = self.safeLayoutBuide.topAnchor.
}
[topLayoutAnchor constraintEqualToAnchor:self.navigationBar.topAnchor].active = YES;

ianegordon (changes requested):

components/NavigationBar/examples/NavigationBarIconsExample.m
romoore: We can't use `-constraintEqualToAnchor:` exclusively until we drop iOS 8 support.  Do you want the following?
1. A path for iOS 8 only using NSLayoutConstraint ...withItem:
2. A second path for iOS 9+ that uses safeAreaInsets if available? We still need the macros to protect older Xcode/base SDK versions from seeing the new symbol.

This would mean we can migrate straight to `-constraintEqualToAnchor` by simply deleting the iOS 8 path later.

components/NavigationBar/examples/NavigationBarIconsExample.m
romoore: How can we drop the first check (`defined (__IPHONE_11_0)`)? Won't that result in an using an undefined symbol in the second check?

romoore:
I'm OK if we want to build a set of functions that simplify some of this logic across MDC. Here are the questions this PR needs to ask:

1. Are layout anchors supported?
2. Are we compiling for iOS 11 SDK **AND** are we running iOS 11?

components/NavigationBar/examples/NavigationBarIconsExample.m
ianegordon: Sorry, I wasn't particularly clear.
Right now we are checking against a single value (__IPHONE_11_0) twice.
We could find ourselves in a position where we have the following error:
```
#if defined(__IPHONE_11_0)  && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_99_3)
```
Although in reality we end up with a dropped character or missing minor version.
By using a macro, we can take a single parameter to check against.
That said, I'm hesitant to create our own macro to add a layer of indirection over well-known pattern.

romoore: That makes sense.  I think we should keep macros in all the places they are needed. I dislike them in general, but they bring attention to areas of code that are particularly gritty (like this example) and deserve attention.

components/NavigationBar/examples/NavigationBarIconsExample.m
ianegordon: iOS 8!!! :shakes_fist:

ianegordon:

PR #1990, opened 2017-09-15
[NavigationBar] Fixed unsafe use of frame.size

### LINKED ISSUE #1486: [NavigationBar] Fix unsafe use of frame.size
In mdc_frameAligned... methods we are using frame.size.height and frame.size.width.  We should not be using the origin or size values directly unless we are sure the frames have been standardized.

Replace frame.size.height with CGRectGetHeight, frame.size.width with CGRectGetWidth, etc. or document that the incoming frames will be standardized.

codeman7:
Uses CGRectGetWidth/Height instead of frame.size.width/height.

Closes #1486

### DIFF
--- a/components/NavigationBar/src/MDCNavigationBar.m
+++ b/components/NavigationBar/src/MDCNavigationBar.m
@@ -449,22 +449,22 @@ static NSString *const MDCNavigationBarTitleAlignmentKey = @"MDCNavigationBarTit
                            alignment:(UIControlContentVerticalAlignment)alignment {
   switch (alignment) {
     case UIControlContentVerticalAlignmentBottom:
-      return CGRectMake(frame.origin.x, CGRectGetMaxY(bounds) - frame.size.height, frame.size.width,
-                        frame.size.height);
+      return CGRectMake(frame.origin.x, CGRectGetMaxY(bounds) - CGRectGetHeight(frame), CGRectGetWidth(frame),
+                        CGRectGetHeight(frame));
 
     case UIControlContentVerticalAlignmentCenter: {
-      CGFloat centeredY = MDCFloor((bounds.size.height - frame.size.height) / 2) + bounds.origin.y;
-      return CGRectMake(frame.origin.x, centeredY, frame.size.width, frame.size.height);
+      CGFloat centeredY = MDCFloor((bounds.size.height - CGRectGetHeight(frame)) / 2) + bounds.origin.y;
+      return CGRectMake(frame.origin.x, centeredY, CGRectGetWidth(frame), CGRectGetHeight(frame));
     }
 
     case UIControlContentVerticalAlignmentTop: {
       // The title frame is vertically centered with the back button but will stick to the top of
       // the header regardless of the header's height.
       CGFloat navigationBarCenteredY =
-          MDCFloor(([self intrinsicContentSize].height - frame.size.height) / 2);
+          MDCFloor(([self intrinsicContentSize].height - CGRectGetHeight(frame)) / 2);
       navigationBarCenteredY = MAX(0, navigationBarCenteredY);
-      return CGRectMake(frame.origin.x, navigationBarCenteredY, frame.size.width,
-                        frame.size.height);
+      return CGRectMake(frame.origin.x, navigationBarCenteredY, CGRectGetWidth(frame),
+                        CGRectGetHeight(frame));
     }
 
     case UIControlContentVerticalAlignmentFill: {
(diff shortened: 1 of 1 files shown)


components/NavigationBar/src/MDCNavigationBar.m
```diff
@@ -449,22 +449,22 @@ - (CGRect)mdc_frameAlignedVertically:(CGRect)frame
                            alignment:(UIControlContentVerticalAlignment)alignment {
   switch (alignment) {
     case UIControlContentVerticalAlignmentBottom:
-      return CGRectMake(frame.origin.x, CGRectGetMaxY(bounds) - frame.size.height, frame.size.width,
-                        frame.size.height);
+      return CGRectMake(frame.origin.x, CGRectGetMaxY(bounds) - CGRectGetHeight(frame), CGRectGetWidth(frame),
```
romoore, line 452: This line is over 100 columns, and needs to be split up.  My suggestion would be to place each argument on a separate line.

```objectivec
return CGRectMake(frame.origin.x,
                  CGRectGetMaxY(bounds) - CGRectGetHeight(frame),
                  CGRectGetWidth(frame),
                  CGRectGetHeight(frame));
```

components/NavigationBar/src/MDCNavigationBar.m
```diff
@@ -449,22 +449,22 @@ - (CGRect)mdc_frameAlignedVertically:(CGRect)frame
...
+                        CGRectGetHeight(frame));
 
     case UIControlContentVerticalAlignmentCenter: {
-      CGFloat centeredY = MDCFloor((bounds.size.height - frame.size.height) / 2) + bounds.origin.y;
-      return CGRectMake(frame.origin.x, centeredY, frame.size.width, frame.size.height);
+      CGFloat centeredY = MDCFloor((bounds.size.height - CGRectGetHeight(frame)) / 2) + bounds.origin.y;
+      return CGRectMake(frame.origin.x, centeredY, CGRectGetWidth(frame), CGRectGetHeight(frame));
```
romoore, line 457: Since we're switching to `CGRectGetWidth` and `CGRectGetHeight`, I believe the access to the origin should switch to
```objectivec
return CGRectMake(CGRectGetMinX(frame), centeredY,
                  CGRectGetWidth(frame), CGRectGetHeight(frame));
```

components/NavigationBar/src/MDCNavigationBar.m
```diff
@@ -449,22 +449,22 @@ - (CGRect)mdc_frameAlignedVertically:(CGRect)frame
...
     case UIControlContentVerticalAlignmentBottom:
-      return CGRectMake(frame.origin.x, CGRectGetMaxY(bounds) - frame.size.height, frame.size.width,
-                        frame.size.height);
+      return CGRectMake(frame.origin.x, CGRectGetMaxY(bounds) - CGRectGetHeight(frame), CGRectGetWidth(frame),
+                        CGRectGetHeight(frame));
 
     case UIControlContentVerticalAlignmentCenter: {
-      CGFloat centeredY = MDCFloor((bounds.size.height - frame.size.height) / 2) + bounds.origin.y;
-      return CGRectMake(frame.origin.x, centeredY, frame.size.width, frame.size.height);
+      CGFloat centeredY = MDCFloor((bounds.size.height - CGRectGetHeight(frame)) / 2) + bounds.origin.y;
```
romoore, line 456: This should use `CGRectGetHeight` for the `bounds`  and `CGRectGetMinY` for the origin; it also needs to be broken-up (too long).

```objectivec
CGFloat centeredY =
    MDCFloor((CGRectGetHeight(bounds) - CGRectGetHeight(frame)) / 2)
    + CGRectGetMinY(bounds);
```

components/NavigationBar/src/MDCNavigationBar.m
```diff
@@ -449,22 +449,22 @@ - (CGRect)mdc_frameAlignedVertically:(CGRect)frame
...
     case UIControlContentVerticalAlignmentTop: {
       // The title frame is vertically centered with the back button but will stick to the top of
       // the header regardless of the header's height.
       CGFloat navigationBarCenteredY =
-          MDCFloor(([self intrinsicContentSize].height - frame.size.height) / 2);
+          MDCFloor(([self intrinsicContentSize].height - CGRectGetHeight(frame)) / 2);
       navigationBarCenteredY = MAX(0, navigationBarCenteredY);
-      return CGRectMake(frame.origin.x, navigationBarCenteredY, frame.size.width,
-                        frame.size.height);
+      return CGRectMake(frame.origin.x, navigationBarCenteredY, CGRectGetWidth(frame),
```
romoore, line 466: We should also switch to using `CGRectGetMinX` if we're going to normalize the width and height.

```objectivec
return CGRectMake(CGRectGetMinX(frame), navigationBarCenteredY, CGRectGetWidth(frame),
                  CGRectGetHeight(frame));
```

components/NavigationBar/src/MDCNavigationBar.m
```diff
@@ -510,12 +510,12 @@ - (CGRect)mdc_frameAlignedHorizontally:(CGRect)frame
       // Place the title as close to the center, shifting it slightly in to the side with more space
       if (leftMidSpaceX >= halfFrameWidth) {
         CGFloat frameMaxX = CGRectGetMinX(rightButtonBar.frame) - titleRightInset;
-        return CGRectMake(frameMaxX - frame.size.width, frame.origin.y, frame.size.width,
-                          frame.size.height);
+        return CGRectMake(frameMaxX - CGRectGetWidth(frame), frame.origin.y, CGRectGetWidth(frame),
```
romoore, line 513: Please use `CGRectGetMinY` here.

```objectivec
return CGRectMake(frameMaxX - CGRectGetWidth(frame), CGRectGetMinY(frame), 
                  CGRectGetWidth(frame), CGRectGetHeight(frame));
```

romoore (changes requested):
I didn't audit the whole file, please make sure anywhere we switch to using `GetWidth` or `GetHeight` that we're also appropriately using `GetMinX` and `GetMinY` and vice-versa.  Non-standardized rects can be tricky. 😰

romoore:
It may be worthwhile to write a few unit tests using non-standardized frame values to verify that the methods are correct wherever possible.

For example, `CGRectMake(20, 15, -20, -30)` is equivalent to `CGRectMake(0, -15, 20, 30)`.

PR #1991, opened 2017-09-15
[NavigationBar] Use CGRect method instead of properties

### LINKED ISSUE #1486: [NavigationBar] Fix unsafe use of frame.size

codeman7:
I opened a new PR because I accidentally changed other files on the previous branch. Use of CGRectGetHeight/Width instead of bounds/frame.size.width/height and CGRectGetMinX/Y instead of bounds/frame.origin.x/y.

Closes #1486


ajsecord (approved):
Thanks!

PR #1992, opened 2017-09-16
[Tabs] fix uppercase title override

schrockblock:
This PR fixes what I believe is a typo in `MDCTabBar#setDisplaysUppercaseTitles`. Currently, instead of setting `_hasDefaultDisplaysUppercaseTitles` to `NO` that method sets `_hasDefaultAlignment` to `NO`, a variable which, as far as I can tell, is unrelated. When I make the submitted change locally, tabs work as expected (namely, titles are not forced to uppercase).


schrockblock:
I signed the CLA!

ajsecord (approved):
Thanks!

PR #1993, opened 2017-09-17
[TextFields] Padded label for placeholder for outlined style.

### LINKED ISSUE #1921: [Textfields] Add horizontal padding to floating titles
floating titles should have 4dp of padding on the left and right of the text.

willlarche:
Closes #1921

### DIFF
--- a/components/TextFields/src/MDCTextInputControllerOutlinedField.m
+++ b/components/TextFields/src/MDCTextInputControllerOutlinedField.m
@@ -17,6 +17,7 @@
 #import "MDCTextInputControllerOutlinedField.h"
 
 #import "MDCTextInputBorderView.h"
+#import "private/MDCPaddedLabel.h"
 #import "private/MDCTextInputControllerDefault+Subclassing.h"
 
 #import "MDCMath.h"
@@ -25,12 +26,14 @@
 
 static const CGFloat MDCTextInputTextFieldOutlinedFullPadding = 16.f;
 static const CGFloat MDCTextInputTextFieldOutlinedNormalPlaceholderPadding = 20.f;
+static const CGFloat MDCTextInputTextFieldOutlinedPlaceholderPadding = 4.f;
 
 static UIRectCorner _roundedCornersDefault = UIRectCornerAllCorners;
 
 @interface MDCTextInputControllerOutlinedField ()
 
 @property(nonatomic, strong) NSLayoutConstraint *placeholderCenterY;
+@property(nonatomic, strong) NSLayoutConstraint *placeholderLeading;
 
 @end
 
@@ -60,6 +63,7 @@ static UIRectCorner _roundedCornersDefault = UIRectCornerAllCorners;
   CGPoint destination = [super floatingPlaceholderDestination];
   CGFloat offset = self.textInput.placeholderLabel.font.lineHeight - self.textInput.placeholderLabel.font.xHeight;
   destination.y = -1 * offset;
+  destination.x -= ((MDCPaddedLabel *)self.textInput.placeholderLabel).padding.horizontal;
   return destination;
 }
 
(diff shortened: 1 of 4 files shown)


components/TextFields/src/private/MDCPaddedLabel.h
```diff
@@ -0,0 +1,42 @@
...
+#import <UIKit/UIKit.h>
+/**
+ Deltas for padding.
+
+ horizontal is the dx and vertical is the dy.
+ */
+struct MDCSymmetricPadding {
+  CGFloat horizontal;
+  CGFloat vertical;
+};
```
ianegordon: Can we not use UIEdgeInsets?
Perhaps with a convenience method: UIEdgeInsetsWithSymetricalPadding(horizontal, vertical)

components/TextFields/src/private/MDCPaddedLabel.m
```diff
@@ -0,0 +1,46 @@
...
+
+#pragma mark - UILabel Overrides
+
+- (void)drawTextInRect:(CGRect)rect {
+  [super drawTextInRect:CGRectInset(rect, self.padding.horizontal, self.padding.vertical)];
+}
+
+#pragma mark - UIView Overrides
+
+- (CGSize)intrinsicContentSize {
```
ianegordon, line 37: Does intrinsicContentSize work with textRectForBounds:limitedToNumberOflines: ?

components/TextFields/src/private/MDCTextInputCommonFundament.m
```diff
@@ -19,6 +19,7 @@
 #import "MDCTextInputArt.h"
 #import "MDCTextInputBorderView.h"
 #import "MDCTextInputCommonFundament.h"
+#import "MDCPaddedLabel.h"
```
ajsecord: Sort

ajsecord (changes requested):
Is this a visual change, if so, can you please add a before/after screenshot?

components/TextFields/src/private/MDCPaddedLabel.h
willlarche: The labels don't have 'real' insets: the left can never be different than the right. And that's by design: it's the only case needed by the class using it and it's a private class.

So I had a conversation with 3 developers asking what would be the most natural way to express these values: CGPoint, CGSize, UIEdgeInsets, Custom struct. After much deliberation, 2:3 vote for custom struct.

willlarche:
The difference visually:

padded label

components/TextFields/src/private/MDCTextInputCommonFundament.m
willlarche: Done!

components/TextFields/src/private/MDCPaddedLabel.m
willlarche: I have every reason to believe it does. That method is called to get the size the _text_ should be which is unchanged by my changes. However, if we want this to have even better support, I should add an overridden sizeThatFits:

components/TextFields/src/private/MDCPaddedLabel.h
ianegordon: ☹️  Creating a custom struct that is almost, but not quite entirely, identical to an existing, commonly used struct strikes me as unfriendly.
Consider that using a UIEdgeInsets might also simplify your documentation:
/**
 The padding to be applied.

  paddingInsets are used to inset the target rect in drawTextInRect and to increase the intrinsicContentSize.
*/
I'll leave it to you.

components/TextFields/src/MDCTextInputControllerOutlinedField.m
```diff
@@ -116,30 +128,49 @@ - (void)updatePlaceholder {
...
+                                                            constant:placeholderConstant];
     self.placeholderCenterY.priority = UILayoutPriorityDefaultHigh;
     self.placeholderCenterY.active = YES;
 
     [self.textInput.placeholderLabel setContentHuggingPriority:UILayoutPriorityDefaultHigh + 1
                                                        forAxis:UILayoutConstraintAxisVertical];
   }
   self.placeholderCenterY.constant = placeholderConstant;
+
+  struct MDCSymmetricPadding padding = ((MDCPaddedLabel *)self.textInput.placeholderLabel).padding;
```
ianegordon: casting is where things often go wrong.
Please move all casts into there own statement / localvar to ease in parsing.  Here an below.
MDCPaddedLabel *paddedLabel = (MDCPaddedLabel *)self.textInput.placeholderLabel;

ianegordon (changes requested):
Please pull the casts into their own statements.
I would argue longer about the struct if this was a public interface.

components/TextFields/src/MDCTextInputControllerOutlinedField.m
willlarche: This is done.

components/TextFields/src/private/MDCPaddedLabel.m
willlarche: Added comment.

ianegordon (approved):

ianegordon:

ajsecord (approved):

PR #1998, opened 2017-09-18
[PageControl] Switch NSMutableArray to Swift Array.

kolodny:
This is mostly a cleanup to use Swift's `Array.map` to create the pages instead of using an `NSMutableArray`

### DIFF
--- a/components/PageControl/examples/PageControlTypicalUseExample.swift
+++ b/components/PageControl/examples/PageControlTypicalUseExample.swift
@@ -31,17 +31,17 @@ class PageControlSwiftExampleViewController: UIViewController, UIScrollViewDeleg
 
   let pageControl = MDCPageControl()
   let scrollView = UIScrollView()
-  let pages: [UIView] = PageControlSwiftExampleViewController.pageColors.enumerated().map {
+  let pageLabels: [UILabel] = PageControlSwiftExampleViewController.pageColors.enumerated().map {
       enumeration in
     let (i, pageColor) = enumeration
-    let page = UILabel()
-    page.text = "Page \(i + 1)"
-    page.font = page.font.withSize(50)
-    page.textColor = UIColor.init(white: 0, alpha: 0.8)
-    page.backgroundColor = pageColor
-    page.textAlignment = NSTextAlignment.center
-    page.autoresizingMask = [.flexibleTopMargin, .flexibleBottomMargin]
-    return page
+    let pageLabel = UILabel()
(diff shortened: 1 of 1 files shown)


components/PageControl/examples/PageControlTypicalUseExample.swift
```diff
@@ -20,47 +20,51 @@ import MaterialComponents.MaterialPageControl
...
+    MDCPalette.cyan.tint700,
+    MDCPalette.cyan.tint300,
+    MDCPalette.cyan.tint500,
+    MDCPalette.cyan.tint700
+  ]
+
   let pageControl = MDCPageControl()
   let scrollView = UIScrollView()
-  let pages = NSMutableArray()
+  let pages: [UIView] = PageControlSwiftExampleViewController.pageColors.enumerated().map {
```
willlarche: Since these are actually labels, let's make it `let pageLabels: [UILabel} = ...` and refer to it as a `pageLabel` in all local variables.

components/PageControl/examples/PageControlTypicalUseExample.swift
```diff
@@ -20,47 +20,51 @@ import MaterialComponents.MaterialPageControl
...
+      enumeration in
+    let (i, pageColor) = enumeration
+    let page = UILabel()
+    page.text = "Page \(i + 1)"
+    page.font = page.font.withSize(50)
+    page.textColor = UIColor.init(white: 0, alpha: 0.8)
```
willlarche: is the `0.8` safe for compiling to a 32-bit machine? (Asking. I don't know answer.)

willlarche (changes requested):

components/PageControl/examples/PageControlTypicalUseExample.swift
kolodny: Done

components/PageControl/examples/PageControlTypicalUseExample.swift
kolodny: Based on [this tool](https://www.h-schmidt.net/FloatConverter/IEEE754.html) it's stored as `0.800000011920928955078125`. [It currently is using that value](https://github.com/material-components/material-components-ios/blob/develop/components/PageControl/examples/PageControlTypicalUseExample.swift#L55)

willlarche: K!

willlarche (approved):

PR #2003, opened 2017-09-19
[TextFields] Pinning clear button to top of input.

willlarche:
Doesn't change position for single line text fields but changes it for multiline. Previous design was for the clear button to follow the underline. Now it will always be 'top right'.


randallli (approved):

randallli:
lgtm. looks like you wanted a review from ian

willlarche:
Tested locally with test_all. Travis too slow.

PR #2011, opened 2017-09-19
[Ink] Call setNeedsLayout on max ripple radius change.

ajsecord:
Fix an ink view bug. When its maxRippleRadius is changed, it needs to call -setNeedsLayout so its ink layer could call -layoutSublayers that re-calculates the ripple frame.

From internal change 169289257


romoore (approved):
LGTM. Do we have steps to reproduce the bug and the fix?

PR #2015, opened 2017-09-20
[Math] Add degrees to radians method to MDCMath

jgunaratne:
Add degrees to radians method to MDCMath

### DIFF
--- a/components/private/Math/src/MDCMath.h
+++ b/components/private/Math/src/MDCMath.h
@@ -32,9 +32,9 @@ static inline CGFloat MDCFabs(CGFloat value) {
 
 static inline CGFloat MDCDegreesToRadians(CGFloat degrees) {
 #if CGFLOAT_IS_DOUBLE
-  return degrees * M_PI / 180.0;
+  return degrees * (CGFloat)M_PI / 180.0;
 #else
-  return degrees * M_PI / 180.f;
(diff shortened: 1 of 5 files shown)


components/private/Math/src/MDCMath.h
```diff
@@ -30,6 +30,10 @@ static inline CGFloat MDCFabs(CGFloat value) {
 #endif
 }
 
+static inline CGFloat MDCDegreesToRadians(CGFloat degrees) {
+  return degrees * M_PI / 180.f;
```
willlarche: Could you make it double if cgfloat is double and float if not?

willlarche (dismissed):

components/private/Math/src/MDCMath.h
ianegordon: Same comment from #2016 
With the explicit 180.f there is a cast if CGFLOAT_IS_DOUBLE.

#if CGFLOAT_IS_DOUBLE
return degrees * M_PI / 180.0;
#else
return degrees * M_PI / 180.f;
#endif

ianegordon (dismissed):
Comment Inline.

ianegordon (approved):
LGTM.
Please merge once Travis is happy.

randallli (approved):

willlarche (approved):

PR #2023, opened 2017-09-22
[TabBar] Vertically center text tab labels

### LINKED ISSUE #1321: [Tabs] Text in tabs with title only is not centered vertically
### Overview
In tabs with title only the text is not properly vertically centered, it's slightly off. 

screen shot 2017-04-13 at 14 15 01

### Version number
23.0.1

romoore:
Text-only tabs should vertically center their labels.

### Before
tabs-center-before-centerline

### After
tabs-center-after-centerline

Closes #1321

### DIFF
--- a/components/Tabs/src/private/MDCItemBarCell.m
+++ b/components/Tabs/src/private/MDCItemBarCell.m
@@ -271,13 +271,13 @@ static const NSTimeInterval kSelectionAnimationDuration = 0.3f;
 
   CGFloat scale = self.window.screen.scale;
   _imageView.bounds = imageBounds;
-  _imageView.center = MDCAlignCenterWithBoundsAndScale(imageCenter, _imageView.bounds, scale);
+  _imageView.center = MDCRoundCenterWithBoundsAndScale(imageCenter, _imageView.bounds, scale);
 
   _badgeLabel.bounds = MDCRectAlignToScale(badgeBounds, scale);
-  _badgeLabel.center = MDCAlignCenterWithBoundsAndScale(badgeCenter, _badgeLabel.bounds, scale);
+  _badgeLabel.center = MDCRoundCenterWithBoundsAndScale(badgeCenter, _badgeLabel.bounds, scale);
 
   _titleLabel.bounds = MDCRectAlignToScale(titleBounds, scale);
-  _titleLabel.center = MDCAlignCenterWithBoundsAndScale(titleCenter, _titleLabel.bounds, scale);
+  _titleLabel.center = MDCRoundCenterWithBoundsAndScale(titleCenter, _titleLabel.bounds, scale);
 }
 
 - (void)tintColorDidChange {
--- a/components/Tabs/src/private/MDCItemBarCell.m
+++ b/components/Tabs/src/private/MDCItemBarCell.m
@@ -20,11 +20,9 @@
 #import "MDCItemBarStyle.h"
 #import "MaterialAnimationTiming.h"
 #import "MaterialInk.h"
+#import "MaterialMath.h"
 #import "MaterialRTL.h"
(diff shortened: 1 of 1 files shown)


brianjmoore (approved):

components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -260,12 +258,12 @@ - (void)layoutSubviews {
...
     } else {
-      // Title only, center vertically.
-      // +1 for slight adjustment to font baseline for centered title labels.
-      const CGFloat centeredTitleYOffset = 1.0f;
-      titleCenter.y = contentBounds.size.height - kBottomPadding - titleSize.height +
-                      centeredTitleYOffset + ((float)titleSize.height / 2.0f);
-      titleCenter.y = roundf((float)titleCenter.y);
+      // UILabels should be aligned to pixel boundaries to avoid anti-aliasing blurriness
+      CGFloat scale = self.window.contentScaleFactor;
+      if (MDCCGFloatEqual(scale, 0)) {
```
ianegordon: When would this be 0?
I'm a big fan of defensive coding, but this check is unnecessary if the system will always return a non-zero value.

components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -260,12 +258,12 @@ - (void)layoutSubviews {
       titleCenter.y = roundf((float)titleCenter.y);
 
     } else {
-      // Title only, center vertically.
-      // +1 for slight adjustment to font baseline for centered title labels.
-      const CGFloat centeredTitleYOffset = 1.0f;
-      titleCenter.y = contentBounds.size.height - kBottomPadding - titleSize.height +
-                      centeredTitleYOffset + ((float)titleSize.height / 2.0f);
-      titleCenter.y = roundf((float)titleCenter.y);
+      // UILabels should be aligned to pixel boundaries to avoid anti-aliasing blurriness
```
ianegordon: We are calculating the center here.
For pixel-alignment do we need to ensure that the label.frame.offset is on a pixel boundary?
If that is the case, consider moving the rounding to after we set the frame below (label.bounds, label.center below).
Do we need to align the badge and image as well?

ianegordon (changes requested):

components/Tabs/src/private/MDCItemBarCell.m
romoore: I can move the rounding code to where the bounds/center are set. I wasn't planning to change the layout of the badge or image in this PR since I was only trying to bring the text-only style up to spec.  I didn't notice misaligned images when I tested, but I'll take another look.

components/Tabs/src/private/MDCItemBarCell.m
romoore: This is 0 when `self.window` is `nil`. One place where this happens in MDCTabBarViewController, which calls `layoutIfNeeded` when setting the selected View Controller.

`TabBarViewControllerExample` &rarr; `-viewDidLoad` &rarr; `-loadTabBar` &rarr; `-[MDCTabBarViewController setSelectedViewController]` &rarr; `-transitionViewsWithoutAnimationFromViewController:toViewController:` &rarr; `-[MDCItemBar layoutSubviews]`

romoore:
Discussed with @ianegordon. Going to push up the MDCMath changes separately first.

romoore:
Blocked by #2042 (adding the Math functions)

romoore:
#2042 is merged, reassigning to @ianegordon for review.

ianegordon (approved):

ianegordon:
Excellent, please merge when Travis is happy.

PR #2028, opened 2017-09-22
[TabBar] Support 2-line labels for "top" tabs

### LINKED ISSUE #2025: [TabBar] Text tabs should support 2-lines of text
### Overview
Tabs (which are not Bottom Navigation) should support 2 lines of text.  The label should wrap text to 2 lines and then shrink the text down to 85% of original size to fit before using truncation rules.  85% uses UILabel's `minimiumScaleFactor` to accomplish the Material Design guideline of dropping from 14 sp to 12 sp font sizes (12/14 = 0.857).

https://material.io/guidelines/components/tabs.html#tabs-specs

### Reproduction steps
1. Open MDC Catalog &rarr; Tab Bar &rarr; Text Tabs
1. Scroll to the right and notice the "very long" tab title - it should be 2 lines long.

### Version number
35.1.0

All/Any

### Related issues
#1321

romoore:
The Tabs component in Material Design indicates that 2 lines are
permitted when the text gets too long to fit on a single line. When
MDCTabBar is used for anything other than Bottom Navigation, it should
support 2 lines of text.  Changing from a "14 sp" to "12 sp" font size
can be approximated using a `minimumScaleFactor` value of 0.85.

tabs-leading

tabs-justified

Closes #2025


romoore:
Marking blocked pending merge of #2023

components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -260,12 +271,11 @@ - (void)layoutSubviews {
       titleCenter.y = roundf((float)titleCenter.y);
 
     } else {
-      // Title only, center vertically.
-      // +1 for slight adjustment to font baseline for centered title labels.
-      const CGFloat centeredTitleYOffset = 1.0f;
-      titleCenter.y = contentBounds.size.height - kBottomPadding - titleSize.height +
-                      centeredTitleYOffset + ((float)titleSize.height / 2.0f);
-      titleCenter.y = roundf((float)titleCenter.y);
+      CGFloat scale = self.window.contentScaleFactor;
```
brianjmoore: Why we pull in the scale factor here is not obvious and threw me for a loop - add comment? Why do we need to do this special rounding here and not most other places?

components/Tabs/src/private/MDCItemBarStyle.h
```diff
@@ -65,6 +65,9 @@
 /** Padding in points between the title and image components, according to the MD spec. */
 @property(nonatomic) CGFloat titleImagePadding;
 
+/** Maximum number of lines of text allowed. */
+@property(nonatomic, assign) NSInteger textNumberOfLines;
```
brianjmoore: Naming: titleNumberOfLines to be consistent with the rest of the methods?

brianjmoore (approved):
Overall looks good! Just a few stylistic comments

components/Tabs/src/private/MDCItemBarCell.m
romoore: We should be doing it everywhere we compute the frame/center/bounds of a view - images and text are the real important ones.  In this case, I'm aligning the center to avoid introducing errors, but I'm not aligning the bounds since it isn't part of this issue. During testing, I didn't see any misaligned tabs.

It does look like the bounds are aligned point-aligned, but not pixel-aligned.  Either way, aligning the center with a point-aligned bounds should result in aligned labels.  For example, if scale=2 and center = (8.5, 13), and the size is (80, 40), then the origin is (-31.5, -7), which is also pixel-aligned.

romoore:
@ianegordon  Even though this merge is blocked pending #2023, I would appreciate your review.

components/Tabs/src/MDCTabBar.m
```diff
@@ -312,6 +312,7 @@ + (MDCItemBarStyle *)defaultStyleForPosition:(UIBarPosition)position
     style.titleFont = [MDCTypography buttonFont];
     style.inkStyle = MDCInkStyleBounded;
     style.titleImagePadding = (kImageTitleSpecPadding + kImageTitlePaddingAdjustment);
+    style.titleNumberOfLines = 2;
```
ianegordon: Why are we choosing whether to wrap base on if the TabBar is top or bottom?
(I couldn't find guidance in the spec)

components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -193,6 +192,18 @@ - (void)applyStyle:(MDCItemBarStyle *)style {
   }
 }
 
+- (void)updateTitleLines {
+  _titleLabel.numberOfLines = _style.titleNumberOfLines;
+  if (_style.titleNumberOfLines == 1) {
+    _titleLabel.adjustsFontSizeToFitWidth = NO;
+    _titleLabel.minimumScaleFactor = 1;
```
ianegordon: This value is ignored if adjustFontSizeToFitWidth.
Consider not changing it here.
If you do want to change, I would recommend going back to the default of 0.0.

components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -193,6 +192,18 @@ - (void)applyStyle:(MDCItemBarStyle *)style {
...
+  } else {
+    _titleLabel.adjustsFontSizeToFitWidth = YES;
+    // 0.85 is based on 12sp/14sp guidelines for single- or double-line text
+    _titleLabel.minimumScaleFactor = (CGFloat)0.85;
```
ianegordon: As we only need this value if adjustFontSizeToFitWidth == YES.
Consider setting this value only once, on init.

components/Tabs/src/private/MDCItemBarStyle.h
```diff
@@ -65,6 +65,9 @@
 /** Padding in points between the title and image components, according to the MD spec. */
 @property(nonatomic) CGFloat titleImagePadding;
 
+/** Maximum number of lines of text allowed. */
+@property(nonatomic, assign) NSInteger titleNumberOfLines;
```
ianegordon: If we are always supposed to support wrapping for text-only items, we could eliminate this property.
If this only applies to text-only tabs, please indicate that in the comment / name.
textOnlyNumberOfLines perhaps?

ianegordon (changes requested):

components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -260,12 +271,12 @@ - (void)layoutSubviews {
       titleCenter.y = roundf((float)titleCenter.y);
 
     } else {
-      // Title only, center vertically.
-      // +1 for slight adjustment to font baseline for centered title labels.
-      const CGFloat centeredTitleYOffset = 1.0f;
-      titleCenter.y = contentBounds.size.height - kBottomPadding - titleSize.height +
-                      centeredTitleYOffset + ((float)titleSize.height / 2.0f);
-      titleCenter.y = roundf((float)titleCenter.y);
+      // UILabels should be aligned to pixel boundaries to avoid anti-aliasing blurriness
```
ianegordon: Similar issue as the previous PR.  This will fail to pixel align on non-even font sizes.
We should round the final label.frame.origin.

ianegordon (changes requested):

components/Tabs/src/MDCTabBar.m
romoore: The Material "Tabs" component is only at the top of a view. The other component is "[Bottom Navigation](https://material.io/guidelines/components/bottom-navigation.html)" and I do not believe they support more than 1 line of text.

components/Tabs/src/private/MDCItemBarStyle.h
romoore: If there's an image, I believe we limit to a single line of text. In that case the spec isn't explicit, but I'm pretty sure the intention is a single line of text once images are present.  Two lines of text are also not permitted for Tabs on the bottom (Botton Nav).

The naming gets wonky, since it's "the maximum possible number of lines of title text depending on other factors."  How about "textOnlyMaximumNumberOfLines" ?

components/Tabs/src/private/MDCItemBarCell.m
romoore: Agreed. This is the previous PR.

components/Tabs/src/private/MDCItemBarCell.m
romoore: Done

components/Tabs/src/private/MDCItemBarCell.m
romoore: Done - setting when the UILabel is created.

components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -185,12 +185,28 @@ - (void)applyStyle:(MDCItemBarStyle *)style {
...
+    [self updateTitleLines];
     [self updateTitleFont];
     [self updateTransformsAnimated:NO];
     [self setNeedsLayout];
   }
 }
 
+- (void)updateTitleLines {
+  // The presence of an image restricts titles to a single line
+  if (_imageView && !_imageView.hidden) {
```
brianjmoore: Nit: Presence of the image view is more of a secondary factor to decide this. The primary source of truth here should be the tab bar style shouldDisplayImage

components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -185,12 +185,28 @@ - (void)applyStyle:(MDCItemBarStyle *)style {
...
+    _titleLabel.numberOfLines = 1;
+  } else {
+    _titleLabel.numberOfLines = _style.textOnlyTitleNumberOfLines;
+  }
+  
+  if (_titleLabel.numberOfLines == 1) {
```
brianjmoore: Nit: I’m not sure splitting this across more lines adds any clarity in this case. Replace with a brief ternary and comment?

brianjmoore (approved):

components/Tabs/src/private/MDCItemBarCell.m
romoore: Done

components/Tabs/src/private/MDCItemBarCell.m
romoore: Done

components/Tabs/src/private/MDCItemBarStyle.h
ianegordon: Got it.
Consider: textOnlyNumberOfLines as it would match UIKit / UILabel.

ianegordon (approved):

ianegordon:
Please consider comment before submit, otherwise LGTM.

components/Tabs/src/private/MDCItemBarStyle.h
romoore: Done.

PR #2042, opened 2017-09-26
[Math] Round CGPoint and align UIView centers

romoore:
Although it is possible to currently align a CGRect (frame) to a
pixel-aligned bounding rectangle, in some cases manipulating the frame
is neither possible nor desirable (such as when the transform is not the
identity).  Adding two new methods to support rounding a CGPoint to a
pixel and aligning the center point of a UIVIew given its bounds.

### DIFF
--- a/components/private/Math/src/MDCMath.h
+++ b/components/private/Math/src/MDCMath.h
@@ -125,3 +125,35 @@ static inline CGRect MDCRectAlignToScale(CGRect rect, CGFloat scale) {
                     MDCCeil((CGRectGetWidth(rect) + adjustWidthHeight.width) * scale) / scale,
                     MDCCeil((CGRectGetHeight(rect) + adjustWidthHeight.height) * scale) / scale);
 }
+
+static inline CGPoint MDCPointAlignToScale(CGPoint point, CGFloat scale) {
+  if (MDCCGFloatEqual(scale, 0)) {
+    return CGPointZero;
+  }
+
+  return CGPointMake(MDCRound(point.x * scale) / scale, MDCRound(point.y * scale) / scale);
+}
+
+/**
+ Align the centerPoint of a view so that its origin is pixel-aligned. Returns @c CGRectZero if
+ @c scale is zero or @c bounds is @c CGRectNull.
+
+ @param center the unaligned center of the view.
+ @param bounds the bounds of the view.
+ @param scale the native scaling factor for pixel alignment.
+
+ @return the center point of the view such that its origin will be pixel-aligned.
+ */
+static inline CGPoint MDCAlignCenterWithBoundsAndScale(CGPoint center,
+                                                       CGRect bounds,
+                                                       CGFloat scale) {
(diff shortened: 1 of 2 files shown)


components/private/Math/src/MDCMath.h
```diff
@@ -125,3 +125,35 @@ static inline CGRect MDCRectAlignToScale(CGRect rect, CGFloat scale) {
...
+ @return the center point of the view such that its origin will be pixel-aligned.
+ */
+static inline CGPoint MDCAlignCenterWithBoundsAndScale(CGPoint center,
+                                                       CGRect bounds,
+                                                       CGFloat scale) {
+  if (MDCCGFloatEqual(scale, 0) || CGRectIsNull(bounds)) {
+    return CGPointZero;
+  }
+
+  CGFloat halfWidth = CGRectGetWidth(bounds) / 2;
```
ianegordon, line 154: CGRectGetMidX() ?

ianegordon (approved):

components/private/Math/src/MDCMath.h
```diff
@@ -125,3 +125,35 @@ static inline CGRect MDCRectAlignToScale(CGRect rect, CGFloat scale) {
...
+ @param center the unaligned center of the view.
+ @param bounds the bounds of the view.
+ @param scale the native scaling factor for pixel alignment.
+
+ @return the center point of the view such that its origin will be pixel-aligned.
+ */
+static inline CGPoint MDCRoundCenterWithBoundsAndScale(CGPoint center,
+                                                       CGRect bounds,
+                                                       CGFloat scale) {
+  if (MDCCGFloatEqual(scale, 0) || CGRectIsNull(bounds)) {
```
ianegordon, line 150: || CGRectIsEmpty(bounds) ?

components/private/Math/src/MDCMath.h
romoore: If the bounds are {(5, 10), (15, 20)} then Width is 15 / 2 = 7.5, while mid-X is 12.5.

components/private/Math/src/MDCMath.h
romoore: If the rect is empty, that implies a zero height and width - should we not still align the point?

romoore: Would you prefer that check to just return `MDCPointRoundWithScale()`?

romoore: Ah, never mind. Although a CGRect is empty, it may still have a height or width (only one needs to be 0). In that case, the math to adjust the center point still needs to take place.

ianegordon: Please disregard, we definitely want to align the point in case of an empty rect.

components/private/Math/src/MDCMath.h
ianegordon: Right, because the bounds may be transformed and the origin may not be 0, 0.
LGTM.

PR #2043, opened 2017-09-26
[NavigationBar] Pixel-align title

### LINKED ISSUE #2039: [NavigationBar] Title label is misaligned on 3x devices
### Overview
Depending on the font size and number of characters, the titleLabel in MDCNavigationBar is not pixel-aligned on 3x devices (iPhone X).

navbar-misaligned-title

### Reproduction steps
1. Open MDC Catalog &rarr; Buttons &rarr; Buttons (Dynamic Type) in a 3x retina simulator
2. Turn on Simulator Debugging option "Color Misaligned Images"
3. Verify no purple/magenta coloring in MDCNavigationBar

### Version number
35.1.0

iPhone X/iOS 11 (Simulator)

romoore:
On 3x devices, centering the Title label was not aligning the origin to
pixels, resulting in blurry text.

### Misaligned
navbar-title-misaligned

### Aligned
navbar-title-aligned

Closes #2039

### DIFF
--- a/components/NavigationBar/src/MDCNavigationBar.m
+++ b/components/NavigationBar/src/MDCNavigationBar.m
@@ -317,8 +317,13 @@ static NSString *const MDCNavigationBarTitleAlignmentKey = @"MDCNavigationBarTit
   CGRect alignedFrame = [self mdc_frameAlignedVertically:titleFrame
                                             withinBounds:textFrame
                                                alignment:titleVerticalAlignment];
-  _titleLabel.frame =
-      [self mdc_frameAlignedHorizontally:alignedFrame alignment:self.titleAlignment];
+  alignedFrame = [self mdc_frameAlignedHorizontally:alignedFrame alignment:self.titleAlignment];
+  CGFloat scale = self.window.screen.scale;
+  if (!MDCCGFloatEqual(0, scale)) {
+    alignedFrame = MDCRectAlignToScale(alignedFrame, scale);
+  }
+
+  _titleLabel.frame = alignedFrame;
   self.titleView.frame = textFrame;
 
   // Button and title label alignment
--- a/components/NavigationBar/src/MDCNavigationBar.m
+++ b/components/NavigationBar/src/MDCNavigationBar.m
@@ -318,12 +318,8 @@ static NSString *const MDCNavigationBarTitleAlignmentKey = @"MDCNavigationBarTit
                                             withinBounds:textFrame
                                                alignment:titleVerticalAlignment];
   alignedFrame = [self mdc_frameAlignedHorizontally:alignedFrame alignment:self.titleAlignment];
-  CGFloat scale = self.window.screen.scale;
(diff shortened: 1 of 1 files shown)


willlarche (approved):
Oh, wow. So much better. Do you really need to check for a 0 scale?

ianegordon (approved):
Ah!...... nothing quite like a simple fix. ;)

components/NavigationBar/src/MDCNavigationBar.m
```diff
@@ -317,8 +317,13 @@ - (void)layoutSubviews {
   CGRect alignedFrame = [self mdc_frameAlignedVertically:titleFrame
                                             withinBounds:textFrame
                                                alignment:titleVerticalAlignment];
-  _titleLabel.frame =
-      [self mdc_frameAlignedHorizontally:alignedFrame alignment:self.titleAlignment];
+  alignedFrame = [self mdc_frameAlignedHorizontally:alignedFrame alignment:self.titleAlignment];
+  CGFloat scale = self.window.screen.scale;
```
ianegordon: Perhaps we want MDC*ToScale() to treat a scale of 0.0 as if it were equal to device.screen.scale?
We could avoid checking for 0.0 every time.
Might be an issue for people outputting to a second screen via airplay?

romoore:
Regarding the 0-check, you aren't the first person to ask about it *cough* Ian *cough*. I think maybe it's better to change the behavior of the function to just return the original CGRect if the scale is 0. It's just pixel-pushing, so if the rect is misaligned it's not going to cause functional errors.

I'll wait for my current PRs to roll through Travis before pushing up the change.

components/NavigationBar/src/MDCNavigationBar.m
romoore: Since this is a C-function, it's difficult to know [which of the screens](https://developer.apple.com/documentation/uikit/uiscreen/1617812-screens) we should use for the scale factor.  It would work as an extension on UIView, though.

romoore: I think between returning an unaligned frame and trying to guess which scale to use, it might make sense to just default to 1x if it's not provided - at least then it'll point-align and be guaranteed to be pixel-aligned.

willlarche: I'm still not convinced this is really a problem and I don't want to clutter up the codebase with a check for it. We use scale all over the place.

romoore: Yes, I think we're in agreement that the check should not be necessary for clients. I'd like to change the function to be safe (it currently returns CGRectNull), so  I hope we're discussing what default behavior to provide.

My recommendation is to return the rect aligned to a 1x scale, though we could just return the original rect instead.

romoore:
Tests were failing on Travis CI because we perform MDCNavBar layout during unit tests. Travis seems to be a headless environment where there is no window available to retrieve the scale factor.

PR #2047, opened 2017-09-26
[Catalog] Fix button examples

romoore:
The FAB in Typical Use was misaligned, resulting in a blurry image. The
Storyboard and Swift example had misaligned buttons due to
safeAreaInsets on iOS 11.

### FAB Before
fab-misaligned

### FAB After
fab-aligned

### Buttons Storyboard and Swift example
misaligned-buttons

### DIFF
--- a/components/Buttons/examples/ButtonsStoryboardAndProgrammatic.swift
+++ b/components/Buttons/examples/ButtonsStoryboardAndProgrammatic.swift
@@ -70,7 +70,7 @@ class ButtonsSwiftAndStoryboardController: UIViewController {
 
     view.addSubview(containerView)
 
-    let viewLayoutGuide : Any = {
+    let viewLayoutGuide: Any = {
 #if swift(>=3.2)
     if #available(iOS 11.0, *) {
       return view.safeAreaLayoutGuide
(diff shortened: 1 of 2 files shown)


romoore:
@codeman7 Hey Cody, I would appreciate if you took a look at my Swift code. Thanks! 😃

codeman7:
@romoore Everything looks good to me just a couple of stylistic things I would change.

romoore:
Thanks, @codeman7 . I don't see any comments, did you perhaps forget to Submit your review?

codeman7 (commented):
I think to follow the Swiftlint standard and the codebase so far we should have. 

let viewLayoutGuide: Any

codeman7:
I also think maybe pulling all of the catalog functions out into an extension would be cleaner. Other than that everything looks good. If you wanted to then we also could minimize viewDidLoad method by pulling out the button setup into its own function and the constraints into their own function. If you did that then you can remove the Swiftlint disable.

romoore:
Opened #2048 (yay power of 2!) to address other clean-up items in the same file.

randallli (approved):

PR #2050, opened 2017-09-26
[Swiftlint][Buttons] Cleans up ButtonsStoryboardAndProgrammatic.swift

### LINKED ISSUE #2048: [Swiftlint] [Buttons] Clean up ButtonsStoryboardAndProgrammatic.swift
### Overview
Swiftlint is disabled on `viewDidLoad` in ButtonsStoryboardAndProgrammatic.swift because of the function length.  The method should be broken-up with a few helpers (buttons, constraints, etc.) and the Catalog By Convention methods should probably be in an extension.

### Version number
35.1.0

codeman7:
Pulls the catalog functions into an extension and the button setup and constraints into their own functions. This makes it so that Swiftlint we no longer throw an error if we enable it for viewDidLoad.  
The constraints are in two separate functions because if they are all in one then we have to disable swiftlint. We can put two variables on each line to fix this problem for all of the constraints that we create.

This closes #2048

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode9
+osx_image: xcode8.3
 sudo: false
 
 # Only do a shallow clone of our Git repo.
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode8.3
+osx_image: xcode9
 sudo: false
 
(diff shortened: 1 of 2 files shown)


romoore:
I think you may have forked off of a branch other than develop.

codeman7:
@romoore Yeah I don't know why those first two commit messages are there, there were no changes for those. But, I used those messages on my other PR currently open.

PR #2053, opened 2017-09-26
[Activity Indicator] Resolve issues caused by restarting activity indicator quickly

jmdetloff:
Ensure MDCActivityIndicator reports to its delegate when stopAnimating is called, even if startAnimating is called again immediately. Additionally, ensure that when stopAnimating is called on a determinate activity indicator and then startAnimating is called immediately afterwards, that the indicator successfully animates in.

### DIFF
--- a/components/ActivityIndicator/src/MDCActivityIndicator.m
+++ b/components/ActivityIndicator/src/MDCActivityIndicator.m
@@ -206,6 +206,9 @@ static const CGFloat kSingleCycleRotation =
 
 - (void)startAnimating {
   if (_animatingOut) {
+    if ([_delegate respondsToSelector:@selector(activityIndicatorAnimationDidFinish:)]) {
+      [_delegate activityIndicatorAnimationDidFinish:self];
+    }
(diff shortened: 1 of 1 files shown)


jmdetloff:
I signed it!

components/ActivityIndicator/src/MDCActivityIndicator.m
```diff
@@ -206,6 +206,7 @@ - (CGSize)intrinsicContentSize {
 
 - (void)startAnimating {
   if (_animatingOut) {
+    [_delegate activityIndicatorAnimationDidFinish:self];
```
ajsecord: You need to check if the delegate has this method—it's optional.

ajsecord (changes requested):

components/ActivityIndicator/src/MDCActivityIndicator.m
jmdetloff: Thanks for the catch!

ajsecord (approved):

Issue #1920, opened 2017-08-28
[TextFields] Unsubscribing KVO may fail before attempting all unsubscribes


romoore:
### Overview
The MDCTextInputControllerDefault method `-unsubscribeFromKVO` uses a try/catch block to hide exceptions from callers. Because the try/catch is around 3 potential sources of thrown exceptions, it is possible that not all calls to `-removeObserver:forKeyPath:` will be attempted.  It would be better for each call to be wrapped separately.

### Version number
33.0.0

### Operating system and device
All/Any

willlarche:
Closed in #2056

PR #2056, opened 2017-09-26
[TextFields] Better custom fonts

willlarche:
Replaces magic KVO code with public API in the same pattern as the recent API for label colors. Do not worry about bloat; this is the last 'thing' that can be moved to public API per label.

### DIFF
--- a/components/TextFields/src/MDCTextInputControllerDefault.m
+++ b/components/TextFields/src/MDCTextInputControllerDefault.m
@@ -89,8 +89,6 @@ static NSString *const MDCTextInputControllerDefaultTrailingUnderlineLabelTextCo
 static NSString *const MDCTextInputControllerDefaultUnderlineViewModeKey =
     @"MDCTextInputControllerDefaultUnderlineViewModeKey";
 
-static NSString *const MDCTextInputControllerDefaultKVOKeyFont = @"font";
-
 static inline UIColor *MDCTextInputDefaultInlinePlaceholderTextColorDefault() {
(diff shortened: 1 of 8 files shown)


willlarche:
Closes #1920

components/TextFields/src/MDCTextInputControllerDefault.m
```diff
@@ -972,6 +952,44 @@ + (void)setInlinePlaceholderColorDefault:(UIColor *)inlinePlaceholderColorDefaul
                                        : MDCTextInputDefaultInlinePlaceholderTextColorDefault();
 }
 
+- (UIFont *)inlinePlaceholderFont {
+  return _inlinePlaceholderFont ?: [self class].inlinePlaceholderFontDefault;
+}
+
+- (void)setInlinePlaceholderFont:(UIFont *)inlinePlaceholderFont {
+  if (![_inlinePlaceholderFont isEqual:inlinePlaceholderFont]) {
+    _inlinePlaceholderFont = inlinePlaceholderFont;
```
randallli, line 961: We are letting the iVar go to nil and assume we will always use the getter?

style guide discourages accessors in some places.
http://google.github.io/styleguide/objcguide.html#avoid-accessors-during-init-and-dealloc

randallli: I guess we are already not following the guidance is this component. We should consider doing so.

randallli (approved):

components/TextFields/src/MDCTextInputControllerDefault.m
willlarche: This will be handled in a future PR.

PR #2060, opened 2017-09-27
[Swiftlint][Buttons] Cleans up ButtonsStoryboardAndProgrammatic.swift 

### LINKED ISSUE #2048: [Swiftlint] [Buttons] Clean up ButtonsStoryboardAndProgrammatic.swift

codeman7:
Pulls the catalog functions into an extension and the button setup and constraints into their own functions. This makes it so that Swiftlint we no longer throw an error if we enable it for viewDidLoad.
The constraints are in two separate functions because if they are all in one then we have to disable swiftlint. We can put two variables on each line to fix this problem for all of the constraints that we create.

I did a new pull request because I was having a conflict after @romoore updated the file.

This closes #2048


willlarche:
We haven't had a discussion internally about upgrading storyboards from Xcode 7 so I'm gonna put a temporary hold on this PR.

willlarche:
Sorry! Nevermind. Misread. this one.

components/Buttons/examples/ButtonsStoryboardAndProgrammatic.swift
```diff
@@ -1,12 +1,12 @@
 /*
  Copyright 2016-present the Material Components for iOS authors. All Rights Reserved.
-
+
```
romoore: Please revert the changes to the copyright notice. If you like, you can make the changes in a follow-up, though it's not necessary. :)

romoore (changes requested):

components/Buttons/examples/ButtonsStoryboardAndProgrammatic.swift
```diff
@@ -1,12 +1,12 @@
 /*
  Copyright 2016-present the Material Components for iOS authors. All Rights Reserved.
-
+ 
  Licensed under the Apache License, Version 2.0 (the "License");
  you may not use this file except in compliance with the License.
  You may obtain a copy of the License at
-
+
```
romoore: Nit: Please remove the added white space on the empty lines throughout.  It will help make the actual changes more obvious as well as keeping the formatting cleaner.

romoore:
Dropping the "iOS 11" label since these changes aren't impacted by iOS 11, they're just a minor refactoring.

components/Buttons/examples/ButtonsStoryboardAndProgrammatic.swift
```diff
@@ -18,67 +18,88 @@ import Foundation
...
-  class func catalogIsPrimaryDemo() -> Bool {
-    return false
-  }
-
+  
   let raisedButton = MDCRaisedButton()
   let flatButton = MDCFlatButton()
   let floatingButton = MDCFloatingButton()
-
+
```
romoore: Formatting: Please remove the extra white space on the empty lines here and below.

romoore (approved):
LGTM

PR #2065, opened 2017-09-28
[Dialogs] Size action buttons during layout

### LINKED ISSUE #1943: [Dialogs] Dialogs look bad when buttons have UIAppearance'd contentEdgeInsets
### Overview
While testing Dialogs in the MDC Catalog, I noticed several of the examples no longer show any text (except an ellipsis) for the button text.  This does not seem to be caused by dynamic type.

It looks like something changed in #1896 (commit [3951775](https://github.com/material-components/material-components-ios/commit/395177518777a355457dc0829903e47483420984)). The bug is present in that commit but not the previous commit.

### Reproduction steps
1. Open MDC Catalog &rarr; Dialogs
1. Open "Demo"
1. Tap on "Programmatic"

### Examples
simulator screen shot sep 1 2017 10 54 03 pm

simulator screen shot sep 1 2017 10 48 03 pm

### Version number
34.0.1

iPhone 5s/iOS 9.3 (Simulator)

romoore:
Buttons were previously being sized during `viewDidLoad`, which caused
problems if UIAppearance caused changes to contentEdgeInsets. Instead,
the buttons should be sized before layout of the dialog's subviews.

Also fixing some examples that were incorrectly sizing buttons before
layout.

Closes #1943

### DIFF
--- a/components/BottomSheet/examples/supplemental/BottomSheetPresenterViewController.m
+++ b/components/BottomSheet/examples/supplemental/BottomSheetPresenterViewController.m
@@ -29,15 +29,19 @@
 
   _button = [[MDCButton alloc] initWithFrame:CGRectZero];
   [_button setTitle:@"Show Bottom Sheet" forState:UIControlStateNormal];
-  [_button sizeToFit];
   _button.autoresizingMask =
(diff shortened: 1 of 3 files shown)


samnm (approved):
Added Ian for Alert stuff.

ianegordon (approved):
Alert update LGTM.

PR #2068, opened 2017-09-29
[FlexibleHeader] Make sure we update the layout when the safe area changes

### LINKED ISSUE #2067: [FlexibleHeader] The height isn't being updated if we're not tracking a scroll view
### Overview
The FlexibleHeader's updateLayout is a no-op when there's no trackingScrollView. Since this component now reacts to changes to the safe area, and adjusts its size accordingly, we need to make sure we update its frame when that happens.

### Reproduction steps
- Open the Catalog app
- Navigate to Collections
- Open an example
- Rotate the device and notice that the flexible header didn't change in size.

### Version number
35.2.0

iPhone X

chuga:
Closes #2067

### DIFF
--- a/components/FlexibleHeader/src/MDCFlexibleHeaderView.m
+++ b/components/FlexibleHeader/src/MDCFlexibleHeaderView.m
@@ -825,7 +825,7 @@ static NSString *const MDCFlexibleHeaderDelegateKey = @"MDCFlexibleHeaderDelegat
     // Even if we're not tracking a scroll view, our layout might have changed due to changes
     // to the safe area insets.
     CGRect bounds = self.bounds;
-    bounds.size.height = _maximumHeight;
+    bounds.size.height = _minimumHeight;
     self.bounds = bounds;
     [self fhv_commitAccumulatorToFrame];
     return;
--- a/components/FlexibleHeader/src/MDCFlexibleHeaderView.m
+++ b/components/FlexibleHeader/src/MDCFlexibleHeaderView.m
@@ -822,13 +822,15 @@ static NSString *const MDCFlexibleHeaderDelegateKey = @"MDCFlexibleHeaderDelegat
 
 - (void)fhv_updateLayout {
   if (!_trackingScrollView) {
-    // Even if we're not tracking a scroll view, our layout might have changed due to changes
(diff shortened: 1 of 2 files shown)


components/FlexibleHeader/src/MDCFlexibleHeaderView.m
```diff
@@ -822,7 +822,15 @@ - (void)fhv_accumulatorDidChange {
 
 - (void)fhv_updateLayout {
   if (!_trackingScrollView) {
-    return;
+    // Even if we're not tracking a scroll view, our minimumHeight might have changed due to new
+    // safe area insets.
+    if (_minimumHeight > 0 && self.bounds.size.height != _minimumHeight) {
```
ianegordon: frame and bounds may have negative values for size.height and size.width.
Use CGRectGetHeight(self.frame) and other CGRectFunc() instead of accessing the struct directly.

components/FlexibleHeader/src/MDCFlexibleHeaderView.m
```diff
@@ -822,7 +822,15 @@ - (void)fhv_accumulatorDidChange {
...
+      CGRect bounds = self.bounds;
+      bounds.size.height = _minimumHeight;
+      self.bounds = bounds;
+      [self fhv_commitAccumulatorToFrame];
+      return;
```
ianegordon: We are moving this return call, should it be outside this if block?

ianegordon (changes requested):

components/FlexibleHeader/src/MDCFlexibleHeaderView.m
chuga: Done

components/FlexibleHeader/src/MDCFlexibleHeaderView.m
chuga: Good catch, done.

randallli (approved):

chuga:
This *did* pass on Travis (see https://travis-ci.org/material-components/material-components-ios/builds/283731076, the error is not an actual test that failed).

I'll go ahead and merge this in.

Issue #2009, opened 2017-09-19
[NavigationBar] Support Safe Area


chuga:
### Overview
Issues visible in the Catalog: The Nav Bar overlaps the status bar / top notch, or sits below it but doesn't extend its background to cover the status bar.

![navbar](https://user-images.githubusercontent.com/4545451/30613715-1b298ac6-9d57-11e7-8d78-47684aa8f8d3.png)
![navbar2](https://user-images.githubusercontent.com/4545451/30613716-1cb843fa-9d57-11e7-9b4b-7eed3e4ffc60.png)

### Version number
35.0.0

### Operating system and device
iOS 11 / iPhone X

chuga:
Another example: ![simulator screen shot - iphone x - 2017-09-27 at 16 56 11](https://user-images.githubusercontent.com/4545451/30937881-2a2ed292-a3a6-11e7-89e4-092ea9b91eef.png)

PR #2070, opened 2017-09-29
[NavigationBar] Fix layoutSubviews to take into account the safe area insets.

chuga:
On iOS 11 need to take into account the safe area insets when laying out the nav bar's subviews.

simulator screen shot - iphone x - 2017-09-29 at 15 22 54

Tested on iPhone X on iOS 11, and iPhone 6 on iOS 11 and iOS 9.3, both in LTR and RTL languages.

Related to #2009

### DIFF
--- a/components/NavigationBar/src/MDCNavigationBar.m
+++ b/components/NavigationBar/src/MDCNavigationBar.m
@@ -278,7 +278,7 @@ static NSString *const MDCNavigationBarTitleAlignmentKey = @"MDCNavigationBarTit
 - (void)layoutSubviews {
   [super layoutSubviews];
 
-  UIEdgeInsets RTLFriendlySafeAreaInsets;
+  UIEdgeInsets RTLFriendlySafeAreaInsets = UIEdgeInsetsZero;
 #if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
   if (@available(iOS 11.0, *)) {
     RTLFriendlySafeAreaInsets =
--- a/components/NavigationBar/src/MDCNavigationBar.m
+++ b/components/NavigationBar/src/MDCNavigationBar.m
@@ -281,6 +281,11 @@ static NSString *const MDCNavigationBarTitleAlignmentKey = @"MDCNavigationBarTit
   CGSize leadingButtonBarSize = [_leadingButtonBar sizeThatFits:self.bounds.size];
   CGRect leadingButtonBarFrame =
       CGRectMake(0, CGRectGetMinY(self.bounds), leadingButtonBarSize.width, leadingButtonBarSize.height);
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    leadingButtonBarFrame.origin.x = self.safeAreaInsets.left;
+  }
+#endif
   _leadingButtonBar.frame = MDCRectFlippedForRTL(leadingButtonBarFrame, CGRectGetWidth(self.bounds),
                                                  self.mdc_effectiveUserInterfaceLayoutDirection);
 
@@ -288,6 +293,11 @@ static NSString *const MDCNavigationBarTitleAlignmentKey = @"MDCNavigationBarTit
   CGRect trailingButtonBarFrame =
       CGRectMake(CGRectGetWidth(self.bounds) - trailingButtonBarSize.width, CGRectGetMinY(self.bounds),
                  trailingButtonBarSize.width, trailingButtonBarSize.height);
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    trailingButtonBarFrame.origin.x -= self.safeAreaInsets.right;
+  }
+#endif
   _trailingButtonBar.frame = MDCRectFlippedForRTL(trailingButtonBarFrame, CGRectGetWidth(self.bounds),
                                                   self.mdc_effectiveUserInterfaceLayoutDirection);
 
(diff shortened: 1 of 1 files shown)


components/NavigationBar/src/MDCNavigationBar.m
```diff
@@ -281,13 +281,23 @@ - (void)layoutSubviews {
   CGSize leadingButtonBarSize = [_leadingButtonBar sizeThatFits:self.bounds.size];
   CGRect leadingButtonBarFrame =
       CGRectMake(0, CGRectGetMinY(self.bounds), leadingButtonBarSize.width, leadingButtonBarSize.height);
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    leadingButtonBarFrame.origin.x = self.safeAreaInsets.left;
```
ianegordon: Do we want to set it equal to safeAreaInsets or += safeAreaInsets?

components/NavigationBar/src/MDCNavigationBar.m
```diff
@@ -296,6 +306,12 @@ - (void)layoutSubviews {
   CGRect textFrame = UIEdgeInsetsInsetRect(self.bounds, textInsets);
   textFrame.origin.x += _leadingButtonBar.frame.size.width;
   textFrame.size.width -= _leadingButtonBar.frame.size.width + _trailingButtonBar.frame.size.width;
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
```
ianegordon, line 320: If we are subtracting the leading/trailingButtonBar against textFrame above, which have already been adjusted by safeAreaInsets above, do we need to repeat the adjustment here?

ianegordon (changes requested):

components/NavigationBar/src/MDCNavigationBar.m
romoore: Since the left/right safeAreaInsets may not be symmetric, do we need to account for RTL/LTR manually in these computations?

components/NavigationBar/src/MDCNavigationBar.m
```diff
@@ -281,13 +281,23 @@ - (void)layoutSubviews {
...
   _leadingButtonBar.frame = MDCRectFlippedForRTL(leadingButtonBarFrame, CGRectGetWidth(self.bounds),
                                                  self.mdc_effectiveUserInterfaceLayoutDirection);
 
   CGSize trailingButtonBarSize = [_trailingButtonBar sizeThatFits:self.bounds.size];
   CGRect trailingButtonBarFrame =
       CGRectMake(CGRectGetWidth(self.bounds) - trailingButtonBarSize.width, CGRectGetMinY(self.bounds),
                  trailingButtonBarSize.width, trailingButtonBarSize.height);
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    trailingButtonBarFrame.origin.x -= self.safeAreaInsets.right;
```
romoore: Same LTR/RTL concern here.

components/NavigationBar/src/MDCNavigationBar.m
```diff
@@ -296,6 +306,12 @@ - (void)layoutSubviews {
   CGRect textFrame = UIEdgeInsetsInsetRect(self.bounds, textInsets);
   textFrame.origin.x += _leadingButtonBar.frame.size.width;
   textFrame.size.width -= _leadingButtonBar.frame.size.width + _trailingButtonBar.frame.size.width;
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    textFrame.origin.x += self.safeAreaInsets.left;
```
romoore, line 321: Same question again for LTR/RTL layout.

components/NavigationBar/src/MDCNavigationBar.m
chuga: leadingButtonBarFrame.origin.x is 0 as per the line above. I can change it to += if you want, but that isn't necessarily a better approach imo.

components/NavigationBar/src/MDCNavigationBar.m
chuga: `textFrame.origin.x += _leadingButtonBar.frame.size.width` doesn't take into account the origin of the _leadingButtonBar frame :/ I'm afraid that CGRectGetMinX on that frame wouldn't be correct as the leadingButtonBar will be on the right for RTL languages.

romoore (commented):
Sorry, forgot to submit my review hours ago!

components/NavigationBar/src/MDCNavigationBar.m
chuga: Not here, as textFrame is never flipped.

chuga:
PTAL

components/NavigationBar/src/MDCNavigationBar.m
romoore: Ack.

romoore (approved):
LGTM

components/NavigationBar/src/MDCNavigationBar.m
```diff
@@ -278,12 +278,24 @@ - (NSInteger)indexOfAccessibilityElement:(id)element {
 - (void)layoutSubviews {
   [super layoutSubviews];
 
+  UIEdgeInsets rtlFriendlySafeAreaInsets;
```
ianegordon: super-nit: Don't lowercase initialisms.
s/rtlFriendlySafeAreaInsets/RTLFriendlySafeAreaInsets/

ianegordon (changes requested):

components/NavigationBar/src/MDCNavigationBar.m
```diff
@@ -278,16 +278,38 @@ - (NSInteger)indexOfAccessibilityElement:(id)element {
 - (void)layoutSubviews {
   [super layoutSubviews];
 
+  UIEdgeInsets RTLFriendlySafeAreaInsets;
```
ianegordon: UIEdgeInsets is a struct not an id, so we need to initialize.
UIEdgeInsets RTLFriendlySafeAreaInsets = UIEdgeInsetsZero;

ianegordon (changes requested):

components/NavigationBar/src/MDCNavigationBar.m
ianegordon: Ack.

ianegordon (approved):

ianegordon:
LGTM, please merge when Travis is happy!

PR #2071, opened 2017-10-01
[TextFields] iOS 11 example bugs fixed

willlarche:


willlarche:
Needs Xcode 8 support. Adding it now.

components/TextFields/examples/TextFieldExample.swift
```diff
@@ -189,6 +189,55 @@ final class TextFieldSwiftExample: UIViewController {
                                                   metrics: nil,
                                                   views: views)
 
+    #if swift(>=4.0)
+      if #available(iOS 11.0, *) {
```
codeman7: I think this should be changed to #if swift(>=3.2) until we move everything to 4.0 as the standard.

See the note.

components/TextFields/examples/TextFieldFilledExample.swift
```diff
@@ -188,6 +188,55 @@ final class TextFieldFilledSwiftExample: UIViewController {
                                                   options: [],
                                                   metrics: nil,
                                                   views: views)
+    #if swift(>=4.0)
```
codeman7: Same as above.

components/TextFields/examples/MultilineTextFieldLegacyExample.m
```diff
@@ -158,6 +157,61 @@ - (void)viewDidLoad {
...
+      [NSLayoutConstraint constraintWithItem:multilineTextFieldCharMaxFullWidth
+                                   attribute:NSLayoutAttributeBottom
+                                   relatedBy:NSLayoutRelationEqual
+                                      toItem:self.scrollView.contentLayoutGuide
+                                   attribute:NSLayoutAttributeBottomMargin
+                                  multiplier:1
+                                    constant:-20]
+    ]];
+  } else {
+    [NSLayoutConstraint activateConstraints:@[
```
romoore, line 183: The two "elses" look the same to me and I see this is a pattern repeated throughout these examples. I would prefer to see either:
1. Make a function/method that handles the "else" case so we don't have 2 updates in the future
2. Make it a single if/else with macros protecting the else?
```objectivec
#if __IPHONE_11
  if (@available) {
    ... do iOS 11 stuff...
  }
  else {
#endif
  ... do iOS 8-10 stuff....
#if __IPHONE_11
  } // @available else
#endif
```

components/TextFields/examples/MultilineTextFieldLegacyExample.m
```diff
@@ -197,6 +251,15 @@ - (void)viewDidLoad {
                     [[note.userInfo valueForKey:UIKeyboardFrameEndUserInfoKey] CGRectValue];
                 self.scrollView.contentInset = UIEdgeInsetsMake(0, 0, CGRectGetHeight(frame), 0);
               }];
+  [[NSNotificationCenter defaultCenter]
+      addObserverForName:UIKeyboardWillChangeFrameNotification
+                  object:nil
+                   queue:[NSOperationQueue mainQueue]
+              usingBlock:^(NSNotification *_Nonnull note) {
```
romoore: Nit: I would prefer to avoid the block syntax with NSNotificationCenter, does it make sense to use a selector instead?

components/TextFields/examples/TextFieldExample.swift
```diff
@@ -189,6 +189,55 @@ final class TextFieldSwiftExample: UIViewController {
...
+                      NSLayoutConstraint(item: message,
+                                         attribute: .bottom,
+                                         relatedBy: .equal,
+                                         toItem: scrollView,
+                                         attribute: .bottomMargin,
+                                         multiplier: 1,
+                                         constant: -20)]
+    }
+      #else
+      constraints += [NSLayoutConstraint(item: name,
```
romoore: Same suggestion as above if possible for swift due to the duplicate elses.

components/TextFields/examples/TextFieldManualLayoutLegacyExample.swift
```diff
@@ -147,11 +152,10 @@ final class TextFieldManualLayoutLegacySwiftExample: UIViewController {
 
   func setupScrollView() {
     view.addSubview(scrollView)
-    scrollView.frame = view.bounds
 
     scrollView.contentSize =
       CGSize(width: scrollView.bounds.width - 2 * LayoutConstants.largeMargin,
-             height: 372.0)
+             height: 500.0)
```
romoore, line 162: Question: Just curious about this magic number? If we're trying to get "bigger than the screen" does it make sense to check the VC's height and double/triple?

components/TextFields/examples/TextFieldOutlinedExample.swift
```diff
@@ -192,6 +192,55 @@ final class TextFieldOutlinedSwiftExample: UIViewController {
...
+            multiplier: 1,
+            constant: 20),
+            NSLayoutConstraint(item: message,
+            attribute: .bottom,
+            relatedBy: .equal,
+            toItem: scrollView.contentLayoutGuide,
+            attribute: .bottomMargin,
+            multiplier: 1,
+            constant: -20)]
+      } else {
```
romoore: Formatting Nit: Did the "+=" make this format oddly?

components/TextFields/examples/MultilineTextFieldLegacyExample.m
```diff
@@ -197,6 +251,15 @@ - (void)viewDidLoad {
...
+                CGRect frame =
+                    [[note.userInfo valueForKey:UIKeyboardFrameEndUserInfoKey] CGRectValue];
+                self.scrollView.contentInset = UIEdgeInsetsMake(0, 0, CGRectGetHeight(frame), 0);
```
romoore: You may need to weakify - I believe `self` will be captured in this block.

romoore (changes requested):
Concern about retain cycle safety in a notification block and a little push-back on the if/else/else structure.  Others are nits.

components/TextFields/examples/TextFieldExample.swift
willlarche: This isn't actually checking for Swift. It's the most appropriate of the compiler conditional functions. A check for Swift 4 will ensure this compile only on Xcode 9 and the else will compile on Xcode 8.

willlarche: https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Statements.html#//apple_ref/doc/uid/TP40014097-CH33-ID538

components/TextFields/examples/MultilineTextFieldLegacyExample.m
willlarche: Well, here's the thing, aside from kitchen sink, this is all educational code so I don't want to get too clever. This is the minimum amount of fancy footwork needed to make it all work.

components/TextFields/examples/MultilineTextFieldLegacyExample.m
willlarche: Since it's educational code, I wanted it to stay as simple as possible. Also, developers are moving to the block syntax for notifications so this will seem familiar to them.

components/TextFields/examples/TextFieldManualLayoutLegacyExample.swift
willlarche: It's not bigger than the screen; it's enough space for all the text fields' heights put together. 

I don't like it anymore than you do but we have users that do their layout this way. :)

components/TextFields/examples/TextFieldOutlinedExample.swift
willlarche: Oh, totally. Thank you.

components/TextFields/examples/MultilineTextFieldLegacyExample.m
willlarche: But this is what I mentioned in my email. We all need to agree on the approach. Put your suggestion in that thread too.

components/TextFields/examples/MultilineTextFieldLegacyExample.m
romoore: Ok. I had learned that the block API was more dangerous, so I've spent a bit of time migrating code away in the past.

components/TextFields/examples/TextFieldManualLayoutLegacyExample.swift
romoore: Got it. Thanks!

components/TextFields/examples/MultilineTextFieldLegacyExample.m
willlarche: No. This one is fine: https://stackoverflow.com/questions/20030873/always-pass-weak-reference-of-self-into-block-in-arc

components/TextFields/examples/TextFieldExample.swift
brianjmoore: Agree with OP here: Checking Swift >=3.2 will do the same thing and will allow Swift 3 code to use iOS 11 API safely.

willlarche:
For true. Changing.

components/TextFields/examples/TextFieldExample.swift
willlarche: @codeman7 and @brianjmoore are right. Changing now.

components/TextFields/examples/MultilineTextFieldLegacyExample.m
romoore: You're right that there isn't a retain cycle here. However, the block will be copied (onto the heap) when the notification center retains it. `self` will be in scope and will be retained by the notification center. As a global singleton, the notification center will retain the example VC until the application exits.  I pulled the PR and tested and was able to confirm that the VC is leaked each time it's pushed and popped.

I believe a safe approach is to register for the notifications in a lifecycle method like `viewWillAppear` and unregister in its sibling `viewDidDisappear`. Moving to the `@selector` API wont help things on iOS 8 (but [it will help on iOS 9+](https://useyourloaf.com/blog/unregistering-nsnotificationcenter-observers-in-ios-9/)).  You would still need to handle unregistering manually.

components/TextFields/examples/MultilineTextFieldLegacyExample.m
romoore: Ok. I put it in the thread, and am OK with this as-is. Thanks!

components/TextFields/examples/MultilineTextFieldLegacyExample.m
willlarche: Ok, great. That actually makes the most sense to me. I guess the example I found in SO was using it in a block that was in local scope, not global scope.

willlarche: I __wish__ this was caught by the static analyzer.

willlarche: I changed them all to selectors and added removeObservers!

components/TextFields/examples/MultilineTextFieldLegacyExample.m
chuga: What I did in a similar situation was to use a separate method for setting up the constraints and return if we do have the iOS 11 API available:
https://github.com/material-components/material-components-ios/pull/2021/files#diff-3f90132dc1e2e304c7961cb3e9d58a40R304

willlarche: Return early. That's an unexpected twist. I think for examples, it would be too easy to miss.

romoore (approved):
No more leaks that I could see. :)

components/TextFields/examples/MultilineTextFieldLegacyExample.m
willlarche: Done.

components/TextFields/examples/TextFieldFilledExample.swift
willlarche: Done.

components/TextFields/examples/TextFieldExample.swift
```diff
@@ -189,6 +193,55 @@ final class TextFieldSwiftExample: UIViewController {
...
+                                         constant: 20),
+                      NSLayoutConstraint(item: message,
+                                         attribute: .bottom,
+                                         relatedBy: .equal,
+                                         toItem: scrollView,
+                                         attribute: .bottomMargin,
+                                         multiplier: 1,
+                                         constant: -20)]
+    }
+      #else
```
ianegordon: nit: This should be aligned 4, not 6.  It may be keying of the #available above.  If this is what the linter / Xcode demands, so be it.
(Here and below)

ianegordon (approved):
I'm hoping we will be able to drop support for Xcode 8 sooner, rather than later and clean up this trickiness.

components/TextFields/examples/TextFieldExample.swift
willlarche: Done

PR #2072, opened 2017-10-02
[PageControl] Fix bug with possible out of bounds array lookup.

kolodny:
To replicate the crashing behavior make the following change in [PageControlTypicalUseExample.swift](https://github.com/material-components/material-components-ios/blob/develop/components/PageControl/examples/PageControlTypicalUseExample.swift#L23-L30):

```diff
-  static let pageColors = [
-    MDCPalette.cyan.tint300,
-    MDCPalette.cyan.tint500,
-    MDCPalette.cyan.tint700,
-    MDCPalette.cyan.tint300,
-    MDCPalette.cyan.tint500,
-    MDCPalette.cyan.tint700
-  ]
+  static let pageColors: [UIColor] = []
```

### DIFF
--- a/components/PageControl/src/MDCPageControl.m
+++ b/components/PageControl/src/MDCPageControl.m
@@ -184,15 +184,17 @@ static inline CGFloat normalizeValue(CGFloat value, CGFloat minRange, CGFloat ma
                                        toEndPoint:endPoint
                                        completion:completionBlock];
   } else {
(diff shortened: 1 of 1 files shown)


randallli:
Sorry for the late review. 
It looks like someone else also has a fix for the same problem.
https://github.com/material-components/material-components-ios/pull/2132
Can you check to make sure that PR also fixes your bug?
If not please reopen.

PR #2073, opened 2017-10-02
[TextFields] Fixing bug of bad intrinsicContentSize.

willlarche:
simulator screen shot - iphone 8 plus - 2017-10-01 at 22 48 07
simulator screen shot - iphone 8 plus - 2017-10-01 at 22 48 02
simulator screen shot - iphone 8 plus - 2017-10-01 at 22 47 57

### DIFF
--- a/components/TextFields/src/private/MDCTextInputCommonFundament.m
+++ b/components/TextFields/src/private/MDCTextInputCommonFundament.m
@@ -132,6 +132,7 @@ static inline UIColor *MDCTextInputUnderlineColor() {
 
     // Initialize elements of UI
     [self setupPlaceholderLabel];
+    [self setupUnderlineView];
     [self setupClearButton];
     [self setupUnderlineLabels];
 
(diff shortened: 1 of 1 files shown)


components/TextFields/src/private/MDCTextInputCommonFundament.m
```diff
@@ -534,20 +533,32 @@ - (void)updateClearButton {
 }
 
 - (void)updateClearButtonConstraints {
-  CGFloat constant = MDCTextInputClearButtonImageSquareWidthHeight * [self clearButtonAlpha];
-  if (self.clearButtonWidth.constant != constant) {
-    self.clearButtonWidth.constant = constant;
-    [self.textInput invalidateIntrinsicContentSize];
+  BOOL shouldInvalidateSize = NO;
+  CGFloat widthConstant = MDCTextInputClearButtonImageSquareWidthHeight * [self clearButtonAlpha];
```
romoore, line 537: Nit/unrelated: This is a bit surprising on first read. Would it make sense in a follow-up CR to check `clearButtonAlpha` for non-zero or zero and then make the multiplier 1/0?

romoore (approved):
LGTM.

PR #2075, opened 2017-10-02
[Tabs] Unit tests for two-line text-only tabs

romoore:
Exposing `titleLabel` in MDCItemBarCell for unit testing so that the
two-line behavior can be tested. This test already caught a bug in the
original PR before submission.

Related to #2025

### DIFF
--- a/components/Tabs/src/private/MDCItemBarCell.m
+++ b/components/Tabs/src/private/MDCItemBarCell.m
@@ -15,6 +15,7 @@
  */
 
 #import "MDCItemBarCell.h"
+#import "MDCItemBarCell+Private.h"
 
 #import "MDCItemBarStringConstants.h"
 #import "MDCItemBarStyle.h"
@@ -52,7 +53,6 @@ const CGFloat kSelectedNavigationImageYOffset = -2;
 static const NSTimeInterval kSelectionAnimationDuration = 0.3f;
 
 @implementation MDCItemBarCell {
-  UILabel *_titleLabel;
   UIImageView *_imageView;
   UILabel *_badgeLabel;
   MDCInkTouchController *_inkTouchController;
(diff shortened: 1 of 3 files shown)


components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -230,7 +230,7 @@ - (void)layoutSubviews {
   imageBounds.size = kImageSize;
   imageCenter.x = CGRectGetMidX(contentBounds);
 
-  CGSize titleSize = [_titleLabel sizeThatFits:contentBounds.size];
+  CGSize titleSize = [self.titleLabel sizeThatFits:contentBounds.size];
```
brianjmoore, line 233: Here and below - why change to property accesses?

Direct use of the backing instance variable before was intentional for consistency and predictability. (I thought that was part of the Obj-C style guide TBH, but I don't see it in there so I guess it's just something that I picked up at some point?).

Elsewhere we can argue the style merits here and decide on a consistent policy for MDC, but for this PR I think that we shouldn't change the code unless it's necessary or there's style guidance that points in a specific direction.

romoore: I believe Apple considers using property accessors to be a [best practice](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/EncapsulatingData/EncapsulatingData.html). Under the section header "Most Properties Are Backed By Instance Variables":

> Although it’s **best practice for an object to access its own properties using accessor methods or dot syntax**, it’s possible to access the instance variable directly from any of the instance methods in a class implementation.  (*Emphasis added*)

Since I was changing to a property to export the value, I thought it was reasonable to update to Apple's recommendation except in the methods called from `init`.  There are a lot of good reasons to use a property in general (KVO, atomicity, custom setter/getter code), but in this case it was just because I was moving from an iVar to a property.   It's also not the only property using accessors (c.f. `image`, `title`, and `badgeValue`).

brianjmoore: Nice quote! SGTM. I would have a hard time arguing against using property accessors, especially given it's also The Way of the Swift. It's also the style I used about 4 years ago.

brianjmoore (approved):

ianegordon (approved):

PR #2078, opened 2017-10-02
[AppBar] Use safeAreaLayoutGuide.topAnchor instead of fixed status bar height

chuga:
This is an iteration on my previous fix. Turns out that constraining the headerStackView to the top of the view using a fixed number isn't enough when the safe area can change due to rotations or making/receiving a call on non-iPhoneX devices. This is a more robust solution as it uses the new safeAreaLayoutGuide API, which changes as the safe area changes.

### DIFF
--- a/components/AppBar/src/MDCAppBar.m
+++ b/components/AppBar/src/MDCAppBar.m
@@ -310,34 +310,33 @@ static NSString *const kMaterialAppBarBundle = @"MaterialAppBar.bundle";
 
 #if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
   if (@available(iOS 11.0, *)) {
-    NSLayoutYAxisAnchor *topAnchor = self.view.safeAreaLayoutGuide.topAnchor;
-    NSLayoutConstraint *topConstraint =
-        [self.headerStackView.topAnchor constraintEqualToAnchor:topAnchor constant:0];
-    topConstraint.active = YES;
+    [self.headerStackView.topAnchor constraintEqualToAnchor:self.view.safeAreaLayoutGuide.topAnchor
+                                                   constant:0].active = YES;
   } else {
 #endif
-  NSLayoutConstraint *topConstraint =
-      [NSLayoutConstraint constraintWithItem:self.headerStackView
-                                   attribute:NSLayoutAttributeTop
-                                   relatedBy:NSLayoutRelationEqual
-                                      toItem:self.view
-                                   attribute:NSLayoutAttributeTop
-                                  multiplier:1
-                                    constant:kPreIOS11StatusBarHeight];
-    topConstraint.active = YES;
+  if (@available(iOS 9.0, *)) {
+    [self.headerStackView.topAnchor constraintEqualToAnchor:self.topLayoutGuide.bottomAnchor
+                                                   constant:0].active = YES;
+  } else {
+    [NSLayoutConstraint constraintWithItem:self.headerStackView
+                                 attribute:NSLayoutAttributeTop
+                                 relatedBy:NSLayoutRelationEqual
+                                    toItem:self.view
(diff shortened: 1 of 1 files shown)


components/AppBar/src/MDCAppBar.m
```diff
@@ -308,24 +308,36 @@ - (void)viewDidLoad {
...
 
-  CGFloat topMargin = kPreIOS11StatusBarHeight;
 #if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
   if (@available(iOS 11.0, *)) {
-    // Starting from iOS 11, the top margin should be the actual status bar height.
-    // This is because the flexible header could be smaller in height when in landscape mode
-    // due to the status bar being hidden by default on some devices (like the iPhone X).
-    topMargin = [UIApplication mdc_safeSharedApplication].statusBarFrame.size.height;
+    NSLayoutYAxisAnchor *topAnchor = self.view.safeAreaLayoutGuide.topAnchor;
+    NSLayoutConstraint *topConstraint =
```
romoore: Super-nit: Consider a one-liner:
```objectivec
[self.headerStackView.topAnchor constraintEqualToAnchor:topAnchor constant:0].active = YES;
```

components/AppBar/src/MDCAppBar.m
```diff
@@ -308,24 +308,36 @@ - (void)viewDidLoad {
...
+#endif
+  NSLayoutConstraint *topConstraint =
+      [NSLayoutConstraint constraintWithItem:self.headerStackView
+                                   attribute:NSLayoutAttributeTop
+                                   relatedBy:NSLayoutRelationEqual
+                                      toItem:self.view
+                                   attribute:NSLayoutAttributeTop
+                                  multiplier:1
+                                    constant:kPreIOS11StatusBarHeight];
+    topConstraint.active = YES;
```
romoore: Consider using `topLayoutGuide.bottomAnchor` for iOS 9/10. I know it creates a third condition (iOS 8), but I believe it should adjust for the in-call status bar size.

https://developer.apple.com/documentation/uikit/uiviewcontroller/1621367-toplayoutguide

romoore (approved):

components/AppBar/src/MDCAppBar.m
chuga: Done

components/AppBar/src/MDCAppBar.m
chuga: Done. It seems to be working fine with the non-iOS 9 as well though.

chuga (commented):
Thanks for reviewing

romoore (approved):
Looks Even Goodlier To Me  (LEGTM)!

romoore (changes requested):
Gah, @available not available in Xcode 8.

chuga:
😓  of course, that would've been too easy.

romoore:
I think it might be better to go back to your original if/else solution. Handling all 3 iOS cases for 2 different SDKs is probably overdoing it at this point.

The original NSConstraint approach adjusted for the in-call status bar during testing on iOS 8.3 (simulator) using Xcode 9.  It also worked for iOS 9.3 (simulator) using Xcode 8.

```objectivec
    [self.headerStackView.topAnchor constraintEqualToAnchor:self.view.safeAreaLayoutGuide.topAnchor
                                                   constant:0].active = YES;
  } else {
#endif
    [NSLayoutConstraint constraintWithItem:self.headerStackView
                                 attribute:NSLayoutAttributeTop
                                 relatedBy:NSLayoutRelationEqual
                                    toItem:self.view
                                 attribute:NSLayoutAttributeTop
                                multiplier:1
                                  constant:kPreIOS11StatusBarHeight].active = YES;
  }
#endif
```

romoore (approved):

PR #2080, opened 2017-10-03
[Examples | Catalog] Updates all components to Swift 4.

codeman7:
Convert all swift code to Swift 4. Should be backward compatible with XCode 8 except for a few things, need to talk to someone on how we want to deal with them. 

Doesn't close any issues.

Had trouble refreshing my fork so had to reclone that's why I have another pull request. Also, before this is merged I have a couple issues.


ianegordon:
Hi @codeman7 !  Thanks for putting this together, but unfortunately we cannot migrate to Swift 4 until we drop support for Xcode 8. 
I've added a tracking bug #2087.

codeman7:
I believe everything still works with XCode 8. Once we drop support then all the #if swift (>=4.0) statements can be dropped.

ianegordon:
Thank you @codeman7 and @willlarche for the PR and making sure it is backwards compatible.

willlarche:
So much, great work! Thank you @codeman7 !!

PR #2088, opened 2017-10-04
[BottomNavigation] Adding bottom navigation component

jgunaratne:
A stylized bottom navigation component.

simulator screen shot - iphone 8 plus - 2017-10-04 at 17 11 47


components/BottomNavigation/src/private/MDCBottomNavigationCell.h
```diff
@@ -0,0 +1,35 @@
...
+
+@property(nonatomic, assign) BOOL selected;
+
+@property(nonatomic, copy) NSString *badgeValue;
+@property(nonatomic, copy) NSString *title;
+
+@property(nonatomic, strong) UIButton *button;
+@property(nonatomic, strong) UIImage *image;
+
+@property(nonatomic, strong) UIColor *badgeColor;
```
ianegordon: UIAppearance these.

components/BottomNavigation/src/MDCBottomNavigationView.h
```diff
@@ -0,0 +1,52 @@
...
+ unselected items are hidden.
+ */
+@interface MDCBottomNavigationView : UIView
+
+/**
+ An array of UITabBarItems that is used to populate bottom navigation view content. The array must
+ contain a minimum of three items with a maxiumum of up to five items. The first item is selected by
+ default.
+ */
+@property(nonatomic, copy, nonnull) NSArray<UITabBarItem *> *navBarItems;
```
ianegordon: To match UITabBar:
@property(nullable, nonatomic, copy) NSArray<UITabBarItem *> *items;        // get/set visible UITabBarItems. default is nil. changes not animated. shown in order

components/BottomNavigation/src/MDCBottomNavigationView.h
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ Selects an item contained in the bottom navigation view.
+
+ @param item UITabBarItem to select from the navBarItems array.
+ */
+- (void)selectItem:(nonnull UITabBarItem *)item;
```
ianegordon: To match UITabBar:
@property(nullable, nonatomic, weak) UITabBarItem *selectedItem; // will show feedback based on mode. default is nil

components/BottomNavigation/src/MDCBottomNavigationView.h
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ Color of selected item. Applies color to items' icons and text.
+ */
+@property(nonatomic, strong, nonnull) UIColor *selectedColor;
+
+/**
+ Color of unselected items. Applies color to items' icons. Text is not displayed in unselected mode.
+ */
+@property(nonatomic, strong, nonnull) UIColor *unselectedColor;
```
ianegordon: To match UITabBar:
@property (nonatomic, readwrite, copy, nullable) UIColor *unselectedItemTintColor NS_AVAILABLE_IOS(10_0) UI_APPEARANCE_SELECTOR;

In stead of selectedColor, consider:
@property (nonatomic, readwrite, copy, nullable) UIColor *selectedItemTintColor NS_AVAILABLE_IOS(10_0) UI_APPEARANCE_SELECTOR;

components/BottomNavigation/src/private/MDCBottomNavigationCell.h
```diff
@@ -0,0 +1,35 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+@interface MDCBottomNavigationCell : UIView
```
ianegordon: If you are not inheriting from UICollectionViewCell, I wouldn't call this MDCXYZCell.
MDCBottomNavigationItemView?

Should this inherit from UIControl instead of UIView since it handles user interaction?

Naming is hard.

components/BottomNavigation/src/private/MDCBottomNavigationCellBadge.m
```diff
@@ -0,0 +1,120 @@
...
+#import "MaterialMath.h"
+
+static const CGFloat kMDCBottomNavigationCellBadgeAngle = 90.f;
+static const CGFloat kMDCBottomNavigationCellBadgeFontSize = 10.f;
+static const CGFloat kMDCBottomNavigationCellBadgeXPadding = 8.f;
+static const CGFloat kMDCBottomNavigationCellBadgeYPadding = 2.f;
+
+@implementation MDCBottomNavigationCellBadge
+
+- (instancetype)init {
```
ianegordon: In general, we typically only implement designated initializers.
So in this case, UIView, it would be initWithFrame: and initWithCoder:

components/BottomNavigation/src/private/MDCBottomNavigationCellBadge.m
```diff
@@ -0,0 +1,120 @@
...
+  self = [super initWithFrame:CGRectZero];
+  if (self) {
+    [self commonInitMDCBottomNavigationCellBadge];
+  }
+  return self;
+}
```
ianegordon: add initwithcoder

components/BottomNavigation/src/private/MDCBottomNavigationCellBadge.m
```diff
@@ -0,0 +1,120 @@
...
+  _xPadding = kMDCBottomNavigationCellBadgeXPadding;
+  _yPadding = kMDCBottomNavigationCellBadgeYPadding;
+
+  _badgeCircleWidth = _badgeValueLabel.bounds.size.width + _xPadding;
+  _badgeCircleHeight = _badgeValueLabel.bounds.size.height + _yPadding;
+
+  if (_badgeCircleWidth < _badgeCircleHeight) {
+    _badgeCircleWidth = _badgeCircleHeight;
+  }
+  self.frame = CGRectMake(self.frame.origin.x,
```
ianegordon: Using CGRect .origin and .size directly is dangerous.
Prefer CGRectGetWidth().

components/BottomNavigation/src/private/MDCBottomNavigationCellBadge.m
```diff
@@ -0,0 +1,120 @@
...
+  [path addLineToPoint:CGPointMake(badgeRadius, 0)];
+
+  [path closePath];
+  self.badgeLayer.path = path.CGPath;
+}
+
+- (void)setBadgeValue:(NSString *)badgeValue {
+  _badgeValue = badgeValue;
+  _badgeValueLabel.text = badgeValue;
+  [self sizeBadge];
```
ianegordon: [self setNeedsLayout] instead of size & draw.

components/BottomNavigation/src/private/MDCBottomNavigationCellBadge.m
```diff
@@ -0,0 +1,120 @@
...
+                          _badgeCircleHeight);
+}
+
+- (void)layoutSubviews {
+  [super layoutSubviews];
+  [self sizeBadge];
+  [self drawBadge];
+}
+
+- (void)drawBadge {
```
ianegordon: This is a bit of a misnomer as we aren't actually drawing anything, we are just calculating and setting the path.

Move this to sizeBadge?

components/BottomNavigation/src/MDCBottomNavigationView.h
```diff
@@ -0,0 +1,52 @@
...
+#import <UIKit/UIKit.h>
+
+/**
+ A bottom navigation view.
+
+ The bottom navigation view is a bar docked at the bottom of the screen with tappable items. Only
+ one item can be selected at at time. The selected item's title text is displayed. Title text for
+ unselected items are hidden.
+ */
+@interface MDCBottomNavigationView : UIView
```
ianegordon: MDCBottomNavigationBar to match UITabBar?

components/BottomNavigation/src/MDCBottomNavigationView.m
```diff
@@ -0,0 +1,258 @@
...
+static NSString *const kMDCBottomNavigationViewBadgeColorString = @"badgeColor";
+static NSString *const kMDCBottomNavigationViewBadgeValueString = @"badgeValue";
+static NSString *const kMDCBottomNavigationViewImageString = @"image";
+static NSString *const kMDCBottomNavigationViewTitleString = @"title";
+static NSString *const kMDCBottomNavigationViewNewString = @"new";
+
+@interface MDCBottomNavigationView ()
+
+@property(nonatomic, strong) NSMutableArray<MDCBottomNavigationCell *> *navBarCells;
+@property(nonatomic, assign) UIUserInterfaceLayoutDirection layoutDirection;
```
ianegordon: I don't know that I would cache this, I would call self.mdc_effectiveUserInterfaceLayoutDirection where appropriate.

components/BottomNavigation/src/MDCBottomNavigationView.m
```diff
@@ -0,0 +1,258 @@
...
+
+- (void)commonMDCBottomNavigationViewInit {
+  self.backgroundColor = [UIColor whiteColor];
+  self.autoresizingMask = (UIViewAutoresizingFlexibleTopMargin | UIViewAutoresizingFlexibleWidth);
+  
+  _selectedColor = [UIColor blackColor];
+  _unselectedColor = [UIColor grayColor];
+  _layoutDirection = self.mdc_effectiveUserInterfaceLayoutDirection;
+
+  // Content in bottom navigation always uses the width of the device portrait orientation width.
```
ianegordon: We really aren't supporting landscape mode?

components/BottomNavigation/src/MDCBottomNavigationView.m
```diff
@@ -0,0 +1,258 @@
...
+      [cell setSelected:NO animated:YES];
+    }
+  }
+  MDCBottomNavigationCell *cell = (MDCBottomNavigationCell *)button.superview;
+  [cell setSelected:YES animated:YES];
+}
+
+- (void)selectItem:(UITabBarItem *)item {
+  NSInteger i = 0;
+  for (UITabBarItem *tabBarItem in self.navBarItems) {
```
ianegordon: If we are using the counter as an index into another array, consider using a for loop with an explicit index.
for (NSInteger itemIndex = 0; itemIndex < self.navBarItems.count; ++itemIndex) {
  if (item == self.navBarItems[itemIndex]) {
    self.navBarCell[itemIndex].selected = YES;
  } else {
    self.navBarCell[itemIndex].selected = NO;
  }
}

components/BottomNavigation/src/MDCBottomNavigationView.m
```diff
@@ -0,0 +1,258 @@
...
+    if (item == tabBarItem) {
+      [self.navBarCells[i] setSelected:YES];
```
ianegordon: Prefer property style methods where possible.
self.navBarCells[i].selected = YES;

ianegordon (dismissed):

components/BottomNavigation/src/private/MDCBottomNavigationCellBadge.h
```diff
@@ -0,0 +1,31 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+@interface MDCBottomNavigationCellBadge : UIView
```
ianegordon: Can we not do this with a UILabel with a cornerRadius?

components/BottomNavigation/src/MDCBottomNavigationView.m
```diff
@@ -0,0 +1,258 @@
...
+  self.autoresizingMask = (UIViewAutoresizingFlexibleTopMargin | UIViewAutoresizingFlexibleWidth);
+  
+  _selectedColor = [UIColor blackColor];
+  _unselectedColor = [UIColor grayColor];
+  _layoutDirection = self.mdc_effectiveUserInterfaceLayoutDirection;
+
+  // Content in bottom navigation always uses the width of the device portrait orientation width.
+  CGSize appSize = [[UIScreen mainScreen] applicationFrame].size;
+  CGFloat minDimension = MIN(appSize.width, appSize.height);
+  CGRect adjustedFrame = CGRectMake(0, 0, minDimension, self.frame.size.height);
```
ianegordon: self.frame could easily be CGRectZero.
Choose a sane default?

components/BottomNavigation/src/MDCBottomNavigationView.m
```diff
@@ -0,0 +1,258 @@
...
+  _navBarContainerView = [[UIView alloc] initWithFrame:adjustedFrame];
```
ianegordon: If we are autosizing based on the parent, just initWithFrame:CGRectZero and have layout deal with sizing?

components/BottomNavigation/src/MDCBottomNavigationView.m
```diff
@@ -0,0 +1,258 @@
...
+  self.navBarContainerView.center = CGPointMake(CGRectGetMidX(self.bounds),
+                                                CGRectGetMidY(self.bounds));
+}
+
+- (void)layoutCellsWithLayoutDirection:(UIUserInterfaceLayoutDirection)layoutDirection {
+  NSInteger numItems = self.navBarItems.count;
+  NSInteger i = 0;
+  CGSize navBarSize = self.navBarContainerView.bounds.size;
+  CGFloat itemWidth = navBarSize.width / numItems;
+  if (layoutDirection == UIUserInterfaceLayoutDirectionRightToLeft) {
```
ianegordon: Most of our client developers will be going through this code in LeftToRight mode, so handle that case first.

Alternatively, consider laying out everything in standard LTR mode and mirror cells if required.

if (self.mdf_effectiveInterface == RTL) {
  for (cell in self.navBarCells) {
    cell.frame = MDFRectMirror(cell.frame, CGRectGetWidth(self.navBarContainerView.frame));
  }
}

components/BottomNavigation/src/MDCBottomNavigationView.h
jgunaratne: Makes sense. Changed name.

components/BottomNavigation/src/MDCBottomNavigationView.h
jgunaratne: Done.

components/BottomNavigation/src/MDCBottomNavigationView.h
jgunaratne: Done.

components/BottomNavigation/src/MDCBottomNavigationView.h
jgunaratne: Done.

components/BottomNavigation/src/MDCBottomNavigationView.m
jgunaratne: Done.

components/BottomNavigation/src/MDCBottomNavigationView.m
jgunaratne: We do support landscape mode. I've updated the comment to clarify. In landscape mode the items should not span the entire width.

components/BottomNavigation/src/MDCBottomNavigationView.m
jgunaratne: Done.

components/BottomNavigation/src/MDCBottomNavigationView.m
jgunaratne: Only using autosizing causes some layout issues when the view controller is pushed in landscape mode, and then the device orientation is changed.

components/BottomNavigation/src/MDCBottomNavigationView.m
jgunaratne: Changed to address LTR first.

components/BottomNavigation/src/MDCBottomNavigationView.m
jgunaratne: Done.

components/BottomNavigation/src/MDCBottomNavigationView.m
jgunaratne: Done.

components/BottomNavigation/src/private/MDCBottomNavigationCell.h
jgunaratne: Done.

components/BottomNavigation/src/private/MDCBottomNavigationCellBadge.m
jgunaratne: Done.

components/BottomNavigation/src/private/MDCBottomNavigationCellBadge.m
jgunaratne: Done.

components/BottomNavigation/src/private/MDCBottomNavigationCellBadge.m
jgunaratne: Done.

components/BottomNavigation/src/private/MDCBottomNavigationCellBadge.m
jgunaratne: Done.

components/BottomNavigation/src/private/MDCBottomNavigationCellBadge.m
jgunaratne: Done.

components/BottomNavigation/src/private/MDCBottomNavigationCell.h
jgunaratne: Renamed this class to MDCBottomNavigationItemView. However, I think this should remain a UIView rather than a UIControl since UIControlStates don't match what's in the design spec.

components/BottomNavigation/src/private/MDCBottomNavigationCellBadge.h
jgunaratne: Replaced with cornerRadius. Originally had a custom layer path to allow more customization by designers.

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,257 @@
...
+ limitations under the License.
+ */
+
+#import "MDCBottomNavigationBar.h"
+
+#import "MaterialMath.h"
+#import "MaterialRTL.h"
+#import "private/MDCBottomNavigationItemView.h"
+
+static const CGFloat kMDCBottomNavigationBarHeight = 72.f;
```
chuga: How does this work on an iPhone X?

jgunaratne: This has been updated to account for the iPhone X bottom inset when necessary.

simulator screen shot - iphone x - 2017-10-10 at 10 07 35
simulator screen shot - iphone x - 2017-10-10 at 10 07 40

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,267 @@
...
+  [self addSubview:_containerView];
+  _itemViews = [NSMutableArray array];
+}
+
+- (void)layoutSubviews {
+  [super layoutSubviews];
+
+  // Accommodate bottom inset for iPhone X.
+  CGFloat bottomInset = 0;
+  if (@available(iOS 11.0, *)) {
```
chuga: Sadly you need to wrap this in `#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
` to not break older versions of Xcode

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,267 @@
...
+    bottomInset = self.safeAreaInsets.bottom;
```
chuga: Do you need to re-layout when `safeAreaInsetsDidChange`?

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,267 @@
...
+  CGFloat heightWithInset = kMDCBottomNavigationBarHeight + bottomInset;
+  self.frame = CGRectMake(0, 0, CGRectGetWidth(superViewRect), heightWithInset);
+  self.center = CGPointMake(CGRectGetWidth(superViewRect) / 2,
+                            CGRectGetHeight(superViewRect) - heightWithInset / 2);
+
+  // The bottom navigation bar always spans the width of the device. However, the container view
+  // that holds the items has a fixed width based on the device's portrait view width. This allows
+  // the items to have a consistent distance from one another independent of device orientation.
+  CGFloat itemContainerWidth = MIN(CGRectGetWidth(superViewRect), CGRectGetHeight(superViewRect));
+  CGFloat containerOffsetY = (CGRectGetWidth(self.bounds) - itemContainerWidth) / 2;
```
chuga: nit: shouldn't this be containerOffsetX?

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,267 @@
...
+- (void)layoutSubviews {
+  [super layoutSubviews];
+
+  // Accommodate bottom inset for iPhone X.
+  CGFloat bottomInset = 0;
+  if (@available(iOS 11.0, *)) {
+    bottomInset = self.safeAreaInsets.bottom;
+  }
+
+  CGRect superViewRect = self.superview.bounds;
```
chuga: Do you need to account for safeAreaInsets.left and .right just in case in the future they're not 0 on portrait?

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,267 @@
...
+  CGFloat heightWithInset = kMDCBottomNavigationBarHeight + bottomInset;
+  self.frame = CGRectMake(0, 0, CGRectGetWidth(superViewRect), heightWithInset);
+  self.center = CGPointMake(CGRectGetWidth(superViewRect) / 2,
+                            CGRectGetHeight(superViewRect) - heightWithInset / 2);
+
+  // The bottom navigation bar always spans the width of the device. However, the container view
```
chuga: Unrelated comment: This behavior is super odd, and not how tab bars work on iOS 😞

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: Yes, should be x. Removed based on comment above.

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: Done.

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: Now calling layoutSubviews when safeAreaInsetsDidChange.

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: Added left and right insets.

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: Removing this behavior for now. Need to check with design about landscape mode.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.h
```diff
@@ -0,0 +1,40 @@
...
+@property(nonatomic, strong) UIButton *button;
+@property(nonatomic, strong) UIImage *image;
+
+@property(nonatomic, strong) UIColor *badgeColor UI_APPEARANCE_SELECTOR;
+@property(nonatomic, strong) UIColor *selectedItemTintColor UI_APPEARANCE_SELECTOR;
+@property(nonatomic, strong) UIColor *unselectedItemTintColor UI_APPEARANCE_SELECTOR;
+
+- (void)setSelected:(BOOL)selected animated:(BOOL)animated;
+
+- (void)setInterimSelected:(BOOL)interimSelected;
```
ianegordon: What is interimSelected / when would I prefer it over selected?

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
```diff
@@ -0,0 +1,229 @@
...
+
+- (void)setBadgeColor:(UIColor *)badgeColor {
+  _badgeColor = badgeColor;
+  self.badge.badgeColor = badgeColor;
+}
+
+- (void)setBadgeValue:(NSString *)badgeValue {
+  _badgeValue = badgeValue;
+  self.badge.badgeValue = badgeValue;
+  if (!_badgeValue) {
```
ianegordon: If badgeValue is nil or .length == 0 set to hidden?

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
```diff
@@ -0,0 +1,229 @@
...
+    self.badge.hidden = YES;
+  } else {
+    self.badge.hidden = NO;
+  }
+}
+
+- (void)setImage:(UIImage *)image {
+  _image = image;
+
+  self.selectedImage = [self colorizeImage:image color:self.selectedItemTintColor];
```
ianegordon: Rather than storing two images I would expect us to use UIImageView.tintColor for this.
self.iconImageView.image = image;
self.iconImageView.tintColor = self.unselectedTintColor;

On setSelected we would update iconImageView.tintColor.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
```diff
@@ -0,0 +1,229 @@
...
+                 font:(UIFont *)font
+     boundingRectSize:(CGSize)boundingRectSize {
+  CGRect rect = [text boundingRectWithSize:boundingRectSize
+                                   options:NSStringDrawingUsesLineFragmentOrigin
+                                attributes:@{ NSFontAttributeName:font }
+                                   context:nil];
+  return rect.size;
+}
+
+- (UIImage *)colorizeImage:(UIImage *)image color:(UIColor *)color {
```
ianegordon: nit: This doesn't refer to self, should likely be a class method.
However, we can probably use imageView.tintColor.  (See comment below)

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
```diff
@@ -0,0 +1,229 @@
...
+
+  self.selectedImage = [self colorizeImage:image color:self.selectedItemTintColor];
+  self.unselectedImage = [self colorizeImage:image color:self.unselectedItemTintColor];
+
+  self.iconImageView.image = self.unselectedImage;
+  [self.iconImageView sizeToFit];
+}
+
+- (void)setTitle:(NSString *)title {
+  _title = title;
```
ianegordon: _title = [title copy]

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
```diff
@@ -0,0 +1,229 @@
...
+}
+
+- (void)layoutSubviews {
+  [super layoutSubviews];
+
+  CGFloat labelHeight = [self sizeForText:self.title
+                                     font:self.label.font
+                         boundingRectSize:self.bounds.size].height;
+  self.label.frame = CGRectMake(0, 0, CGRectGetWidth(self.bounds), labelHeight);
+  [self setSelected:self.selected];
```
ianegordon: This call is suspicious.
We are essentially calling: self.selected = self.selected.

Move all layout code into this method and have setSelected: call setNeedsLayout?

ianegordon (changes requested):

components/BottomNavigation/src/MDCBottomNavigationBar.h
```diff
@@ -0,0 +1,70 @@
...
+
+ The bottom navigation bar is docked at the bottom of the screen with tappable items. Only one item
+ can be selected at at time. The selected item's title text is displayed. Title text for unselected
+ items are hidden.
+ */
+@interface MDCBottomNavigationBar : UIView
+
+/**
+ Configures when item titles should be displayed.
+ Default is MDCBottomNavigationBarTitleHideStateUnselect.
```
ianegordon: Consider:
Default behavior is to only show the title of the selected item.
(MDCBottomNavigationBarTitleHideStateUnselect doesn't exist)

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,286 @@
...
+                 0,
+                 CGRectGetWidth(superViewRect) - insets.left - insets.right,
+                 kMDCBottomNavigationBarHeight);
+  [self layoutItemViewsWithLayoutDirection:self.mdc_effectiveUserInterfaceLayoutDirection];
+}
+
+- (void)layoutItemViewsWithLayoutDirection:(UIUserInterfaceLayoutDirection)layoutDirection {
+  NSInteger numItems = self.items.count;
+  CGSize navBarSize = self.containerView.bounds.size;
+  CGFloat itemWidth = navBarSize.width / numItems;
```
ianegordon: Use CGRectGetWidth()

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,286 @@
...
+  self.containerView.frame =
+      CGRectMake(insets.left,
+                 0,
+                 CGRectGetWidth(superViewRect) - insets.left - insets.right,
+                 kMDCBottomNavigationBarHeight);
+  [self layoutItemViewsWithLayoutDirection:self.mdc_effectiveUserInterfaceLayoutDirection];
+}
+
+- (void)layoutItemViewsWithLayoutDirection:(UIUserInterfaceLayoutDirection)layoutDirection {
+  NSInteger numItems = self.items.count;
```
ianegordon: Early return if self.items.count == 0?

ianegordon (changes requested):

components/BottomNavigation/src/MDCBottomNavigationBar.h
jgunaratne: Done.

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: Done.

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: Done.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.h
jgunaratne: I've updated the name. This is intended to show the circle during tap down, but before tap up.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
jgunaratne: Moved layout code out of setSelected.

components/BottomNavigation/src/MDCBottomNavigationBar.h
```diff
@@ -0,0 +1,105 @@
...
+/**
+ The maximum width of the navigation bar item container in landscape mode.
+ Default is 320.
+ */
+@property(nonatomic, assign) CGFloat maxLandscapeContainerWidth;
+
+/**
+ Display font used for item titles.
+ */
+@property(nonatomic, copy, nullable) UIFont *itemTitleFont;
```
ianegordon: add UI_APPEARANCE_SELECTOR

components/BottomNavigation/src/MDCBottomNavigationBar.h
```diff
@@ -0,0 +1,105 @@
...
+/**
+ Selected item in the bottom navigation bar.
+ */
+@property(nonatomic, weak, nullable) UITabBarItem *selectedItem;
+
+/**
+ Color of selected item. Applies color to items' icons and text.
+ */
+@property (nonatomic, readwrite, copy, nullable) UIColor *selectedItemTintColor
+    NS_AVAILABLE_IOS(10_0) UI_APPEARANCE_SELECTOR;
```
ianegordon: Only available in iOS 10?
C & P Error?

components/BottomNavigation/src/MDCBottomNavigationBar.h
```diff
@@ -0,0 +1,105 @@
...
+ An array of UITabBarItems that is used to populate bottom navigation bar content. The array must
+ contain a minimum of three items with a maxiumum of up to five items.
+ */
+@property(nonatomic, copy, nullable) NSArray<UITabBarItem *> *items;
+
+/**
+ The maximum width of the navigation bar item container in landscape mode.
+ Default is 320.
+ */
+@property(nonatomic, assign) CGFloat maxLandscapeContainerWidth;
```
ianegordon: I would expect this to be tied to the horizontal size category and idiom in the traitCollections.
Even if a single value is sufficient, do we have to make this public?

components/BottomNavigation/src/MDCBottomNavigationBar.h
```diff
@@ -0,0 +1,105 @@
...
+/**
+ Color of selected item. Applies color to items' icons and text.
+ */
+@property (nonatomic, readwrite, copy, nullable) UIColor *selectedItemTintColor
+    NS_AVAILABLE_IOS(10_0) UI_APPEARANCE_SELECTOR;
+
+/**
+ Color of unselected items. Applies color to items' icons. Text is not displayed in unselected mode.
+ */
+@property (nonatomic, readwrite, copy, nullable) UIColor *unselectedItemTintColor
```
ianegordon: UIColor (and other immutable classes) are typically strong, not copy.

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,354 @@
...
+  [self layoutItemViewsWithLayoutDirection:self.mdc_effectiveUserInterfaceLayoutDirection];
+}
+
+- (CGSize)sizeThatFits:(CGSize)size {
+  UIEdgeInsets insets = [self getUIEdgeInsets];
+  CGFloat heightWithInset = kMDCBottomNavigationBarHeight + insets.bottom;
+  return CGSizeMake(size.width, heightWithInset);
+}
+
+- (void)layoutLandscapeModeWithSize:(CGSize)size width:(CGFloat)width {
```
ianegordon: Having a size and a width parameter is a bit confusing.
Perhaps itemSize and containerWidth  would offer more insight?  (Assuming that is what they represent)

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,354 @@
...
+  [super layoutSubviews];
+
+  CGSize size = self.bounds.size;
+  if (UIDeviceOrientationIsLandscape([UIDevice currentDevice].orientation)) {
+    [self layoutLandscapeModeWithSize:size width:self.maxLandscapeContainerWidth];
+  } else {
+    [self sizeContainerViewItemsDistributed:YES withSize:size width:size.width];
+    [self showTitleBelowItem:YES];
+  }
+  [self layoutItemViewsWithLayoutDirection:self.mdc_effectiveUserInterfaceLayoutDirection];
```
ianegordon: We are sending an ivar as a parameter to an instance method, which is a bit of a warning flag.
Consider removing the parameter completely (simpler!) and accessing it where it is needed in the layoutItemViews method.

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,354 @@
...
+    if (layoutDirection == UIUserInterfaceLayoutDirectionLeftToRight) {
+      itemView.frame = CGRectMake(i * itemWidth, 0, itemWidth, navBarHeight);
+    } else {
+      itemView.frame =
+          CGRectMake(navBarWidth - (i + 1) * itemWidth, 0,  itemWidth, navBarHeight);
+    }
+  }
+}
+
+- (void)showTitleBelowItem:(BOOL)titleBelowIcon {
```
ianegordon: Could this be a property?

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,354 @@
...
+      [itemView setTitle:change[kMDCBottomNavigationBarNewString]];
+    }
+  }
+}
+
+#pragma mark - Touch handlers
+
+- (void)didTouchDownButton:(UIButton *)button {
+  MDCBottomNavigationItemView *itemView = (MDCBottomNavigationItemView *)button.superview;
+  [itemView setCircleHighlightHidden:NO];
```
ianegordon: Prefer dot syntax when invoking properties.
itemView.cicleHighlightHidden = NO;

In Xcode, run Edit -> Convert -> To Modern Objective-C Syntax on your new files will likely fix these automagically.

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,354 @@
...
+}
+
+- (void)didTouchUpInsideButton:(UIButton *)button {
+  for (MDCBottomNavigationItemView *itemView in self.itemViews) {
+    if (itemView.button != button) {
+      [itemView setSelected:NO animated:YES];
+    }
+  }
+  MDCBottomNavigationItemView *itemView = (MDCBottomNavigationItemView *)button.superview;
```
ianegordon: Might be simpler to bump the following two lines into the loop.

if (itemView.button == button) {
 // Newly selected item
 [itemView setSelected:YES animated:YES];
 itemView.circleHighlightHideen = YES;
} else {
  // De-select all other items
  [itemView setSelected:NO animated:YES]
}

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,354 @@
...
+      [self showTitleBelowItem:NO];
+      break;
+    case MDCBottomNavigationBarLandscapeItemModeCluster:
+      [self sizeContainerViewItemsDistributed:NO withSize:size width:width];
+      [self showTitleBelowItem:YES];
+      break;
+  }
+}
+
+- (UIEdgeInsets)getUIEdgeInsets {
```
ianegordon: super-nit: This is not an Objective-c~ish method name.  Objective-C rarely uses get.  
This is also a invoking safeAreaInsets.
Consider declaring as a property :
@property (readonly) UIEdgeInsets mdc_safeAreaInsets;

Naming is hard!

components/BottomNavigation/src/private/MDCBottomNavigationItemBadge.h
```diff
@@ -0,0 +1,30 @@
...
+@interface MDCBottomNavigationItemBadge : UIView
+
+@property(nonatomic, assign) CGFloat badgeCircleWidth;
+@property(nonatomic, assign) CGFloat badgeCircleHeight;
+@property(nonatomic, assign) CGFloat xPadding;
+@property(nonatomic, assign) CGFloat yPadding;
+
+@property(nonatomic, copy) NSString *badgeValue;
+@property(nonatomic, strong) UIColor *badgeColor UI_APPEARANCE_SELECTOR;
+@property(nonatomic, strong) UILabel *badgeValueLabel;
```
ianegordon: readonly?

components/BottomNavigation/src/private/MDCBottomNavigationItemView.h
```diff
@@ -0,0 +1,42 @@
...
+
+@interface MDCBottomNavigationItemView : UIView
+
+@property(nonatomic, assign) BOOL titleBelowIcon;
+@property(nonatomic, assign) BOOL selected;
+@property(nonatomic, assign) MDCBottomNavigationBarTitleHideState titleHideState;
+
+@property(nonatomic, copy) NSString *badgeValue;
+@property(nonatomic, copy) NSString *title;
+@property(nonatomic, copy) UIFont *itemTitleFont;
```
ianegordon: s/copy/strong/

components/BottomNavigation/src/private/MDCBottomNavigationItemView.h
```diff
@@ -0,0 +1,42 @@
...
+@property(nonatomic, strong) UIButton *button;
+@property(nonatomic, strong) UIImage *image;
+
+@property(nonatomic, strong) UIColor *badgeColor UI_APPEARANCE_SELECTOR;
+@property(nonatomic, strong) UIColor *selectedItemTintColor UI_APPEARANCE_SELECTOR;
+@property(nonatomic, strong) UIColor *unselectedItemTintColor UI_APPEARANCE_SELECTOR;
+
+- (void)setSelected:(BOOL)selected animated:(BOOL)animated;
+
+- (void)setCircleHighlightHidden:(BOOL)circleHighlightHidden;
```
ianegordon: @property?

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
```diff
@@ -0,0 +1,282 @@
...
+  self.unselectedImage = [self colorizeImage:image color:self.unselectedItemTintColor];
+
+  self.iconImageView.image = self.unselectedImage;
+  [self.iconImageView sizeToFit];
+}
+
+- (void)setTitle:(NSString *)title {
+  _title = [title copy];
+  self.label.text = _title;
+  [self.label sizeToFit];
```
ianegordon: When we set self.label.text don't we implicitly call setNeedsLayout which should call sizeToFit?
Can we drop the sizeToFit?

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
```diff
@@ -0,0 +1,282 @@
...
+}
+
+- (void)setItemTitleFont:(UIFont *)itemTitleFont {
+  _itemTitleFont = itemTitleFont;
+  self.label.font = itemTitleFont;
```
ianegordon: We may or may not get a setNeedsLayout call for setting self.label.font.
If we do not get one implicitly , we likely want setNeedsLayout over sizeToFit.

ianegordon (changes requested):

components/BottomNavigation/src/MDCBottomNavigationBar.h
jgunaratne: No, this doesn't necessarily need to be public. Removing this from the public API.

components/BottomNavigation/src/MDCBottomNavigationBar.h
jgunaratne: Done.

components/BottomNavigation/src/MDCBottomNavigationBar.h
jgunaratne: Yep, removed.

components/BottomNavigation/src/MDCBottomNavigationBar.h
jgunaratne: Done.

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: Done.

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: Yes, agree. Updated the arg names so it's a bit more clear.

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: Updated this to better follow Obj-C conventions.

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: Yes. Updated.

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: When I run Edit -> Convert -> To Modern Objective-C Syntax in Xcode 9. It tells me to use the set syntax for some reason (see below). Wrong setting?

screen shot 2017-10-12 at 10 40 19 am
screen shot 2017-10-12 at 10 45 22 am

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: Done.

components/BottomNavigation/src/private/MDCBottomNavigationItemBadge.h
jgunaratne: Yes, there is only one instance that should be read.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.h
jgunaratne: Done.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
jgunaratne: Removed. Originally had this just to be safe.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
jgunaratne: Done.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.h
jgunaratne: Done.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
jgunaratne: In the process of doing research on imageView.tintColor. tintColor does not appear to be affecting icons that are only black with 1.f alpha.

components/BottomNavigation/src/MDCBottomNavigationBar.h
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ Selected item in the bottom navigation bar.
+ */
+@property(nonatomic, weak, nullable) UITabBarItem *selectedItem;
+
+/**
+ Display font used for item titles.
+ */
+@property(nonatomic, copy, nullable) UIFont *itemTitleFont UI_APPEARANCE_SELECTOR;
```
ianegordon: s/copy/strong/

components/BottomNavigation/examples/BottomNavigationTypicalUseExample.m
```diff
@@ -0,0 +1,106 @@
...
+      [[UITabBarItem alloc] initWithTitle:@"Search"
+                                    image:[UIImage imageNamed:@"Search"]
+                                      tag:0];
+  tabBarItem4.badgeValue = @"88";
+  UITabBarItem *tabBarItem5 =
+      [[UITabBarItem alloc] initWithTitle:@"Birthday"
+                                    image:[UIImage imageNamed:@"Cake"]
+                                      tag:0];
+  tabBarItem5.badgeValue = @"999+";
+  tabBarItem5.badgeColor = [MDCPalette cyanPalette].accent700;
```
ianegordon: This is an iOS 10 property.

ianegordon (dismissed):

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: Disregard. Screens are reversed. Made updates.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
jgunaratne: Removing this method. Fixed how tintColor is applied to UIImageView.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
jgunaratne: Done. Now using tintColor. Needed to use imageWithRenderingMode.

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,372 @@
...
+- (void)layoutSubviews {
+  [super layoutSubviews];
+
+  CGSize size = self.bounds.size;
+  if (UIDeviceOrientationIsLandscape([UIDevice currentDevice].orientation)) {
+    [self layoutLandscapeModeWithBottomNavSize:size
+                                containerWidth:self.maxLandscapeClusterContainerWidth];
+  } else {
+    [self sizeContainerViewItemsDistributed:YES withBottomNavSize:size containerWidth:size.width];
+    self.titleBelowItem = YES;
```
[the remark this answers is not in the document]
jgunaratne: In this mode the designers want the title positions to change: adjacent title in landscape, and below title portrait. This similar to how UINavBar behaves in iOS 11 (e.g. Photos app).

components/BottomNavigation/examples/BottomNavigationTypicalUseExample.m
jgunaratne: Added check.

components/BottomNavigation/src/MDCBottomNavigationBar.h
jgunaratne: Done.

jgunaratne: Done.

ianegordon:
Xcode 8 needs a guard around some of our examples.
Looking at VO now.

ianegordon:
We can implement our VO similar to UITabBar.

When moving around the tab bar, entire tabbaritems are selected, not individual labels or badges.

VO reads:
"selected" - (optional) if the item is selected
tabbaritem.title
tabbaritem.badgeValue " items" - (optional) if the badge value is set and a number
-OR-
tabbaritem.badgeValue - (optional) if the badge value is set and not a number
"tab"
items.index + 1 " of " items.count - where items are the list of tabbaritems in the tabbar

ianegordon (dismissed):
Please update the VO behavior to match UITabBar.

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,414 @@
...
+  if (_items == items) {
+    return;
+  }
+  NSAssert(items.count >= 3, @"Need to have at least 3 items in navigation bar.");
+  NSAssert(items.count <= 5, @"Navigation bar has a maximum of 5 items.");
+
+  _items = items;
+
+  [self removeBottomNavigationitemViews];
+  [self removeObserversFromTabBarItems];
```
ianegordon: This, and the removeBottomNavigationItemViews above need to happen before the assignment (_items = items) above.  (Or else we drop the old ones on the floor and try to remove the newly assigned on).

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,414 @@
...
+- (void)setItems:(NSArray<UITabBarItem *> *)items {
+  if (_items == items) {
+    return;
+  }
+  NSAssert(items.count >= 3, @"Need to have at least 3 items in navigation bar.");
+  NSAssert(items.count <= 5, @"Navigation bar has a maximum of 5 items.");
+
+  _items = items;
+
+  [self removeBottomNavigationitemViews];
```
ianegordon: This method is a single loop that is only called from here.  Consider simplifying by removing this method completely and just having the loop with a comment.

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -0,0 +1,414 @@
...
+#pragma mark - Setters
+
+- (void)setItems:(NSArray<UITabBarItem *> *)items {
+  if (_items == items) {
+    return;
+  }
+  NSAssert(items.count >= 3, @"Need to have at least 3 items in navigation bar.");
+  NSAssert(items.count <= 5, @"Navigation bar has a maximum of 5 items.");
+
+  _items = items;
```
ianegordon: This property is annotated copy so this should be:
_items = [items copy];

ianegordon (dismissed):

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: Done.

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: Done.

components/BottomNavigation/src/MDCBottomNavigationBar.m
jgunaratne: Done.

jgunaratne:
Tested a11y on iPhone 8 Plus and resolved issues related to iPhone X.

components/BottomNavigation/src/MDCBottomNavigationBar.h
```diff
@@ -0,0 +1,99 @@
...
+ can be selected at at time. The selected item's title text is displayed. Title text for unselected
+ items are hidden.
+ */
+@interface MDCBottomNavigationBar : UIView
+
+/**
+ Configures when item titles should be displayed.
+ Default is MDCBottomNavigationBarTitleVisibilitySelected.
+ */
+@property(nonatomic, assign) MDCBottomNavigationBarTitleHideState titleHideState;
```
romoore: Suggestion: `titleVisibility`

components/BottomNavigation/src/MDCBottomNavigationBar.h
```diff
@@ -0,0 +1,105 @@
...
+  // Items are distributed using the entire width of the device. Titles are centered below icons.
+  MDCBottomNavigationBarDistributionEqual = 0,
+
+  // Items are distributed using the entire width of the device. Titles are positioned adjacent to
+  // icons.
+  MDCBottomNavigationBarDistributionEqualAdjacentTitles = 1,
+
+  // Items are tightly clustered together. Titles are positioned below icons.
+  MDCBottomNavigationBarDistributionCluster = 2
+};
```
ianegordon: super-nit / consider:
MDCBottomNavigatioBarAlignment {
MDCBottomNavigatioBarAlignmentJustified
MDCBottomNavigatioBarAlignmentCentered
MDCBottomNavigatioBarAlignmentCenteredWithAdjacentTitles
}

.... naming is hard.

components/BottomNavigation/src/MaterialBottomNavigation.bundle/Resources/en.lproj/MaterialBottomNavigation.strings
```diff
@@ -0,0 +1,3 @@
+/* Accessibility hint preposition for item number i of n. */
+"MaterialBottomNavigationItemCountAccessibilityHint" = "of";
```
ianegordon: I believe this is insufficient.  I seem to recall certain languages may reorder this.  So were we might say "2 of 5", another language might be "of 5 item 2".  So the order of the parameters may be reversed.

components/BottomNavigation/src/MaterialBottomNavigation.bundle/Resources/en.lproj/MaterialBottomNavigation.strings
```diff
@@ -0,0 +1,3 @@
...
+"MaterialBottomNavigationTabElementAccessibilityLabel" = "tab";
```
ianegordon: Add a comment explaining what a tab is or a link to our spec.

components/BottomNavigation/src/private/MDCBottomNavigationItemBadge.m
```diff
@@ -0,0 +1,91 @@
...
+                          _badgeCircleHeight);
+  self.badgeValueLabel.center = CGPointMake(CGRectGetMidX(self.bounds), CGRectGetMidY(self.bounds));
+
+  CGFloat badgeRadius = CGRectGetMidY(self.bounds);
+  self.layer.cornerRadius = badgeRadius;
+  self.layer.backgroundColor = self.badgeColor.CGColor;
+}
+
+- (void)setBadgeValue:(NSString *)badgeValue {
+  _badgeValue = badgeValue;
```
ianegordon: If we always store this is the label, can we get rid of _badgeValue and have the self.badgeValue getter return self.badgeValueLabel.text?

components/BottomNavigation/src/private/MDCBottomNavigationItemBadge.m
```diff
@@ -0,0 +1,91 @@
...
+  _badgeValueLabel.isAccessibilityElement = NO;
+  [self addSubview:_badgeValueLabel];
+}
+
+- (void)layoutSubviews {
+  [super layoutSubviews];
+  [self sizeBadge];
+}
+
+- (void)sizeBadge {
```
ianegordon: super-nit: layout views calls super then invokes this method.  Consider simplifying by moving the contents of sizeBadge into layoutSubviews and removing the separate method.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
```diff
@@ -0,0 +1,339 @@
...
+- (instancetype)initWithCoder:(NSCoder *)aDecoder {
+  self = [super initWithCoder:aDecoder];
+  if (self) {
+    [self commonMDCBottomNavigationItemViewInit];
+  }
+  return self;
+}
+
+- (void)commonMDCBottomNavigationItemViewInit {
+  self.autoresizingMask = (UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight);
```
ianegordon: This is set based on whether the view was build programmatically or in storyboards.  Do we need to override?

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
```diff
@@ -0,0 +1,339 @@
...
+
+  _titleBelowIcon = YES;
+  _selectedItemTintColor = [UIColor blackColor];
+  _unselectedItemTintColor = [UIColor grayColor];
+
+  _iconImageView = [[UIImageView alloc] initWithFrame:CGRectZero];
+  _iconImageView.isAccessibilityElement = NO;
```
ianegordon: A-ha!  isAccessibilityElement = NO seems like the magic that got all the pieces working together.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
```diff
@@ -0,0 +1,339 @@
...
+                      iconImageViewCenter.y - CGRectGetMidX(self.iconImageView.bounds));
+      self.label.textAlignment = NSTextAlignmentRight;
+    }
+  }
+}
+
+- (CGSize)sizeForText:(NSString *)text
+                 font:(UIFont *)font
+     boundingRectSize:(CGSize)boundingRectSize {
+  CGRect rect = [text boundingRectWithSize:boundingRectSize
```
ianegordon: super-nit / style-wars:
This method essentially just returns the result of another method and is only called from one place.  Do we need this to be its own method?  If so, it doesn't use self and should be a class method.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
```diff
@@ -0,0 +1,339 @@
...
+                                attributes:@{ NSFontAttributeName:font }
+                                   context:nil];
+  return rect.size;
+}
+
+- (NSString *)accessibilityLabelWithTitle:(NSString *)title {
+  NSMutableArray *labelComponents = [NSMutableArray array];
+
+  // Use untransformed title as accessibility label to ensure accurate reading.
+  NSString *titleComponent = title;
```
ianegordon: This local variable just masks title.  Consider:
if (title.length > 0) {
  [labelComponents addObject:title];
}

components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -205,7 +205,7 @@ - (void)updateWithItem:(UITabBarItem *)item
   self.title = item.title;
   self.image = item.image;
   self.badgeValue = item.badgeValue;
-  self.accessibilityIdentifier = item.accessibilityIdentifier;
+  //self.accessibilityIdentifier = item.accessibilityIdentifier;
```
ianegordon: comment?

ianegordon (dismissed):

components/BottomNavigation/src/MaterialBottomNavigation.bundle/Resources/en.lproj/MaterialBottomNavigation.strings
romoore: +1. I believe the preferred approach is to keep the entire phrase together for context, something like,
```
/* Accessibility hint for the item number. Used when blah blah... */
"tab %1$d of %2$d"
```

components/Tabs/src/private/MDCItemBarCell.m
jgunaratne: Reverted.

components/BottomNavigation/src/MaterialBottomNavigation.bundle/Resources/en.lproj/MaterialBottomNavigation.strings
jgunaratne: Updated to allow for localization similar to MDCTabBar.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
jgunaratne: Removed. The autoresizing masks don't seem to be necessary.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
jgunaratne: Yep, this is one of the secrets to get VO to focus on the right UI elements.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
jgunaratne: Eliminated. Now calling boundingRectWithSize directly.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
jgunaratne: Done.

ianegordon (dismissed):
LGTM, aside from:

The @availability errors that need to be wrapped (See the Travis logs on xcode 8.3).
Can you add a Swift + Storyboard example?

I would pull in @brianjmoore as a fellow Tab enthusiast to get a +1.

components/BottomNavigation/src/private/MDCBottomNavigationItemBadge.m
jgunaratne: Done.

ianegordon (approved):
LGTM.
Please open an issue to add a Swift and Storyboard example.

PR #2089, opened 2017-10-04
[Examples | Catalog] Updates all components to Swift 4.

codeman7:
Convert all swift code to Swift 4.

Doesn't close any issues.

Had trouble refreshing my fork so had to reclone that's why I have another pull request. Also, before this is merged I have a couple issues.


willlarche (changes requested):
Thank you! I thought we wanted >=3.2 you said. Why 4 now?

codeman7:
Once we switch to 4.0 being default then it goes to 4.0 but while 3.0 is still the default code then you do 3.2. Does that make sense?

codeman7:
When the Swift 4 compiler is working with Swift 3 code, it identifies its language version as 3.2. As a result, you can use conditional compilation blocks like #if swift(>=3.2) to write code that’s compatible with multiple versions of the Swift compiler.

willlarche:
Will this cause all files to be compiled in Swift 4? Or is that a framework
/ module setting?

codeman7:
This would make it so we switch the Swift language version to 4.0. If we decide to merge this then if someone uses XCode 8, they need to go and set the 'Use Legacy Swift Language Version' to 'No' in build settings or it won't build.

willlarche:
This is great. So what's missing is setting the target for MDCCatalog to actually build with Swift 4.0. BUT, how do we adjust that by xcode?

codeman7:
I don't think you can have it different settings for different versions of XCode. I think you have to manually do it. Maybe, we should just wait until we drop support for XCode 8 to merge this.

willlarche:
No. I think it’s good even it’s just to be read by potential users. Most
users copy and paste.

codeman7:
But, then other devs will have to go and change the Swift version manually if they want to use MDC-iOS for XCode 8.

willlarche:
Sorry, I meant keep target set to Swift 3 but merge in this swift 4 code even tho it won't be run in testing until we drop Xcode 8 support.

willlarche (approved):

codeman7:
Should I change all the checks for Swift 4.0 to 3.2?

willlarche:
I'm still not understanding why we'd want it to be 3.2 if the code is 4.

codeman7:
Because we would tell Xcode that we are only using Swift 3, so it makes Swift 4 as Swift 3.2. I am pretty sure Swift 3.2 and 4.0 are the same.

willlarche:
And what's the consequence of leaving as 4?

codeman7:
All the Swift 4.0 code will never be executed.

willlarche (changes requested):
Just tested on 8. It will skip the 3.2 line. Great!
Change them to 3.2 and I think we are done!
Thanks for your patience.

codeman7:
@willlarche I think everything should be good now.

willlarche:
Missing quote from outside conversation:

> So when you switch all the checks to Swift 3.2 then all the renamed APIs are not renamed yet. So that takes out the point of having every #if swift(>=3.2) check. So until support for XCode 8 is dropped then there is no point in having the PR because nothing changes.

Issue #1774, opened 2017-08-10
[Themes] Themes is missing a README.md


ajsecord:

jgunaratne:
Fixed in https://github.com/material-components/material-components-ios/pull/2094

PR #2094, opened 2017-10-05
[Themes] Add documentation for material themes

jgunaratne:
Add documentation for material themes.

Resolves: https://github.com/material-components/material-components-ios/issues/1774


willlarche (approved):
I wish the status bar were white in the picture but otherwise perfect!

PR #2095, opened 2017-10-05
[Buttons] Restore alpha when re-enabled

### LINKED ISSUE #1964: MDCButton ignores/overwrites alpha.
If you try to set the alpha of a button the button will get set it back to 1.0. 

Internal logic uses the enabled/disabled state to set the alpha of the button.

 `   self.alpha = self.enabled ? 1.0f : _disabledAlpha;`
https://github.com/material-components/material-components-ios/blob/cda52f4e6b29f0ed72a4de47974c603970bfe142/components/Buttons/src/MDCButton.m#L643

romoore:
When re-enabled, MDCButton previously set its alpha avalue to 1 instead
of the previously-set value.

Closes #1964

### DIFF
--- a/components/Buttons/src/MDCButton.m
+++ b/components/Buttons/src/MDCButton.m
@@ -42,6 +42,7 @@ static NSString *const MDCButtonUppercaseTitleKey = @"MDCButtonShouldCapitalizeT
 // Previous value kept for backwards compatibility.
 static NSString *const MDCButtonUnderlyingColorHintKey = @"MDCButtonUnderlyingColorKey";
 static NSString *const MDCButtonDisableAlphaKey = @"MDCButtonDisableAlphaKey";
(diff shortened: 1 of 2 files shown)


components/Buttons/src/MDCButton.m
```diff
@@ -402,6 +410,9 @@ - (void)touchesCancelled:(NSSet *)touches withEvent:(UIEvent *)event {
 #pragma mark - UIControl methods
 
 - (void)setEnabled:(BOOL)enabled {
+  if (self.enabled && !enabled) {
+    _enabledAlpha = self.alpha;
```
samnm: IMO you should write to _enabledAlpha on setAlpha:

samnm (changes requested):

components/Buttons/src/MDCButton.m

samnm (approved):

ianegordon (approved):

randallli (approved):

PR #2100, opened 2017-10-05
[TextFields] Fix broken MDCMultilineTextField build for iOS 10.

mpetrov:
It fails to build in it's current state because  setAdjustsFontForContentSizeCategory: is defined twice; once inside the ifdef block and once outside. The version in the ifdef also fails to compile since UIView does not conform to UIContentSizeCategoryAdjusting.

### DIFF
--- a/components/TextFields/src/MDCMultilineTextField.m
+++ b/components/TextFields/src/MDCMultilineTextField.m
@@ -474,13 +474,6 @@ static NSString *const MDCMultilineTextFieldTrailingViewModeKey =
 
 #pragma mark - Properties Implementation
 
-#if defined(__IPHONE_10_0) && (__IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_10_0)
-- (void)setAdjustsFontForContentSizeCategory:(BOOL)adjustsFontForContentSizeCategory {
-  [super setAdjustsFontForContentSizeCategory:adjustsFontForContentSizeCategory];
-  [self mdc_setAdjustsFontForContentSizeCategory:adjustsFontForContentSizeCategory];
-}
-#endif
-
 - (BOOL)adjustsFontForContentSizeCategory {
   return self.mdc_adjustsFontForContentSizeCategory;
 }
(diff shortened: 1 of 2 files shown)


willlarche:
Is this specific to a version of iOS 10? Our Travis builds on iOS 10.3.1 with no problem: https://travis-ci.org/material-components/material-components-ios

willlarche:
Oh, nevermind. i see. This is a target that's 10+

mpetrov:
Correct, it's a 10+ *target* issue.

ianegordon:
I recommend renaming adjustsFontForContentSizeCategory to mdc_adjustsFontForContentSizeCategory until our minimum iOS versions includes UIContentSizeCategoryAdjusting.
See MDCFeatureHighlightViewController.h for an example.
Otherwise, collisions will collide.

willlarche (changes requested):
The problem isn't the first one but the second one. It's a typo. `setAdjustsFontForContentSizeCategory` outside of iOS 10+ should be cut.

components/TextFields/src/MDCMultilineTextField.m
```diff
@@ -474,13 +474,6 @@ - (void)updateBorder {
 
 #pragma mark - Properties Implementation
 
-#if defined(__IPHONE_10_0) && (__IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_10_0)
```
willlarche, line 477: Since all this is leftover from when we tried to keep the superclass as UITextView, go ahead and delete adjusts and setAdjusts also.

willlarche (changes requested):

components/TextFields/src/MDCMultilineTextField.m
mpetrov: Done

willlarche (approved):

willlarche:
Pending Travis.

willlarche:
In `MultilineTests.swift`, delete the following:

```
    if #available(iOS 10, *) {
      XCTAssertEqual(textField.mdc_adjustsFontForContentSizeCategory,
                     textField.adjustsFontForContentSizeCategory)
    }
```

willlarche (changes requested):
See the comment I left.

willlarche (approved):

PR #2102, opened 2017-10-05
[Checklist] Status spreadsheet link

willlarche:


ianegordon:
I was unable to visit the link from my external device.

ianegordon (dismissed):
Please changes permissions to allow view-only access from the public.

willlarche:
Should be good now.

randallli (approved):
what is the transposed tab? is it supposed to be all our components? looks like it is just 3 right now. Do you need to rerun a script?

willlarche:
Transposed is a tab I made for @ianegordon  that shows the same info inverted so you can do filtering in different ways. I'll clear the filter he's left there.

PR #2103, opened 2017-10-05
[AppBar] Fix app bar's top constraint that is buggy on iOS 11.0

### LINKED ISSUE #2096: [AppBar] Catalog examples cannot tap Bar Buttons
### Overview
The AppBar examples (except for Storyboard) have non-tappable buttons in the NavBar/Flexible Header view.  If you scroll the AppBar so that it goes up (toward the Status Bar), the buttons become tappable.

Some views have the navigation elements in the wrong position.

simulator screen shot - iphone se - 2017-10-06 at 02 51 52

### Reproduction steps
1. Open MDC Catalog &rarr; App Bar &rarr; Demo
2. Tap on the Back button in the App Bar

#### Expected Result
View Controller is popped

#### Actual Result
No effect

### Version number
35.2.0

iPhone SE/iOS 11.0 (Simulator), Xcode 9 (GM)

chuga:
The app bar's safe area is incorrectly set on iOS 11.0 by the OS. This is now fixed on 11.1, but until we can depend on that being the min iOS 11 version, we have to use this workaround -_-

Fixes #2096

### DIFF
--- a/components/AppBar/src/MDCAppBar.m
+++ b/components/AppBar/src/MDCAppBar.m
@@ -318,7 +318,7 @@ static NSString *const kMaterialAppBarBundle = @"MaterialAppBar.bundle";
     // Note: We should switch to using self.view.safeAreaLayoutGuide.topAnchor once 11.1 is
     // the min iOS 11 we support. 11.0 has a bug where the Safe Area is incorrectly calculated for
     // this view.
-    topMargin = [UIApplication mdc_safeSharedApplication].statusBarFrame.size.height;
+    topMargin = CGRectGetMaxY([UIApplication mdc_safeSharedApplication].statusBarFrame);
   }
 #endif
   _verticalConstraint = [NSLayoutConstraint constraintWithItem:self.headerStackView
@@ -356,7 +356,7 @@ static NSString *const kMaterialAppBarBundle = @"MaterialAppBar.bundle";
     // We only get the actual status bar height if we're on iOS 11, because previously we were not
     // adjusting the header height to make it smaller when the status bar is hidden.
     _verticalConstraint.constant =
-        [UIApplication mdc_safeSharedApplication].statusBarFrame.size.height;
+        CGRectGetMaxY([UIApplication mdc_safeSharedApplication].statusBarFrame);
   }
 #endif
 }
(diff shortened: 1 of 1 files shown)


components/AppBar/src/MDCAppBar.m
```diff
@@ -308,22 +310,25 @@ - (void)viewDidLoad {
...
-                               attribute:NSLayoutAttributeTop
-                              multiplier:1
-                                constant:kPreIOS11StatusBarHeight].active = YES;
-#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+    // We only get the actual status bar height if we're on iOS 11, because previously we were not
+    // adjusting the header height to make it smaller when the status bar is hidden.
+    // Note: We should switch to using self.view.safeAreaLayoutGuide.topAnchor once 11.1 is
+    // the min iOS 11 we support. 11.0 has a bug where the Safe Area is incorrectly calculated for
+    // this view.
+    topMargin = [UIApplication mdc_safeSharedApplication].statusBarFrame.size.height;
```
romoore: Nit: I think it's better to use the maxY of the status Bar in case its frame does not start at 0?  The maxY is the offset into the window where the headerStackView should start.

```objectivec
CGRectGetMaxY([UIApplication mdc_safeSharedApplication].statusBarFrame)
```

components/AppBar/src/MDCAppBar.m
```diff
@@ -343,6 +348,19 @@ - (void)viewWillAppear:(BOOL)animated {
...
 
+- (void)viewWillLayoutSubviews {
+  [super viewWillLayoutSubviews];
+
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    // We only get the actual status bar height if we're on iOS 11, because previously we were not
+    // adjusting the header height to make it smaller when the status bar is hidden.
+    _verticalConstraint.constant =
+        [UIApplication mdc_safeSharedApplication].statusBarFrame.size.height;
```
romoore: Same suggestion here

romoore (approved):

components/AppBar/src/MDCAppBar.m
chuga: Done (even though I've never seen a status bar frame that doesn't start at 0 :shrug:

components/AppBar/src/MDCAppBar.m
chuga: Done

Issue #1953, opened 2017-09-07
[Catalog] A few Tab Bar examples are visually broken


chuga:
Looks like something went wrong when theming was implemented.

Examples:
![simulator screen shot sep 7 2017 12 07 18 pm](https://user-images.githubusercontent.com/4545451/30173313-3585524c-93c5-11e7-9d61-7397bbc40632.png)
![simulator screen shot sep 7 2017 12 07 27 pm](https://user-images.githubusercontent.com/4545451/30173316-37485fca-93c5-11e7-83b5-8010aa173793.png)

romoore:
I think we should decide how much of the catalog will be themed versus providing a backdrop against which to display themed components.

PR #2104, opened 2017-10-05
[TabBar] Remove tint color from themer

jgunaratne:
Using the tab bar themer causes some visual issues since it is setting tint color. This removes tint color from the tab bar themer. Design will need to provide more guidance on how to theme tab bars appropriately.

Addresses issue: https://github.com/material-components/material-components-ios/issues/1953


chuga (approved):

romoore:
Is there another way to update the `_itemBar.backgroundColor` inside MDCTabBar.m ? Should we update examples to have each one of them setting the `barTintColor` individually?

PR #2108, opened 2017-10-05
[AppBar] Example corrections.

### LINKED ISSUE #1156: [App Bar] ObjC Interface Builder Example missing storyboard
### Overview
There's only one storyboard in the resources and it's set to use the Swift IB example as its initial view controller's class.

### Reproduction steps
1. Put a breakpoint in AppBarInterfaceBuilderExample.m in viewDidLoad (or anywhere.)
1. Run the catalog and tap on App Bar -> Interface Builder

It will open the storyboard but never hit the breakpoints. 

The AppBarInterfaceBuilderExample class has the catalogStoryboardName returning the Swift storyboard which then loads with the swift VC class.

willlarche:
Closes #1156 .

### DIFF
--- a/components/AppBar/examples/AppBarInterfaceBuilderExampleController.m
+++ b/components/AppBar/examples/AppBarInterfaceBuilderExampleController.m
@@ -46,6 +46,7 @@
 }
 
 - (void)commonAppBarInterfaceBuilderExampleSetup {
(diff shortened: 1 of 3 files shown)


ianegordon (approved):
Verified breakpoints work as expected on an iPhone X Sim @ 11.0.

PR #2113, opened 2017-10-06
[Checklist] Add color themer as checklist item

jgunaratne:
Color themers should be part of every component that can be colorized. Adding section in the checklist to note this.


willlarche (dismissed):
It's great! It's just missing the steps you need to do and the final `1. Enter YES, NO or N/A`

PR #2120, opened 2017-10-06
[ThumbTrack] Remove unused property and method

### LINKED ISSUE #1216: Remove MDCThumbTrack interpolateOnOffColors property
This property was removed, then re-added to unbreak an internal client.

Once the internal client is no longer utilizing this property we should remove this property.

samnm:
Closes #1216.

This property is no longer used internally. See cl/171303772.


ianegordon (approved):
Less is more!

PR #2124, opened 2017-10-06
[Dialogs] Migrate to  MDFInternationalization

ianegordon:
Replace private/RTL functions with the MDFI18n versions.


samnm (approved):
Make sure you `gem update psych`. I think thats why it's adding quotation marks in the Podfile.lock.

PR #2132, opened 2017-10-08
[PageControl] Fixed crasher when resetting -numberOfPages to 0

jpoumailloux:
This is otherwise supported by UIPageControl, which MDCPageControl aims to be a drop-in replacement for.

Crasher details:
Test Case '-[PageControlExampleTests testResetNumberOfPagesToZero]' started.
~/material-components-ios/components/PageControl/tests/unit/PageControlExampleTests.m:139: error: -[PageControlExampleTests testResetNumberOfPagesToZero] : failed: caught "NSRangeException", "*** -[__NSArrayM objectAtIndexedSubscript:]: index 0 beyond bounds for empty array"
(
	0   CoreFoundation                      0x000000010e3ed03b __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x000000010da81f41 objc_exception_throw + 48
	2   CoreFoundation                      0x000000010e42ce2c _CFThrowFormattedException + 194
	3   CoreFoundation                      0x000000010e41e714 -[__NSArrayM objectAtIndexedSubscript:] + 148
	4   MaterialComponents                  0x0000000128271441 -[MDCPageControl resetControl] + 2193
	5   MaterialComponents                  0x000000012826e489 -[MDCPageControl setNumberOfPages:] + 281

### DIFF
--- a/components/PageControl/tests/unit/PageControlExampleTests.m
+++ b/components/PageControl/tests/unit/PageControlExampleTests.m
@@ -134,12 +134,19 @@
   // Given
   MDCPageControl *pageControl = [[MDCPageControl alloc] init];
   pageControl.numberOfPages = 3;
+  NSException *exception;
 
   // When
-  pageControl.numberOfPages = 0;
-  pageControl.currentPage = 0;
+  @try {
+    pageControl.numberOfPages = 0;
+    pageControl.currentPage = 0;
+  }
+  @catch (NSException *e) {
+    exception = e;
+  }
 
-  // Then, no crash
+  // Then
+  XCTAssertNil(exception, @"PageControl crashed with exception: %@", exception);
 }
 
 @end
--- a/components/PageControl/tests/unit/PageControlExampleTests.m
+++ b/components/PageControl/tests/unit/PageControlExampleTests.m
@@ -130,4 +130,15 @@
   XCTAssertEqual(pageControl.currentPage, page);
 }
 
+- (void)testResetNumberOfPagesToZero {
+  // Given
+  MDCPageControl *pageControl = [[MDCPageControl alloc] init];
+  pageControl.numberOfPages = 3;
+
+  // When
+  pageControl.numberOfPages = 0;
+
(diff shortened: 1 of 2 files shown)


components/PageControl/tests/unit/PageControlExampleTests.m
```diff
@@ -130,4 +130,16 @@ - (void)testCurrentPageGetsUpdatedWhenOffsetIsChanged {
   XCTAssertEqual(pageControl.currentPage, page);
 }
 
+- (void)testResetNumberOfPagesToZero {
```
romoore, line 133: Glad to see a covering test for the change. :) My only request would be that we explicitly check for no exception.  Something like this:

```objectivec
- (void)testResetNumberOfPagesToZero {
  // Given
  MDCPageControl *pageControl = [[MDCPageControl alloc] init];
  pageControl.numberOfPages = 3;
  NSException *exception;

  // When
  @try {
    pageControl.numberOfPages = 0;
    pageControl.currentPage = 0;
  }
  @catch (NSException *e) {
    exception = e;
  }

  // Then
  XCTAssertNil(exception, @"PageControl crashed with exception: %@", exception);
}
```

romoore (changes requested):
Thanks @jpoumailloux, the change looks good to me.  Just one small request in the unit test.

components/PageControl/tests/unit/PageControlExampleTests.m
randallli: +1

randallli (approved):

components/PageControl/tests/unit/PageControlExampleTests.m
jpoumailloux: Thanks for the review & feedback! I've incorporated the suggested change, which I think is a cleaner way to do it, to the unit test.

romoore (approved):
LGTM! Thanks for the PR. 👍

PR #2141, opened 2017-10-10
[Catalog] Commit updated project.pbxprojs

samnm:
Running `pod install` makes some changes to the xcode projects.


ianegordon:
Do we all need to update to the latest cocoapods?

samnm:
We should, but we don't have to. Unfortunately the newest cocoapods version surfaced some warnings/errors so we should be on the latest version to resolve those.

ianegordon:
Restarted broken Travis Build.

ianegordon (approved):
Cocoapods update.
Travis is happy.
LGTM.

PR #2142, opened 2017-10-10
[ShadowElevations] Change shadow elevatoins to functions instead of an enum

### LINKED ISSUE #2105: [ShadowElevations] ShadowElevations should be an extensible enum for Swift

codeman7:
This would allow users to extend ShadowElevations with a function

`extension MDCShadowElevation {
   static func random() -> CGFloat {
     return 762.0
   }
}`

Closes #2105

### DIFF
--- a/catalog/MDCCatalog/MDCCatalogCollectionViewCell.swift
+++ b/catalog/MDCCatalog/MDCCatalogCollectionViewCell.swift
@@ -17,6 +17,7 @@ limitations under the License.
 import UIKit
 
 import MaterialComponents.MaterialTypography
+import MaterialComponents.MaterialShadowElevations
 
 class MDCCatalogCollectionViewCell: UICollectionViewCell {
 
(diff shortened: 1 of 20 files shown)


codeman7:
I still think this can be improved so if you guys like it please don't merge yet.

willlarche:
Thank you! But we're going to go with an extensible struct. See #2105

PR #2148, opened 2017-10-11
[Tabs] Add standard Material Design shadow shadow to MDCTabBar

### LINKED ISSUE #2112: [Tabs] MDCTabBar should have an 8dp shadow
### Overview

According to the material design specification, a bottom navigation bar (like the tab bar), should have an 8dp shadow, but MDCTabBar currently has no shadow.

### Reproduction steps
Create an instance of MDCTabBarViewController.  Note that the contained MDCTabBar view has no shadow.

### Version number
35.3.0

iOS 11 / iPhone X (simulator)

Icycle:
This change adds a Menu-height shadow layer to the MDCTabBar, and ensures that the
tab bar is always at the frontmost view in MDCTabBarViewController.

closes #2112

### DIFF
--- a/components/Tabs/src/MDCTabBarViewController.m
+++ b/components/Tabs/src/MDCTabBarViewController.m
@@ -131,6 +131,7 @@ const CGFloat MDCTabBarViewControllerAnimationDuration = 0.3f;
     if (selectedViewController) {
       self.tabBar.selectedItem = selectedViewController.tabBarItem;
     }
+    [self setNeedsStatusBarAppearanceUpdate];
   }
 }
 
@@ -241,5 +242,14 @@ const CGFloat MDCTabBarViewControllerAnimationDuration = 0.3f;
   }
 }
 
+#pragma mark - UIViewController status bar
+
+- (nullable UIViewController *)childViewControllerForStatusBarStyle {
+  return _selectedViewController;
+}
+
+- (nullable UIViewController *)childViewControllerForStatusBarHidden {
+  return _selectedViewController;
+}
 
 @end
--- a/components/Tabs/src/MDCTabBarViewController.m
+++ b/components/Tabs/src/MDCTabBarViewController.m
@@ -50,7 +50,7 @@ const CGFloat MDCTabBarViewControllerAnimationDuration = 0.3f;
 }
 
 - (void)commonMDCTabBarShadowViewInitialization {
-  MDCShadowLayer *shadowLayer =  (MDCShadowLayer *)self.layer;
+  MDCShadowLayer *shadowLayer = (MDCShadowLayer *)self.layer;
   [shadowLayer setElevation:MDCShadowElevationMenu];
 }
 
--- a/components/Tabs/src/MDCTabBarViewController.m
+++ b/components/Tabs/src/MDCTabBarViewController.m
@@ -16,7 +16,45 @@
 
 #import "MDCTabBarViewController.h"
 
+#import "MaterialShadowElevations.h"
+#import "MaterialShadowLayer.h"
+
 const CGFloat MDCTabBarViewControllerAnimationDuration = 0.3f;
+
+/**
+ * View to host shadow for the tab bar.
+ */
+@interface MDCTabBarShadowView : UIView
+@end
+
+@implementation MDCTabBarShadowView
+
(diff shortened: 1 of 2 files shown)


brianjmoore:
I think this needs a different approach. MDCTabBar is a relatively low-level component meant to be versatile and used in a variety of situations, including top tabs and bottom navigation, so we can't  add an always-on shadow directly onto it without limiting its usefulness.

In the top navigation case, it can be used as a subview of a MDCAppBar and in this case the flexible header is responsible for the shadow. We do not want MDCTabBar to have its own shadow in this case. This is also the primary use case for MDCTabBar.

For bottom navigation, we use MDCTabBarViewController, which really should have a shadow (tbh I thought it did). I see two possible approaches here:
* Have MDCTabBarViewController provide the shadow in a separate view and place it under the tab bar. This is what I have done for custom tab bar controllers in production, and I've found that it has certain advantages because it lets the tab bar controller manage certain behavior (like tab bar hiding) in a way that specifically takes the shadow into account.
* Add API to MDCTabBar that lets clients turn on a shadow that's built into the tab bar. This would have to be off by default so as to not break existing clients. I'm not sure if this case is common enough to warrant a new API though: most use cases for the tab bar are with higher-level components that provide a custom shadow.

brianjmoore (changes requested):
See comment above.

Icycle:
I think I like your first suggestion, of have MDCTabBarViewController create and manage a shadow view just behind the actual tab bar.  I've tested the approach in my own app and it looks and works great, so I'll update the PR with that approach shortly.

Icycle:
PR updated with the new approach.  PTAL

brianjmoore (approved):
It looks like some other files got pulled into this PR that will need to be reverted. I reviewed the change to MDCTabBarController.m and it looks good. Thanks!

Icycle:
I think the other files got pulled in when I was trying to sync my branch to head.  Is there any action item for me, or will you revert the files?

brianjmoore:
I don't know. @willlarche - does this PR look safe to merge?

Icycle:
A spot check seems to indicate that the diffs in everything but MDCTabBarController.m are spurious and represent changes that were recently submitted and got pulled into my branch when I synced to head.  I'm still not very comfortable in the GitHub workflow, so it's possible I've messed something up.  I'm not sure if this PR is safe to merge as is, if there is somewhat for you or me to "clean up" the PR, or if it should be abandoned and a new PR submitted from a clean branch?

romoore:
I did a local rebase (and squash) and it looks like the merge will be fine.  @Icycle you may wish to rebase with develop just to clean up the PR before submission.

```bash
# update branch develop to the MDC upstream repo (not your fork). Rename "material" if you have another name for remote
git checkout develop
git pull material develop
# switch back to your PR branch and rebase
git checkout fix2111
git rebase develop
```

You will need to resolve conflicts. Just edit the files to have the correct source code (look for `<<<` or `>>>`). When you have saved the resolved versions, you can `git add` to stage and `git rebase continue`.  The only conflict appears to be in `MDCTabBarViewController` where you bring the subviews to the front. I believe the correct resolution is to keep both statements?

Once that's all up to date, you will need to force-push the branch to GitHub.

```bash
git push origin fix2111 -f
```

Icycle:
OK.  I think the PR is all cleaned up now.  Thanks for the help!

components/Tabs/src/MDCTabBarViewController.m
```diff
@@ -16,8 +16,46 @@
...
+- (nullable instancetype)initWithCoder:(NSCoder *)aDecoder {
+  self = [super initWithCoder:aDecoder];
+  if (self) {
+    [self commonMDCTabBarShadowViewInitialization];
+  }
+  return self;
+}
+
+- (void)commonMDCTabBarShadowViewInitialization {
+  MDCShadowLayer *shadowLayer =  (MDCShadowLayer *)self.layer;
```
romoore: Nit: spacing after the `=`

romoore (approved):
LGTM, just a small formatting nit.

Icycle:
Nit fixed

romoore:
Thanks, @Icycle . I think this is ready to merge once Travis is happy.

PR #2155, opened 2017-10-11
[Tabs] Always call MDCTabBarControllerDelegate methods when a tab is tapped

### LINKED ISSUE #2154: [Tabs] MDCTabBarViewController should notify delegate when tapping already selected view controller
### Overview
MDCTabBarViewController will only call the  MDCTabBarViewController delegate methods -tabBarController:shouldSelectViewController: and -tabBarController:didSelectViewController: if the newly selected view controller is different than the currently selected view controller.  This behavior is different than the behavior of UITabBarController (since iOS 3.0).

Calling the delegate in the already-selected tab case allows the client to reset the tab to some default state, such as popping to the top of a deep navigation stack, or scrolling to the top of a scroll view, and in seen in apps such as Instagram, Google Translate, and Newsstand.

### Reproduction steps
Create a client application with MDCTabBarViewController and two more more tabs.  Tap on an already selected tab, and confirm that the delegate callbacks are not invoked.

### Version number
 35.3.0

iPhone SE / iOS 11

### Related issues
N/A

Icycle:
To more closely align with UITabBarController behavior, MDCTabBarController will invoke
the -tabBarController:shouldSelectViewController: and -tabBarController:didSelectViewController: delegate methods any time a tab is tapped, even if the selected view controller is unchanged.

closes #2154

### DIFF
--- a/components/Tabs/src/MDCTabBarViewController.h
+++ b/components/Tabs/src/MDCTabBarViewController.h
@@ -71,6 +71,9 @@ IB_DESIGNABLE
  If you provide this method, you can control whether tapping on a tab bar item actually
  switches to that viewController. If not provided, MDCTabBarViewController will always switch.
 
+ The tab bar controller calls this method regardless of whether the selected view controller
+ changed.
+
  You can also use this method as a willSelectViewController.
  */
 - (BOOL)tabBarController:(nonnull MDCTabBarViewController *)tabBarController
@@ -80,6 +83,9 @@ IB_DESIGNABLE
  Called when the user taps on a tab bar item. Not called for programmatic selection.
  MDCTabBarViewController will call your delegate once it has responded to the user's tap
  by changing the selected view controller.
+
+ The tab bar controller calls this method regardless of whether the selected view controller
+ changed.
  */
 - (void)tabBarController:(nonnull MDCTabBarViewController *)tabBarController
  didSelectViewController:(nonnull UIViewController *)viewController;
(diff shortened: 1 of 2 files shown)


components/Tabs/src/MDCTabBarViewController.h
```diff
@@ -71,6 +71,9 @@ IB_DESIGNABLE
  If you provide this method, you can control whether tapping on a tab bar item actually
  switches to that viewController. If not provided, MDCTabBarViewController will always switch.
 
+ The tab bar controller calls this method regardless of whether the selected view controller
+ changed.
```
romoore: I think I misread this comment when I first went through it.  If I understand this PR correctly, I think the following phrasing might be clearer to readers.  Please feel free to edit. :)

> @note The tab bar controller will call this method even when the tapped tab bar item is the currently-selected tab bar item.

components/Tabs/src/MDCTabBarViewController.h
```diff
@@ -80,6 +83,9 @@ IB_DESIGNABLE
  Called when the user taps on a tab bar item. Not called for programmatic selection.
  MDCTabBarViewController will call your delegate once it has responded to the user's tap
  by changing the selected view controller.
+
+ The tab bar controller calls this method regardless of whether the selected view controller
+ changed.
```
romoore: Same suggestion here:

> The tab bar controller will call this method even if the selected view controller was the same before and after selection.

romoore (changes requested):

brianjmoore (approved):

components/Tabs/src/MDCTabBarViewController.h
Icycle: I pulled this text directly from UITabBarController's docs, but I think your suggestion is clearer.  I'll update this PR in just a moment.

romoore (approved):
👍

ianegordon (approved):
LGTM.
@romoore, please merge when Travis is happy.  (You were here first 😄  )

Issue #2007, opened 2017-09-19
[Dialog] Support Safe Area


chuga:
### Overview
Issues visible in the Catalog: The padding needs to be adjusted to make sure it fits into the safe area

![simulator screen shot - iphone x - 2017-09-19 at 15 51 57](https://user-images.githubusercontent.com/4545451/30612463-aa8bae24-9d52-11e7-8776-5fe4e021dae4.png)

### Version number
35.0.0

### Operating system and device
iOS 11 / iPhone X

romoore:
Related to #1973

PR #2165, opened 2017-10-12
 [Dialogs] Respect safe area insets

ianegordon:
Adjust the presented controller's frame based on the safe area insets of the container view.

### DIFF
--- a/components/Dialogs/src/MDCDialogPresentationController.m
+++ b/components/Dialogs/src/MDCDialogPresentationController.m
@@ -198,23 +198,20 @@ static UIEdgeInsets MDCDialogEdgeInsets = {24, 20, 24, 20};
   }
 
   // For pre iOS 11 devices, we are assuming a safeAreaInset of UIEdgeInsetsZero
-  UIEdgeInsets presentingSafeAreaInsets = UIEdgeInsetsZero;
+  UIEdgeInsets containerSafeAreaInsets = UIEdgeInsetsZero;
 #if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
   if (@available(iOS 11.0, *)) {
-    // Do not use presentingViewController.additionalSafeAreaInsets since we want to center the
-    // presented view in the 'entire' presenting controller view.
-    // KM : ??? Consider using the application's keyView / UIWindow?
-    presentingSafeAreaInsets = self.presentingViewController.view.safeAreaInsets;
+    containerSafeAreaInsets = self.containerView.safeAreaInsets;
   }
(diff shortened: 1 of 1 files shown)


ianegordon:
Closes #2007

components/Dialogs/src/MDCDialogPresentationController.m
```diff
@@ -197,9 +197,26 @@ - (CGSize)sizeForChildContentContainer:(id<UIContentContainer>)container
     return CGSizeZero;
   }
 
+  // For pre iOS 11 devices, we are assuming a safeAreaInset of UIEdgeInsetsZero
```
chuga: Do we need to react to viewSafeAreaInsetsDidChange?

components/Dialogs/src/MDCDialogPresentationController.m
```diff
@@ -197,9 +197,26 @@ - (CGSize)sizeForChildContentContainer:(id<UIContentContainer>)container
...
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    containerSafeAreaInsets = self.containerView.safeAreaInsets;
+  }
+#endif
+
+  // Take the larger of the Safe Area insets and the Material specified insets.
+  CGFloat standardWidthInset = MDCDialogEdgeInsets.left + MDCDialogEdgeInsets.right;
+  CGFloat safeAreaWidthInset = containerSafeAreaInsets.left + containerSafeAreaInsets.right;
+  CGFloat widthInset = MAX(standardWidthInset, safeAreaWidthInset);
```
samnm: Since the safeAreaInsets can be asymmetric I don't believe your centering logic will work correctly anymore.

samnm (dismissed):

ianegordon:
Before:
broken

After:
fixed

components/Dialogs/src/MDCDialogPresentationController.m
ianegordon: Good catch.  Split inset code out from sizing code to allow handling this case.

ianegordon:
Properly handling horizontal insets when placing dialog.
Note that the safe area insets are rendered below in red.

fixedwithvisiblesafe

components/Dialogs/src/MDCDialogPresentationController.m
ianegordon: This initially appears tricky.  
View controllers have viewSafeAreaInsetsDidChange, but presentation controllers do not.  I'll explore further.

samnm (approved):

PR #2169, opened 2017-10-12
[TextFields] Implementing textFieldShouldClear:.

### LINKED ISSUE #2062: [TextFields] textFieldShouldClear is not invoked when clear button is tapped
The `UITextFieldDelegate` method `textFieldShouldClear` should be invoked when the clear button is tapped on an MDCTextField.  

https://developer.apple.com/documentation/uikit/uitextfielddelegate/1619594-textfieldshouldclear

willlarche:
Closes #2062 

Also added an equivalent to MDCMultilineTextField.

### DIFF
--- a/components/TextFields/src/MDCMultilineTextField.h
+++ b/components/TextFields/src/MDCMultilineTextField.h
@@ -18,6 +18,7 @@
 
 #import "MDCTextInput.h"
 
+@protocol MDCMultilineTextInputDelegate;
 @protocol MDCMultilineTextInputLayoutDelegate;
 
 /**
@@ -28,6 +29,9 @@
 
 /** A mirror of the same property that already exists on UITextField, UITextView, and UILabel. */
 @property(nonatomic, assign) BOOL adjustsFontForContentSizeCategory;
+
+/** An optional delegate for useful methods not included in UITextViewDelegate.*/
+@property(nonatomic, nullable, weak) IBOutlet id<MDCMultilineTextInputDelegate> multilineDelegate;
 
 /**
  The delegate for changes to preferred content size.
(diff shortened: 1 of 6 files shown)


components/TextFields/src/MDCMultilineTextField.h
```diff
@@ -29,6 +30,9 @@
 /** A mirror of the same property that already exists on UITextField, UITextView, and UILabel. */
 @property(nonatomic, assign) BOOL adjustsFontForContentSizeCategory;
 
+/** An optional delegate for useful methods not included in UITextViewDelegate.*/
+@property(nonatomic, nullable, weak) IBOutlet id<MDCMultilineTextInputDelegate> multilineDelegate;
```
romoore, line 34: Follow-up PR: Unit tests for encoding/decoding?

romoore (approved):

components/TextFields/src/MDCMultilineTextField.h
willlarche: I couldn’t figure out how. I don’t have any classes that conform.

romoore: You could create a custom test class that conforms and encode/decode. If you wanted to be super-sneaky, slap those both into an NSArray, encode/decode the array and verify that the relationship still holds.

willlarche: Do I need to make that class in another file or can I do it in the unit test file at the top?

romoore: I put them at the top of the test file until I have more than one. Then it's a test Supplemental?

Issue #1015, opened 2016-12-13
MDCPageControl crashes with `-initWithFrame:CGRectZero`


ajsecord:
From internal report b/33557156:

> The MDCPageControl seems to only do internal calculations inside commonMDCPageControlInit which is only called during init.  So if the page control is not initialized with at least the final size, then MDCPageControl crashes upon the first scrollViewDid scroll callback because track length is incorrectly negative.  Basically it looks like MDCPageControl doesn't support resizing.
> 
> It is expected to be able to resize UIViews without crashing the app.  The documentation on https://material-ext.appspot.com/mdc-ios-preview/components/PageControl/ is not sufficient because it describe where "myFrame" should come from for "Page control configuration".

![screen shot 2016-12-12 at 6 15 17 pm](https://cloud.githubusercontent.com/assets/742162/21141118/75da58ea-c109-11e6-86a4-b594fe11a107.png)

PR #2197, opened 2017-10-16
[PageControl] Fix bug with OOB array lookup when calling scrollViewDidScroll.

kolodny:
The crash referenced in https://github.com/material-components/material-components-ios/pull/2072#issue-261964446 still applies even after the fix of #2132

This PR covers the invalid lookup in the `scrollViewDidScroll` method; tests included.

### DIFF
--- a/components/PageControl/src/MDCPageControl.m
+++ b/components/PageControl/src/MDCPageControl.m
@@ -276,12 +276,7 @@ static inline CGFloat normalizeValue(CGFloat value, CGFloat minRange, CGFloat ma
                                                mediaTimingFunction:animation.timingFunction];
                    });
 
-  } else if (scrolledPercentage >= 0 && scrolledPercentage <= 1) {
-
-    if (_numberOfPages == 0) {
-      return;
-    }
-
+  } else if (scrolledPercentage >= 0 && scrolledPercentage <= 1 && _numberOfPages > 0) {
     // Update active indicator position.
     CGFloat transformX = scrolledPercentage * _trackLength;
     if (!_isDeferredScrolling) {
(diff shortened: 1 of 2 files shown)


components/PageControl/src/MDCPageControl.m
```diff
@@ -277,6 +277,11 @@ - (void)scrollViewDidScroll:(UIScrollView *)scrollView {
                    });
 
   } else if (scrolledPercentage >= 0 && scrolledPercentage <= 1) {
+
+    if (_numberOfPages == 0) {
```
ianegordon: If this in a general guard statement, move to the beginning of the method?

If not, consider adding it in the if clause:
else if (scrolledPercentage >= 0 && scrolledPercentage <= 1 && _numberOfPages > 0)

A return in the middle of a method can easily get lost.

components/PageControl/src/MDCPageControl.m
```diff
@@ -277,6 +277,11 @@ - (void)scrollViewDidScroll:(UIScrollView *)scrollView {
                    });
 
   } else if (scrolledPercentage >= 0 && scrolledPercentage <= 1) {
+
```
ianegordon: Remove empty line?

ianegordon (changes requested):

components/PageControl/src/MDCPageControl.m
kolodny: Done

components/PageControl/src/MDCPageControl.m
kolodny: Added it to the if clause above

ianegordon (approved):

ianegordon:
Many thanks! I will merge once Travis is happy.

romoore:
Closed #1015
Closed #2072

PR #2202, opened 2017-10-17
[Tabs] Fix a bug that where an item cell title can be missing

### LINKED ISSUE #2201: [Tabs] Item cell title can be missing after an appearance change.
### Overview
If you have an MDCTabBar that was initially created with appearance MDCTabBarItemAppearanceImages, and then later changed to MDCTabBarItemAppearanceTitledImages, this can result in tabs missing their titles. This is caused by an ordering issue in MDCItemBarCell.m between when the label's text property is set and the hidden property.

I will submit a pull request to resolve this.

csilv:
After an appearance change, tab titles might be missing. This is caused by an ordering issue in MDCItemBarCell.m where -updateDisplayedTitle will only set the .text property if the label is already visible. But when the style is changed, the visibility of the label is set after -updateDisplayedTitle is called. This can be resolved by checking _style.shouldDisplayTitle instead.

closes #2201

### DIFF
--- a/components/Tabs/src/private/MDCItemBarCell.m
+++ b/components/Tabs/src/private/MDCItemBarCell.m
@@ -563,9 +563,7 @@ static const NSTimeInterval kSelectionAnimationDuration = 0.3f;
 }
 
 - (void)updateDisplayedTitle {
-  if (_style.shouldDisplayTitle) {
-    _titleLabel.text = [[self class] displayedTitleForTitle:_title style:_style];
-  }
+  _titleLabel.text = [[self class] displayedTitleForTitle:_title style:_style];
 }
 
 @end
--- a/components/Tabs/src/private/MDCItemBarCell.m
+++ b/components/Tabs/src/private/MDCItemBarCell.m
@@ -563,7 +563,7 @@ static const NSTimeInterval kSelectionAnimationDuration = 0.3f;
 }
 
 - (void)updateDisplayedTitle {
-  if (!_titleLabel.hidden) {
+  if (_style.shouldDisplayTitle) {
     _titleLabel.text = [[self class] displayedTitleForTitle:_title style:_style];
   }
 }
--- a/components/Tabs/tests/unit/MDCItemBarCellTests.m
+++ b/components/Tabs/tests/unit/MDCItemBarCellTests.m
@@ -39,12 +39,12 @@
   style.shouldDisplayTitle = YES;
 
   // When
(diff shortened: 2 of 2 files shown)


chuga (approved):
lgtm, thanks for adding a test

components/Tabs/tests/unit/MDCItemBarCellTests.m
```diff
@@ -59,4 +59,29 @@ - (void)testTitleNumberOfLines {
...
+  MDCItemBarStyle *iconAndTextStyle = [[MDCItemBarStyle alloc] init];
+  iconAndTextStyle.shouldDisplayImage = YES;
+  iconAndTextStyle.shouldDisplayTitle = YES;
+
+  // Create a cell and set the style before settting the image/title. That is the order items will
+  // be configured in the app runtime.
+  MDCItemBarCell *cell = [[MDCItemBarCell alloc] initWithFrame:CGRectZero];
+  [cell applyStyle:iconOnlyStyle];
+  [cell setImage:[UIImage imageNamed:@"TabBarDemo_ic_info"]];
+  [cell setTitle:@"A title"];
```
ianegordon: Super-nit: Prefer invoking properties if they have been defined.
cell.image = [UIImage imageNamed:@"info"];
cell.title = @"A title";

There is an automated tool under Xcode -> Edit -> Convert -> Modern Obj-C Syntax

components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -563,7 +563,7 @@ - (void)updateDisplayedImage {
 }
 
 - (void)updateDisplayedTitle {
-  if (!_titleLabel.hidden) {
+  if (_style.shouldDisplayTitle) {
     _titleLabel.text = [[self class] displayedTitleForTitle:_title style:_style];
```
ianegordon: If we are using the .hidden property, can we drop the if clause and always set _titleLabel.text?

ianegordon (changes requested):

components/Tabs/src/private/MDCItemBarCell.m
csilv: Done.

components/Tabs/tests/unit/MDCItemBarCellTests.m
csilv: Done.

ianegordon (approved):

ianegordon:
Thank you @csilv !

PR #2214, opened 2017-10-17
[Dialogs] Set Dialog example button to open material.io

ianegordon:
This change will assist in diagnosing a problem with our dialog presentation controller not staying consistent.

https://stackoverflow.com/questions/46791185/modal-transition-not-working-after-returning-back-to-the-app

### DIFF
--- a/components/Dialogs/examples/supplemental/DialogsTypicalUseSupplemental.m
+++ b/components/Dialogs/examples/supplemental/DialogsTypicalUseSupplemental.m
@@ -87,6 +87,49 @@ static NSString * const kReusableIdentifierItem = @"cell";
   self.view.backgroundColor = [UIColor whiteColor];
 
   _dismissButton = [[MDCFlatButton alloc] init];
+  [_dismissButton setTitle:@"Dismiss" forState:UIControlStateNormal];
+  [_dismissButton setTitleColor:[UIColor blackColor] forState:UIControlStateNormal];
+  _dismissButton.autoresizingMask =
+      UIViewAutoresizingFlexibleTopMargin | UIViewAutoresizingFlexibleLeftMargin |
+      UIViewAutoresizingFlexibleRightMargin | UIViewAutoresizingFlexibleBottomMargin;
+  [_dismissButton addTarget:self
+                     action:@selector(dismiss:)
+           forControlEvents:UIControlEventTouchUpInside];
+
+  [self.view addSubview:_dismissButton];
+}
+
+- (void)viewWillLayoutSubviews {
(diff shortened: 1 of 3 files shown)


samnm:
It's possible the stackoverflow issue depends on the behavior where the opened website triggers the app opening again, so this may not expose the issue.

Also do we really want to land this? It changes all the Alert example dismiss buttons to open material.io? Or is it just one example.

ianegordon:
I'd like to be able to point the original questioner at this example in the catalog.  AKA Works for us, can you add more detail?

Right now it affects two examples, but I'll pull it into its own example.

components/Dialogs/examples/supplemental/DialogsTypicalUseSupplemental.m
```diff
@@ -110,7 +111,8 @@ - (CGSize)preferredContentSize {
 }
 
 - (IBAction)dismiss:(id)sender {
-  [self.presentingViewController dismissViewControllerAnimated:YES completion:NULL];
+  NSURL *testURL = [NSURL URLWithString:@"https://www.material.io"];
+  [[UIApplication mdc_safeSharedApplication] performSelector:@selector(openURL:) withObject:testURL];
```
willlarche: I don't know much about mdc_safeSharedApplication. Is this so it's safe in an extension?

willlarche (approved):

components/Dialogs/examples/supplemental/DialogsTypicalUseSupplemental.m
samnm: Yeah thats correct. In extensions you can't get access to [UIApplication sharedApplication], so `mdc_safeSharedApplication` returns nil inside extensions.

willlarche:
I agree with @ianegordon 's approach. I've been adding a lot of cases into my examples in the hopes they become UI tests.

components/Dialogs/examples/supplemental/DialogsTypicalUseSupplemental.m
willlarche: In that case, don't hate me, but a tiny comment explaining that would be helpful since it's educational code.

ianegordon:
Pulled into its own view controller.
Added comment per @willlarche request.
PTAL

samnm (approved):

willlarche:
I've looked at the Travis. This has passed.

haroldogtf:
@ianegordon I add this code to reproduce the Issue: https://github.com/haroldogtf/DialogTransition-Issue

PR #2215, opened 2017-10-17
[Chips] API review

samnm:
This is the API review for the core part of MDC Chips. Since Chips are predominately planned to be used through UICollectionViews there is not a lot of complexity. The integration with MDCTextFields will arrive later.

### DIFF
--- a/components/Chips/examples/ChipsTypicalUseViewController.m
+++ b/components/Chips/examples/ChipsTypicalUseViewController.m
@@ -1,0 +1,159 @@
+/*
+ Copyright 2017-present the Material Components for iOS authors. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import "ChipsTypicalUseViewController.h"
+
+#import "MaterialChips.h"
+
+@interface ChipModel : NSObject
+@property(nonatomic, strong) NSString *title;
+@property(nonatomic, strong) UIImage *image;
+@property(nonatomic, strong) UIImage *selectedImage;
+@property(nonatomic, strong) UIView *accessoryView;
+@end
+
+static inline ChipModel *MakeModel(NSString *title,
+                              UIImage *image,
+                              UIImage *selectedImage,
+                              UIView *accessoryView) {
+  ChipModel *chip = [[ChipModel alloc] init];
+  chip.title = title;
+  chip.image = image;
+  chip.selectedImage = selectedImage;
+  chip.accessoryView = accessoryView;
+  return chip;
+}
+
+@interface ChipsTypicalUseViewController () <UICollectionViewDelegate, UICollectionViewDataSource>
+@property(nonatomic, strong) NSArray<ChipModel *> *model;
+@end
+
+@implementation ChipsTypicalUseViewController {
+  BOOL _popRecognizerDelaysTouches;
+}
+
++ (NSArray *)catalogBreadcrumbs {
+  return @[ @"Chips", @"Chips" ];
+}
+
++ (NSString *)catalogDescription {
+  return @"Chips!";
+}
+
++ (BOOL)catalogIsPrimaryDemo {
+  return YES;
+}
+
+- (instancetype)init {
+  MDCAlignedCollectionViewFlowLayout *layout = [[MDCAlignedCollectionViewFlowLayout alloc] init];
+  layout.horizontalAlignment = MDCCollectionHorizontalAlignmentLeft;
+  layout.minimumInteritemSpacing = 10.0f;
+  layout.estimatedItemSize = CGSizeMake(60, 33);
+
+  self = [super initWithCollectionViewLayout:layout];
+  if (self) {
+    self.editing = YES;
+  }
+  return self;
+}
+
+- (void)loadView {
+  [super loadView];
+
+  self.collectionView.backgroundColor = [UIColor lightGrayColor];
+  self.collectionView.delaysContentTouches = NO;
+  self.collectionView.contentInset = UIEdgeInsetsMake(20, 20, 20, 20);
+  [self.collectionView registerClass:[MDCChipCollectionViewCell class]
+          forCellWithReuseIdentifier:@"Cell"];
+}
+
+- (void)viewDidAppear:(BOOL)animated {
+  [super viewDidAppear:animated];
+
+  _popRecognizerDelaysTouches =
+      self.navigationController.interactivePopGestureRecognizer.delaysTouchesBegan;
+  self.navigationController.interactivePopGestureRecognizer.delaysTouchesBegan = NO;
+}
+
+- (void)viewWillDisappear:(BOOL)animated {
+  self.navigationController.interactivePopGestureRecognizer.delaysTouchesBegan =
+  _popRecognizerDelaysTouches;
+}
+
+- (NSInteger)collectionView:(UICollectionView *)collectionView
+     numberOfItemsInSection:(NSInteger)section {
+  return self.model.count;
+}
+
+- (__kindof UICollectionViewCell *)collectionView:(UICollectionView *)collectionView
+                           cellForItemAtIndexPath:(NSIndexPath *)indexPath {
+  MDCChipCollectionViewCell *cell =
+      [collectionView dequeueReusableCellWithReuseIdentifier:@"Cell" forIndexPath:indexPath];
+
+  ChipModel *model = self.model[indexPath.row];
+  MDCChipView *chip = cell.chip;
+  chip.titleLabel.text = model.title;
+  chip.imageView.image = model.image;
+  chip.selectedImageView.image = model.selectedImage;
+  chip.accessoryView = model.accessoryView;
+
+  return cell;
+}
+
+- (NSArray *)model {
+  if (!_model) {
+    _model = @[
+      MakeModel(@"Chip", [self faceImage], nil, nil),
+      MakeModel(@"Chip", [self faceImage], nil, [self deleteButton]),
+      MakeModel(@"Chip", nil, nil, [self deleteButton]),
+      MakeModel(@"Chip", [self faceImage], [self doneImage], [self deleteButton]),
+      MakeModel(@"Chip", nil, [self doneImage], [self deleteButton]),
+    ];
+  }
+  return _model;
+}
+
+- (UIImage *)faceImage {
+  UIImage *image = [UIImage imageNamed:@"ic_face"
+                              inBundle:[NSBundle bundleForClass:[self class]]
+         compatibleWithTraitCollection:nil];
+  return image;
+}
+
+- (UIImage *)doneImage {
+  UIImage *image = [UIImage imageNamed:@"ic_done"
+                              inBundle:[NSBundle bundleForClass:[self class]]
+         compatibleWithTraitCollection:nil];
+  return image;
+}
+
+- (UIButton *)deleteButton {
+  UIImage *image = [UIImage imageNamed:@"ic_check_circle_18pt"
+                              inBundle:[NSBundle bundleForClass:[self class]]
+         compatibleWithTraitCollection:nil];
+  image = [image imageWithRenderingMode:UIImageRenderingModeAlwaysTemplate];
+
+  UIButton *button = [[UIButton alloc] initWithFrame:CGRectZero];
+  button.tintColor = [UIColor colorWithWhite:0 alpha:0.7];
+  [button setImage:image forState:UIControlStateNormal];
+
+  return button;
+}
+
+@end
+
+@implementation ChipModel
+@end
--- a/components/Chips/src/MDCAlignedCollectionViewFlowLayout.h
+++ b/components/Chips/src/MDCAlignedCollectionViewFlowLayout.h
@@ -1,0 +1,41 @@
+/*
+ Copyright 2017-present the Material Components for iOS authors. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+typedef enum : NSUInteger {
+  MDCCollectionHorizontalAlignmentLeading,
+  MDCCollectionHorizontalAlignmentTrailing,
+  MDCCollectionHorizontalAlignmentLeft,
+  MDCCollectionHorizontalAlignmentRight
+} MDCCollectionHorizontalAlignment;
+
+/*
+ MDCAlignedCollectionViewFlowLayout is an aligned version of UICollectionViewFlowLayout.
+
+ Use exactly as you would a standard UICollectionViewLayout. Set minimumInteritemSpacing to control
+ spacing between cells.
+ */
+@interface MDCAlignedCollectionViewFlowLayout : UICollectionViewFlowLayout
+
+/*
+ The horizontal alignment of the layout.
+
+ Defaults to MDCCollectionHorizontalAlignmentLeading.
+*/
+@property(nonatomic, assign) MDCCollectionHorizontalAlignment horizontalAlignment;
+
+@end
--- a/components/Chips/src/MDCAlignedCollectionViewFlowLayout.h
+++ b/components/Chips/src/MDCAlignedCollectionViewFlowLayout.h
@@ -16,12 +16,12 @@
 
 #import <UIKit/UIKit.h>
 
-typedef enum : NSUInteger {
+typedef NS_ENUM(NSInteger, MDCCollectionHorizontalAlignment) {
   MDCCollectionHorizontalAlignmentLeading,
   MDCCollectionHorizontalAlignmentTrailing,
   MDCCollectionHorizontalAlignmentLeft,
   MDCCollectionHorizontalAlignmentRight
-} MDCCollectionHorizontalAlignment;
+};
 
 /*
  MDCAlignedCollectionViewFlowLayout is an aligned version of UICollectionViewFlowLayout.
--- a/components/Chips/src/MDCChipCollectionViewCell.h
+++ b/components/Chips/src/MDCChipCollectionViewCell.h
@@ -23,6 +23,11 @@
  */
 @interface MDCChipCollectionViewCell : UICollectionViewCell
 
+/*
+ The chip view.
+
+ The chip's bounds will be set to fill the collection view cell.
+ */
 @property(nonatomic, nonnull, strong, readonly) MDCChipView *chipView;
 
 /*
--- a/components/Chips/src/MDCChipCollectionViewCell.h
+++ b/components/Chips/src/MDCChipCollectionViewCell.h
@@ -25,4 +25,12 @@
 
 @property(nonatomic, nonnull, strong, readonly) MDCChipView *chipView;
 
+/*
+ Animates the chip view every time the bounds change. Defaults to NO.
+
+ Set this to YES to animate chip view when it resizes. If the chip view changes size upon selection
+ this should be set to YES.
+ */
+@property(nonatomic, assign) BOOL alwaysAnimateCellResize;
+
 @end
--- a/components/Chips/src/MDCChipCollectionViewCell.h
+++ b/components/Chips/src/MDCChipCollectionViewCell.h
@@ -23,6 +23,6 @@
  */
 @interface MDCChipCollectionViewCell : UICollectionViewCell
 
-@property(nonatomic, nonnull, strong, readonly) MDCChipView *chip;
+@property(nonatomic, nonnull, strong, readonly) MDCChipView *chipView;
 
 @end
--- a/components/Chips/src/MDCChipCollectionViewCell.h
+++ b/components/Chips/src/MDCChipCollectionViewCell.h
@@ -1,0 +1,28 @@
+/*
+ Copyright 2017-present the Material Components for iOS authors. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+@class MDCChipView;
+
+/*
+ A collection view cell containing a single MDCChipView.
+ */
+@interface MDCChipCollectionViewCell : UICollectionViewCell
+
+@property(nonatomic, nonnull, strong, readonly) MDCChipView *chip;
+
+@end
--- a/components/Chips/src/MDCChipView.h
+++ b/components/Chips/src/MDCChipView.h
@@ -72,7 +72,7 @@
  UIContentSizeCategory is changed.
 
  This property is modeled after the adjustsFontForContentSizeCategory property in the
- UIConnectSizeCategoryAdjusting protocol added by Apple in iOS 10.0.
+ UIContentSizeCategoryAdjusting protocol added by Apple in iOS 10.0.
 
  If set to YES, this button will base its text font on MDCFontTextStyleButton.
 
--- a/components/Chips/src/MDCChipView.h
+++ b/components/Chips/src/MDCChipView.h
@@ -36,7 +36,7 @@
 @property(nonatomic, readonly, nonnull) IBInspectable UIImageView *imageView;
 
 /*
- A UIImageView that leads the title label. Appears above the imageView.
+ A UIImageView that leads the title label. Appears in front of the imageView.
 
  This image view is typically used to show some icon that denotes the chip as selected, such as a
  tick. If imageView has no image then the chip will require resizing when selected or deselected to
--- a/components/Chips/src/MDCChipView.h
+++ b/components/Chips/src/MDCChipView.h
@@ -1,0 +1,169 @@
+/*
+ Copyright 2017-present the Material Components for iOS authors. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+@interface MDCChipView : UIControl
+
+/*
+ A UIImageView that leads the title label.
+
+ Only visible when the image is non-nil.
+ */
+@property(nonatomic, readonly, nonnull) UIImageView *imageView;
+
+/*
+ A UIImageView that leads the title label. Appears above the imageView.
+
+ Only visible when the chip is selected and the image is non-nil.
+ */
+@property(nonatomic, readonly, nonnull) UIImageView *selectedImageView;
+
+/*
+ A UIView that trails the title label.
+
+ It will be given a size based on the value returned from sizeThatFits:.
+ */
+@property(nonatomic, strong, nullable) UIView *accessoryView;
+
+/*
+ The title label.
+
+ @note The title color is controlled by setTitleColor:forState:.
+ */
+@property(nonatomic, readonly, nonnull) UILabel *titleLabel;
+
+/*
+ Padding around the image view. Only used if the image view has a non-nil image.
+
+ Defaults to (0, 0, 0, 0).
+ */
+@property(nonatomic, assign) UIEdgeInsets imagePadding UI_APPEARANCE_SELECTOR;
+
+/*
+ Padding around the accessory view. Only used if the accessory view is non-nil.
+
+ Defaults to (0, 0, 0, 0).
+ */
+@property(nonatomic, assign) UIEdgeInsets accessoryPadding UI_APPEARANCE_SELECTOR;
+
+/*
+ Padding around the title.
+
+ Defaults to (8, 12, 8, 12).
+ */
+@property(nonatomic, assign) UIEdgeInsets titlePadding UI_APPEARANCE_SELECTOR;
+
+/*
+ Indicates whether the chip should automatically update its font when the device’s
+ UIContentSizeCategory is changed.
+
+ This property is modeled after the adjustsFontForContentSizeCategory property in the
+ UIConnectSizeCategoryAdjusting protocol added by Apple in iOS 10.0.
+
+ If set to YES, this button will base its text font on MDCFontTextStyleButton.
+
+ Default value is NO.
+ */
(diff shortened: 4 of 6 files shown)


components/Chips/src/MDCChipCollectionViewCell.h
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+#import <UIKit/UIKit.h>
+
+@class MDCChipView;
+
+/*
+ A collection view cell containing a single MDCChipView.
+ */
+@interface MDCChipCollectionViewCell : UICollectionViewCell
```
jverkoey, line 24: Can this be private?

samnm: I don't think so. The intention is for people to create their own UICollectionView and manage the data source themselves.

jverkoey: Ah, interesting. Can you include an example of usage in the PR? I'm not able to clearly see how this component is meant to be used.

samnm: I've pushed up an example that shows one use case.

components/Chips/examples/ChipsTypicalUseViewController.m
```diff
@@ -0,0 +1,159 @@
...
+#import "MaterialChips.h"
+
+@interface ChipModel : NSObject
+@property(nonatomic, strong) NSString *title;
+@property(nonatomic, strong) UIImage *image;
+@property(nonatomic, strong) UIImage *selectedImage;
+@property(nonatomic, strong) UIView *accessoryView;
+@end
+
+static inline ChipModel *MakeModel(NSString *title,
```
willlarche, line 28: I don't usually see inline functions used to make view models of objects. I think this would be more expected just as an init with parameters or a convenience class method with parameters. But that's not crazy important right now.

components/Chips/src/MDCAlignedCollectionViewFlowLayout.h
```diff
@@ -0,0 +1,41 @@
...
+
+typedef enum : NSUInteger {
+  MDCCollectionHorizontalAlignmentLeading,
+  MDCCollectionHorizontalAlignmentTrailing,
+  MDCCollectionHorizontalAlignmentLeft,
+  MDCCollectionHorizontalAlignmentRight
+} MDCCollectionHorizontalAlignment;
+
+/*
+ MDCAlignedCollectionViewFlowLayout is an aligned version of UICollectionViewFlowLayout.
```
ajsecord: I wasn't sure what "aligned" meant at first. Add a second sentence "UICollectionViewFlowLayout only provides justified alignment of its cells, MDCAlignedCollectionViewFlowLayout adds flush left and flush right alignments."? https://en.wikipedia.org/wiki/Typographic_alignment

Might consider renaming to MDCFlushAligned*, but that's a nit.

components/Chips/src/MDCChipCollectionViewCell.h
ajsecord, line 24: If MDCChipCollectionViewCell remains public, document how it contains the chip view, e.g. "The contained MDCChipView is resized to fill the cell." (or whatever is true).

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,169 @@
...
+ Only visible when the image is non-nil.
+ */
+@property(nonatomic, readonly, nonnull) UIImageView *imageView;
+
+/*
+ A UIImageView that leads the title label. Appears above the imageView.
+
+ Only visible when the chip is selected and the image is non-nil.
+ */
+@property(nonatomic, readonly, nonnull) UIImageView *selectedImageView;
```
ajsecord: Annoying question: no `imageView:forState:`? I'm assuming not, but you should probably document. :)

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,169 @@
...
+ */
+- (CGFloat)elevationForState:(UIControlState)state;
+
+/*
+ Sets the elevation for a particular control state.
+
+ @param elevation The elevation to set.
+ @param state The state to set.
+ */
+- (void)setElevation:(CGFloat)elevation forState:(UIControlState)state UI_APPEARANCE_SELECTOR;
```
ajsecord: Elevations are now `MDCShadowElevation` instead of `CGFloat`.

ajsecord (changes requested):

components/Chips/src/MDCChipView.h
romoore: +1 To the documentation request. Is the intention to limit the use of focused (tvOS)/highlighted(iOS) state changes to the images?

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,169 @@
...
+
+ @param state The state.
+ @return The background color.
+ */
+- (nullable UIColor *)backgroundColorForState:(UIControlState)state;
+
+/*
+ A color used as the chip's @c backgroundColor.
+
+ If left unset or reset to nil for a particular control state, then a default blue color is used.
```
romoore: I would prefer that we follow UIKit's typical `*:forState:` conventions here to avoid surprising users.  Is there a reason why we're departing from [UIKit's](https://developer.apple.com/documentation/uikit/uibutton/1623997-setimage?language=objc) [convention](https://developer.apple.com/documentation/uikit/uibutton/1623993-settitlecolor?language=objc) of [falling back to the value for `.Normal`](https://developer.apple.com/documentation/uikit/uibutton/1624018-settitle?language=objc) if a particular state isn't matched?  I feel like the default blue can be achieved through UIAppearance rather than a baked-in default.

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,169 @@
...
+    BOOL mdc_adjustsFontForContentSizeCategory UI_APPEARANCE_SELECTOR;
+
+/*
+ A color used as the chip's @c backgroundColor for @c state.
+
+ @param state The state.
+ @return The background color.
+ */
+- (nullable UIColor *)backgroundColorForState:(UIControlState)state;
+
```
romoore: Based on my reading the documentation for `setBackgroundColor:forState:`, this return value should be `nonnull` since a blue color will be returned if no other match can be found.  However, I would prefer to return `nil` if no values for any state have been set.

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,169 @@
...
+- (void)setBackgroundColor:(nullable UIColor *)backgroundColor forState:(UIControlState)state
+    UI_APPEARANCE_SELECTOR;
+
+/*
+ Returns the border color for a particular control state.
+
+ @param state The control state to retrieve the elevation.
+ @return The border color for the requested state.
+ */
+- (nullable UIColor *)borderColorForState:(UIControlState)state;
```
romoore: Same suggestion to follow UIKit's convention of falling-back to `.Normal`.

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,169 @@
...
+ @param backgroundColor The background color.
+ @param state The state.
+ */
+- (void)setBackgroundColor:(nullable UIColor *)backgroundColor forState:(UIControlState)state
+    UI_APPEARANCE_SELECTOR;
+
+/*
+ Returns the border color for a particular control state.
+
+ @param state The control state to retrieve the elevation.
```
romoore: Nit: s/elevation/border color

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,169 @@
...
+ @param borderColor The border color.
+ @param state The state.
+ */
+- (void)setBorderColor:(nullable UIColor *)borderColor forState:(UIControlState)state
+    UI_APPEARANCE_SELECTOR;
+
+/*
+ Returns the border width for a particular control state.
+
+ @param state The control state to retrieve the elevation.
```
romoore: Nit: s/elevation/border width

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,169 @@
...
+ @note The title color is controlled by setTitleColor:forState:.
+ */
+@property(nonatomic, readonly, nonnull) UILabel *titleLabel;
+
+/*
+ Padding around the image view. Only used if the image view has a non-nil image.
+
+ Defaults to (0, 0, 0, 0).
+ */
+@property(nonatomic, assign) UIEdgeInsets imagePadding UI_APPEARANCE_SELECTOR;
```
romoore: Are `imagePadding`, `titlePadding`, and `accessoryPadding` included or excluded in the value returned by `intrinsicContentSize`?

components/Chips/src/MDCChipView.h
romoore: Just curious: If users call this with a negative elevation, are we bounding the value to some internal min/max?

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,169 @@
...
+ */
+- (nullable UIColor *)titleColorForState:(UIControlState)state;
+
+/*
+ Sets the title color for a particular control state.
+
+ @param titleColor The title color.
+ @param state The state.
+ */
+- (void)setTitleColor:(nullable UIColor *)titleColor forState:(UIControlState)state
```
romoore: Same suggestion to follow [UIButton's convention](https://developer.apple.com/documentation/uikit/uibutton/1623993-settitlecolor?language=objc) when handling values for state.

components/Chips/src/MDCAlignedCollectionViewFlowLayout.h
jverkoey: This isn't necessarily correct:

> UICollectionViewFlowLayout only provides justified alignment of its cells

By default flow layout is flush left.

components/Chips/src/MDCAlignedCollectionViewFlowLayout.h
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+#import <UIKit/UIKit.h>
+
+typedef enum : NSUInteger {
+  MDCCollectionHorizontalAlignmentLeading,
+  MDCCollectionHorizontalAlignmentTrailing,
+  MDCCollectionHorizontalAlignmentLeft,
+  MDCCollectionHorizontalAlignmentRight
+} MDCCollectionHorizontalAlignment;
```
jverkoey: Consider using `UIControlContentHorizontalAlignment` instead of recreating the API.

jverkoey (changes requested):

components/Chips/src/MDCAlignedCollectionViewFlowLayout.h
samnm: Jeff that isn't correct. UICollectionViewFlowLayout does not left align cells. It only provides a justified layout. Hence the need for this class.

components/Chips/src/MDCAlignedCollectionViewFlowLayout.h
samnm: I think it makes more sense to use a custom type here. We will never have a `fill` horizontalAlignment.

components/Chips/src/MDCAlignedCollectionViewFlowLayout.h
jverkoey: Hah! Well I'll be darned - I'd always read `minimumInteritemSpacing` as simply `iteritemSpacing`. Is showing how long since I've used a flow layout from scratch haha.

tumblr_inline_orc8s2sita1tlruqr_540

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,169 @@
...
+ @param state The state.
+ */
+- (void)setBackgroundColor:(nullable UIColor *)backgroundColor forState:(UIControlState)state
+    UI_APPEARANCE_SELECTOR;
+
+/*
+ Returns the border color for a particular control state.
+
+ @param state The control state to retrieve the elevation.
+ @return The border color for the requested state.
```
willlarche: Document the default too.

willlarche (approved):

components/Chips/src/MDCChipCollectionViewCell.h
```diff
@@ -0,0 +1,28 @@
...
+
+@property(nonatomic, nonnull, strong, readonly) MDCChipView *chip;
```
brianjmoore: Probably better to just call this "chipView" so its type is clear (much like you wouldn't call an image view property "image")

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,169 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+@interface MDCChipView : UIControl
```
brianjmoore: This will be documented in the real implementation, I assume?

components/Chips/src/MDCChipView.h
brianjmoore: "Padding" isn't a concept that UIKit APIs use as far as I can tell - would it change the meaning to hew closer to UIKit's established conventions and call these "imageInsets", "titleInsets" and "accessoryInsets"?

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,169 @@
...
+ Defaults to (8, 12, 8, 12).
+ */
+@property(nonatomic, assign) UIEdgeInsets titlePadding UI_APPEARANCE_SELECTOR;
+
+/*
+ Indicates whether the chip should automatically update its font when the device’s
+ UIContentSizeCategory is changed.
+
+ This property is modeled after the adjustsFontForContentSizeCategory property in the
+ UIConnectSizeCategoryAdjusting protocol added by Apple in iOS 10.0.
```
brianjmoore: *Content

brianjmoore (commented):
This looks great. It's a super clean design!

components/Chips/src/MDCChipCollectionViewCell.h
jverkoey: It may be beneficial to also make MDCChipCollectionViewCell a final class if there isn't any particularly novel logic contained within.

jverkoey: If we remove this cell, I expect that there'd be relatively minor changes to client code: they'd have to implement their own cell with a chip inside in addition to managing the collection view.

Given that we're already expecting them to create and manage their own collection view, would it be reasonable for us to remove this cell class so that we have one less API to maintain?

components/Chips/src/MDCAlignedCollectionViewFlowLayout.h
```diff
@@ -0,0 +1,41 @@
...
+
+/*
+ MDCAlignedCollectionViewFlowLayout is an aligned version of UICollectionViewFlowLayout.
+
+ Use exactly as you would a standard UICollectionViewLayout. Set minimumInteritemSpacing to control
+ spacing between cells.
+ */
+@interface MDCAlignedCollectionViewFlowLayout : UICollectionViewFlowLayout
```
jverkoey: Can we move this flow layout to its own component? This would allow this component to focus solely on representation of the chip view, with clients being able to choose how they integrate it in their app.

We'll probably still recommend using collection views, in which case we'd encourage them to use this layout component in our examples.

This also allows clients to make use of this aligned flow layout without pulling in the chip componentry.

components/Chips/src/MDCChipCollectionViewCell.h
willlarche: I would kind of expect a protocol and we have one class that implements it. Or you can roll your own.

jverkoey: @willlarche Who is the intended client of the protocol?

willlarche: @jverkoey whomever is doing the layout calculations. I don't know much about custom layouts but I would have guessed that it would need to know some standard information about what it's laying out since these are dynamically sized. Maybe flow layout does that for you automatically? Or I know you can use autolayout to get size but that requires, often, a cached cell for sizing.

I guess I want some help understanding how complicated we are expecting the layout to be above what comes free. And if it is very complicated, would it be helpful to have a common API, like from a protocol, that you could make the views conform to to give the layout machine the info it needs but still allow anyone to make their own?

samnm: The cell has a few pieces of logic that are unintuitive and if we expected clients to recreate it every time I suspect they would make a number of common mistakes. For example collection views manage highlighted and selected state for the cells so ideally we pipe that through to the Chip. I don't think providing a Cell is too providing too much API and it would be a mistake to exclude it.

As for a protocol for layout calculations, UICollectionView API already supports everything we need for that. If you want self sizing cells you use self sizing cells, if you want explicitly sized cells you use explicit sizes.

components/Chips/examples/ChipsTypicalUseViewController.m
```diff
@@ -0,0 +1,159 @@
...
+  layout.estimatedItemSize = CGSizeMake(60, 33);
+
+  self = [super initWithCollectionViewLayout:layout];
+  if (self) {
+    self.editing = YES;
+  }
+  return self;
+}
+
+- (void)loadView {
```
ianegordon, line 73: My understanding is that loadView is typically only overwritten if you are creating your view and its hierarchy directly.
Since we aren't doing that, can we move this to viewDidLoad?

components/Chips/examples/ChipsTypicalUseViewController.m
```diff
@@ -0,0 +1,159 @@
...
+  [super viewDidAppear:animated];
+
+  _popRecognizerDelaysTouches =
+      self.navigationController.interactivePopGestureRecognizer.delaysTouchesBegan;
+  self.navigationController.interactivePopGestureRecognizer.delaysTouchesBegan = NO;
+}
+
+- (void)viewWillDisappear:(BOOL)animated {
+  self.navigationController.interactivePopGestureRecognizer.delaysTouchesBegan =
+  _popRecognizerDelaysTouches;
```
ianegordon, line 93: nit: ident

components/Chips/examples/ChipsTypicalUseViewController.m
```diff
@@ -0,0 +1,159 @@
...
+      MakeModel(@"Chip", [self faceImage], [self doneImage], [self deleteButton]),
+      MakeModel(@"Chip", nil, [self doneImage], [self deleteButton]),
+    ];
+  }
+  return _model;
+}
+
+- (UIImage *)faceImage {
+  UIImage *image = [UIImage imageNamed:@"ic_face"
+                              inBundle:[NSBundle bundleForClass:[self class]]
```
ianegordon, line 131: I know we cannot, but I wish we could return Ponch and Jon for the faces.

components/Chips/src/MDCAlignedCollectionViewFlowLayout.h
```diff
@@ -0,0 +1,41 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+typedef enum : NSUInteger {
```
ianegordon: NS_ENUM(NSInteger, MDCCollectionHorizontalAlignment)

components/Chips/src/MDCChipCollectionViewCell.h
ianegordon: I think there is utility in a public MDCChipCell, but agree we should indicate it is not to be subclassed.

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,169 @@
...
+ Only visible when the chip is selected and the image is non-nil.
+ */
+@property(nonatomic, readonly, nonnull) UIImageView *selectedImageView;
+
+/*
+ A UIView that trails the title label.
+
+ It will be given a size based on the value returned from sizeThatFits:.
+ */
+@property(nonatomic, strong, nullable) UIView *accessoryView;
```
ianegordon: Exposing a readwrite view property means that we are going to have to cleanup / rebuild the hierarchy in every dequeCell / prepareCellForReuse.  Is there any way we can do this without a writable view?

If we have selectedImageView, do we need selectedAccessoryView?
It would mimic UITableViewCell's editingAccessoryView.

components/Chips/src/MDCChipView.h
samnm: Fixed. This value will be piped directly to MDCShadowLayer so it will act exactly as that does.

components/Chips/src/MDCChipView.h
samnm: This behaves differently to `imageView:forState:` might. This image view is only visible in the selected state and appears above the standard `imageView`. I'm open to suggestions as to how to improve the doc string and/or property name.

components/Chips/src/MDCChipView.h
samnm: I intend to follow UIKit's typical xForState patterns. The doc string is wrong.

components/Chips/examples/ChipsTypicalUseViewController.m
samnm: Yes this example will change before landing. I've only included it for context as to how the API will be used.

components/Chips/src/MDCChipView.h
romoore: So would it be possible for a chip to have both imageViews visible and on-screen at the same time? I think I'm having trouble understanding why there is a separate UIImageView rather than just changing the displayed image.

samnm: Yes the selectedImageView appears on top of the standard imageView.

components/Chips/src/MDCChipView.h
samnm: If you have any suggestions for other ways to handle this I'm all ears, but this appears to be the best compromise.

We don't need selectedAccessoryView. `selectedImageView` Serves a very specific purpose in reflecting what is described in the Chip designs.

components/Chips/src/MDCChipView.h
samnm: Fixed.

components/Chips/src/MDCChipView.h
samnm: I thought about this a bit and at the time decided "padding" made the most sense. In UIKit insets aren't always synonymous with padding so it isn't ideal IMO. For example UIButton titleEdgeInsets aren't used in sizing (but contentEdgeInsets are). That said, I'm open to discussion.

components/Chips/src/MDCChipCollectionViewCell.h
samnm: Fixed.

components/Chips/src/MDCChipView.h
samnm: They are included in `intrinsicContentSize` and `sizeThatFits`.

romoore: Awesome. I think it's worth adding this to the document as is done for [UIButton.contentEdgeInsets](https://developer.apple.com/documentation/uikit/uibutton/1624036-contentedgeinsets).

> The chip uses this property to determine intrinsicContentSize and sizeThatFits(_:).

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,170 @@
...
+
+/*
+ The title label.
```
bhamiltoncx: Is this for gray placeholder text in the text input field which disappears when text is entered? Or is it a different title?

If it's for a different title, we'll also want a placeholder label.

bhamiltoncx (commented):
(Please ignore my earlier comment about placeholder text, I misunderstood what this pull request included.)

Just to understand, this PR is just for a read-only chips UI, correct?

Clients are likely to also need an editable chips UI, with things like:

1) A cell which can grow/shrink in height as chips are entered
2) Validation and normalization APIs for text input
3) Placeholder text when no chips have yet been entered

willlarche:
@bhamiltoncx Hey, Ben! Don't worry. That functionality is planned to be included. We just do things in waves of PRs to make it easier for code reviewers. Stay tuned.

In the meanwhile, I'd start getting familiar with MDC TextFields since they will power the chips text inputting. If you have any questions about them, feel free to reach out to me directly. I'm the owner.

bhamiltoncx:
Thanks for clarifying! I'll take a look at `MDCTextFields` as well.

/cc @saidinwot since he'll likely be interested in this PR as well as follow-ups.

willlarche:
And to clarify, the chips will be backed, most likely, by the MDCMultilineTextField class and will support all the controllers that support multiline.

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,170 @@
...
+#import <UIKit/UIKit.h>
+
+@interface MDCChipView : UIControl
+
+/*
+ A UIImageView that leads the title label.
+
+ Only visible when the image is non-nil.
+ */
+@property(nonatomic, readonly, nonnull) UIImageView *imageView;
```
ianegordon: IB_Inspectable

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,170 @@
...
+
+/*
+ A UIImageView that leads the title label. Appears above the imageView.
+
+ Only visible when the chip is selected and the image is non-nil.
+ */
+@property(nonatomic, readonly, nonnull) UIImageView *selectedImageView;
```
ianegordon: IB_Inspectable

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,170 @@
...
+ It will be given a size based on the value returned from sizeThatFits:.
+ */
+@property(nonatomic, strong, nullable) UIView *accessoryView;
+
+/*
+ The title label.
+
+ @note The title color is controlled by setTitleColor:forState:.
+ */
+@property(nonatomic, readonly, nonnull) UILabel *titleLabel;
```
ianegordon: IB_Inspectable

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,170 @@
...
+@property(nonatomic, readwrite, setter=mdc_setAdjustsFontForContentSizeCategory:)
+    BOOL mdc_adjustsFontForContentSizeCategory UI_APPEARANCE_SELECTOR;
+
+/*
+ A color used as the chip's @c backgroundColor for @c state.
+
+ @param state The control state.
+ @return The background color.
+ */
+- (nullable UIColor *)backgroundColorForState:(UIControlState)state;
```
ianegordon, line 107: IB_Inspectable

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,170 @@
...
+/*
+ A color used as the chip's @c backgroundColor.
+
+ If left unset or reset to nil for a particular control state, then a default blue color is used.
+
+ @param backgroundColor The background color.
+ @param state The control state.
+ */
+- (void)setBackgroundColor:(nullable UIColor *)backgroundColor forState:(UIControlState)state
+    UI_APPEARANCE_SELECTOR;
```
ianegordon, line 118: Is the expectation that is color with be consistent across the app (UI_APPEARANCE_SELECTOR) or set per-instance (IB_Inspectable)?
Here and the other setX:forState:

ianegordon (changes requested):

components/Chips/src/MDCChipCollectionViewCell.h
samnm: Making it final would be nice, but if people subclass MDCChipView I would like them to easily be able to use their subclass in tandem with MDCChipCollectionViewCell. Any suggestions that work well with both? `@property(nonatomic) Class chipClass;`?

components/Chips/src/MDCChipView.h
samnm: I've expanded the comment to make it more clear.

components/Chips/src/MDCChipView.h
samnm: Oops, added a comment.

components/Chips/src/MDCAlignedCollectionViewFlowLayout.h
samnm: We should eventually move the layout to its own component. I think the best plan of attack might be to make this MDCChipCollectionViewLayout, and then build out a more full featured and generic Aligned layout later.

components/Chips/src/MDCChipView.h
samnm: Consistent across container views I think. Definitely not per-instance.

components/Chips/src/MDCChipView.h
samnm: I don't think this can be IBInspectable. I have set the view properties to be IBInspectable.

components/Chips/src/MDCChipView.h
samnm: Good idea. Done.

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,190 @@
...
+@property(nonatomic, readonly, nonnull) IBInspectable UIImageView *imageView;
+
+/*
+ A UIImageView that leads the title label. Appears above the imageView.
+
+ This image view is typically used to show some icon that denotes the chip as selected, such as a
+ tick. If imageView has no image then the chip will require resizing when selected or deselected to
+ account for the changing visibility of selectedImageView.
+
+ Only visible when the chip is selected and the image is non-nil.
```
jverkoey: > and the **image** is non-nil.

Do you mean `imageView` or `.image` on selectedImageView? If the latter then I think that's implied. If it's the former then `imageView` would be more clear.

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,190 @@
...
+
+/*
+ A UIImageView that leads the title label.
+
+ Only visible when the image is non-nil.
+ */
+@property(nonatomic, readonly, nonnull) IBInspectable UIImageView *imageView;
+
+/*
+ A UIImageView that leads the title label. Appears above the imageView.
```
jverkoey: > Appears **in front** of the imageView.

"above" implies y-space.

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,190 @@
...
+ account for the changing visibility of selectedImageView.
+
+ Only visible when the chip is selected and the image is non-nil.
+ */
+@property(nonatomic, readonly, nonnull) IBInspectable UIImageView *selectedImageView;
+
+/*
+ A UIView that trails the title label.
+
+ It will be given a size based on the value returned from sizeThatFits:.
```
jverkoey, line 49: Is the chip view's sizeThatFits: a function of the accessoryView's sizeThatFits? E.g. if accessoryView has a size of 100x100, will the chip's also be at least 100x100?

If not, how is the accessory view's sizeThatFits result taken into account? This may be beneficial to document here as an expansion of the current statement.

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,190 @@
...
+ UIContentSizeCategory is changed.
+
+ This property is modeled after the adjustsFontForContentSizeCategory property in the
+ UIContentSizeCategoryAdjusting protocol added by Apple in iOS 10.0.
+
+ If set to YES, this button will base its text font on MDCFontTextStyleButton.
+
+ Default value is NO.
+ */
+@property(nonatomic, readwrite, setter=mdc_setAdjustsFontForContentSizeCategory:)
```
jverkoey, line 98: Why the mdc_ prefix?

jverkoey (changes requested):
Primarily documentation modification requests - looking good!

components/Chips/src/MDCChipView.h
samnm: .image on selectedImageView. The salient point is that the selectedImageView is hidden when it doesn't have an image but I guess that doesn't need to be communicated?

components/Chips/src/MDCChipView.h
samnm: The chips sizeThatFits is a function of all of its subviews sizeThatFits.

components/Chips/src/MDCChipView.h
samnm: Thats the pattern we've used for this same property across all of MDC. @ianegordon may have some input here.

components/Chips/src/MDCChipView.h
jverkoey: Yeah I think it's expected that if there's no image then nothing will be shown.

components/Chips/src/MDCChipView.h
samnm: Done.

components/Chips/src/MDCChipView.h
samnm: Sounds good. Removed (and reordered a little).

jverkoey (approved):

samnm:
I'm going to wrap this up Monday evening so if you have any final comments please add them :)

PR #2229, opened 2017-10-18
DO NOT SUBMIT. Example for #2230

sigito:
This is an example for #2230.


willlarche:
If you don't need Travis tests because this is a diff test or something, let us know and we can cancel them to conserve resources.

romoore:
Canceled - this is just example code to demonstrate the visual alignment.

sigito:
@willlarche , sorry, would be more careful next time.

willlarche:
It’s all good! It’s something new for us to limit Travis builds.

randallli (changes requested):
DO NOT SUBMIT

I am marking it "request changes" to remove it from my "needs review" queue.

components/NavigationBar/examples/NavigationBarWithFlatButtonItemExample.m
```diff
@@ -0,0 +1,91 @@
+/*
+ Copyright 2016-present the Material Components for iOS authors. All Rights Reserved.
```
ianegordon, line 2: C&P Error? New files should have the current year.
Here and below.

components/private/Icons/icons/ic_add/src/MaterialIcons+ic_add.h
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+#import <UIKit/UIKit.h>
+
+#import "MaterialIcons.h"
+
+// This file was automatically generated by running ./scripts/sync_icons.sh
+// Do not modify directly.
+
+@interface MDCIcons (ic_add)
```
ianegordon, line 24: This add a symbol to MDC that is not actually used by our components.  I would expect image this to be a part of the example, not pulled from our private icons component.
One of the goals of the examples is to allow people to easily utilize the code in their own projects.  Relying on a private API makes this a bit trickier.

ianegordon (changes requested):

willlarche:
Is this still in play?

sigito:
@willlarche, this is just an example, so surely we can remove it.

Also, I might have misunderstood @romoore in the issues #2230. Was it a green light to actually align the icon with the button text? I remember having a private chat with @romoore, where he was saying that maybe a designer's opinion is necessary since Material spec does not specify if a material button can have text and icon at the same time.

components/NavigationBar/examples/NavigationBarWithFlatButtonItemExample.m
sigito: It was just code copied from another example.

1. Since I do not plan to submit this code, I guess it's not a big issue.
1. Existing code might have old year in copyright section. Is this an issue?

components/private/Icons/icons/ic_add/src/MaterialIcons+ic_add.h
sigito: Fair enough, but since I just wanted to demonstrate the issue, I didn't think of the cleanest solution.

willlarche:
Ok! I'm closing this then. If you need, you can also always run a compare on GitHub. Go to the branch in code view and the compare button will appear on the top right.

screen shot 2017-10-30 at 10 26 52 am

willlarche:
And I'd reach out to @romoore directly again (on discord?) and get that straightened out. Let me know if you need anything else.

And thank you for being a fabulous contributor!

sigito:
All I need is to know if it is expected to be supported or no. I would be happy to help if any changes would be necessary though.

Thanks, I appreciate it.

romoore:
I've reached out internally to our design team. I will update #2230 when I get a go/no-go decision.  Sorry for the delay/confusion! 🙁

PR #2233, opened 2017-10-19
[AppBarExample] Load bundle image with UIImage constructor

### LINKED ISSUE #2217: [Catalog] Crash when selecting AppBar -> Imagery (Swift) example
### Overview
Thread 1: Fatal error: Unexpectedly found nil while unwrapping an Optional value

### Reproduction steps
 - Launch catalog
 - Select AppBar -> Imagery (swift) example

### Version number
 - 36.3.0

sigito:
Fixes #2217.


ianegordon (approved):
Tested locally on an iPhone X Sim @ 11.0 and an iPhone 4s Sim @ 8.1.
Thanks @sigito !

PR #2234, opened 2017-10-19
[BottomAppBar] Account for iPhone X insets

jgunaratne:
Ensure bottom app bar renders properly on iPhone X and use insets.

Addresses issue: https://github.com/material-components/material-components-ios/issues/2226

simulator screen shot - iphone x - 2017-10-19 at 15 46 41

simulator screen shot - iphone x - 2017-10-19 at 15 46 40


ianegordon:
This PR will adjust the size for safe areas, but without a complementary call from its view controller when safe area insets change it will not get sized correctly.
Added #2259 to investigate the best way to handle this.

ianegordon (approved):

PR #2239, opened 2017-10-20
[BottomNavigation] Add color themer to bottom navigation bar

jgunaratne:
Add color themer to bottom navigation bar.

Pending merge of: https://github.com/material-components/material-components-ios/pull/2088

simulator screen shot - iphone 8 - 2017-10-20 at 17 14 51


ianegordon:
Blocked on #2088
Will revisit as soon as BottomNavigationBar lands.

ianegordon:
@romoore could you take a look at this when BottomNavigationBar lands?

components/BottomNavigation/src/ColorThemer/MDCBottomNavigationBarColorThemer.h
```diff
@@ -0,0 +1,35 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import "MaterialThemes.h"
+#import "MaterialBottomNavigation.h"
+
+/**
+ Used to apply a color scheme to theme MDCBottomNavigationBar.
```
romoore: I think it would be good to provide a description of what the themer will do (and what it won't do). For example, does it apply `tintColor`, `backgroundColor`, selected item color, etc.?  Are there any colors that aren't changed/updated that clients should expect to change manually?

components/BottomNavigation/src/ColorThemer/MDCBottomNavigationBarColorThemer.h
```diff
@@ -0,0 +1,35 @@
...
+@interface MDCBottomNavigationBarColorThemer : NSObject
+
+/**
+ Applies a color scheme to theme a MDCBottomNavigationBar. Use a UIAppearance proxy to apply a color
+ scheme to all instances of MDCBottomNavigationBar.
+
+ @param colorScheme The color scheme to apply to MDCBottomNavigationBar.
+ @param bottomNavigationBar A MDCBottomNavigationBar instance to apply a color scheme.
+ */
++ (void)applyColorScheme:(NSObject<MDCColorScheme> *)colorScheme
```
romoore, line 33: Does this need to be `NSObject` or can it just be `id`?

components/BottomNavigation/src/ColorThemer/MDCBottomNavigationBarColorThemer.m
```diff
@@ -0,0 +1,28 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import "MDCBottomNavigationBarColorThemer.h"
+
+@implementation MDCBottomNavigationBarColorThemer
+
++ (void)applyColorScheme:(NSObject<MDCColorScheme> *)colorScheme
+   toBottomNavigationBar:(MDCBottomNavigationBar *)bottomNavigationBar {
```
romoore: Nit: Same indentation concern.

components/BottomNavigation/src/ColorThemer/MDCBottomNavigationBarColorThemer.h
```diff
@@ -0,0 +1,35 @@
...
+   toBottomNavigationBar:(MDCBottomNavigationBar *)bottomNavigationBar;
```
romoore: Nit: I think this needs to be indented 1 more space. It's at 3, but I believe it should be 4.

romoore (changes requested):
If we have a Swift example, would love to see it themed as well. If not, we can add it later.

components/BottomNavigation/src/ColorThemer/MDCBottomNavigationBarColorThemer.m
jgunaratne: Done.

components/BottomNavigation/src/ColorThemer/MDCBottomNavigationBarColorThemer.h
jgunaratne: Done.

components/BottomNavigation/src/ColorThemer/MDCBottomNavigationBarColorThemer.h
jgunaratne: All of the other themers use NSObject<MDCColorScheme>. Would be good to have them all be consistent.

components/BottomNavigation/src/ColorThemer/MDCBottomNavigationBarColorThemer.h
jgunaratne: I've elaborated on how this affects the various color properties of the bottom navigation bar. The client shouldn't have to change any colors manually when using the themer.

components/BottomNavigation/src/ColorThemer/MDCBottomNavigationBarColorThemer.h
romoore: Ack. Sounds like a good reason. :)

romoore (approved):

components/BottomNavigation/src/ColorThemer/MDCBottomNavigationBarColorThemer.h
jgunaratne: Also added theming to the Swift example as suggested.

PR #2240, opened 2017-10-21
[BottomDrawer] Initial API review

willlarche:
Attempting to do "as little as possible" and use existing UIKit API.

Much to be borrowed from Bottom Sheet and Dialogs.


components/BottomDrawer/src/MDCBottomDrawerPresentationController.h
```diff
@@ -0,0 +1,38 @@
...
+ */
+@interface MDCBottomDrawerPresentationController: NSObject
+    <UIAdaptivePresentationControllerDelegate>
+
+/** Is the presented view full screen? */
+@property(nonatomic, assign, readonly, getter=isPresentedViewFullScreen) BOOL
+    presentedViewFullScreen;
+
+/** Is the presented view's preferredContentSize.height completely visible? */
+@property(nonatomic, assign, readonly, getter=isPresentedViewFullyExtended) BOOL
```
romoore, line 35: Does this return `YES` or `NO` if the height extends into the safe area?

components/BottomDrawer/src/MDCBottomDrawerTransitionController.h
```diff
@@ -0,0 +1,40 @@
...
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+@class MDCBottomAppBarView;
+
+// TODO: (larche) Cite spec article.
+/**
+ A controller for scrubbable scrolling (aka dragging) vertical pan gesture-based mnodal transitions
```
romoore: Nit/spelling: s/mnodal/modal

romoore (commented):
Do you have an example file to demonstrate how you think clients will call the API?

components/BottomDrawer/src/MDCBottomDrawerPresentationController.h
willlarche: Regardless of how tall the view is, or what the safe area insets are, if it is fully extended and has no more content below, this returns `YES`.

romoore: Ack.

components/BottomDrawer/src/MDCBottomDrawerPresentationController.h
```diff
@@ -0,0 +1,38 @@
...
+
+#import <UIKit/UIKit.h>
+
+// TODO: (larche) Cite spec article.
+/**
+ A controller for modally presenting view controllers from the bottom of the screen either partial
+ or full screen.
+
+ Also provides a dark, translucent scrim for easy dismissal and free rounding of corners on the
+ presented view that "heal" if the presented view is full screen.
```
heinberg, line 25: Do we want to make the rounding effect optional (i.e. add a BOOL for whether we want it to round corners? Or is this part of the official spec?

components/BottomDrawer/src/MDCBottomDrawerTransitionController.h
```diff
@@ -0,0 +1,40 @@
...
+@interface MDCBottomDrawerTransitionController: UIPercentDrivenInteractiveTransition
+    <UIViewControllerTransitioningDelegate>
+
+/**
+ Initializes the controller and adds a pan gesture recognizer to the bottom app bar.
+
+ @param bottomAppBar The bottom app bar that initiates and reacts to the presentation of the
+ drawer.
+ */
+- (nonnull instancetype)initWithBottomAppBar:(nonnull MDCBottomAppBarView *)bottomAppBar
```
heinberg: It looks like this will always require an MDC app bar. Is it possible that some apps will wan this to launch off a custom view? Is there anything strictly requiring this?

components/BottomDrawer/src/MDCBottomDrawerPresentationController.h
```diff
@@ -0,0 +1,38 @@
...
+ */
+@interface MDCBottomDrawerPresentationController: NSObject
```
heinberg: Sorry for my lack of knowledge - how will the user indicate what UIViewController should be hosted inside the controller?

components/BottomDrawer/src/MDCBottomDrawerPresentationController.h
ianegordon: How do we expect these will be used?
They may be useful from a development / debugging viewpoint, but I'm wondering if we would want clients changing behavior based on these values.

components/BottomDrawer/src/MDCBottomDrawerPresentationController.h
ianegordon: I would expect an XYZNameHerePresentationController to be of type UIPresentationController, not a delegate.

Naming is hard.

components/BottomDrawer/src/MDCBottomDrawerTransitionController.h
ianegordon: Who will own this view?  I would expect it to be a subview of the presenting controller (MDCBottomAppBarController?).  Will we require the presented view controller to be of a certain type?

ianegordon (changes requested):

components/BottomDrawer/src/MDCBottomDrawerTransitionController.h
willlarche: Thank you.

components/BottomDrawer/src/MDCBottomDrawerPresentationController.h
willlarche: It's part of the official spec right now but to be conservative, I'll build in a BOOL in private API that we could always expose in the future.

components/BottomDrawer/src/MDCBottomDrawerTransitionController.h
willlarche: `MDCBottomAppBarController` will indeed own the view and the drawer. So agreed with @ajsecord and @jgunaratne in a working session.

components/BottomDrawer/src/MDCBottomDrawerPresentationController.h
willlarche: It's been confirmed that there is behavior (motion) that depends on the content knowing if it's fully presented. I'm exploring if that can just be handled by querying the percent of the percent based transition.

components/BottomDrawer/src/MDCBottomDrawerPresentationController.h
willlarche: That's a typo. Thank you. It's of type `UIPresentationController` now.

willlarche: That will be done in the API for `MDCAppBarViewController` which actually puts all these pieces together. That's owned by @jgunaratne .

willlarche:
Rebase confused GitHub and CLA bot. Closing and reopening.

PR #2241, opened 2017-10-21
[Scripts] Added a script to check that umbrella header are used between components

### LINKED ISSUE #1592: Ensure that we are using umbrella headers between components.
Right now we can and do import various public headers of components from within other components. We should always import umbrella headers instead. This could be done via script.

sigito:
The script checks Objective-C files in _src_ and _examples_ directories. It reads all of the imports in the file and applies next logic:
1. skip Objective-C framework headers;
1. skip MDF and Motion headers;
1. check that the header is not an umbrella header of the component itself.
   Note: this check is only applied to the main code of the component. _examples_ and sub-modules code (i.e., `ColorThemer`) has to use umbrella headers for all component imports.
1. skip imports of the processed component files;
1. check that every imported header is an umbrella header.

Also, the script splits the code onto sub-modules. If a directory in _src_ starts with a capital letter, it is treated as a separate module.

Closes #1592.

### DIFF
--- a/contributing/checklist.md
+++ b/contributing/checklist.md
@@ -462,6 +462,12 @@ Material Components for iOS is built primarily for adoption with CocoaPods. Ther
 1. Enter YES or NO
 
 
+### Use umbrella header between components
+
+
+Material Components should always use umbrella headers to access API of other components.
+
+
 ## Running the checklist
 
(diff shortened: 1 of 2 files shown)


ianegordon (approved):
Very helpful, thank you @sigito !

PR #2242, opened 2017-10-21
[Catalog] Pad trailing edge inside.

sigito:
Before:
iphone8 11_before

After:
iphone8 11_after

Tested: 
* iPhone X (iOS 11.0) Sim;
* iPhone 8 (iOS 11.0) Sim;
* iPhone 7 (iOS 10.3.1) Sim;
* iPhone 5 (iOS 8.4) Sim.


romoore (approved):
LGTM!

Issue #1475, opened 2017-06-02
Migrate to MDFInternationalization


ianegordon:
Migrate from our private/MDCInternationalization methods to MDFInternationalization.

The API is nearly identical.

We can use MDFTextAccessibility as an example of how to import an external dependency.

sigito:
Hi, @ianegordon. Indeed the APIs are almost the same. However there are some minor internal implementation differences in `UIImage+MaterialRTL` between:
1) `MDCRTLFlippedImage` vs `MDFRTLFlippedImage`;
2) `mdc_imageFlippedForRightToLeftLayoutDirection` vs `mdf_imageWithHorizontallyFlippedOrientation`.

Do you think this needs extra attention?

Also `MDCRectFlippedForRTL` function receives a layout direction, while `MDFRectFlippedHorizontally` does not. An additional check in the code would be necessary before calling.

I can replace the usage of `MaterialRTL.h`. Do you think it should be done per component or is it fine to do it all together? Also are there tests for RTL in place already or should I rely on visual testing (check manually)?

Total mapping:

| MaterialRTL  | MDFInternationalization |
| --- | --- | 
| **MDCRTL.h** | **MDFRTL.h** |
| `MDCAutoresizingFlexibleLeadingMargin` | `MDFLeadingMarginAutoresizingMaskForLayoutDirection` |
| `MDCAutoresizingFlexibleTrailingMargin` | `MDFTrailingMarginAutoresizingMaskForLayoutDirection` |
| `MDCRectFlippedForRTL` | `mdc_effectiveUserInterfaceLayoutDirection` + `MDFRectFlippedHorizontally` |
| `MDCInsetsMakeWithLayoutDirection` | `MDFInsetsMakeWithLayoutDirection` |
| | `MDFInsetsFlippedHorizontally` |
| **UIImage+MaterialRTL.h** | **UIImage+MaterialRTL.h** |
| `mdc_imageFlippedForRightToLeftLayoutDirection` | `mdf_imageWithHorizontallyFlippedOrientation` |
| **UIView+MaterialRTL.h** | **UIView+MaterialRTL.h** |
| `mdc_semanticContentAttribute` | `mdf_semanticContentAttribute` |
| `mdc_effectiveUserInterfaceLayoutDirection` | `mdf_effectiveUserInterfaceLayoutDirection` |
[...]

PR #2246, opened 2017-10-21
[ActivityIndicator] Migrated to MDFi18n.

sigito:
Partially addresses #1475.

### DIFF
--- a/MaterialComponents.podspec
+++ b/MaterialComponents.podspec
@@ -44,9 +44,9 @@ Pod::Spec.new do |s|
       sss.source_files = "components/#{ss.base_name}/src/*.{h,m}", "components/#{ss.base_name}/src/private/*.{h,m}"
       sss.resources = ["components/#{ss.base_name}/src/Material#{ss.base_name}.bundle"]
 
+      ss.dependency "MDFInternationalization"
       ss.dependency "MaterialComponents/Palettes"
       ss.dependency "MaterialComponents/private/Application"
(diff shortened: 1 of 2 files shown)


ianegordon (dismissed):
Excellent!

Please add MDFI18n as a dependency and remove private/RTL in the Podspec.  (Here and in the other migration PRs)

sigito:
@ianegordon, I have updated a Podspec file in this PR only. Could you take a look if it's alright, so I would do the same in other and future RPs? 

Also, how could I catch a problem of not making these changes in a Podspec file? `pod update` worked OK and it built just fine in Xcode. Is there a different process to test pods configurations?

ianegordon:
Hi @sigito  My understanding is that since one of our components pulls in MDFI18n (Dialogs), any of our components can use it since they are all pulled into the same project.  This would break if a client tried to pull in only ActivityIndicator, without Dialogs (and transitively MDFI18n).  Unfortunately I don't know of an easy test.  Perhaps in the future it would make sense to have a target for each component.

ianegordon (approved):

ianegordon:
HI @sigito .  Thanks for putting this together.  Sorry for the delay, but I wanted to land the latest release of MDFI18n internally before merging this.

Your change to the podspec is exactly what we need for the other PRs.  Remove the ss.dependency on private/RTL and add a dependency on MDFInternationalization.  Can you make this change to the other PRs and I'll merge them as well?

PR #2249, opened 2017-10-21
[Catalog] RTL for settings dialog.

sigito:
Before:
simulator screen shot - iphone 8 - 2017-10-22 at 00 06 30
After:
after


romoore (approved):
LGTM. Thanks for the fix!

PR #2252, opened 2017-10-21
[Buttons] Deprecate defaultContentEdgeInsets

### LINKED ISSUE #2250: [Buttons] Deprecate -defaultContentEdgeInsets
### Overview
MDCButton should use the `contentEdgeInsets` property instead of depending on subclasses to implement the method, `defaultContentEdgeInsets`.  All MDCButton subclasses should stop using the deprecated method and instead set their contentEdgeInsets either directly or via UIAppearance proxy.

### Version number
37.0.0

Any/all

#1896, #1948, #1943 

---
#### Internal data
- Associated internal bug: [b/117179461](http://b/117179461)

romoore:
Subclases should no longer use defaultContentEdgeInsets and should
instead set the property directly. This change is need to clients can
customize MDCButton more to suit their needs.

Closes #2250

### DIFF
--- a/components/BottomAppBar/src/MDCBottomAppBarView.m
+++ b/components/BottomAppBar/src/MDCBottomAppBarView.m
@@ -243,6 +243,7 @@ static const int kMDCButtonAnimationDuration = 200;
 
 - (void)layoutSubviews {
   [super layoutSubviews];
+  [self.floatingButton sizeToFit];
       self.floatingButton.center =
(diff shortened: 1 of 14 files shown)


samnm (changes requested):
This PR is the same change as #1896 which had to be rolled back in #1948 due to internal conflicts. We have to resolve those conflicts before landing this change.

romoore:
Closing this but leaving the Issue open until the internal clients can accommodate this change.

PR #2254, opened 2017-10-22
[Buttons] Add properties minimumSize, maximumSize

### LINKED ISSUE #2253: [Buttons] Create minimumSize and maximumSize properties
### Overview
Material guidelines specify that some buttons have minimum dimensions while others (like floating buttons) have maximum dimensions.  Add two new properties to support these requirements.

If these properties are set, they should be used to bound `sizeThatFits`.

https://material.io/guidelines/components/buttons.html#buttons-style
https://material.io/guidelines/components/buttons-floating-action-button.html

### Version number
37.0.0

Any/All

romoore:
Material Guidelines specify that buttons should have minimum sizes, and
in some cases maximum sizes (e.g., floating buttons).

Closes #2253


romoore:
Going to determine whether this change is still required as-is. Marking "planning changes" for now.

ianegordon (approved):
LGTM.
Should this have any impact on our custom handling of pointInside:withEvent: ?

romoore:
@ianegordon In general, I think the "tappable area" and the "alignment rect" should be semi-independent. If I set a `maximumSize` value smaller than the minimum "tappable area", the pointInside logic should continue to work as it does.  However, if I also set a custom `hitAreaInsets`, then the button should respect that instead of the Material guidelines recommendations.

romoore:
Looks like this change is worth making. Removing Planning Changes and Needing Research labels.

randallli (approved):

samnm (approved):
Looks good!

components/Buttons/src/MDCButton.m
```diff
@@ -274,6 +286,8 @@ - (void)commonMDCButtonInit {
 
   // Default content insets
   self.contentEdgeInsets = [self defaultContentEdgeInsets];
+  _minimumSize = CGSizeZero;
+  _maximumSize = CGSizeZero;
```
randallli, line 290: wont this make all the buttons invisible?

randallli (changes requested):

components/Buttons/src/MDCButton.m
romoore: No because the sizes are only enforced if they are greater than zero.  See [`sizeThatFits:`](https://github.com/material-components/material-components-ios/pull/2254/files/ba86df99749819e91ecac5840712d1a05aefe7c2#diff-f89cec76e933099b540c38f711df8c53R403) below.

components/Buttons/src/MDCButton.h
```diff
@@ -73,6 +73,24 @@
...
+@property(nonatomic, assign) CGSize minimumSize UI_APPEARANCE_SELECTOR;
+
+/**
+ The maximum size of the button’s alignment rect. If either the height or width are non-positive
+ (negative or zero), they will be ignored and that axis will adjust to its content size. Setting a
+ maximum size may result in image clipping or text truncation.
+
+ Defaults to CGSizeZero.
+ */
+@property(nonatomic, assign) CGSize maximumSize UI_APPEARANCE_SELECTOR;
```
randallli, line 90: what happens if we set the max to smaller than the min? or vice versa

romoore: Then undefined behavior will result.  I'm not sure that we should add logic to handle that situation.

samnm (approved):
No idea why it doubled up on reviews... Odd github bug?

romoore:
For some reason it also does not list @samnm as a Reviewer.

components/Buttons/src/MDCButton.h
randallli: looks like flexible header does clamping
https://github.com/material-components/material-components-ios/blob/develop/components/FlexibleHeader/src/MDCFlexibleHeaderView.m#L1331

randallli (changes requested):

components/Buttons/src/MDCButton.h
romoore: We discussed this offline. I believe we agreed to keep the implementation as-is for now and add clamping if needed in the future.

randallli (approved):

PR #2256, opened 2017-10-22
[Buttons] Deprecate -cornerRadius method

### LINKED ISSUE #2255: [Buttons] Deprecate `-cornerRadius` method
### Overview
MDCButton should deprecate its `-cornerRadius` method used for subclassing. Instead, callers should directly write to the button's layer's `cornerRadius` property.  Methods internal to MDCButton that previously depended on `-cornerRadius` should change to the layer's `cornerRadius` property.

### Version number
37.0.0

Any/All

### Related issues
Related to #1896, #1948

romoore:
Buttons should no longer use the subclassing method `-cornerRadius` and
should instead depend on the backing layer's `cornerRadius` property.

Closes #2255

Originally part of #1896

### DIFF
--- a/components/Buttons/src/MDCButton.m
+++ b/components/Buttons/src/MDCButton.m
@@ -354,12 +354,12 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
 - (void)layoutSubviews {
   [super layoutSubviews];
   self.layer.shadowPath = [self boundingPath].CGPath;
+  if ([self respondsToSelector:@selector(cornerRadius)]) {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
-  if ([self didOverrideMethod:@selector(cornerRadius)]) {
     self.layer.cornerRadius = [self cornerRadius];
-  }
 #pragma clang diagnostic pop
+  }
 
   // Center unbounded ink view frame taking into account possible insets using contentRectForBounds.
   if (_inkView.inkStyle == MDCInkStyleUnbounded) {
@@ -791,12 +791,14 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
 
 - (UIBezierPath *)boundingPath {
   CGFloat cornerRadius = self.layer.cornerRadius;
+
+  if ([self respondsToSelector:@selector(cornerRadius)]) {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
-  if ([self didOverrideMethod:@selector(cornerRadius)]) {
     cornerRadius = [self cornerRadius];
-  }
 #pragma clang diagnostic pop
+  }
+
 
   return [UIBezierPath bezierPathWithRoundedRect:self.bounds cornerRadius:cornerRadius];
 }
@@ -808,25 +810,6 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
 - (BOOL)shouldHaveOpaqueBackground {
   BOOL isFlatButton = MDCCGFloatIsExactlyZero([self elevationForState:UIControlStateNormal]);
   return !isFlatButton;
-}
-
-- (BOOL)didOverrideMethod:(SEL)selector {
-  Class superclass = [self superclass];
-  IMP selfIMP = [self methodForSelector:selector];
-  BOOL isOverriden = NO;
-
-  while (superclass) {
-    isOverriden = [superclass instancesRespondToSelector:selector]
-        && selfIMP != [superclass instanceMethodForSelector:selector];
-
-    if (isOverriden) {
-      break;
-    }
-
-    superclass = [superclass superclass];
-  }
-
-  return isOverriden;
 }
 
 - (void)updateAlphaAndBackgroundColorAnimated:(BOOL)animated {
--- a/components/Buttons/src/MDCButton.m
+++ b/components/Buttons/src/MDCButton.m
@@ -62,6 +62,7 @@ static NSString *const MDCButtonShadowColorsKey = @"MDCButtonShadowColorsKey";
 // https://material.io/guidelines/layout/metrics-keylines.html#metrics-keylines-touch-target-size
 static const CGFloat MDCButtonMinimumTouchTargetHeight = 48;
 static const CGFloat MDCButtonMinimumTouchTargetWidth = 48;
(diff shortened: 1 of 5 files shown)


components/Buttons/src/MDCButton.m
```diff
@@ -768,6 +779,25 @@ - (BOOL)shouldHaveOpaqueBackground {
   return !isFlatButton;
 }
 
+- (BOOL)checkIfSubclassDidOverrideMethod:(SEL)selector {
```
jverkoey: `didOverrideMethod`

jverkoey (dismissed):
I'm a bit nervous about the fact that `self.layer.shadowPath` implicitly depends on `self.layer.cornerRadius`.

Prior to this change it wasn't easy to change the cornerRadius without subclassing, meaning that we weren't necessarily supporting a setter API for cornerRadius. Now that we are encouraging using a settable API, I can imagine someone not being aware that setting a cornerRadius also requires calling `setNeedsLayout`.

What are your thoughts on how to encourage clients to best use this new approach?

romoore:
@jverkoey Thanks for the feedback. I don't believe we really need to do anything with clients regarding this approach. Not having an API for a button's corner radius is the actual goal (as I understand it). It should be pretty straightforward to change the layer's `cornerRadius` now that it's not being overridden in `layoutSubviews`. 

None of the example code has had to be updated as a result of this change. What is the specific concern that you have regarding `cornerRadius` and `shadowPath`?

components/Buttons/src/MDCButton.m

romoore:
Is this what you're worried about - that changing the cornerRadius while on-screen will result in weird effects?

button layer cornerradius async

It might make sense for MDCShadowLayer to call `setNeedsLayout` on itself when its `cornerRadius` property changes.  Here's the result (the shadow is one rendering pass behind the cornerRadius).

```objectivec
- (void)setCornerRadius:(CGFloat)cornerRadius {
  CGFloat currentRadius = super.cornerRadius;
  if (!MDCCGFloatEqual(cornerRadius, currentRadius)) {
    [self setNeedsLayout];
  }
}

```

button layer cornerradius setneedslayout

## Updated
As an FYI, here's the same effect by modifying the return value of MDCButton's `-cornerRadius`.

```objectivec
- (void)changeButton {
  dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.5 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
    self.mdc_cornerRadius += 1;
    if (self.mdc_cornerRadius > 15) {
      self.mdc_cornerRadius = 0;
    }
    [self setNeedsLayout];
    [self changeButton];
  });
}

- (CGFloat)cornerRadius {
  return self.mdc_cornerRadius;
}
```

mdcbutton cornerradius async

jverkoey:
Yep! That was the situation I was nervous about. Are you happy with either of those two approaches?

Would a third approach be possible where we update `layer.shadowPath` when the cornerRadius changes?

If none of the above feel like adequate solutions, at a minimum let's document the fact that `layer.shadowPath` is dependent on the value of `layer.cornerRadius` (if we aren't already).

romoore:
We should document the effect. I really feel like it's an edge case to change the `cornerRadius` of a button's layer while it's visible on-screen.  The only way I was able to achieve the change to both the view's layer and the shadowPath in the same layout pass was to make the change within MDCButton (to retrieve the boundingPath) and disable CoreAnimation's implicit animations.

For the gifs above, I hacked MDCButton to constantly change its layer's `cornerRadius` property in a dispatch loop. I may have missed something, but the only way I could find to make this change atomically would be within MDCButton. We could expose a new MDCButton property, `cornerRadius`, which performs the animation-disabled updates. However, doing so would be contrary to our long-term trend toward simpler objects.

I would vote on the side of documentation and call-out the visual artifact at this point. If in the future this becomes too much of a problem, we can revisit adding the functionality to MDCButton.

romoore:
Also, here's the best rendering I got by making the CA changes in MDCButton directly.

mdcbutton cornerradius disableanimations

samnm (dismissed):
This PR is the same change as #1896 which had to be rolled back in #1948 due to internal conflicts. We have to resolve those conflicts before landing this change.

romoore:
I'm going to close this PR but leave the related issue open until we can make the required internal changes.

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -105,15 +105,13 @@ - (CGSize)intrinsicContentSize {
   }
 }
 
-#pragma mark - Subclassing
-
-- (UIBezierPath *)boundingPath {
-  return [UIBezierPath bezierPathWithOvalInRect:self.bounds];
+- (void)layoutSubviews {
+  // We have to set cornerRadius before laying out subviews so that the boundingPath is correct.
+  self.layer.cornerRadius = CGRectGetHeight(self.bounds) / 2;
```
randallli, line 110: I thought the whole point was to let clients set the corner radius and that we would not erase their value. What about subclasses of MDCFloatingButton?

romoore: I spoke with @samnm on Friday. We determined that `cornerRadius` should not be problematic for internal clients as originally authored by @samnm and reproduced in this PR. The previous approach (using a new property) should not be necessary as an intermediate step. 

Simply removing the behavior (but allowing subclasses to override for now) is a cleaner, direct step in the correct direction.

jverkoey (approved):

samnm (approved):

components/Buttons/src/MDCFloatingButton.m
randallli: sgtm

randallli (approved):

willlarche:
Is this ready to merge?

components/Buttons/src/MDCFloatingButton.m
yarneo: @romoore sorry to revive an old PR, but any reason why here it is `CGRectGetHeight` whereas it is replacing `CGRectGetWidth`

romoore: The width in .expanded mode can be very large, while the height is the dimension that generates the "pill shape".  The same height can be used when the frame is a square to generate a circlular FAB. 

Is this code causing a problem?

yarneo: not as of yet, i am just working on shapes now and wanted to understand the thought behind the calculation and if we mistakenly updated it differently. Thanks for the info!

PR #2261, opened 2017-10-23
[Tabs] Internal fixes to account for the Safe Area insets.

### LINKED ISSUE #2012: [Tabs] Support Safe Area

chuga:
~~Consists of two things: sizeThatFits takes into account any bottom safe area inset the tab bar might have, and~~ a few fixes to the collection view itself to make sure our content is laid out correctly.

Closes #2012

### DIFF
--- a/components/Tabs/src/MDCTabBar.m
+++ b/components/Tabs/src/MDCTabBar.m
@@ -272,9 +272,16 @@ static MDCItemBarAlignment MDCItemBarAlignmentForTabBarAlignment(MDCTabBarAlignm
 }
 
 - (CGSize)intrinsicContentSize {
-  return _itemBar.intrinsicContentSize;
+  CGSize size = _itemBar.intrinsicContentSize;
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    if (![[self class] isTopTabsForPosition:_barPosition]) {
+      size.height += self.safeAreaInsets.bottom;
+    }
+  }
+#endif
+  return size;
 }
-
 
 - (CGSize)sizeThatFits:(CGSize)size {
   size = [_itemBar sizeThatFits:size];
@@ -286,6 +293,16 @@ static MDCItemBarAlignment MDCItemBarAlignmentForTabBarAlignment(MDCTabBarAlignm
   }
 #endif
   return size;
+}
+
+- (void)safeAreaInsetsDidChange {
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    [super safeAreaInsetsDidChange];
+  }
+#endif
+  [self invalidateIntrinsicContentSize];
+  [self setNeedsLayout];
 }
 
 - (void)didMoveToWindow {
--- a/components/Tabs/src/MDCTabBar.m
+++ b/components/Tabs/src/MDCTabBar.m
@@ -280,7 +280,9 @@ static MDCItemBarAlignment MDCItemBarAlignmentForTabBarAlignment(MDCTabBarAlignm
   size = [_itemBar sizeThatFits:size];
 #if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
   if (@available(iOS 11.0, *)) {
-    size.height += self.safeAreaInsets.bottom;
+    if (![[self class] isTopTabsForPosition:_barPosition]) {
+      size.height += self.safeAreaInsets.bottom;
+    }
   }
 #endif
   return size;
--- a/components/Tabs/src/MDCTabBar.m
+++ b/components/Tabs/src/MDCTabBar.m
@@ -130,6 +130,7 @@ static MDCItemBarAlignment MDCItemBarAlignmentForTabBarAlignment(MDCTabBarAlignm
 
   // Create item bar.
   _itemBar = [[MDCItemBar alloc] initWithFrame:self.bounds];
+  _itemBar.autoresizingMask = (UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight);
   _itemBar.delegate = self;
   _itemBar.alignment = MDCItemBarAlignmentForTabBarAlignment(_alignment);
   [self addSubview:_itemBar];
--- a/components/Tabs/src/MDCTabBar.m
+++ b/components/Tabs/src/MDCTabBar.m
@@ -301,7 +301,9 @@ static MDCItemBarAlignment MDCItemBarAlignmentForTabBarAlignment(MDCTabBarAlignm
     [super safeAreaInsetsDidChange];
   }
 #endif
+  // The intrinsic content size depends on the Safe Area insets so we invalidate it when it changes.
   [self invalidateIntrinsicContentSize];
+
   [self setNeedsLayout];
   [self setNeedsUpdateConstraints];
 }
--- a/components/Tabs/src/MDCTabBar.m
+++ b/components/Tabs/src/MDCTabBar.m
@@ -272,40 +272,11 @@ static MDCItemBarAlignment MDCItemBarAlignmentForTabBarAlignment(MDCTabBarAlignm
 }
 
 - (CGSize)intrinsicContentSize {
-  CGSize size = _itemBar.intrinsicContentSize;
-#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
-  if (@available(iOS 11.0, *)) {
-    if (![[self class] isTopTabsForPosition:_barPosition]) {
-      size.height += self.safeAreaInsets.bottom;
-    }
-  }
-#endif
-  return size;
+  return _itemBar.intrinsicContentSize;
 }
 
 - (CGSize)sizeThatFits:(CGSize)size {
-  size = [_itemBar sizeThatFits:size];
-#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
-  if (@available(iOS 11.0, *)) {
-    if (![[self class] isTopTabsForPosition:_barPosition]) {
-      size.height += self.safeAreaInsets.bottom;
-    }
-  }
-#endif
-  return size;
-}
-
-- (void)safeAreaInsetsDidChange {
-#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
-  if (@available(iOS 11.0, *)) {
-    [super safeAreaInsetsDidChange];
-  }
-#endif
-  // The intrinsic content size depends on the Safe Area insets so we invalidate it when it changes.
-  [self invalidateIntrinsicContentSize];
-
-  [self setNeedsLayout];
-  [self setNeedsUpdateConstraints];
+  return [_itemBar sizeThatFits:size];
 }
 
 - (void)didMoveToWindow {
--- a/components/Tabs/src/MDCTabBar.m
+++ b/components/Tabs/src/MDCTabBar.m
@@ -117,6 +117,7 @@ static MDCItemBarAlignment MDCItemBarAlignmentForTabBarAlignment(MDCTabBarAlignm
 }
 
 - (void)commonMDCTabBarInit {
+  self.clipsToBounds = YES;
   _barPosition = UIBarPositionAny;
   _hasDefaultItemAppearance = YES;
   _hasDefaultAlignment = YES;
@@ -129,12 +130,18 @@ static MDCItemBarAlignment MDCItemBarAlignmentForTabBarAlignment(MDCTabBarAlignm
 
   // Create item bar.
   _itemBar = [[MDCItemBar alloc] initWithFrame:self.bounds];
-  _itemBar.autoresizingMask = (UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight);
   _itemBar.delegate = self;
   _itemBar.alignment = MDCItemBarAlignmentForTabBarAlignment(_alignment);
   [self addSubview:_itemBar];
 
   [self updateItemBarStyle];
+}
+
+- (void)layoutSubviews {
+  [super layoutSubviews];
+
+  CGSize sizeThatFits = [_itemBar sizeThatFits:self.bounds.size];
+  _itemBar.frame = CGRectMake(0, 0, sizeThatFits.width, sizeThatFits.height);
 }
 
 #pragma mark - Public
@@ -177,7 +184,7 @@ static MDCItemBarAlignment MDCItemBarAlignmentForTabBarAlignment(MDCTabBarAlignm
     _barTintColor = barTintColor;
 
     // Update background color.
-    _itemBar.backgroundColor = barTintColor;
+    self.backgroundColor = barTintColor;
   }
 }
 
@@ -268,8 +275,15 @@ static MDCItemBarAlignment MDCItemBarAlignmentForTabBarAlignment(MDCTabBarAlignm
   return _itemBar.intrinsicContentSize;
 }
 
+
 - (CGSize)sizeThatFits:(CGSize)size {
-  return [_itemBar sizeThatFits:size];
+  size = [_itemBar sizeThatFits:size];
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    size.height += self.safeAreaInsets.bottom;
+  }
+#endif
+  return size;
 }
 
 - (void)didMoveToWindow {
--- a/components/Tabs/src/MDCTabBar.m
+++ b/components/Tabs/src/MDCTabBar.m
@@ -303,6 +303,7 @@ static MDCItemBarAlignment MDCItemBarAlignmentForTabBarAlignment(MDCTabBarAlignm
 #endif
   [self invalidateIntrinsicContentSize];
   [self setNeedsLayout];
+  [self setNeedsUpdateConstraints];
 }
 
 - (void)didMoveToWindow {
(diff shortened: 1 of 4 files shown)


chuga:
The difference between this PR and #2236 is that here I dropped the changes to instrinsicContentSize because there's no way to make that work with the Interface Builder example we have in the catalog. So instead of doing that, in the BottomNavigationBarExample I explicitly set a height constraint that uses sizeThatFits

components/Tabs/src/MDCTabBar.m
```diff
@@ -268,8 +275,15 @@ - (CGSize)intrinsicContentSize {
   return _itemBar.intrinsicContentSize;
 }
 
+
 - (CGSize)sizeThatFits:(CGSize)size {
-  return [_itemBar sizeThatFits:size];
+  size = [_itemBar sizeThatFits:size];
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    size.height += self.safeAreaInsets.bottom;
```
brianjmoore: This should probably be conditional on the bar position? I think it wouldn't make sense to add the bottom insets for top tabs (and vice-versa).

components/Tabs/src/MDCTabBarViewController.m
```diff
@@ -241,7 +246,7 @@ - (nullable UIViewController *)controllerWithView:(nullable UIView *)view {
 
 - (void)updateLayout {
   CGRect bounds = self.view.bounds;
-  CGFloat tabBarHeight = [[_tabBar class] defaultHeightForItemAppearance:_tabBar.itemAppearance];
+  CGFloat tabBarHeight = [_tabBar sizeThatFits:bounds.size].height;
```
brianjmoore: We need to set the tab bar's frame to self.bounds before calling sizeThatFits: to ensure its safe area insets are consistent with the view controller's.

components/Tabs/src/MDCTabBar.m
chuga: self.safeAreaInsets is conditional on the bar position

romoore: It's possible for a parent VC to set safeAreaInsets that overlap the tabs even if they aren't in the bottom position. In that case, does it still make sense to increase the height?

components/Tabs/src/MDCTabBarViewController.m
chuga: No, the idea is that depending on where the _tabBar is positioned, this will returned the right size. The first time this is called the height will not be adjusted to account for the safe area inset as the _tabBar currently doesn't have a frame. But every subsequent call will. And since layout happens at least twice, this works as expected.

components/Tabs/src/MDCTabBar.m
chuga: I don't know, you mean by setting additionalSafeAreaInsets? What would be the use case? Since this is a tab bar, and it has to account for the "normal" safe area inset at the bottom for the home indicator (if it's in that position), I'd say yes, it has to expand to account for any safe area inset. If the parent VC wants to add even more, I don't see how that'd change this behavior.

romoore: If the tabbar were scrolling up along with some content, wouldn't the safeAreaInsets be set?  Sorry for any confusion, I was trying to back-up @brianjmoore and his suggestion to check the bar position.

chuga: I'm not following the use case you're describing. So to answer the original suggestion again: "This should probably be conditional on the bar position?" -> yes, it should and it is, as safeAreaInsets is already conditional on the position.

chuga: Discussed offline, I'll add a check using isTopTabsForPosition just as extra safety measure I guess

chuga:
@brianjmoore The MDCItemBarFlowLayout uses self.collectionView.bounds in multiple places, is that something that should be updated to use the adjusted bounds that accounts for the safe area insets (the new|adjustedCollectionViewWidth|)? My guess is that it should.

chuga:
I've confirmed that MDCItemBarFlowLayout needs to be fixed as well. Will be looking into that.

ianegordon:
Notes that @romoore may have some insights from his work on ##1982.
I would ping him for a +1 when you are ready.

chuga:
He's already included in this PR

romoore:
@ianegordon  This PR is for TabBar, the closed one was for the VC, which included things like adding the tabbar's frame as part of the safeAreaInsets on the child VCs.

ianegordon:
@romoore are TabBar and TabBarController separate enough concepts that they can be handled in different PRs?  I would expect the changes would need to land simultaneously to avoid conflicting with each other.

If this is not the case, please do re-open the previous PR and we'll revisit.

chuga:
This is for both.

romoore:
If this is for both, then I would recommend that the TabBarViewController update its child view controllers to include additionalSafeAreaInsets that define the position of the bottom TabBar.

I've updated the subject of this PR to reflect that it covers the entire Tabs component, not just TabBar.

Here's Apple's example of how to do it. https://developer.apple.com/documentation/uikit/uiview/positioning_content_relative_to_the_safe_area

DO NOT use your view's `safeAreaInsets` as the starting value, start with UIEdgeInsetsZero and just add the TabBar's height to the `.bottom` value. You can then pass that to each child VC's `additionalSafeAreaInsets`.    Here's how the example would look if they did it correctly:

```swift
override func viewDidAppear(_ animated: Bool) {
   var newSafeArea = UIEdgeInsets.zero // NOT! view.safeAreaInsets

   // Adjust the safe area to accommodate 
   //  the width of the side view.
   if let sideViewWidth = sideView?.bounds.size.width {
      newSafeArea.right += sideViewWidth
   }

   // Adjust the safe area to accommodate 
   //  the height of the bottom view.
   if let bottomViewHeight = bottomView?.bounds.size.height {
      newSafeArea.bottom += bottomViewHeight
   }

   // Adjust the safe area insets of the 
   //  embedded child view controller.
   let child = self.childViewControllers[0]
   child.additionalSafeAreaInsets = newSafeArea
}
```

chuga:
Great, thanks Rob. If you happen to have some spare time, this PR accepts contributions! I'm currently swamped with Flexible Header stuff :(

jmdetloff:
Compiled thoughts from our conversation a few moments ago:

You've removed the call to updateConstraints within safeAreaInsetsDidChange because it triggered an infinite layout loop in some cases. You also removed the addition of the safeAreaInsets to the intrinsicContentSize because without the ability to trigger a second layout pass the MDCTabBar will not lay out with the added height since the safeAreaInsets is not available on the first pass.

With that in mind, I think we should remove safeAreaInsets from sizeThatFits too. By removing it from intrinsicContentSize we're already mandating that clients not using manual layout account for the safe area insets in the parent view or view controller, they might as well handle this case when doing manual layout as well. 

Additionally, even using manual layout sizeThatFits may fail to contain safeAreaInsets in some cases. The example in the catalog app works because of 2 special circumstances:

A. The tab bar is initialized to fills its parent's bounds, so it initially overlaps the safeAreaInsets
B. The view controller is performing manual layout in viewDidLayoutSubviews

Because of both of these factors the view has safeAreaInsets set before its laid out, and it has that safeAreaInsets populated by the time the view controller performs its layout code. However this isn't a good pattern to mandate, as it requires clients to:

A. Perform layout code within viewDidLoad as well as viewDidLayoutSubviews.
B. Never do layout in viewWillLayoutSubviews.
C. This wont work at all for views that are children of other views. viewDidLayoutSubviews allows the parent in this case to do manual layout after the entire layout pass has finished. A parent view does not have the same option, it performs layout in layoutSubviews- the middle of the layout pass.

So, I think this approach leaves open a bunch of ways for clients to trip themselves up - cases where sizeThatFits will not report the right height on the first layout pass, and where the user will need to trigger a second layout pass with [self.superview setNeedsLayout] within safeAreaInsetsDidChange - but even that approach wont be effective, since if the autolayout engine is running on that view hierarchy subtree then the setNeedsLayout will be ineffective, meaning users will really need to call layoutSubviews directly, putting us in the same place as before - vulnerable to infinite layout loops.

I'm planning on taking another pass at GOOTabBar with this info in mind, and can either contribute or put up another PR mirroring that change if that would be helpful (although I will be OOO tomorrow)

chuga:
That'd be super helpful. We need to see how both implementations look like if we want to compare them. Thanks for putting that feedback in writing btw.

romoore:
@chuga How do you prefer contributions? Do you want a patch file in a comment or do you want to add write access to your fork?

chuga:
@romoore As an admin, you should have write access to this branch in my fork. Let me know if that's not the case

components/Tabs/src/private/MDCItemBar.m
```diff
@@ -109,14 +112,19 @@ - (void)commonItemBarInit {
...
   collectionView.backgroundColor = [UIColor clearColor];
-  collectionView.clipsToBounds = YES;
+  collectionView.clipsToBounds = NO;
   collectionView.scrollsToTop = NO;
   collectionView.showsHorizontalScrollIndicator = NO;
   collectionView.showsVerticalScrollIndicator = NO;
+
+#if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
+  if (@available(iOS 11.0, *)) {
+    collectionView.contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentAlways;
```
jmdetloff: We risk introducing vertical scrolling in our tabs if we do this. If the client lays out the tab bar smaller than its sizeThatFits, instead of the item bar just being clipped it will become vertically scrollable. Thoughts on that? I don't think scrolling is necessarily strictly worse than clipping, it just feels a little weird.

chuga: Both sound like a bug on their end tbh. What if we use UIScrollViewContentInsetAdjustmentScrollableAxes instead?

components/Tabs/src/MDCTabBar.m
```diff
@@ -280,7 +280,9 @@ - (CGSize)sizeThatFits:(CGSize)size {
   size = [_itemBar sizeThatFits:size];
 #if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
   if (@available(iOS 11.0, *)) {
-    size.height += self.safeAreaInsets.bottom;
+    if (![[self class] isTopTabsForPosition:_barPosition]) {
```
jmdetloff: Sorry, I missed the conversation regarding why this is needed. Why do we need to check this?

components/Tabs/src/private/MDCItemBar.m
jmdetloff: I think just automatic (the default) will do what we need here, it'll scroll horizontally and not vertically.

When I say clipping, I simply mean if we make the containing view smaller than our UICollectionView with clipsToBounds set to YES (set above) it'll clip. Thats I think the behavior we would want.

jmdetloff:
Here's a paste containing the changes I would recommend, based off the rationale described in my comment above:

https://paste.googleplex.com/5724248198873088

chuga:
Assigning it to @ianegordon to make a call

ianegordon:
Looking.

chuga:
Added back the intrinsicContentSize for consistency with sizeThatFits.

ianegordon:
Based on our offline discussion:
We will update sizeThatFits: and intrinsicContentSize to include safeAreaInsets.bottom. (This should work regardless of where the tab bar is positioned)
We may have to adjust TabBarItems frames to hug the top of the TabBar.
We will match the 'broken' UITabBar behavior if the tab bar is on the bottom on an iPhone X.  There is an unanswered question about why UITabBar is not getting properly resized outside of a UITabBarController.
We will not call layoutSubview or updateConstraints from safeAreaInsetsDidChange.
We may call setNeedsLayout from layoutSubviews.

chuga:
@brianjmoore Any chance you could help me out with MDCItemBarFlowLayout and RTL? I made a couple of changes but the layout is behaving weirdly on load (works fine when I select one of the items).

brianjmoore:
I haven't patched this in to debug yet, but I would be curious what the behavior is if you just use UICollectionViewFlowLayout as-is. MDCItemBarFlowLayout is just a workaround for some iOS 9-era bugs, so switching it off might isolate if the issue is in the layout or elsewhere. It also might be that we don't need it anymore on 11+, I don't know.

chuga:
Thanks @brianjmoore, I'll give that a try.
@ianegordon What's the final decision on sizeThatFits?

ianegordon:
Based on Apple's documentation and recommendations, I think we should *not* include safeAreaInsets in our control's intrinsic size and sizeThatFits: and leave it to the View Controller.

PROs
This matches Apple's recommendations.  (IntrinsicContentSize should not take into account a view's place in hierarchy, only its content)
It is simpler.

CONs
This is not (currently) what Apple is doing.   UITabBar may be pulling safeAreaInsets from its  UIWindow.  However, UINavigationBar does not take into account safeArea, but rather adds a shim UIView.

That said, I've gone back and forth, more than once, on my preferred strategy.  I think we should go with the simpler implementation for now until the devices arrive.

If possible, consider escalation to get a recommendation from Apple.
If possible, we should be able to adjust our strategy once we have devices in hand.  I expect this will not be possible, but flexibility would be useful.

chuga:
ACK.
I do have to say, in case it isn't clear from this discussion already, that doing what UITabBar does (and that is, take into account the safe area insets) feels like the right thing to do.
I'll undo that part of this PR, and assign a new issue to @jmdetloff to implement that change.

chuga:
For reference, here's the new issue: https://github.com/material-components/material-components-ios/issues/2302

chuga:
After being able to repro the RTL glitches with a vanilla UICollectionViewFlowLayout, I created a new issue to look into that and not block this one on it https://github.com/material-components/material-components-ios/issues/2303. Looks like a bug on UIKit, but worth exploring more.

chuga:
Ok folks, this is now ready for review.

chuga:
@romoore re: https://github.com/material-components/material-components-ios/pull/2261#issuecomment-339104493 it looks like MDCTabBarViewController doesn't give its child VCs the entire frame: https://github.com/material-components/material-components-ios/blob/660c3fb066097f8fc5a42fbda8448ce195f6499e/components/Tabs/src/MDCTabBarViewController.m#L249 Since there's no overlap with the tab bar, I don't think we should be setting any extra safe area insets. Let me know if I'm wrong.

chuga:
Friendly ping, this is ready for review.

romoore:
@chuga  Thanks for checking. I see that the behavior of MDCTabBarViewController differs from UITabBarController in this respect. I've opened #2311 to change this behavior.

components/Tabs/src/MDCTabBar.m
```diff
@@ -177,7 +185,7 @@ - (void)setBarTintColor:(UIColor *)barTintColor {
     _barTintColor = barTintColor;
 
     // Update background color.
-    _itemBar.backgroundColor = barTintColor;
+    self.backgroundColor = barTintColor;
```
romoore, line 188: This looks like an unrelated change and not related to the safe area. Is there a reason we want to set the background color of the entire view rather than just the itemBar?

components/Tabs/src/private/MDCItemBar.m
```diff
@@ -394,7 +414,7 @@ - (CGSize)collectionView:(UICollectionView *)collectionView
   // Divide justified items evenly across the view.
   if (_alignment == MDCItemBarAlignmentJustified) {
     NSInteger count = [self collectionView:_collectionView numberOfItemsInSection:0];
-    size.width = _collectionView.bounds.size.width / MAX(count, 1);
+    size.width = self.adjustedCollectionViewWidth / MAX(count, 1);
```
romoore: Nit: Same feedback about method invocation here and below.

components/Tabs/src/private/MDCItemBar.m
```diff
@@ -275,13 +284,15 @@ - (void)layoutSubviews {
   _collectionView.frame = bounds;
 
   // Update collection metrics if the size has changed.
-  if (!CGSizeEqualToSize(bounds.size, _lastSize)) {
+  if (!CGSizeEqualToSize(bounds.size, _lastSize) ||
+      self.adjustedCollectionViewWidth != _lastAdjustedCollectionViewWidth) {
```
romoore: Nit: As a method, this should be invoked using the "brackets" syntax `[self adjustedCollectionViewWidth]`.

components/Tabs/src/private/MDCItemBar.m
```diff
@@ -71,6 +71,9 @@ @implementation MDCItemBar {
   /// Size of the view at last layout, for deduplicating changes.
   CGSize _lastSize;
 
+  /// Width of the collection view accounting for Safe Area insets at last layout.
```
romoore: Super-nit: "Safe Area insets" should probably just be "safeAreaInsets" since it's a property name.

components/Tabs/src/private/MDCItemBar.m
```diff
@@ -776,7 +812,7 @@ - (void)prepareLayout {
   // Apply global content size padding.
   if (shouldPadContentSizeForRTL) {
     _isPaddingCollectionViewContentSize = YES;
-    _paddedCollectionViewContentSize = self.collectionView.bounds.size;
+    _paddedCollectionViewContentSize = self.adjustedCollectionViewBounds.size;
```
romoore: Nit: Same feedback about invoking this method as well.

romoore (changes requested):

components/Tabs/src/MDCTabBar.m
chuga: Yeah, the frame of the MDCTabBar can (and should be) taller than the _itemBar to leave some space for the home indicator.

chuga:
Done. Thanks for the review, Rob.

components/Tabs/src/MDCTabBar.m
romoore: Ack. Make sense to me. Aside from the nits, everything LGTM!

romoore (approved):

PR #2272, opened 2017-10-25
[BottomAppBar] Set nav bar frame on layout

jgunaratne:
Sets nav bar frame on layout. Fixes an issue where button is not tappable despite frame appearing to be correct size.

### DIFF
--- a/components/BottomAppBar/src/MDCBottomAppBarView.m
+++ b/components/BottomAppBar/src/MDCBottomAppBarView.m
@@ -238,8 +238,8 @@ static const int kMDCButtonAnimationDuration = 200;
 
 - (void)layoutSubviews {
   [super layoutSubviews];
-      self.floatingButton.center =
-  [self getFloatingButtonCenterPositionForWidth:CGRectGetWidth(self.bounds)];
+  self.floatingButton.center =
+      [self getFloatingButtonCenterPositionForWidth:CGRectGetWidth(self.bounds)];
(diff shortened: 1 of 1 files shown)


components/BottomAppBar/src/MDCBottomAppBarView.m
```diff
@@ -92,11 +92,7 @@ - (void)addFloatingButton {
 }
 
 - (void)addNavBar {
-  CGRect navBarFrame = CGRectMake(0,
-                                  kMDCBottomAppBarYOffset,
-                                  self.bounds.size.width,
-                                  self.bounds.size.height);
-  _navBar = [[MDCNavigationBar alloc] initWithFrame:navBarFrame];
+  _navBar = [[MDCNavigationBar alloc] initWithFrame:CGRectZero];
   _navBar.autoresizingMask = UIViewAutoresizingFlexibleWidth;
```
chuga, line 96: do you still need this?

components/BottomAppBar/src/MDCBottomAppBarView.m
```diff
@@ -245,6 +241,12 @@ - (void)layoutSubviews {
       self.floatingButton.center =
```
chuga: looks like we need to fix the indentation here

chuga (approved):

components/BottomAppBar/src/MDCBottomAppBarView.m
jgunaratne: Fixed.

PR #2274, opened 2017-10-25
[TextFields] Major improvements to RTL

### LINKED ISSUE #1843: [TextFields] Placeholder layout is broken in RTL with an overlay view
The width is insufficient. Works when no leadingView or trailingView.
 iOS 8.1. iPhone 5S
simulator screen shot - iphone 6 - 2017-08-18 at 16 20 18

### LINKED ISSUE #2237: [TextFields] TrailingView does not work in RTL
### Overview
The documentation for the MDCTrailingView trailingView property indicates that it will display on the left when in RTL mode.  However, while the trailingView displays properly in LTR mode, when in RTL mode, it doesn't show at all.

It looks like the trailingView property should be updated to check the layout direction, similar to the implementation of the trailingViewMode property.

### Reproduction steps
- Create a view with the trailingView set.  Change language to an RTL language.  The trailingView is not visible.

### Version number
37.0.0

willlarche:
Closes #1843 
Closes #2237 

Replaces old RTL support with MDFInternationalization.

### DIFF
--- a/MaterialComponents.podspec
+++ b/MaterialComponents.podspec
@@ -460,6 +460,7 @@ Pod::Spec.new do |s|
 
       sss.dependency "MDFInternationalization"
       sss.dependency "MaterialComponents/private/Math"
+      sss.dependency "MDFInternationalization"
     end
     ss.subspec "ColorThemer" do |sss|
       sss.ios.deployment_target = '8.0'
--- a/components/TextFields/src/MDCTextField.m
+++ b/components/TextFields/src/MDCTextField.m
@@ -19,9 +19,9 @@
 #import "MDCTextInputCharacterCounter.h"
 #import "private/MDCTextInputCommonFundament.h"
 
-#import "MDFInternationalization.h"
 #import "MaterialMath.h"
 #import "MaterialTypography.h"
+#import "MDFInternationalization.h"
 
 static NSString *const MDCTextFieldCursorColorKey = @"MDCTextFieldCursorColorKey";
 static NSString *const MDCTextFieldFundamentKey = @"MDCTextFieldFundamentKey";
@@ -121,7 +121,6 @@ static const CGFloat MDCTextInputEditingRectRightViewPaddingCorrection = -2.f;
     copy.trailingView = [self.trailingView copy];
   }
   copy.trailingViewMode = self.trailingViewMode;
-
 
   return copy;
 }
--- a/components/TextFields/src/MDCTextField.m
+++ b/components/TextFields/src/MDCTextField.m
@@ -307,26 +307,36 @@ static const CGFloat MDCTextInputEditingRectRightViewPaddingCorrection = -2.f;
 }
 
 - (UIView *)trailingView {
+  if ([self shouldManuallyEnforceRightToLeftLayoutForOverlayViews] &&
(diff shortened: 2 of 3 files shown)


MaterialComponents.podspec
```diff
@@ -459,7 +459,7 @@ Pod::Spec.new do |s|
       sss.dependency "MaterialComponents/Typography"
 
       sss.dependency "MaterialComponents/private/Math"
-      sss.dependency "MaterialComponents/private/RTL"
+      sss.dependency "MDFInternationalization"
```
sigito: Since this component does not depend on MDCRTL anymore, you should also remove the rest of `MatiralRTL.h` imports in:
* `MDCMultilineTextField.m`
* `MDCTextInputControllerFilled.m`
* `MDCTextInputControllerFullWidth.m`
* `MDCTextInputControllerLegacyDefault.m`
* `MDCTextInputControllerLegacyFullWidth.m`

willlarche: I think yours covers that. We’ll merge it first.

components/TextFields/src/MDCTextField.m
```diff
@@ -282,26 +282,36 @@ - (UILabel *)trailingUnderlineLabel {
 }
 
 - (UIView *)trailingView {
+  if ([self shouldManuallyEnforceRightToLeftLayoutForOverlayViews] &&
+      self.mdf_effectiveUserInterfaceLayoutDirection == UIUserInterfaceLayoutDirectionRightToLeft) {
```
ianegordon: Should this be an && or and || ? 
If we are iOS 9+, and self.mdf_effectiveUIDirection == RTL, do we not want to return self.leftView?

components/TextFields/src/MDCTextField.m
```diff
@@ -621,14 +638,32 @@ - (void)textFieldDidEndEditing:(__unused NSNotification *)note {
...
+#pragma mark - RTL
+
+// TODO: (larche) remove when we drop iOS 8
+// Prior to iOS 9 RTL was not automatically applied, so we need to apply fixes manually.
+- (BOOL)shouldManuallyEnforceRightToLeftLayoutForOverlayViews {
+  BOOL manuallyEnforceRTL = YES;
+
+  NSOperatingSystemVersion iOS9Version = {9, 0, 0};
+  NSProcessInfo *processInfo = [NSProcessInfo processInfo];
+  if ([processInfo respondsToSelector:@selector(isOperatingSystemAtLeastVersion:)] &&
```
ianegordon, line 677: Ah, now I understand our conversation determining iOS version.

components/TextFields/src/MDCTextField.m
ianegordon: Followup: I understand now.  Consider adding a comment:
// On iOS 8 we must manually exchange leftView and rightView in RTL mode.

ianegordon (approved):
Please verify (ideally including screenshots) of before and after on in RTL mode of an iOS 8 Sim and an iOS 9+ Sim.

components/TextFields/src/MDCTextField.m
willlarche: I've added a comment on each of these now that says that.

PR #2289, opened 2017-10-26
[TextFields] Fixes bugs on multiline IB support

### LINKED ISSUE #2278: [TextFields] Multiline text field causes crash if added to IB file
**Problem**
Creating an instance of MDCMultilineTextField in Storyboard in Xcode 9.0 leads to 'NSGenericException', specifically regarding the NSLayoutConstraints on the embedded UITextView within the MDCMultilineTextField.

**Steps to reproduce** (XCode 9.0)
1. Start a new XCode Project of the 'Single View App' type
2. Install the MaterialComponents pod (`pod init`, add "pod 'MaterialComponents', '~> 38.0'" to the podfile, `pod install`)
3. Check that MaterialComponents installed correctly and open the new workspace file
4. Drag and drop a new UIView onto the default View Controller
5. Change the Custom Class of the new UIView in the Identity Inspector to be MDCMultilineTextField
6. Add four constraints between the four edges of the MDCMultilineTextField view and the main view
7. Run. NSGenericException occurs stating that it cannot activate constraints.

**Underlying cause:**
When MDCMultilineTextField's `updateConstraints` method is called, there are two UITextViews. One UITextView is a subview of the MDCMultilineTextField (which you can find by using NSLog to print out the value self.subviews). The other UITextView is the one stored in `self.textView` of the MDCMultilineTextField. This second one is not a subview of the MDCMultilineTextField, so trying to activate the new constraint(s) in the `updateConstraints` method fails.

The `_textView` property is initialised twice during the `initWithCoder` method. One UITextView is created during the `[[MDCTextInputCommonFundament alloc] initWithTextInput:self];` method. Then another UITextView is created later in the `initWithCoder` method, just after the line `_multilineDelegate = [aDecoder decodeObjectForKey:MDCMultilineTextFieldMultilineDelegateKey];`. This second UITextView is never added as a subview.

**Workaround:**
Comment out the line with `_textView = [[UITextView alloc] initWithFrame:CGRectZero]`, immediately following the line `_multilineDelegate = ...`.

Running now leads to no exception, and the multiline text field works as expected.

willlarche:
Closes #2278

### DIFF
--- a/components/TextFields/examples/TextFieldInterfaceBuilderExample.m
+++ b/components/TextFields/examples/TextFieldInterfaceBuilderExample.m
@@ -20,22 +20,22 @@
 
 @import MaterialComponents.MaterialTextFields;
 
-@interface TextFieldInterfaceBuilderLegacyExample () <UITextFieldDelegate>
+@interface TextFieldInterfaceBuilderExample () <UITextFieldDelegate>
 
 @property(weak, nonatomic) IBOutlet MDCTextField *firstTextField;
-@property(nonatomic, strong) MDCTextInputControllerLegacyDefault *firstController;
+@property(nonatomic, strong) MDCTextInputControllerFilled *firstController;
 @property(weak, nonatomic) IBOutlet MDCTextField *lastTextField;
-@property(nonatomic, strong) MDCTextInputControllerLegacyDefault *lastController;
+@property(nonatomic, strong) MDCTextInputControllerFilled *lastController;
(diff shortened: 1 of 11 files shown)


ajsecord (approved):

willlarche:
Well, this borked the NSCoder unit tests. Working on fix.

willlarche:
@ajsecord , there have been changes. Please re-approve.

ajsecord (approved):

PR #2300, opened 2017-10-27
[NavigationBarExamples] Swapped private/RTL with MDFi18n

sigito:
I really don't know how it slipped through. The last piece before `private/RTL` is ready to be removed.

#1475.

### DIFF
--- a/components/NavigationBar/examples/NavigationBarLayoutExample.m
+++ b/components/NavigationBar/examples/NavigationBarLayoutExample.m
@@ -55,7 +55,7 @@
 
   UIBarButtonItem *backButtonItem = [[UIBarButtonItem alloc]
       initWithImage:[[[MDCIcons imageFor_ic_arrow_back]
-                        mdf_imageFlippedForRightToLeftLayoutDirection]
+                        mdf_imageWithHorizontallyFlippedOrientation]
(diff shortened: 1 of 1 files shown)


components/NavigationBar/examples/NavigationBarLayoutExample.m
```diff
@@ -19,11 +19,10 @@
 //
 //
 
+#import "MDFInternationalization.h"
```
willlarche: Move down to bottom of list to keep it alphabetized.

willlarche (changes requested):

components/NavigationBar/examples/NavigationBarLayoutExample.m
sigito: Aren't all capital letter are in front of lower ones according to the ObjC style guide?

willlarche: Oh, possibly! Can you cite?

sigito: I struggle to actually find it. Moved the import down :)

willlarche: Thank you!

willlarche (approved):

PR #2305, opened 2017-10-29
[Catalog] @objc annotating all CatalogByConvention funcs

mpetrov:
CatalogByConvention funcs are implicitly `@objc` in Swift3.

In Swift4, this will no longer be the case and could cause runtime failures.


willlarche (approved):
Thank you!

PR #2317, opened 2017-10-31
[FlexibleHeader] Add a 'hideViewWhenShifted' API to the flexible header

chuga:
This PR introduces a new API that will allow client teams to easily hide the content of the flexible header when its shift behavior is set to MDCFlexibleHeaderShiftBehaviorEnabled. This is fully opt-in to prevent regressions.

Closes https://github.com/material-components/material-components-ios/issues/2127.

ezgif-4-081414844b

### DIFF
--- a/components/AppBar/examples/AppBarTypicalUseExample.m
+++ b/components/AppBar/examples/AppBarTypicalUseExample.m
@@ -39,6 +39,9 @@
     // Optional: Change the App Bar's background color and tint color.
     UIColor *color = [UIColor colorWithWhite:0.2 alpha:1];
     _appBar.headerViewController.headerView.backgroundColor = color;
+    _appBar.headerViewController.headerView.shiftBehavior = MDCFlexibleHeaderShiftBehaviorEnabled;
+    [_appBar.headerViewController.headerView hideViewWhenShifted:_appBar.headerStackView];
+
     _appBar.navigationBar.tintColor = [UIColor whiteColor];
     _appBar.navigationBar.titleTextAttributes = @{
                                             NSForegroundColorAttributeName : [UIColor whiteColor],
--- a/components/FlexibleHeader/src/MDCFlexibleHeaderView.h
+++ b/components/FlexibleHeader/src/MDCFlexibleHeaderView.h
@@ -335,6 +335,15 @@ IB_DESIGNABLE
 @property(nonatomic) MDCFlexibleHeaderShiftBehavior shiftBehavior;
 
 /**
+ Hides the view by changing its alpha when the header shifts. Note that this only happens when the
+ header shifting behavior is set to MDCFlexibleHeaderShiftBehaviorEnabled.
+ */
+- (void)hideViewWhenShifted:(nonnull UIView *)view;
+
+/** Stops hiding the view when the header shifts. */
+- (void)stopHidingViewWhenShifted:(nonnull UIView *)view;
+
+/**
  If shiftBehavior is enabled, this property affects the manner in which the Header reappears when
  pulling content down in the tracking scroll view.
 
--- a/components/FlexibleHeader/src/MDCFlexibleHeaderView.m
+++ b/components/FlexibleHeader/src/MDCFlexibleHeaderView.m
@@ -30,6 +30,11 @@ static const CGFloat kFlexibleHeaderDefaultHeight = 56;
 // The maximum default opacity of the shadow.
 static const float kDefaultVisibleShadowOpacity = 0.4f;
 
+// The threshold in which the _viewsToHideWhenShifted should be fully hidden. 0.5 means the views
+// are completely hidden when the header has shifted half of its content height upwards. This should
+// never be 0.
+static const float kContentHidingThreshold = 0.5f;
+
 // This length defines the moment at which the shadow will be fully visible as the header shifts
 // on-screen.
 static const CGFloat kShadowScaleLength = 8;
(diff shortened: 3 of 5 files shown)


components/FlexibleHeader/src/MDCFlexibleHeaderView.m
```diff
@@ -776,17 +785,24 @@ - (void)fhv_accumulatorDidChange {
...
   CGFloat boundedAccumulator = MIN([self fhv_accumulatorMax], _shiftAccumulator);
 
+  if (_shiftBehavior == MDCFlexibleHeaderShiftBehaviorEnabled) {
+    CGFloat contentHeight = self.computedMinimumHeight - MDCDeviceTopSafeAreaInset();
+    CGFloat hideThreshold = kContentHidingThreshold;
+    CGFloat alpha = MAX(contentHeight - boundedAccumulator / hideThreshold, 0) / contentHeight;
+    for (UIView *view in _viewsToHideWhenShifted) {
+      view.alpha = alpha;
+    }
+  }
```
jverkoey, line 804: This logic may be simplified if you use the phase properties and move this logic to after `fhv_recalculatePhase` is invoked in `fhv_commitAccumulatorToFrame`. You should then be able to use the `_scrollPhasePercentage` to set the view alpha.

jverkoey (changes requested):

components/AppBar/examples/AppBarTypicalUseExample.m
```diff
@@ -39,6 +39,9 @@ - (id)init {
     // Optional: Change the App Bar's background color and tint color.
     UIColor *color = [UIColor colorWithWhite:0.2 alpha:1];
     _appBar.headerViewController.headerView.backgroundColor = color;
+    _appBar.headerViewController.headerView.shiftBehavior = MDCFlexibleHeaderShiftBehaviorEnabled;
+    [_appBar.headerViewController.headerView hideViewWhenShifted:_appBar.headerStackView];
```
jverkoey, line 43: Should this be in the AppBar's logic?

chuga: I thought about that, and it would be nice to make this the default, but I don't think we can as it will change the current behavior... WDYT?

jverkoey: Ah true. I wager that we may want this to be the default behavior, so perhaps in a follow-up PR we can introduce an opt-in API where the content is faded out when enabled?

components/FlexibleHeader/src/MDCFlexibleHeaderView.m
chuga: I don't think I can use `_scrollPhasePercentage` for this, take a look:

ezgif-4-c113d5b396

jverkoey: Hrm, weird. The documented behavior is for shifting is supposed to be: "0 is unshifted, 1.0 is fully shifted off-screen". Seems there might be a bug in the shift phase calculations?

jverkoey: The configurator demo looks like it's setting the values as expected....I'm not sure why it would be working differently here.

chuga: Weird indeed, I'm seeing similar numbers in the Configurator demo

jverkoey: Must be something recent? I can't repro the odd numbers on the latest internal cut of MDC.

components/AppBar/examples/AppBarTypicalUseExample.m
AlimoRabbani: Opt-in API on the AppBar would be nice. Thanks for preparing this PR so quickly!

components/FlexibleHeader/src/MDCFlexibleHeaderView.m
chuga: I just rebased and this is still happening, `_scrollPhasePercentage` doesn't go from 0 to 1. Can I go ahead with the current implementation?

jverkoey: 👍

jverkoey (approved):

components/FlexibleHeader/src/MDCFlexibleHeaderView.h
```diff
@@ -335,6 +335,15 @@ IB_DESIGNABLE
 @property(nonatomic) MDCFlexibleHeaderShiftBehavior shiftBehavior;
 
 /**
+ Hides the view by changing its alpha when the header shifts. Note that this only happens when the
+ header shifting behavior is set to MDCFlexibleHeaderShiftBehaviorEnabled.
+ */
+- (void)hideViewWhenShifted:(nonnull UIView *)view;
```
ianegordon, line 341: Can we make this a BOOL property?

ianegordon (changes requested):

components/FlexibleHeader/src/MDCFlexibleHeaderView.h
chuga: huh?

jverkoey: https://www.youtube.com/watch?v=kfVsfOSbJY0

ianegordon (approved):
Discussed offline.  A BOOL to control this behavior is a much better fit in AppBar.  Details / implementation to be determined in a future PR.

chuga:
Thanks for the reviews, I'll add a simple API to the AppBar in a follow up PR.

PR #2318, opened 2017-11-01
[FlexibleHeader] Update Swift snippets in README

mpetrov:
Updated to modern Swift syntax, as well as using some more idiomatic constructs for protocol conformance and nil checking.


ajsecord (approved):
Thanks!

PR #2322, opened 2017-11-01
[Tabs] API review: selection indicator templates 

brianjmoore:
This is an API review for adding support for custom selection indicators to MDC Tabs, with an initial focus on path-based indicators. This will allow clients to define whatever shape they want to replace the default underline indicator, including shapes that are based on the size and location of the tab's title/image/title+image.

This design is intended to leave room for future plans. In particular, future revisions will add support for indicator images, which would be accomplished by adding an optional image property to `MDCTabBarIndicatorAttributes`.

This PR includes a brief usage example for illustration purposes only. The actual implementation will provide a full, independent custom indicator example.


brianjmoore:
To be clear: As this is only the API, it should not actually be merged after approval.

jverkoey:
Can you add a screenshot of the new behavior?

brianjmoore:
Here's a screenshot of the included example, which adds an oval behind the tab's content.

simulator screen shot - iphone 8 - 2017-11-01 at 14 43 34

jverkoey:
Neat. I like the UICollectionViewItem-esque API design here.

Will changes to the tab selection animate the selected path?

components/Tabs/src/MDCTabBarIndicatorTemplate.h
```diff
@@ -0,0 +1,36 @@
+/*
+ Copyright 2016-present the Material Components for iOS authors. All Rights Reserved.
```
jverkoey: Please update copyright headers here and throughout.

brianjmoore:
Yep, the provided path will animate between states using CAShapeLayer's normal tweening. My rough prototype is at https://github.com/brianjmoore/material-components-ios/tree/selectiontemplate/components if you want to check it out.

components/Tabs/src/MDCTabBarIndicatorTemplate.h
```diff
@@ -0,0 +1,36 @@
...
+/*
+ Template for indicator content which defines how the indicator changes appearance in response to
+ changes in its context.
+ */
+@protocol MDCTabBarIndicatorTemplate <NSObject>
+
+/**
+ Returns an attributes object that describes how the indicator should appear in a given context.
+
+ If unimplemented, the default is an empty attributes object.
```
jverkoey: This can't be unimplemented - do we need this line of documentation?

components/Tabs/src/MDCTabBarIndicatorContext.h
```diff
@@ -0,0 +1,41 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+/** Information about the context in which a tab bar indicator will be displayed. */
+@interface MDCTabBarIndicatorContext : NSObject
```
jverkoey: Can this be a protocol publicly? Making it a protocol gives us more flexibility to adjust the underlying implementation without affecting clients. I'm assuming that clients will never need to instantiate an instance of this type.

jverkoey (changes requested):

romoore (commented):
Is there a reason why we would want indicators to be something other than a "shape" sitting below the content within a tab? For example, do we want to support circles enclosing the tab content, translucent vector art overlaying it, etc.?  Without understanding the use cases for such customizable behavior, I'm thinking that having a basic ShapeGenerator that the TabBar uses to render a shaped underline might be better.

components/Tabs/examples/TabBarTextOnlyExample.m
```diff
@@ -72,6 +87,7 @@ - (void)loadTabBar {
   // rely on the UIAppearance proxy.
   self.tabBar.selectedItemTintColor = [UIColor whiteColor];
   self.tabBar.tintColor = [UIColor colorWithWhite:1.0f alpha:0.5f];
```
jgunaratne, line 90: Tangentially related: any plans to move away from tintColor in MDCTabBar and instead use color properties that have clearer semantic meaning (e.g. barBackgroundColor, disabledTabColor, etc.)? @ianegordon has mentioned tintColor applied to some UIKit components can have unpredictable colorization results.

components/Tabs/src/MDCTabBarIndicatorTemplate.h
brianjmoore: Done

components/Tabs/src/MDCTabBarIndicatorTemplate.h
brianjmoore: Removed

components/Tabs/src/MDCTabBarIndicatorContext.h
brianjmoore: Yeah, that makes sense. Changed.

components/Tabs/examples/TabBarTextOnlyExample.m
brianjmoore: No plans at the moment and they would be independent of this work, but I agree it would help add some uniformity to the API. Should be as easy as adding a selectionIndicatorTintColor property.

brianjmoore:
@romoore The mocks that defined the requirements for this design include selection indicators that are not underlines below the tab content, such as a round-rect pill shape behind the text. There are also designs that show a bitmap under the tab's content.

At its core this design is just a shape generation scheme, but it's written to be flexible to future visual designs. If we were to have generalized shape definition facilities in MDC in the future, they could be developed orthogonally to this and plugged in straightforwardly as a new template type.

romoore:
@brianjmoore Thanks for the reply. Given the need to support images and shaped highlights, would it make sense to have the Indicator Template return a UIView for the provided context rather than the IndicatorAttributes?

brianjmoore:
@romoore Custom views were in fact the original design, but I decided to go for something more limited for a few reasons:
* In practice, implementing indicator views that update and animate their contents in a predictable and robust way is difficult and subtle. Requiring clients to perform this work would be onerous and placing this complexity in MDCTabBar would increase MDC’s support burden. Templates let us make it easy for clients to implement a pretty wide variety of pretty behaviors without much difficulty.
* Integration with UIAppearance was more awkward. UIAppearance support is what led me to value-semantic template objects in the first place, and at that point it didn't feel appropriate to have the template generate a full-on view due to the potential for breakage (see above) and misuse.
* More of a tertiary concern, but logic for custom indicator views written for iOS are less portable to other platforms than custom path-or-image generating templates would be.

One neat thing about templates is that they would still allow full custom indicator view support to be built on top in the future if we wanted to provide ultimate flexibility. The view API is then pretty straightforward with just a way to apply MDCTabBarIndicatorAttributes provided by a template.

romoore:
@brianjmoore That sounds good to me. Thanks for replying and explaining the rationale for the design. 👍  Looks good to me!

romoore (approved):
General API looks good to me.

brianjmoore:
Thanks for the feedback everyone! Let's give this review one more day and close it out at 12:00 eastern time on 2017-11-8.

jverkoey (approved):

jgunaratne (approved):

brianjmoore:
It appears all comments have been addressed. Thanks for the review!

willlarche:
Woo!!!

PR #2326, opened 2017-11-02
[Catalog] Fixed some swift Imports

randallli:
Changed some imports to be at the component level instead of all of material components.


catalog/MDCCatalog/AppDelegate.swift
```diff
@@ -17,7 +17,21 @@ limitations under the License.
 import UIKit
 
 import CatalogByConvention
-import MaterialComponents
+
```
willlarche, line 20: Why do we prefer 14 imports to 1?

randallli: Internally we do not have a global target that includes everything MDC.

willlarche: Why is this an issue now?

randallli: This is to get our catalog working on internal build.

jverkoey (approved):

PR #2351, opened 2017-11-04
[TextFields] Fix multiline placeholder position and transform bug

### LINKED ISSUE #2121: [TextFields] Floating label in MDCTextInputControllerLegacyDefault is sometimes indented.
### Overview
Occasionally, the floating label for the placeholder is showing as indented.  I believe this happens with both the MDCTextInputControllerDefault & MDCTextInputControllerLegacyDefault.  It appears to be related to setting the MDCTextField's text programatically.

### Reproduction steps
 Set both the placeholder & text programatically. The floating label will be indented (both when field is active & inactive).
image

### Version number
 35.3.0

iOS 10.3 iPhone 7 Plus (simulator)

willlarche:
closes #2121 

Includes tiny bits of organization too.

### DIFF
--- a/components/TextFields/examples/TextFieldExample.swift
+++ b/components/TextFields/examples/TextFieldExample.swift
@@ -153,6 +153,7 @@ final class TextFieldSwiftExample: UIViewController {
     scrollView.addSubview(message)
     let messageController = MDCTextInputControllerDefault(textInput: message)
     message.textView?.delegate = self
+    message.text = "This is where you could put a multi-line message like an email."
     message.multilineDelegate = self
     messageController.placeholderText = "Message"
     allTextFieldControllers.append(messageController)
--- a/components/TextFields/src/MDCTextInputControllerDefault.m
+++ b/components/TextFields/src/MDCTextInputControllerDefault.m
@@ -570,8 +570,8 @@ static UITextFieldViewMode _underlineViewModeDefault = UITextFieldViewModeWhileE
   if (isToUp) {
     UIOffset offset = [self floatingPlaceholderOffset];
     UIEdgeInsets insets = self.textInput.textInsets;
-
-    CGFloat horizontalLeading = insets.left - offset.horizontal;
+    CGFloat leadingConstant = [self floatingPlaceholderAnimationConstraintLeadingConstant:insets
+                                                                                   offset:offset];
     if (!self.placeholderAnimationConstraintLeading) {
       self.placeholderAnimationConstraintLeading =
           [NSLayoutConstraint constraintWithItem:self.textInput.placeholderLabel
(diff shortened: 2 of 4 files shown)


components/TextFields/src/MDCTextInputControllerDefault.m
```diff
@@ -620,6 +620,23 @@ - (void)updatePlaceholderAnimationConstraints:(BOOL)isToUp {
   }
 }
 
+- (BOOL)needsUpdatePlaceholderAnimationConstraintsToUp {
```
randallli, line 623: Nonblocking comment: What is "up" in this case? I feel like you explain this to me once but now I forgot again. Perhaps a comment would provide context?

randallli: I noticed that it is a private property. Let us put the comment there.

randallli (approved):

components/TextFields/examples/TextFieldExample.swift
```diff
@@ -153,6 +153,7 @@ final class TextFieldSwiftExample: UIViewController {
     scrollView.addSubview(message)
     let messageController = MDCTextInputControllerDefault(textInput: message)
     message.textView?.delegate = self
+    message.text = "This is where you could put a multi-line message like an email."
```
romoore, line 156: Can we demo something with line breaks like `\r\n`?

components/TextFields/src/MDCTextInputControllerDefault.m
```diff
@@ -653,6 +670,18 @@ - (UIOffset)floatingPlaceholderOffset {
   return UIOffsetMake(horizontal, vertical);
 }
 
+- (CGFloat)floatingPlaceholderAnimationConstraintLeadingConstant:(UIEdgeInsets)textInsets
```
romoore, line 673: If the UIEdgeInsets don't need to be swapped for RTL, could you please document that you expect them to arrived with leading mapped to left, and trailing mapped to right? It would reduce my RTL-related anxiety. :)

components/TextFields/src/MDCTextInputControllerLegacyDefault.m
```diff
@@ -542,12 +537,19 @@ - (void)movePlaceholderToUp:(BOOL)isToUp {
       }];
 }
 
+- (BOOL)isPlaceholderUp {
+  return self.placeholderAnimationConstraints.count > 0 &&
+  !CGAffineTransformEqualToTransform(self.textInput.placeholderLabel.transform,
```
romoore, line 542: Nit/Indentation: The second row should be indented 4 spaces.

romoore (changes requested):

components/TextFields/examples/TextFieldExample.swift
willlarche: Done!

components/TextFields/src/MDCTextInputControllerDefault.m
willlarche: I've removed it as a property since I really just need the getter. And I've commented on that section.

components/TextFields/src/MDCTextInputControllerLegacyDefault.m
willlarche: Done.

components/TextFields/src/MDCTextInputControllerDefault.m
willlarche: Done!

components/TextFields/examples/TextFieldExample.swift
willlarche: DOne

PR #2357, opened 2017-11-06
[BottomNavigation] Fix example using iOS 10 API

### LINKED ISSUE #2355: [BottomNavigation] BottomNavigationTypicalUseExample uses unavailable API UITabBarItem.badgeColor
### Overview
The example code is checking for the SDK at compile time but not at runtime. It needs an additional @available check.

```objectivec
// - (void)commonBottomNavigationTypicalUseExampleInit {
// ...
    tabBarItem5.badgeColor = [MDCPalette cyanPalette].accent700;
  }
#endif
```

romoore:
The typical use example uses an iOS 10-only API and needs an @available check
to silence warnings.

Closes #2355

### DIFF
--- a/components/BottomNavigation/examples/BottomNavigationTypicalUseExample.m
+++ b/components/BottomNavigation/examples/BottomNavigationTypicalUseExample.m
@@ -76,7 +76,12 @@
                                       tag:0];
   tabBarItem5.badgeValue = @"999+";
 #if defined(__IPHONE_10_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_10_0)
-  tabBarItem5.badgeColor = [MDCPalette cyanPalette].accent700;
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wpartial-availability"
+  if ([tabBarItem5 respondsToSelector:@selector(badgeColor)]) {
+    tabBarItem5.badgeColor = [MDCPalette cyanPalette].accent700;
+  }
+#pragma clang diagnostic pop
 #endif
   _bottomNavBar.items = @[ tabBarItem1, tabBarItem2, tabBarItem3, tabBarItem4, tabBarItem5 ];
   _bottomNavBar.selectedItem = tabBarItem2;


components/BottomNavigation/examples/BottomNavigationTypicalUseExample.m
```diff
@@ -76,7 +76,9 @@ - (void)commonBottomNavigationTypicalUseExampleInit {
                                       tag:0];
   tabBarItem5.badgeValue = @"999+";
 #if defined(__IPHONE_10_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_10_0)
-  tabBarItem5.badgeColor = [MDCPalette cyanPalette].accent700;
+  if (@available(iOS 10.0, *)) {
```
jgunaratne: I believe when I added this either Travis or Xcode (compiling for 8.4) complained. Please double-check this works with both.

romoore: Thanks. My approach doesn't work since @available isn't *available* until the iOS 11 SDK. Switched to a respondsToSelector instead.  Fixes the crash on iOS 8.1 simulators.

components/BottomNavigation/examples/BottomNavigationTypicalUseExample.m
```diff
@@ -76,7 +76,12 @@ - (void)commonBottomNavigationTypicalUseExampleInit {
                                       tag:0];
   tabBarItem5.badgeValue = @"999+";
 #if defined(__IPHONE_10_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_10_0)
-  tabBarItem5.badgeColor = [MDCPalette cyanPalette].accent700;
+#pragma clang diagnostic push
```
willlarche, line 79: I'm of the opinion that this isn't exactly what we want to be doing. Instead someone should go thru and add @available all over the place.

romoore: @available is only available on iOS 11+, so we would have to do a macro check for iOS 11 inside the macro check for iOS 10 to see if we can use @available. If we can't use @available, we'd still need to check if the selector is defined since it was introduced in iOS 10.

jgunaratne (approved):

PR #2362, opened 2017-11-06
[Slider]Create default light and dark color schemes for MDCSlider

jmdetloff:

### DIFF
--- a/components/Slider/src/ColorThemer/MDCSliderColorThemer.m
+++ b/components/Slider/src/ColorThemer/MDCSliderColorThemer.m
@@ -15,6 +15,10 @@
  */
 
 #import "MDCSliderColorThemer.h"
+#import "MaterialPalettes.h"
+
+static const CGFloat kSliderThemerLightAlpha = 0.26f;
+static const CGFloat kSliderThemerDarkAlpha = 0.3f;
 
 @implementation MDCSliderColorThemer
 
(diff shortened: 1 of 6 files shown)


components/Slider/src/MDCSlider.m
```diff
@@ -91,20 +91,56 @@ - (void)commonMDCSliderInit {
...
   _thumbTrack.trackOffColor =
       trackBackgroundColor ? trackBackgroundColor : [[self class] defaultTrackOffColor];
-  ;
 }
 
 - (UIColor *)trackBackgroundColor {
   return _thumbTrack.trackOffColor;
 }
 
+- (void)setDisabledColor:(UIColor *)disabledColor {
```
randallli, line 107: lets write a unit test for this.

components/Slider/src/MDCSlider.h
```diff
@@ -152,6 +161,15 @@ IB_DESIGNABLE
  */
 @property(nonatomic, assign, getter=isThumbHollowAtStart) BOOL thumbHollowAtStart;
 
+/**
+ * A default color scheme for sliders displayed on light backgrounds. This color scheme is applied
+ * by default to new MDCSlider instances.
+ */
++ (nonnull MDCBasicColorScheme *)defaultSliderLightColorScheme;
```
randallli: Can we move these into the color themer so the base component doesn't depend on the NDCBasicColorScheme?

randallli (changes requested):

components/Slider/src/MDCSlider.m
jmdetloff: Done

components/Slider/src/ColorThemer/MDCSliderColorThemer.m
```diff
@@ -15,13 +15,46 @@
  */
 
 #import "MDCSliderColorThemer.h"
+#import "MaterialPalettes.h"
```
randallli, line 18: do we need to add a cocoapod dep?

randallli (approved):

components/Slider/src/ColorThemer/MDCSliderColorThemer.m
jmdetloff: Added!

Issue #1221, opened 2017-02-24
[Palettes] Replace hex to RGB functions w/ dependency to Palettes and its function


willlarche:
### Overview
Several pods have a hex to RGB function. There's already one in the Palettes tho. Let's migrate to that and make Palettes a common lightweight dependency.

### Reproduction steps

See:
37:MDCPalettes.m:
static inline UIColor *ColorFromRGB(uint32_t rgbValue) 

65:MDCButton.m:
static inline UIColor *MDCColorFromRGB(uint32_t rgbValue)

36:MDCCollectionInfoBarView.m:
static inline UIColor *ColorFromRGB(uint32_t rgbValue) 

30:MDCProgressView.m:
static inline UIColor *MDCColorFromRGB(uint32_t rgbValue) 

33:MDCSlider.m:
static inline UIColor *MDCColorFromRGB(uint32_t rgbValue)

sigito:
@willlarche, regarding `Snackbar`. It has a few RGB encoded colors. However, some of them are black colors with different alpha. Should they be replaced to `UIColor.blackColor#colorWithAlphaComponent:`?

Also, according to the [spec](https://material.io/guidelines/components/snackbars-toasts.html#snackbars-toasts-specs) a default background color is `0x323232`, which is not in [the color palette](https://material.io/guidelines/style/color.html#color-color-palette). Any ideas on how to approach this?

romoore:
No longer desired. See conversation in #2962

PR #2369, opened 2017-11-06
[Slider] Use MDCPalettes for default thumb track color

sigito:
Partially addresses #1221 and #1222.

### DIFF
--- a/components/Slider/src/MDCSlider.m
+++ b/components/Slider/src/MDCSlider.m
@@ -17,6 +17,7 @@
 #import "MDCSlider.h"
 
 #import "private/MDCSlider_Subclassable.h"
+#import "MaterialPalettes.h"
 #import "MaterialThumbTrack.h"
 
 static const CGFloat kSliderDefaultWidth = 100.0f;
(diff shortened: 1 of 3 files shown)


randallli (approved):

randallli:
I would like @jmdetloff to review as well.

components/Slider/src/MDCSlider.m
```diff
@@ -329,10 +323,6 @@ - (void)thumbTrackTouchCanceled:(__unused MDCThumbTrack *)thumbTrack {
   [self sendActionsForControlEvents:UIControlEventTouchCancel];
 }
 
-+ (UIColor *)defaultColor {
```
randallli, line 332: just an fyi: we may have set this up for subclass use. See as we are discouraging that im ok with removing it.

sigito: Sure, but I didn't see any uses of this method, so removed. It is also not in the [documentation](https://github.com/material-components/material-components-ios/blob/develop/components/Slider/README.md).

ianegordon:
Hi @sigito Can you rebase on top of origin/develop to eliminate the conflict and force push?
Once CI is happy we can land.

sigito:
Done.

jmdetloff (approved):

PR #2372, opened 2017-11-07
[BottomNavigation]  Add protocol for responding to item selection

jgunaratne:
Add protocol for responding to user initiated item selection, similar to MDCTabBar.


brianjmoore:
Why the difference from UITabBar/MDCTabBar's behavior where it is only called for user-initiated changes? I think that detail is fairly important to the API to avoid unintentional side-effects when setting the selected item in code and to differentiate programmatic changes form user-initiated ones.

jgunaratne:
I've updated the protocol to be consistent with UITabBar/MDCTabBar's behavior. The delegate methods are only called when the user initiates changes.

brianjmoore (approved):
Awesome! LGTM

Issue #2350, opened 2017-11-03
[Tabbar] TabBar has incorrect height inside MDCTabBarController


egoldfarb:
### Overview
When using titles + images style, MDCTabBarController sets its TabBar's frame to 72
It should always be using 56 since it's always at bottom position, but position isn't taken into account when calculating the frame.

MDCTabBarController calls +MDCTabBar.defaultHeightForItemAppearance; however, that method doesn't accept bar position as an argument.
The implementation of defaultHeightForItemAppearance in MDCTabBar uses UIBarPositionAny, which then results in the assumption that the TabBar is in top position rather than bottom, and incorrect values are returned.

### Version number
38.1.1

### Operating system and device
iOS 11.1, iPhone 7

PR #2375, opened 2017-11-08
[Tabs] Use bottom position when determining the bottom tab bar's height

brianjmoore:
Fix for https://github.com/material-components/material-components-ios/issues/2350.

This PR adds missing API that allows clients to determine the preferred height for a tab bar given a position and style, then updates MDCTabBarViewController to use it.

This also fixes a small bug triggered by this change where MDCTabBar wouldn't update its layout at the right times and could leave the item bar with the wrong height.

### DIFF
--- a/components/Tabs/src/MDCTabBar.h
+++ b/components/Tabs/src/MDCTabBar.h
@@ -35,6 +35,10 @@
 IB_DESIGNABLE
 @interface MDCTabBar : UIView <UIBarPositioning>
 
+/** The default height for the tab bar with a given position and item appearance. */
++ (CGFloat)defaultHeightForBarPosition:(UIBarPosition)position
+                        itemAppearance:(MDCTabBarItemAppearance)appearance;
+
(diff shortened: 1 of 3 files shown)


chuga (approved):

ianegordon (changes requested):
Thanks @brianjmoore .  We are waiting for a response from Design on how we should handle this in BottomNavBar.  I expect we want to do the same here.  I'll update as soon as we hear back.

brianjmoore:
I'm not sure Design should need to get involved. This was 100% a bug in our implementation.

ianegordon (approved):

ianegordon:
Caught up offline.  This is unrelated to the portrait -vs- landscape sizing question on BottomAppBar.

chuga:
I just re-started the build now that https://github.com/material-components/material-components-ios/pull/2385 has been merged in.

chuga:
Travis is happy. Merging.

PR #2384, opened 2017-11-09
[Tabs] Selection indicator templates

brianjmoore:
This is the initial implementation of selection indicator templates for `MDCTabBar`, supporting indicator shapes. It adds API, documentation, and an example of the new feature.

simulator screen shot - iphone 8 - 2017-11-09 at 09 06 03

Indicator templates allow `MDCTabBar` clients to define how the selected tab is visually indicated. `MDCTabBar` transitions between indicator appearances automatically as the user's selection changes. The default indicator remains a fixed-height underline across the entire tab's bounds, but templates can provide arbitrary paths within the cell's bounds and are provided contextual information that allows placement of these paths relative to the tab's content.

API review: https://github.com/material-components/material-components-ios/pull/2322


components/Tabs/examples/TabBarIndicatorTemplateExample.swift
```diff
@@ -0,0 +1,119 @@
...
+
+  lazy var appBar: MDCAppBar = self.setupAppBar()
+
+  lazy var tabBar: MDCTabBar = {
+    let tabBar = MDCTabBar()
+    tabBar.alignment = .justified
+
+    tabBar.items = [
+      UITabBarItem(title: "Fly", image: nil, tag:0),
+      UITabBarItem(title: "Sleep", image: nil, tag:0),
```
romoore: Can we mix all three tab types in the example just to show how the template will automatically change depending on the tab content?
* Text-only
* Image-only
* Text and Image

components/Tabs/examples/TabBarIndicatorTemplateExample.swift
```diff
@@ -0,0 +1,119 @@
...
+    ]
+
+    let blue = MDCPalette.blue.tint500
+    tabBar.tintColor = blue
+    tabBar.inkColor = blue.withAlphaComponent(0.15)
+
+    tabBar.barTintColor = UIColor.white
+    tabBar.itemAppearance = .titles
+    tabBar.selectedItemTintColor = UIColor.black.withAlphaComponent(0.87)
+    tabBar.unselectedItemTintColor = UIColor.black.withAlphaComponent(0.38)
```
romoore: If possible, can we rely on the catalog themers for color here? Do we need to override the colors specifically for this example?

components/Tabs/examples/supplemental/TabBarIndicatorTemplateExampleSupplemental.swift
```diff
@@ -0,0 +1,97 @@
+/*
+ Copyright 2016-present the Material Components for iOS authors. All Rights Reserved.
```
romoore: s/2016/2017

components/Tabs/examples/supplemental/TabBarIndicatorTemplateExampleSupplemental.swift
```diff
@@ -0,0 +1,97 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+// swiftlint:disable function_body_length
+// swiftlint:disable line_length
+
```
romoore: Why disable these? I don't see any particularly-long functions or lines below.

components/Tabs/examples/supplemental/TabBarIndicatorTemplateExampleSupplemental.swift
```diff
@@ -0,0 +1,97 @@
...
+import MaterialComponents.MaterialButtons
+
+extension TabBarIndicatorTemplateExample {
+
+  func setupAlignmentButton() -> MDCRaisedButton {
+    let alignmentButton = MDCRaisedButton()
+
+    alignmentButton.setTitle("Change Alignment", for: .normal)
+    alignmentButton.setTitleColor(.white, for: .normal)
```
romoore: Can we leave the buttons the way they're themed in the catalog?

components/Tabs/examples/supplemental/TabBarIndicatorTemplateExampleSupplemental.swift
```diff
@@ -0,0 +1,97 @@
...
+                       relatedBy: .equal,
+                       toItem: self.view,
+                       attribute: .centerX,
+                       multiplier: 1,
+                       constant: 0).isActive = true
+    NSLayoutConstraint(item: alignmentButton,
+                       attribute: .bottom,
+                       relatedBy: .equal,
+                       toItem: self.view,
+                       attribute: .bottom,
```
romoore: Will this automatically align to the view's safeAreaLayoutGuide.bottom if present?

components/Tabs/examples/supplemental/TabBarIndicatorTemplateExampleSupplemental.swift
```diff
@@ -0,0 +1,97 @@
...
+  func setupAppBar() -> MDCAppBar {
+    let appBar = MDCAppBar()
+
+    self.addChildViewController(appBar.headerViewController)
+    appBar.navigationBar.backgroundColor = UIColor.white
+
+    appBar.headerViewController.headerView.backgroundColor = UIColor.white
+    appBar.headerViewController.headerView.minMaxHeightIncludesSafeArea = false
+    appBar.headerViewController.headerView.minimumHeight = 56 + 72
+    appBar.headerViewController.headerView.tintColor = MDCPalette.blue.tint500
```
romoore: Same color request here. Can we rely on the Catalog themers instead of local overrides?

components/Tabs/examples/supplemental/TabBarIndicatorTemplateExampleSupplemental.swift
```diff
@@ -0,0 +1,97 @@
...
+
+  func setupAppBar() -> MDCAppBar {
+    let appBar = MDCAppBar()
+
+    self.addChildViewController(appBar.headerViewController)
+    appBar.navigationBar.backgroundColor = UIColor.white
+
+    appBar.headerViewController.headerView.backgroundColor = UIColor.white
+    appBar.headerViewController.headerView.minMaxHeightIncludesSafeArea = false
+    appBar.headerViewController.headerView.minimumHeight = 56 + 72
```
romoore: Nit: Please collapse `56 + 72` to `128` and add a comment to explain why the number was selected. I don't actually know why we're setting to 128. 😓

romoore (changes requested):

components/Tabs/examples/supplemental/TabBarIndicatorTemplateExampleSupplemental.swift
brianjmoore: Unfortunately not for these. Most of the tab bar examples create their own app bar because they're demoing app bar-related stuff, and the catalog themes don't appear to theme custom app bars. Almost all of the tab bar examples are pretty broken, and fixing this is outside of the scope of this work: https://github.com/material-components/material-components-ios/issues/2408

components/Tabs/examples/TabBarIndicatorTemplateExample.swift
brianjmoore: Done. Though it's not perfect. See https://github.com/material-components/material-components-ios/issues/2408/

components/Tabs/examples/TabBarIndicatorTemplateExample.swift
brianjmoore: Good idea. Done

components/Tabs/examples/supplemental/TabBarIndicatorTemplateExampleSupplemental.swift
brianjmoore: Done

components/Tabs/examples/supplemental/TabBarIndicatorTemplateExampleSupplemental.swift
brianjmoore: Yeah just copy-pasted

components/Tabs/examples/supplemental/TabBarIndicatorTemplateExampleSupplemental.swift
brianjmoore: Yep

components/Tabs/examples/supplemental/TabBarIndicatorTemplateExampleSupplemental.swift
brianjmoore: Updated

components/Tabs/examples/supplemental/TabBarIndicatorTemplateExampleSupplemental.swift
brianjmoore: Done

components/Tabs/examples/supplemental/TabBarIndicatorTemplateExampleSupplemental.swift
romoore: Ack.

components/Tabs/examples/TabBarIndicatorTemplateExample.swift
romoore: Ack.

romoore (approved):

components/Tabs/src/MDCTabBarIndicatorAttributes.m
```diff
@@ -0,0 +1,53 @@
...
+- (NSString *)description {
+  return [NSString stringWithFormat:@"%@ path:%@", [super description], _path];
+}
+
+- (BOOL)isEqual:(id)object {
+  if (![object isKindOfClass:[self class]]) {
+    return NO;
+  }
+
+  MDCTabBarIndicatorAttributes *otherAttributes = object;
```
ianegordon, line 40: nit: Add explicit cast?

components/Tabs/src/MDCTabBarIndicatorAttributes.m
```diff
@@ -0,0 +1,53 @@
...
+
+  if ((_path != otherAttributes.path) && ![_path isEqual:otherAttributes]) {
```
ianegordon: ![_path isEqual:otherAttributes.path]

components/Tabs/src/MDCTabBar.m
```diff
@@ -98,6 +100,9 @@ + (void)initialize {
   [[[self class] appearance] setUnselectedItemTintColor:[UIColor colorWithWhite:1.0 alpha:0.7f]];
   [[[self class] appearance] setInkColor:[UIColor colorWithWhite:1.0 alpha:0.7f]];
   [[[self class] appearance] setBarTintColor:nil];
+
+  id<MDCTabBarIndicatorTemplate> template = [[MDCTabBarUnderlineIndicatorTemplate alloc] init];
+  [[[self class] appearance] setSelectionIndicatorTemplate:template];
```
ianegordon: If a property is defined, prefer to use it over the methods.
[[self class] appearance].selectionIndicatorTemplate = template;
Xcode > Edit > Convert > Modernize Objective-c Syntax will do this for you automagically.

components/Tabs/src/private/MDCItemBar.m
```diff
@@ -1,5 +1,5 @@
 /*
- Copyright 2016-present the Material Components for iOS authors. All Rights Reserved.
+ Copyright 2017-present the Material Components for iOS authors. All Rights Reserved.
```
ianegordon: This is not a new file.
Unless we made a mistake and this file was actually created in 2017, we never change copyrights.
If we did make a mistake to our copyright, please make this change in a separate PR.

components/Tabs/src/private/MDCItemBar.m
```diff
@@ -282,13 +281,15 @@ - (void)layoutSubviews {
   const CGRect bounds = self.bounds;
 
   _collectionView.frame = bounds;
+  [_collectionView sendSubviewToBack:_selectionIndicator];
```
ianegordon: I typically would expect to only see sizing / frame changes in layoutSubviews.
Please add a comment on why we reordering our subview's hierarchy.

components/Tabs/src/private/MDCTabBarIndicatorContext+Private.h
```diff
@@ -0,0 +1,31 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+#import "MDCTabBarIndicatorContext.h"
+
+/// Concrete implementation of a tab indicator context.
+@interface MDCTabBarIndicatorContext : NSObject <MDCTabBarIndicatorContext>
```
ianegordon: We have defined an object that implements a protocol with the exact same name?
Consider @interface MDCTabBarIndicatorPrivateContext ?
Naming is hard.

components/Tabs/src/private/MDCTabBarIndicatorView.m
```diff
@@ -0,0 +1,94 @@
...
+  if (self) {
+    // Fill the indicator with the shape.
+    _shapeView = [[MDCTabBarIndicatorShapeView alloc] initWithFrame:self.bounds];
+    _shapeView.autoresizingMask =
+        UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
+    [self addSubview:_shapeView];
+  }
+  return self;
+}
+
```
ianegordon: If we implement one designated initializer, we likely need all designated initializers.
Do we need to implement initWithCoder: ?
Confirm this works as expected in a sample?

ianegordon (changes requested):

components/Tabs/src/MDCTabBar.m
brianjmoore: It looks like these were written this way because [[self class] appearance] returns id, which doesn't permit unknown property lookup but does allow arbitrary message sends. Works with explicitly-written class names though, which I think should be fine considering this is +initialize.

components/Tabs/src/MDCTabBarIndicatorAttributes.m
brianjmoore: Interesting, why? Explicitly casting loses a tiny bit of compile-time safety I think. If the type of `object` were to change, the implicit `id` downcast would fail but the explicit cast would continue to compile despite being wrong.

components/Tabs/src/MDCTabBarIndicatorAttributes.m
brianjmoore: *headdesk*

components/Tabs/src/private/MDCItemBar.m
brianjmoore: yep, that was changed in error

components/Tabs/src/private/MDCItemBar.m
brianjmoore: Added comment here and moved it to where I think it actually belongs

components/Tabs/src/private/MDCTabBarIndicatorContext+Private.h
brianjmoore: Yeah, likely not wise to overlap class/protocol names. Couldn't come up with a much better name

components/Tabs/src/private/MDCTabBarIndicatorView.m
brianjmoore: We shouldn't need it, but implemented anyway. Only the top-level MDCTabBar should need to support NSCoding, as it directly instantiates all of its children

components/Tabs/src/MDCTabBarIndicatorAttributes.m
ianegordon: LGTM.
Sorry about that, old habits from a previous that prohibited implicit casting.

ianegordon (approved):

ianegordon:

PR #2389, opened 2017-11-09
[Chips] Chip component

samnm:
This is the new Chips component. It's important to note that Chips+TextField integration will arrive at a later time.

There are two non-blocking bugs that can occur when Chips resize:
* Ink appears glitchy
* Shadows fade from one size to another


components/Chips/README.md
```diff
@@ -0,0 +1,124 @@
...
+## Usage
+
+### Importing
+
+Before using Chips, you'll need to import them:
+
+<!--<div class="material-code-render" markdown="1">-->
+#### Swift
+``` swift
+import MaterialComponents
```
romoore: Should this be `import MaterialComponents.MaterialChips`?

components/Chips/README.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+#### Objective-C
+
+``` objc
+#import "MaterialChips.h"
+```
+<!--</div>-->
+
+### Ink splash animation
```
romoore: On material.io, [the Ink component documentation](https://material.io/components/ios/catalog/ink/) refers to "ripples" and not "splashes." Can we use the same "ripple" name for consistency?

components/Chips/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Selected Image View
+In order to make it as clear as possible a chip has been selected, you can optionally set the image
+of the `selectedImageView`. This image will only appear when the chip is selected. If you have a
+image set on the standard `imageView`, then the `selectedImageView` will appear on top. Otherwise
+you'll need to resize the chip to show the selected image. See the Filter chip example to see this
+in action.
+
+### Padding
+There are 4 `padding` properties which control how a chip is laid out. One for each of the chip's
```
romoore, line 84: I think it may be worth expanding this section a bit more to talk about either (or both):
* How padding works within the chip.
* How padding differs from insets/outsets in UIKit.

components/Chips/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Examples
+
+### Create a single Chip
+
+<!--<div class="material-code-render" markdown="1">-->
+#### Swift
+``` swift
+let chipView = MDCChipView()
+chipView.customTitleColor = UIColor.gray
```
romoore: Is `customTitleColor` correct? If yes, how does it differ from `setTitleColor(for:)`?

components/Chips/README.md
```diff
@@ -0,0 +1,124 @@
...
+chipView.title = "Tap me"
+chipView.setTitleColor(UIColor.red, for: .selected)
```
romoore: Is there a default `titleColor` for the `.normal` state? Black?

components/Chips/examples/ChipsChoiceExampleViewController.m
```diff
@@ -0,0 +1,81 @@
...
+  [_collectionView registerClass:[MDCChipCollectionViewCell class]
+          forCellWithReuseIdentifier:@"Cell"];
+
+  [self.view addSubview:_collectionView];
+}
+
+- (void)viewWillLayoutSubviews {
+  [super viewWillLayoutSubviews];
+
+  CGFloat height = 33;
```
romoore: Nit: Remove `height` and just inline the `33` in the `CGRectMake` calls.

components/Chips/examples/ChipsChoiceExampleViewController.m
```diff
@@ -0,0 +1,81 @@
...
+
+- (__kindof UICollectionViewCell *)collectionView:(UICollectionView *)collectionView
+                           cellForItemAtIndexPath:(NSIndexPath *)indexPath {
+  MDCChipCollectionViewCell *cell =
+      [collectionView dequeueReusableCellWithReuseIdentifier:@"Cell" forIndexPath:indexPath];
+  cell.chipView.titleLabel.text = self.titles[indexPath.row];
+  return cell;
+}
+
+- (NSMutableArray *)titles {
```
romoore: `NSArray *` instead of `NSMutableArray *`?

components/Chips/examples/ChipsChoiceExampleViewController.m
```diff
@@ -0,0 +1,81 @@
...
+  if (!_titles) {
+    _titles = [NSMutableArray arrayWithArray:@[
+      @"The Bronx",
+      @"Brooklyn",
+      @"Manhattan",
+      @"Queens",
+      @"Staten Island",
+    ]];
+  }
+  return _titles;
```
romoore: And then return `[_titles copy]`?

components/Chips/examples/ChipsCollectionExampleViewController.m
```diff
@@ -0,0 +1,99 @@
...
+
+- (__kindof UICollectionViewCell *)collectionView:(UICollectionView *)collectionView
+                           cellForItemAtIndexPath:(NSIndexPath *)indexPath {
+  MDCChipCollectionViewCell *cell =
+      [collectionView dequeueReusableCellWithReuseIdentifier:@"Cell" forIndexPath:indexPath];
+  cell.chipView.titleLabel.text = self.titles[indexPath.row];
+  return cell;
+}
+
+- (NSMutableArray *)titles {
```
romoore: Same change to `NSArray *`

components/Chips/examples/ChipsCollectionExampleViewController.m
```diff
@@ -0,0 +1,99 @@
...
+      @"Farm-to-table",
+      @"XOXO",
+      @"Chillwave",
+      @"Fanny",
+      @"Pack",
+      @"Master",
+      @"Cleanse",
+      @"Small",
+      @"Batch",
+      @"Salvia",
```
romoore: [Salvia divinorum](https://en.wikipedia.org/wiki/Salvia_divinorum)? The psychotropic plant?

components/Chips/examples/ChipsFilterExampleViewController.m
```diff
@@ -0,0 +1,108 @@
...
+    didSelectItemAtIndexPath:(NSIndexPath *)indexPath {
+  [collectionView performBatchUpdates:nil completion:nil];
+}
+
+- (void)collectionView:(UICollectionView *)collectionView
+    didDeselectItemAtIndexPath:(NSIndexPath *)indexPath {
+  [collectionView performBatchUpdates:nil completion:nil];
+}
+
+- (NSMutableArray *)titles {
```
romoore: Same `NSArray *` suggestion

components/Chips/examples/ChipsSizingExampleViewController.m
```diff
@@ -0,0 +1,81 @@
...
+ limitations under the License.
+ */
+
+#import "ChipsExamplesSupplemental.h"
+
+#import "MaterialChips.h"
+
+@implementation ChipsSizingExampleViewController {
+  MDCChipView *_chipView;
+  UISlider *_widthSlider;
```
romoore: `MDCSlider` instead of `UISlider`?

components/Chips/examples/ChipsSizingExampleViewController.m
```diff
@@ -0,0 +1,81 @@
...
+  [super viewWillLayoutSubviews];
+
+  CGSize sliderSize = [_widthSlider sizeThatFits:self.view.bounds.size];
+  _widthSlider.frame = CGRectMake(20, 140, self.view.bounds.size.width - 40, sliderSize.height);
+  _heightSlider.frame =
+      CGRectMake(20, 140 + sliderSize.height + 20, self.view.bounds.size.width - 40, sliderSize.height);
+}
+
+- (void)widthSliderChanged:(UISlider *)slider {
+  CGRect frame = _chipView.frame;
```
romoore: Super-nit: Call `CGRectStandardize`?

```objectivec
CGRect frame = CGRectStandardize(_chipView.frame);
```

components/Chips/examples/ChipsSizingExampleViewController.m
```diff
@@ -0,0 +1,81 @@
...
+  frame.size.width = slider.value;
+  _chipView.frame = frame;
+  [_chipView layoutIfNeeded];
+}
+
+- (void)heightSliderChanged:(UISlider *)slider {
+  CGRect frame = _chipView.frame;
```
romoore: Same-nit: `CGRectStandardize`

components/Chips/examples/ChipsTypicalUseViewController.m
```diff
@@ -0,0 +1,106 @@
...
+}
+
+- (void)collectionView:(UICollectionView *)collectionView
+    didSelectItemAtIndexPath:(NSIndexPath *)indexPath {
+  [collectionView performBatchUpdates:nil completion:nil];
+}
+
+- (CGSize)collectionView:(UICollectionView *)collectionView
+                  layout:(UICollectionViewLayout*)collectionViewLayout
+  sizeForItemAtIndexPath:(NSIndexPath *)indexPath {
```
romoore, line 83: Nit/indentation: I think this should be 4 spaces from the left.

components/Chips/examples/supplemental/ChipsExamplesSupplemental.h
```diff
@@ -0,0 +1,80 @@
...
+
+@class MDCChipView;
+@class ChipModel;
+
+@interface ExampleChipCollectionViewController : UICollectionViewController
+@end
+
+@interface ChipsChoiceExampleViewController : UIViewController
+    <UICollectionViewDelegate, UICollectionViewDataSource>
+@property(nonatomic, strong) NSMutableArray<NSString *> *titles;
```
romoore: Change to `NSArray *`?  Here and below.
```objectivec
@property(nonatomic, copy) NSArray<NSString *> *titles;
```

components/Chips/examples/supplemental/ChipsExamplesSupplemental.m
```diff
@@ -0,0 +1,191 @@
...
+@interface ChipsChoiceExampleViewController (Supplemental)
+@end
+
+@implementation ChipsChoiceExampleViewController (Supplemental)
+
++ (NSArray *)catalogBreadcrumbs {
+  return @[ @"Chips", @"Choice" ];
+}
+
++ (NSString *)catalogDescription {
```
romoore: `catalogDescription` is only used on the primary demo.

components/Chips/examples/supplemental/ChipsExamplesSupplemental.m
```diff
@@ -0,0 +1,191 @@
...
+
+@interface ChipsChoiceExampleViewController (Supplemental)
+@end
+
+@implementation ChipsChoiceExampleViewController (Supplemental)
+
++ (NSArray *)catalogBreadcrumbs {
+  return @[ @"Chips", @"Choice" ];
+}
+
```
romoore, line 80: Each of the examples should have ~~`catalogIsPrimaryIsPrimaryDemo`~~ `catalogIsPrimaryDemo` defined.

components/Chips/src/MDCChipCollectionViewCell.h
```diff
@@ -0,0 +1,48 @@
...
+ this should be set to YES.
+ */
+@property(nonatomic, assign) BOOL alwaysAnimateCellResize;
+
+/*
+ Creates an MDCChipView for use in the collection view cell.
+
+ Override this method to return a custom MDCChipView subclass.
+ */
+- (nonnull MDCChipView *)createChipView;
```
romoore, line 46: Is there a reason to prefer subclass override instead of `initWith...chipView:`?

components/Chips/src/MDCChipCollectionViewCell.h
```diff
@@ -0,0 +1,48 @@
...
+ */
+@property(nonatomic, nonnull, strong, readonly) MDCChipView *chipView;
+
+/*
+ Animates the chip view every time the bounds change. Defaults to NO.
+
+ Set this to YES to animate chip view when it resizes. If the chip view changes size upon selection
+ this should be set to YES.
+ */
+@property(nonatomic, assign) BOOL alwaysAnimateCellResize;
```
romoore: `BOOL` property getter should be `isAlwaysAnimateCellResize`. 

"Naming is hard". Suggestion:
```objectivec
@property(nonatomic, assign, getter=isCellResizeAnimated) BOOL cellResizeAnimated;
```

components/Chips/src/MDCChipCollectionViewCell.m
```diff
@@ -0,0 +1,133 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import "MDCChipCollectionViewCell.h"
+
+#import "MaterialChips.h"
+
+#import "MDCChipView+Private.h"
```
romoore: Please explicitly use the `private/` subdirectory to support internal builds.

```objectivec
#import "private/MDCChipView+Private.h"
```

components/Chips/src/MDCChipCollectionViewCell.m
```diff
@@ -0,0 +1,133 @@
...
+
+@implementation MDCChipCollectionViewCell
+
+- (instancetype)init {
+  if (self = [super init]) {
```
romoore: I think this initializer should be removed. I believe neither `UICollectionViewCell` nor `UICollectionReusableView` define `init` and the designated initializer for `UIView` is `initWithFrame:`.

If there's a reason to keep it, please call the designated initializer instead.

components/Chips/src/MDCChipCollectionViewCell.m
```diff
@@ -0,0 +1,133 @@
...
+  if (self = [super initWithFrame:rect]) {
+    [self commonMDCChipCollectionViewCellInit];
+  }
+  return self;
+}
+
+- (void)setBounds:(CGRect)bounds {
+  [super setBounds:bounds];
+
+  _chipView.selected = self.selected;
```
romoore: Why do we need to update the `selected` state when the bounds changes?  Please add a comment.

components/Chips/src/MDCChipCollectionViewCell.m
```diff
@@ -0,0 +1,133 @@
...
+  BOOL animated = NO;
+  if (self.alwaysAnimateCellResize) {
+    BOOL isFrameNonzero = !CGRectIsEmpty(_chipView.frame) && !CGRectIsEmpty(self.bounds);
+    BOOL isFrameDifferent = !CGRectEqualToRect(_chipView.frame, self.bounds);
+
+    animated = isFrameNonzero && isFrameDifferent;
+  }
+
+  if (animated) {
+    [UIView animateWithDuration:0.25 animations:^{
```
romoore, line 70: Is there a Material Motion spec for the animations?

components/Chips/src/MDCChipCollectionViewCell.m
```diff
@@ -0,0 +1,133 @@
...
+
+- (CGSize)intrinsicContentSize {
+  return [_chipView intrinsicContentSize];
+}
+
+- (void)setSelected:(BOOL)selected {
+  [super setSelected:selected];
+
+  if (self.alwaysAnimateCellResize && [_chipView willChangeSizeWithSelectedValue:selected]) {
+    [self setNeedsLayout];
```
romoore: I think this also requires a comment. Am I correct in guessing that you're relying on a bounds change here to do a `selected` state change up in `setBounds`?

components/Chips/src/MDCChipCollectionViewCell.m
```diff
@@ -0,0 +1,133 @@
...
+  } else {
+    _chipView.selected = selected;
+  }
+}
+
+- (void)setHighlighted:(BOOL)highlighted {
```
romoore, line 103: Do we need to pass through the values for `UIControlStateFocused` and `UIControlStateEnabled`?

components/Chips/src/MDCChipCollectionViewFlowLayout.m
```diff
@@ -0,0 +1,51 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import "MDCChipCollectionViewFlowLayout.h"
+
+static inline CGRect CGRectAlignToRect(CGRect rect, CGRect staticRect, CGFloat padding) {
```
romoore: Please document this method. I don't really understand what the goal is. Some unit tests might help as well.

components/Chips/src/MDCChipCollectionViewFlowLayout.m
```diff
@@ -0,0 +1,51 @@
...
+    UICollectionViewLayoutAttributes *attrs = [layoutAttributes[i] copy];
+    UICollectionViewLayoutAttributes *prevAttrs = customLayoutAttributes[i - 1];
+
+    if (CGRectGetMinY(prevAttrs.frame) == CGRectGetMinY(attrs.frame)) {
+      attrs.frame = CGRectAlignToRect(attrs.frame, prevAttrs.frame, self.minimumInteritemSpacing);
+    }
+    [customLayoutAttributes addObject:attrs];
+  }
+
+  return customLayoutAttributes;
```
romoore: Return an `NSArray *`,  `[customLayoutAttributes copy]`.

components/Chips/src/MDCChipCollectionViewFlowLayout.m
```diff
@@ -0,0 +1,51 @@
...
+ limitations under the License.
+ */
+
+#import "MDCChipCollectionViewFlowLayout.h"
+
+static inline CGRect CGRectAlignToRect(CGRect rect, CGRect staticRect, CGFloat padding) {
+  return CGRectMake(CGRectGetMaxX(staticRect) + padding,
+                    CGRectGetMinY(rect),
+                    rect.size.width,
+                    rect.size.height);
```
romoore: Please use CGRectGet* methods here.

components/Chips/src/MDCChipCollectionViewFlowLayout.m
```diff
@@ -0,0 +1,51 @@
...
+
+@implementation MDCChipCollectionViewFlowLayout
+
+- (NSArray<UICollectionViewLayoutAttributes *> *)layoutAttributesForElementsInRect:(CGRect)rect {
+  NSArray *layoutAttributes = [super layoutAttributesForElementsInRect:rect];
+  NSMutableArray *customLayoutAttributes =
+      [NSMutableArray arrayWithCapacity:layoutAttributes.count];
+
+  if (layoutAttributes.count > 0) {
+    UICollectionViewLayoutAttributes *attrs = layoutAttributes[0];
```
romoore: Suggestion, remove the local variable by changing to:
```objectivec
[customLayoutAttributes addObject:layoutAttributes.firstObject]
```

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,204 @@
...
+ A UIImageView that leads the title label.
+ */
+@property(nonatomic, readonly, nonnull) IBInspectable UIImageView *imageView;
+
+/*
+ A UIImageView that leads the title label. Appears in front of the imageView. Only visible when the
+ chip is selected.
+
+ This image view is typically used to show some icon that denotes the chip as selected, such as a
+ tick. If imageView has no image then the chip will require resizing when selected or deselected to
```
romoore: s/tick/check mark ?

When I read this my brain first showed me the blood-sucking animal. 😝

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,204 @@
...
+ UIContentSizeCategory is changed.
+
+ This property is modeled after the adjustsFontForContentSizeCategory property in the
+ UIContentSizeCategoryAdjusting protocol added by Apple in iOS 10.0.
+
+ If set to YES, this button will base its text font on MDCFontTextStyleButton.
+
+ Default value is NO.
+ */
+@property(nonatomic, readwrite, setter=mdc_setAdjustsFontForContentSizeCategory:)
```
romoore, line 115: Do we want (and are we able) to conform to [UIContentSizeCategoryAdjusting](https://developer.apple.com/documentation/uikit/uicontentsizecategoryadjusting?language=objc)? UIControls don't seem to conform, but should the chip?

components/Chips/src/MDCChipView.h
```diff
@@ -0,0 +1,204 @@
...
+- (nullable UIColor *)backgroundColorForState:(UIControlState)state;
+
+/*
+ A color used as the chip's @c backgroundColor.
+
+ Defaults to blue.
+
+ @param backgroundColor The background color.
+ @param state The control state.
+ */
```
romoore, line 136: Nit/Docs: I think it would be good to document the "fall-back" behavior here. In the same way that UIButton does.

https://developer.apple.com/documentation/uikit/uibutton/1624018-settitle?language=objc

components/Chips/src/private/MDCChipView+Private.h
```diff
@@ -0,0 +1,25 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import "MDCChipView.h"
+
+@interface MDCChipView (Private)
```
romoore, line 19: Do we want to make this a class-continuation category to ensure the compiler checks for missing definitions?

components/Chips/src/MDCChipView.m
```diff
@@ -0,0 +1,556 @@
...
+
+#import "MaterialInk.h"
+#import "MaterialMath.h"
+#import "MaterialShadowLayer.h"
+#import "MaterialShadowElevations.h"
+#import "MaterialTypography.h"
+
+#define RGBCOLOR(r, g, b) \
+    [UIColor colorWithRed:(r) / 255.0f green:(g) / 255.0f blue:(b) / 255.0f alpha:1]
+#define HEXCOLOR(hex) RGBCOLOR((((hex) >> 16) & 0xFF), (((hex) >> 8) & 0xFF), ((hex)&0xFF))
```
romoore: Please convert these macros to static inline functions. [We recently converted many such macros to functions](https://github.com/material-components/material-components-ios/issues/1682)

components/Chips/src/MDCChipView.m
```diff
@@ -0,0 +1,556 @@
...
+
+static const NSInteger MDCChipBackgroundColor = 0xEBEBEB;
```
romoore: Can we use palette colors for these defaults?

components/Chips/src/MDCChipView.m
```diff
@@ -0,0 +1,556 @@
...
+
+@interface UIColor (MDCChipView)
+- (UIColor *)darken:(CGFloat)percent;
+- (UIColor *)lighten:(CGFloat)percent;
+@end
+
+static CGRect CGRectVerticallyCentered(CGRect rect, UIEdgeInsets padding, CGFloat height) {
+  CGFloat viewHeight = CGRectGetHeight(rect) + padding.top + padding.bottom;
+  CGRect frame = CGRectOffset(rect, 0, (height - viewHeight) / 2);
+  return MDCRectAlignToScale(frame, [UIScreen mainScreen].scale);
```
romoore, line 81: Pass in the scale? Otherwise this may misalign on secondary screens where the scale differs from the main screen scale.

components/Chips/src/MDCChipView.m
```diff
@@ -0,0 +1,556 @@
...
+}
+
+static inline CGRect MDCChipBuildFrame(UIEdgeInsets insets,
```
romoore, line 84: Please use uioffset since you only want the vertical/horizontal. This helps avoid RTL/LTR issues in the function.

components/Chips/src/MDCChipView.m
```diff
@@ -0,0 +1,556 @@
...
+static inline CGSize CGSizeShrinkWithInsets(CGSize size, UIEdgeInsets edgeInsets) {
+  return CGSizeMake(size.width - UIEdgeInsetsHorizontal(edgeInsets),
+                    size.height - UIEdgeInsetsVertical(edgeInsets));
+}
+
+@interface MDCChipView ()
+@property(nonatomic, readonly) CGRect contentRect;
+@property(nonatomic, readonly, strong) MDCShadowLayer *layer;
+@property(nonatomic, readonly) BOOL showImageView;
+@property(nonatomic, readonly) BOOL showSelectedImageView;
```
romoore, line 114: Please add getters for these `BOOL`s.  Naming would be easier if they followed the `hidden` pattern in UIKit.

```objective
@property(nonatomic, readonly, getter=isShowImageView) BOOL showImageView;
@property(nonatomic, readonly, getter=isShowSelectedImageView) BOOL showSelectedImageView;
```

components/Chips/src/MDCChipView.m
```diff
@@ -0,0 +1,556 @@
...
+}
+
+@dynamic layer;
+
++ (Class)layerClass {
+  return [MDCShadowLayer class];
+}
+
+- (instancetype)init {
+  if (self = [super init]) {
```
romoore: Please consider removing or else please call UIVIew's `initWithFrame:` as the designated initializer.

components/Chips/src/MDCChipView.m
```diff
@@ -0,0 +1,556 @@
...
+  return self;
+}
+
+- (instancetype)initWithFrame:(CGRect)frame {
+  if (self = [super initWithFrame:frame]) {
+    [self commonMDCChipViewInit];
+  }
+  return self;
+}
+
```
romoore: `initWithCoder:`?

components/Chips/src/MDCChipView.m
```diff
@@ -0,0 +1,556 @@
...
+#pragma mark - Layout
+
+- (void)layoutSubviews {
+  _inkView.frame = self.bounds;
+  _imageView.frame = [self imageViewFrame];
+  _selectedImageView.frame = [self selectedImageViewFrame];
+  _accessoryView.frame = [self accessoryViewFrame];
+  _titleLabel.frame = [self titleLabelFrame];
+
+  _selectedImageView.alpha = self.showSelectedImageView ? 1 : 0;
```
romoore, line 432: Is there a reason to use `.alpha` instead of `.hidden`?

components/Chips/src/MDCChipView.m
```diff
@@ -0,0 +1,556 @@
...
+}
+
+#pragma mark - Layout
+
+- (void)layoutSubviews {
+  _inkView.frame = self.bounds;
+  _imageView.frame = [self imageViewFrame];
+  _selectedImageView.frame = [self selectedImageViewFrame];
+  _accessoryView.frame = [self accessoryViewFrame];
+  _titleLabel.frame = [self titleLabelFrame];
```
romoore, line 430: Please use property accessors unless there's a need to access the backing iVar here.

components/Chips/src/MDCChipView.m
```diff
@@ -0,0 +1,556 @@
...
+- (CGRect)imageViewFrame {
+  return [self frameForImageView:self.imageView visible:self.showImageView];
+}
+
+- (CGRect)selectedImageViewFrame {
+  return [self frameForImageView:self.selectedImageView visible:self.showSelectedImageView];
+}
+
+- (CGRect)frameForImageView:(UIImageView *)imageView visible:(BOOL)visible {
+  CGRect frame = CGRectMake(self.contentRect.origin.x, CGRectGetMidY(self.contentRect), 0, 0);
```
romoore: Please use CGRectGetMinX(self.contentRect) if possible.

romoore (changes requested):

components/Chips/src/MDCChipView.h
ianegordon: UIContentSizeCategoryAdjusting was added in iOS 10.
Original plan (for all Dynamic Type supporting components) was to re-implement as mdc_setAdjustForXYZ.  When we support iOS 10 as a minimum, we can explicitly add the protocol, rename the method, add a deprecated shim method with the old name.

components/Chips/src/MDCChipCollectionViewCell.h
samnm: `BOOL` property getter should only start with `is` "if the name of a declared property is expressed as an adjective".

https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingIvarsAndTypes.html#//apple_ref/doc/uid/20001284-BAJGIIJE

IMO alwaysAnimateCellResize is clearer than cellResizeAnimated.

components/Chips/src/MDCChipCollectionViewCell.h
samnm: UICollectionViews control cell initialization so we can't prescribe a custom initializer.

components/Chips/src/MDCChipCollectionViewCell.m
samnm: No,  UICollectionViewCell is not a UIControl. We only pass through the common API.

components/Chips/src/private/MDCChipView+Private.h
samnm: Class-continuation categories behave differently to named categories?

romoore: As far as I know, whatever is declared in a class-continuation category has to be defined in the main implementation. If it's defined in a named category, then you can fail to define it and the compiler won't complain.  It's very possible I've just missed the compiler warnings/errors that it emits in those cases.

For example, when I define a (Testing) category that declares some method, if I delete that method from the implementation the compiler won't warn me. The runtime just crashes when the selector is called.

components/Chips/src/MDCChipCollectionViewCell.m
romoore: Ack. Thanks!

components/Chips/src/MDCChipCollectionViewCell.h
romoore: Got it.

components/Chips/src/MDCChipCollectionViewCell.h
romoore: Good reference link, thanks!

components/Chips/src/private/MDCChipView+Private.h
samnm: All of the methods are implemented inside of `@implementation MDCChipView (Private)` and removing any method implementation causes a compilation error.

components/Chips/src/MDCChipView.m
samnm: This function is specific to layout in MDCChipView. Using UIOffset would only complicate it IMO. It specifically maps the `xPadding` values to layout. The other parts of the UIEdgeInsets are used elsewhere.

We may either need to document that UIEdgeInsets.left is leading and .right is trailing, but this implementation works very well with RTL.

components/Chips/src/MDCChipView.m
samnm: IMO these properties have the correct setter.

Sharing the relevant link again: https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingIvarsAndTypes.html#//apple_ref/doc/uid/20001284-BAJGIIJE

components/Chips/src/MDCChipView.m
samnm: We are animating the alpha.

components/Chips/src/MDCChipView.m
samnm: Why?

components/Chips/src/MDCChipView.m
samnm: Done.

components/Chips/README.md
samnm: `padding` seems pretty self explanatory to me but if you have any concrete suggestions I'm happy to add them.

components/Chips/examples/supplemental/ChipsExamplesSupplemental.m
samnm: Done.

components/Chips/src/MDCChipCollectionViewCell.m
samnm: No.

components/Chips/src/MDCChipView.m
samnm: I'll open a bug after submission. There would be added complexity around making sure sizing cells work correctly.

components/Chips/src/MDCChipView.h
samnm: Done.

samnm:
All comments should have been addressed. @romoore @ianegordon PTAL.

components/Chips/examples/ChipsFilterExampleViewController.m
```diff
@@ -0,0 +1,108 @@
...
+  _collectionView.dataSource = self;
+  _collectionView.delegate = self;
+  _collectionView.allowsMultipleSelection = YES;
+  _collectionView.backgroundColor = [UIColor whiteColor];
+  _collectionView.delaysContentTouches = NO;
+  _collectionView.contentInset = UIEdgeInsetsMake(4, 8, 4, 8);
+  [_collectionView registerClass:[MDCChipCollectionViewCell class]
+      forCellWithReuseIdentifier:@"Cell"];
+
+  _sizingChip = [[MDCChipView alloc] init];
```
ianegordon, line 42: Prefer designated initializer, initWithFrame:CGRectZero.

components/Chips/src/MDCChipCollectionViewCell.h
ianegordon: Split the difference, resizeAnimated?  (getter=isResizeAnimated)
'Cell' is redundant and 'always' is implicit.

Naming is definitely hard.

components/Chips/src/MDCChipCollectionViewCell.m
```diff
@@ -0,0 +1,130 @@
...
+#import "private/MDCChipView+Private.h"
+
+@implementation MDCChipCollectionViewCell
+
+- (instancetype)initWithFrame:(CGRect)rect {
+  if (self = [super initWithFrame:rect]) {
+    [self commonMDCChipCollectionViewCellInit];
+  }
+  return self;
+}
```
ianegordon: Add initWithCoder.

components/Chips/src/MDCChipCollectionViewCell.m
```diff
@@ -0,0 +1,130 @@
...
+
```
ianegordon: Consider moving commonXYZInit directly under the init methods.

components/Chips/src/MDCChipCollectionViewCell.m
```diff
@@ -0,0 +1,130 @@
...
+  }
+}
+
+- (void)commonMDCChipCollectionViewCellInit {
+  _chipView = [self createChipView];
+  [self.contentView addSubview:_chipView];
+}
+
+- (MDCChipView *)createChipView {
+  return [[MDCChipView alloc] init];
```
ianegordon: Use designated initializer.

components/Chips/src/MDCChipCollectionViewCell.m
```diff
@@ -0,0 +1,130 @@
...
+
+- (void)setSelected:(BOOL)selected {
+  [super setSelected:selected];
+
+  if (self.alwaysAnimateCellResize && [_chipView willChangeSizeWithSelectedValue:selected]) {
+    // Since MDCChipView can resize when it is selected, if a resize will occur we need to delay our
+    // selected state update until our next layout.
+    [self setNeedsLayout];
+  } else {
+    _chipView.selected = selected;
```
ianegordon, line 99: Do we need to setSelected in both cases / move outside the if clause?

components/Chips/src/MDCChipView.m
```diff
@@ -0,0 +1,626 @@
...
+
+// Creates a UIColor from a 24-bit RGB color encoded as an integer.
+static inline UIColor *MDCColorFromRGB(uint32_t rgbValue) {
+  return [UIColor colorWithRed:((CGFloat)((rgbValue & 0xFF0000) >> 16)) / 255
+                         green:((CGFloat)((rgbValue & 0x00FF00) >> 8)) / 255
+                          blue:((CGFloat)((rgbValue & 0x0000FF) >> 0)) / 255
+                         alpha:1];
+}
+
+static const uint32_t MDCChipBackgroundColor = 0xEBEBEB;
```
ianegordon: If possible, pull this from Palette.

ianegordon: If possible, pull this from Palette.

components/Chips/src/MDCChipView.m
```diff
@@ -0,0 +1,626 @@
...
+
+@dynamic layer;
+
++ (Class)layerClass {
+  return [MDCShadowLayer class];
+}
+
+- (instancetype)init {
+  return [self initWithFrame:CGRectZero];
+}
```
ianegordon: Is this the default implementation of UIControl?
I wouldn't expect to see anything other than the designated initializers overwritten.

components/Chips/src/MDCChipView.m
```diff
@@ -0,0 +1,626 @@
...
+      _backgroundColors[@(UIControlStateSelected)] = selected;
+    }
+    _borderColors = [NSMutableDictionary dictionary];
+    _borderWidths = [NSMutableDictionary dictionary];
+
+    _elevations = [NSMutableDictionary dictionary];
+    _elevations[@(UIControlStateNormal)] = @(0);
+    _elevations[@(UIControlStateHighlighted)] = @(MDCShadowElevationRaisedButtonPressed);
+    _elevations[@(UIControlStateHighlighted | UIControlStateSelected)] =
+    @(MDCShadowElevationRaisedButtonPressed);
```
ianegordon: indent

components/Chips/src/MDCChipView.m
```diff
@@ -0,0 +1,626 @@
...
+
+- (UIColor *)inkColor {
+  return _inkView.inkColor;
+}
+
+- (void)mdc_setAdjustsFontForContentSizeCategory:(BOOL)adjusts {
+  _mdc_adjustsFontForContentSizeCategory = adjusts;
+  if (_mdc_adjustsFontForContentSizeCategory) {
+    UIFont *font = [UIFont mdc_preferredFontForMaterialTextStyle:MDCFontTextStyleButton];
+    self.titleLabel.font = font;
```
ianegordon, line 265: sizeToFit?
setNeedsLayout?

components/Chips/src/MDCChipView.m
```diff
@@ -0,0 +1,626 @@
...
+  self.layer.cornerRadius = cornerRadius;
+  self.layer.shadowPath =
+      [UIBezierPath bezierPathWithRoundedRect:self.bounds cornerRadius:cornerRadius].CGPath;
+}
+
+- (CGRect)contentRect {
+  return CGRectMake(self.contentPadding.left,
+                    self.contentPadding.top,
+                    self.bounds.size.width - UIEdgeInsetsHorizontal(self.contentPadding),
+                    self.bounds.size.height - UIEdgeInsetsVertical(self.contentPadding));
```
ianegordon: Is this method not just UIInsetsInsetRect() ?

components/Chips/src/MDCChipView.m
```diff
@@ -0,0 +1,626 @@
...
+    frame = MDCChipBuildFrame(_imagePadding,
+                              selectedSize,
+                              CGRectGetMinX(self.contentRect),
+                              CGRectGetHeight(self.frame));
+  }
+  return frame;
+}
+
+- (CGRect)accessoryViewFrame {
+  if (!_accessoryView) {
```
ianegordon: Or if _accessoryView.hidden ?

ianegordon (changes requested):

components/Chips/examples/ChipsSizingExampleViewController.m
romoore: Since the frame rect may be non-standard, setting the width may cause it to shift locations.
For (extreme) example, 
```objectivec
CGRect frame = _chipView.frame; // (95, 95, -80, -80) == (15, 15, 80, 80)
frame.size.width = slider.value; // (95, 95, 100, -80) == (95, 15, 100, 80)
_chipView.frame = frame; // Has now moved 80 points in the X-axis
```

components/Chips/examples/supplemental/ChipsExamplesSupplemental.m
```diff
@@ -28,15 +28,15 @@
 
 static UIImage *FaceImage() {
```
romoore: Rename to MaskImage ?

components/Chips/src/MDCChipCollectionViewFlowLayout.m
```diff
@@ -31,7 +39,8 @@ @implementation MDCChipCollectionViewFlowLayout
       [NSMutableArray arrayWithCapacity:layoutAttributes.count];
 
   if (layoutAttributes.count > 0) {
-    UICollectionViewLayoutAttributes *attrs = layoutAttributes[0];
+    UICollectionViewLayoutAttributes *attrs = [layoutAttributes[0] copy];
+    attrs.frame = CGRectLeftAlign(attrs.frame);
```
romoore: If we're laying-out the collection for RTL, do we need to swap the alignment? I would expect the order of chips to be reversed in the UI but the same in the underlying model.

components/Chips/src/MDCChipCollectionViewFlowLayout.m
```diff
@@ -40,12 +49,15 @@ @implementation MDCChipCollectionViewFlowLayout
     UICollectionViewLayoutAttributes *prevAttrs = customLayoutAttributes[i - 1];
 
     if (CGRectGetMinY(prevAttrs.frame) == CGRectGetMinY(attrs.frame)) {
-      attrs.frame = CGRectAlignToRect(attrs.frame, prevAttrs.frame, self.minimumInteritemSpacing);
+      attrs.frame =
+          CGRectLeftAlignToRect(attrs.frame, prevAttrs.frame, self.minimumInteritemSpacing);
```
romoore: Same RTL question.

components/Chips/src/MDCChipCollectionViewFlowLayout.m
romoore: Do you need to reset the `origin.x` to zero even when the first element in the `rect` would not be the first element in the collection?

components/Chips/src/MDCChipView.m
```diff
@@ -99,67 +119,97 @@ + (Class)layerClass {
...
-- (void)commonMDCChipViewInit {
-  if (!_backgroundColors) {
-    // _backgroundColors may have already been initialized by setting the backgroundColor setter.
-    UIColor *normal = HEXCOLOR(MDCChipBackgroundColor);
-    UIColor *disabled = [normal lighten:MDCChipDisabledLightenPercent];
-    UIColor *selected = [normal darken:MDCChipSelectedDarkenPercent];
+- (instancetype)initWithCoder:(NSCoder *)aDecoder {
+  if (self = [super initWithCoder:aDecoder]) {
+    _imageView = [aDecoder decodeObjectForKey:MDCChipImageViewKey];
+    _selectedImageView = [aDecoder decodeObjectForKey:MDCChipSelectedImageViewKey];
```
romoore: Do you want to check for the key to exist before overwriting the variable?

components/Chips/src/MDCChipView.m
```diff
@@ -99,67 +119,97 @@ + (Class)layerClass {
...
 - (instancetype)init {
-  if (self = [super init]) {
-    [self commonMDCChipViewInit];
-  }
-  return self;
+  return [self initWithFrame:CGRectZero];
 }
 
 - (instancetype)initWithFrame:(CGRect)frame {
   if (self = [super initWithFrame:frame]) {
```
romoore: Nit: Although this is absolutely Apple's idiosyncratic approach to initializers, most of the code in MDC-iOS follows the internal 2-line approach:

```objectivec
if (self) {
```

romoore (changes requested):
A few more comments.

components/Chips/README.md
romoore: For me, the fact that it's called "padding" and not "insets" implies that it behaves differently from the typical `.insets` properties on UIView types.  For example, will setting the padding after setting the frame cause the chip to expand outward (grow larger) or the internal subviews to become smaller and possibly clip the content?

With `UIButton.contentEdgeInsets`, the insets cause the content to be moved AFTER being positioned within the center of the button. Is it the same for padding? Does negative padding cause the content to shrink or to be moved in the opposite direction?

components/Chips/examples/ChipsFilterExampleViewController.m
samnm: Why? Surely initializing a view with init is common enough that they are interchangeable.

components/Chips/src/MDCChipCollectionViewCell.m
samnm: No, the the first case `selected` is updated inside `layoutSubviews`. Does the comment need clarification?

components/Chips/examples/ChipsFilterExampleViewController.m
ianegordon: One would hope, but we can't assume :/
Check out the following:
https://blog.twitter.com/engineering/en_us/a/2014/how-to-objective-c-initializer-patterns.html

samnm: Thats sorta my point. We can't assume clients will always use initWithFrame so shouldn't we explicitly support views initializing with init?

samnm: That doc seems to imply that UIView init is guaranteed to call UIView initWithFrame.

components/Chips/src/MDCChipCollectionViewCell.h
samnm: I renamed it to alwaysAnimateResize. IMO 'always' is not implicit (see alwaysBounceHorizontal on UIScrollView).

components/Chips/src/MDCChipView.m
samnm: sizeToFit is called in contentSizeCategoryDidChange. This is identical to MDCButton - happy to fix if its wrong I can't say I know exactly what should happen here.

components/Chips/src/MDCChipView.m
samnm: It is! Changed to use UIEdgeInsetsInsetRect.

components/Chips/src/MDCChipView.m
samnm: Done.

components/Chips/src/MDCChipView.m
samnm: Added todo.

components/Chips/README.md
samnm: Added a note making that behavior super explicit.

samnm:
All the comments (bar the init/initWithFrame comments) have been resolved. PTAL.

components/Chips/examples/ChipsFilterExampleViewController.m
ianegordon: Convinced.

In an init method, require a designated initializer from the *immediate* super class.  
Outside of init methods, call a designated initializer from *any* super class.

components/Chips/src/MDCChipCollectionViewCell.m
ianegordon: Comment explains.
I was comparing with the method below, which is almost, but not quite the same as this.
Disregard.

components/Chips/src/MDCChipView.m
ianegordon: setNeedsLayout is the one.

ianegordon (approved):

ianegordon:
Please merge when Travis @ @romoore are happy.

romoore (approved):

PR #2392, opened 2017-11-10
[BottomSheet] Safe Area fixes

### LINKED ISSUE #2005: [BottomSheet] Support Safe Area
### Overview
Issues visible in the Catalog: Full screen is unusable because of the top notch. Potential issues with the Home indicator and the content. On portrait, the content doesn’t adhere to the safe area.

bottomsheetlandscape

bottomsheetfullscreen

### Version number
35.0.0

iOS 11 / iPhoneX

chuga:
- Adjusts the max height to prevent the content from going below the top safe area inset.
- Sets the scroll view's bottom inset to at least  the bottom safe area inset.
- Stops ignoring the contentInset when calculating the content size of the scroll view.
- Adjusts preferredSheetHeight to include the bottom safe area inset when used internally.

Closes #2005

### DIFF
--- a/components/BottomSheet/src/private/MDCSheetContainerView.m
+++ b/components/BottomSheet/src/private/MDCSheetContainerView.m
@@ -166,7 +166,7 @@ static const CGFloat kSheetBounceBuffer = 150.0f;
 
     CGRect scrollViewFrame = CGRectStandardize(self.sheet.scrollView.frame);
     scrollViewFrame.size = CGSizeMake(scrollViewFrame.size.width,
-                                      self.frame.size.height - self.safeAreaInsets.top);
+                                      CGRectGetHeight(self.frame) - self.safeAreaInsets.top);
     self.sheet.scrollView.frame = scrollViewFrame;
   }
 #endif
--- a/components/BottomSheet/src/private/MDCSheetContainerView.m
+++ b/components/BottomSheet/src/private/MDCSheetContainerView.m
@@ -164,7 +164,7 @@ static const CGFloat kSheetBounceBuffer = 150.0f;
     contentInset.bottom = MAX(contentInset.bottom, self.safeAreaInsets.bottom);
     self.sheet.scrollView.contentInset = contentInset;
 
-    CGRect scrollViewFrame = self.sheet.scrollView.frame;
+    CGRect scrollViewFrame = CGRectStandardize(self.sheet.scrollView.frame);
     scrollViewFrame.size = CGSizeMake(scrollViewFrame.size.width,
                                       self.frame.size.height - self.safeAreaInsets.top);
     self.sheet.scrollView.frame = scrollViewFrame;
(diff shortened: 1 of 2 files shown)


samnm (approved):
🎉💯🔥

chuga:
🙌

components/BottomSheet/src/private/MDCSheetContainerView.m
```diff
@@ -215,14 +249,20 @@ - (CGFloat)truncatedPreferredSheetHeight {
 
 // Returns the maximum allowable height that the sheet can be dragged to.
 - (CGFloat)maximumSheetHeight {
-  CGSize size = self.bounds.size;
-  CGFloat scrollViewContentHeight = self.sheet.scrollView.contentSize.height;
+  CGFloat boundsHeight = self.bounds.size.height;
```
ianegordon: CGRectGetHeight() since bounds may not be standardized.

ianegordon (changes requested):

components/BottomSheet/src/private/MDCSheetContainerView.m
chuga: Done (we were already using self.bounds.size.height fwiw).
PTAL

components/BottomSheet/src/private/MDCSheetContainerView.m
```diff
@@ -137,6 +145,25 @@ - (void)traitCollectionDidChange:(UITraitCollection *)previousTraitCollection {
...
+  if (@available(iOS 11.0, *)) {
+    [super safeAreaInsetsDidChange];
+
+    _preferredSheetHeight = self.originalPreferredSheetHeight + self.safeAreaInsets.bottom;
+
+    UIEdgeInsets contentInset = self.sheet.scrollView.contentInset;
+    contentInset.bottom = MAX(contentInset.bottom, self.safeAreaInsets.bottom);
+    self.sheet.scrollView.contentInset = contentInset;
+
+    CGRect scrollViewFrame = self.sheet.scrollView.frame;
```
ianegordon: Standardize before accessing the struct directly.
CGRect scrollViewFrame = CGRectStandardize(self.sheet.scrollView.frame);

ianegordon (changes requested):

components/BottomSheet/src/private/MDCSheetContainerView.m
chuga: What's the reasoning behind that?

ianegordon: We are directly accessing the struct directly below.
To ensure we are getting valid value out of a bounds / frame, we can either use the accessor functions like CGRectGetWidth() / Height() / etc. -OR- we can CGRectStandardize() the frame and access the standardized struct values directly.

Note that Swift eliminates all of this by only returning standardized values.  Our bright, safe Swifty future awaits! 😄

components/BottomSheet/src/private/MDCSheetContainerView.m
ianegordon: Indeed, we definitely haven't been as strict as we should have been in the past, but if we are touching the code.....

chuga:
RIP my place in the Travis queue 😔
PTAL

ianegordon (approved):

PR #2409, opened 2017-11-13
[Slider] Correct colors for slider default light color scheme

jmdetloff:


ianegordon (changes requested):
Could you drop in a Before / After screenshot?
This will likely break a lot of internal tests, so we want to make sure we have the right rendering.

randallli (approved):

randallli:
This only landed in the previous release with the incorrect values. But a screenshot would be nice.

ianegordon (approved):

ianegordon:
Fix for original import.

PR #2412, opened 2017-11-13
[LibraryInfo] Add new LibraryInfo component (attempt #2)

ajsecord:
This is identical to #2401, my git history got badly munged, so this is just a fresh copy.

### DIFF
--- a/components/LibraryInfo/tests/unit/LibraryInfoTests.swift
+++ b/components/LibraryInfo/tests/unit/LibraryInfoTests.swift
@@ -1,0 +1,39 @@
+/*
+ Copyright 2017-present the Material Components for iOS authors. All Rights Reserved.
+ 
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+ 
+ http://www.apache.org/licenses/LICENSE-2.0
+ 
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


jverkoey (approved):

ianegordon:

components/LibraryInfo/tests/unit/LibraryInfoTests.swift
```diff
@@ -0,0 +1,39 @@
...
+  func testVersionFormat() {
+    // Given
+
+    // This regex pattern does the following:
+    // Accept: "42.0.0", etc.
+    // Reject: "0.0.0", "1.2", "1", "-1.2.3", "Hi, I'm a version 1.2.3", "1.2.3 is my version", etc.
+    //
+    // Note the major version must be >= 1 since "0.0.0" is used as the version when something goes
+    // wrong in the underlying code and we're in Release mode.
+    let pattern = "^[1-9]+\\.[0-9]+\\.[0-9]+$"
```
ianegordon, line 31: Is this string intended to be Semantic Version compliant.  If so, we may have 1.2.3-alpha, 4.5.6-rc, 5.4.3-rc.99.

ianegordon (approved):
Please see note about semantic versioning.  If it doesn't apply, we are set.

components/LibraryInfo/tests/unit/LibraryInfoTests.swift
ajsecord: Oooh, or 1.2.3-alpha+buildnum, too. 

No, this is intended to test that we're returning our API in the header, which just specifies "X.Y.Z". If we change that then we can update this test.

PR #2422, opened 2017-11-14
[Tabs] Remove incorrect layout optimization

brianjmoore:
The optimization where we skipped updating the layout for certain alignment modes was invalid and caused many bugs. Removing it fixes those bugs and doesn't seem to have much of a downside given we already don't update unless the bounds changes. I have verified that none of the tab bar examples are broken after this change.

This also removes a state variable that was not used at all.


ianegordon (approved):

PR #2429, opened 2017-11-15
Replace unnecessary imports with forward declarations

sigito:
Partially addresses #1523.

### DIFF
--- a/components/ActivityIndicator/examples/ActivityIndicatorExample.m
+++ b/components/ActivityIndicator/examples/ActivityIndicatorExample.m
@@ -16,10 +16,9 @@
 
 #import <UIKit/UIKit.h>
 
-#import "ActivityIndicatorExampleSupplemental.h"
-
 #import "MaterialActivityIndicator.h"
 #import "MaterialPalettes.h"
+#import "supplemental/ActivityIndicatorExampleSupplemental.h"
 
 #define MDC_CATALOG_BLACK [UIColor colorWithWhite:0.1f alpha:1]
 #define MDC_CATALOG_GREY  [UIColor colorWithWhite:0.9f alpha:1]
--- a/components/ActivityIndicator/examples/supplemental/ActivityIndicatorExampleSupplemental.m
+++ b/components/ActivityIndicator/examples/supplemental/ActivityIndicatorExampleSupplemental.m
@@ -19,9 +19,8 @@
  instructions. It is not necessary to import this file to use Material Components for iOS.
  */
 
-#import <Foundation/Foundation.h>
-
 #import "ActivityIndicatorExampleSupplemental.h"
+
 #import "MaterialTypography.h"
 
 #define MDC_CATALOG_GREEN [UIColor colorWithRed:0 green:0xe6/255.0f blue:0x76/255.0f alpha:1]
--- a/components/AnimationTiming/examples/supplemental/AnimationTimingExampleSupplemental.m
+++ b/components/AnimationTiming/examples/supplemental/AnimationTimingExampleSupplemental.m
@@ -14,9 +14,8 @@
  limitations under the License.
  */
 
-#import <Foundation/Foundation.h>
-
 #import "AnimationTimingExampleSupplemental.h"
+
 #import "MaterialTypography.h"
 
 const CGFloat kTopMargin = 16.f;
--- a/components/Buttons/examples/supplemental/ButtonsTypicalUseSupplemental.m
+++ b/components/Buttons/examples/supplemental/ButtonsTypicalUseSupplemental.m
@@ -19,9 +19,8 @@
  instructions. It is not necessary to import this file to use Material Components for iOS.
  */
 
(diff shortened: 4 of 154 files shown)


willlarche:
Running extra tests internally. Then it will be unblocked.

components/ActivityIndicator/examples/ActivityIndicatorExample.m
```diff
@@ -16,10 +16,9 @@
 
 #import <UIKit/UIKit.h>
 
-#import "ActivityIndicatorExampleSupplemental.h"
-
 #import "MaterialActivityIndicator.h"
 #import "MaterialPalettes.h"
+#import "supplemental/ActivityIndicatorExampleSupplemental.h"
```
randallli, line 21: nice! the supplementa/ is required for bazel build.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.h
```diff
@@ -16,7 +16,7 @@
 
 #import <UIKit/UIKit.h>
 
-#import "MaterialFeatureHighlight.h"
+#import "MDCFeatureHighlightView.h"
```
randallli, line 19: I think this is an incorrect transform.

components/FlexibleHeader/examples/supplemental/FlexibleHeaderWrappedSupplemental.m
```diff
@@ -21,8 +21,6 @@
 
 #import "FlexibleHeaderWrappedSupplemental.h"
 
-#import "MaterialFlexibleHeader.h"
```
randallli, line 24: interesting. how did you detect that the component wasn't being used?

components/Ink/examples/InkTypicalUse.m
```diff
@@ -17,8 +17,7 @@
 #import <UIKit/UIKit.h>
 
 #import "MaterialInk.h"
-
```
randallli: If possible, I think we want to preserve the space between the import categories.

randallli (approved):
LGTM As long as this passes test.

randallli:
testing it in cl/175877403.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.h
sigito: I would disagree for a few reasons:
1. Since this private header extends the API of the `MDCFeatureHighlightView`, it should include its header directly.
1. `MaterialFeatureHighlight` is own component header. Components should not include own headers.

components/FlexibleHeader/examples/supplemental/FlexibleHeaderWrappedSupplemental.m
sigito: I have no idea. It's been so long ago.

components/ActivityIndicator/examples/supplemental/ActivityIndicatorExampleSupplemental.m
```diff
@@ -19,9 +19,8 @@
  instructions. It is not necessary to import this file to use Material Components for iOS.
  */
 
-#import <Foundation/Foundation.h>
```
randallli, line 22: we are using nsstring so we should keep this.

components/AnimationTiming/examples/supplemental/AnimationTimingExampleSupplemental.m
```diff
@@ -14,9 +14,8 @@
  limitations under the License.
  */
 
-#import <Foundation/Foundation.h>
```
randallli, line 17: still need this.
in fact we need CoreGraphics too since we are using CGSize CGFloat et al

components/Buttons/examples/supplemental/ButtonsTypicalUseSupplemental.m
```diff
@@ -19,9 +19,8 @@
  instructions. It is not necessary to import this file to use Material Components for iOS.
  */
 
-#import <Foundation/Foundation.h>
```
randallli, line 22: lets double check all of these.

randallli (changes requested):
Can we sync with develop since there are some tests failing that have been fixed there.

randallli (approved):

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.h
ianegordon: I agree with @sigito , but happy to discuss if you feel strongly.

ianegordon:
Waiting for a +1 from kokoro.

ianegordon:
Tested locally.  Archiving and UITests on the sim.
kokoro is in flux, so I'm going to land.
@sigito Huge thanks for all this cleanup and your patience!
🎆 👍

randallli:
tests pass for me too.

PR #2440, opened 2017-11-15
[Themer] Updating the protocol generic syntax.

willlarche:
Just to be more expected.

### DIFF
--- a/catalog/MaterialCatalog/MDCCatalogTiles.h
+++ b/catalog/MaterialCatalog/MDCCatalogTiles.h
@@ -19,43 +19,43 @@
 @protocol MDCColorScheme;
 
 /** A function that will perform drawing operations in @c frame. */
-typedef void (*MDCDrawFunc)(CGRect frame, NSObject<MDCColorScheme> *colorScheme);
+typedef void (*MDCDrawFunc)(CGRect frame, id<MDCColorScheme> colorScheme);
 
 /** Render the drawing operations in @c drawFunc into a new image. */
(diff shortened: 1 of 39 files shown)


ajsecord:
Big change—what's the benefit?

willlarche:
The functionality doesn’t change at least. No code implementing these
functions needs to change.

The benefit is that this is a more expected API. When the api uses ‘id’,
the message to the developer is ‘this stands in for an object whose type is
unknown.’ But when the api used is ‘NSObject*’, the subtext is “I’m looking
for the type NSObject” which is specific and misleading. It could cause
someone to think twice about what they are doing and these are protocols we
want people to use. They shouldn’t hesitate. The new API matches the syntax
Apple, and the general community, and us in the majority of places, use.

Moreover, by making the protocol inherit from the NSObject, we are adhering
to a best practice.

From:
https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithProtocols/WorkingwithProtocols.html

It’s best practice to define your protocols to conform to the NSObject
protocol (some of the NSObject behavior is split from its class interface
into a separate protocol; the NSObject class adopts the NSObject protocol).

ajsecord (approved):

jgunaratne (approved):

romoore (approved):
Yes, please! I've wondered why we require the NSObject base class. :)

ajsecord:
Build breakage is from #2400, not this PR...

PR #2444, opened 2017-11-16
[AnimationTiming] Update animation timing README and examples

codeman7:
This pull request updates the examples of animation timing to match the new APIs. Changes made with padding are to make sure animation fits on iPhone 4s. Also, an update to the documentation to the match new APIs.

#2370 is blocking this.


ajsecord (approved):
Thanks!

PR #2445, opened 2017-11-16
[Snackbar] Add nullability annotations to types defined in MDCSnackbarMessage.h

bendeming:
* Adds nullability annotations to `MDCSnackbarMessage` and
  `MDCSnackbarMessageAction`.
* Convenience initializers for `MDCSnackbarMessage` specify `nonnull`
  in line with comments for `text` and `attributedText`
  ("Either @c text or @c attributedText must be set.").
* Made each of the aforementioned properties `nullable` because it's
  possible to create an instance of the class where they are nil
  (e.g., `-init` is available). Same logic applies to properties on
  `MDCSnackbarMessageAction`. Future work could be to make `-init`
  unavailable and introduce designated initializers.

### DIFF
--- a/components/Snackbar/src/MDCSnackbarMessage.h
+++ b/components/Snackbar/src/MDCSnackbarMessage.h
@@ -42,7 +42,7 @@ extern const NSTimeInterval MDCSnackbarMessageDurationMax;
  This attribute can be set over any range of @c attributedText and that text will have the proper
  font applied.
  */
-extern NSString *const MDCSnackbarMessageBoldAttributeName;
+extern NSString * __nonnull const MDCSnackbarMessageBoldAttributeName;
 
 /**
  Represents a message to unobtrusively show to the user.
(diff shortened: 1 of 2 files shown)


jverkoey (approved):
Thank you!

bendeming:
Rebased against develop to pick up the build fix (#2449).

components/Snackbar/src/MDCSnackbarMessage.h
```diff
@@ -61,52 +61,52 @@ extern NSString *const MDCSnackbarMessageBoldAttributeName;
...
-@property(nonatomic, strong) MDCSnackbarMessageAction *action;
+@property(nonatomic, strong, nullable) MDCSnackbarMessageAction *action;
 
 /**
  The color used for button text on the snackbar in normal state.
 
  Default is white.
  */
-@property(nonatomic, strong) UIColor *buttonTextColor;
+@property(nonatomic, strong, nullable) UIColor *buttonTextColor;
```
ianegordon, line 102: Should this be null_resettable since we have a default value?

components/Snackbar/src/MDCSnackbarMessage.h
```diff
@@ -61,52 +61,52 @@ extern NSString *const MDCSnackbarMessageBoldAttributeName;
...
 
 /**
  The color used for button text on the snackbar in highlighted state.
 
  Default is white.
  */
-@property(nonatomic, strong) UIColor *highlightedButtonTextColor;
+@property(nonatomic, strong, nullable) UIColor *highlightedButtonTextColor;
```
ianegordon, line 109: Should this be null_resettable since we have a default value?

ianegordon (dismissed):
Hi @bendeming , thanks for the PR!  What do you think about the comments on nullable -vs- null_resettable?

PR #2450, opened 2017-11-16
WIP!! [Buttons] Prototype for Extended FAB

romoore:
This is a prototype/demo implementation of the Extended FAB.  The headers are ready for an initial review. The code has NOT been refactored/revised/cleaned-up and is full of nits/inefficiencies.

Feel free to discuss the public interfaces, but try to avoid getting caught-up in the details of the implementation.

### DIFF
--- a/components/Buttons/examples/ButtonsContentEdgeInsetsExample.m
+++ b/components/Buttons/examples/ButtonsContentEdgeInsetsExample.m
@@ -53,7 +53,8 @@
   self.flatButton.contentEdgeInsets = UIEdgeInsetsMake(64, 64, 0, 0);
   self.raisedButton.contentEdgeInsets = UIEdgeInsetsMake(0, 0, 64, 64);
   [self.floatingActionButton setContentEdgeInsets:UIEdgeInsetsMake(40, 40, 0, 0)
-                                         forShape:MDCFloatingButtonShapeDefault];
+                                          forType:MDCFloatingButtonTypeDefault
+                                             mode:MDCFloatingButtonModeNormal];
 
   [self updateInkStyle:self.inkBoundingSwitch.isOn ? MDCInkStyleBounded : MDCInkStyleUnbounded];
 }
--- a/components/Buttons/examples/ButtonsContentEdgeInsetsExample.m
+++ b/components/Buttons/examples/ButtonsContentEdgeInsetsExample.m
@@ -52,7 +52,8 @@
   self.flatButton.inkColor = [UIColor colorWithWhite:1.0 alpha:0.1];
   self.flatButton.contentEdgeInsets = UIEdgeInsetsMake(64, 64, 0, 0);
   self.raisedButton.contentEdgeInsets = UIEdgeInsetsMake(0, 0, 64, 64);
-  self.floatingActionButton.contentEdgeInsets = UIEdgeInsetsMake(40, 40, 0, 0);
+  [self.floatingActionButton setContentEdgeInsets:UIEdgeInsetsMake(40, 40, 0, 0)
+                                         forShape:MDCFloatingButtonShapeDefault];
 
   [self updateInkStyle:self.inkBoundingSwitch.isOn ? MDCInkStyleBounded : MDCInkStyleUnbounded];
 }
--- a/components/Buttons/examples/ButtonsTypicalUse.m
+++ b/components/Buttons/examples/ButtonsTypicalUse.m
@@ -148,9 +148,11 @@
 
       }
       if (self.floatingButton.shape >= MDCFloatingButtonShapeExtendedLeadingIcon) {
+        [self.floatingButton setImage:[UIImage imageNamed:@"Plus"] forState:UIControlStateNormal];
         [self.floatingButton setTitle:@"Extended" forState:UIControlStateNormal];
-      } else {
-        [self.floatingButton setTitle:nil forState:UIControlStateNormal];
+      }  else {
+        [self.floatingButton setImage:nil forState:UIControlStateNormal];
+        [self.floatingButton setTitle:@"New" forState:UIControlStateNormal];
       }
       [self.floatingButton sizeToFit];
     }];
--- a/components/Buttons/examples/ButtonsTypicalUse.m
+++ b/components/Buttons/examples/ButtonsTypicalUse.m
@@ -134,14 +134,19 @@
 - (void)didTap:(id)sender {
   NSLog(@"%@ was tapped.", NSStringFromClass([sender class]));
   if (sender == self.floatingButton) {
-    [self.floatingButton
-          collapse:YES
-        completion:^{
-          dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(1 * NSEC_PER_SEC)),
-                         dispatch_get_main_queue(), ^{
-                           [self.floatingButton expand:YES completion:nil];
-                         });
-        }];
+    [UIView animateWithDuration:0.25 animations:^{
+      self.floatingButton.shape += 1;
+      if (self.floatingButton.shape > MDCFloatingButtonShapeExtendedTrailingIcon) {
+        self.floatingButton.shape = 0;
+      }
+      if (self.floatingButton.shape >= MDCFloatingButtonShapeExtendedLeadingIcon) {
+        [self.floatingButton setTitle:@"Extended" forState:UIControlStateNormal];
+      } else {
+        [self.floatingButton setTitle:nil forState:UIControlStateNormal];
+      }
+      [self.floatingButton sizeToFit];
+    }];
+
   }
 }
 
--- a/components/Buttons/examples/ButtonsTypicalUse.m
+++ b/components/Buttons/examples/ButtonsTypicalUse.m
@@ -21,6 +21,7 @@
 
 @interface ButtonsTypicalUseViewController ()
 @property(nonatomic, strong) MDCFloatingButton *floatingButton;
+@property(nonatomic, assign) BOOL styleChanged;
 @end
 
 @implementation ButtonsTypicalUseViewController
@@ -136,18 +137,29 @@
   NSLog(@"%@ was tapped.", NSStringFromClass([sender class]));
   if (sender == self.floatingButton) {
     [UIView animateWithDuration:0.25 animations:^{
-      self.floatingButton.shape += 1;
-      if (self.floatingButton.shape > MDCFloatingButtonShapeExtendedTrailingIcon) {
-        self.floatingButton.shape = 0;
-        static dispatch_once_t onceToken;
-        dispatch_once(&onceToken, ^{
+      self.floatingButton.mode += 1;
+      if (self.floatingButton.mode > MDCFloatingButtonModeExtended) {
+        self.floatingButton.mode = 0;
+        self.floatingButton.imagePosition += 1;
+        if (self.floatingButton.imagePosition > MDCFloatingButtonImagePositionTrailing) {
+          self.floatingButton.imagePosition = 0;
+          self.floatingButton.contentEdgeInsetsFlippedForTrailingImagePosition
+            = !self.floatingButton.contentEdgeInsetsFlippedForTrailingImagePosition;
+        }
+
+        if (!self.styleChanged) {
           MDCBasicColorScheme *scheme = [[MDCBasicColorScheme alloc] initWithPrimaryColor:UIColor.purpleColor
                                                                            secondaryColor:UIColor.orangeColor];
           [ExampleFloatingButtonThemer applyToButton:self.floatingButton withColorScheme:scheme];
-        });
-
+          self.styleChanged = YES;
+        }
       }
-      if (self.floatingButton.shape >= MDCFloatingButtonShapeExtendedLeadingIcon) {
+      NSLog(@"Mode: %ld, Pos: %ld, Flip: %d\nCEI: %@",
+            self.floatingButton.mode,
+            self.floatingButton.imagePosition,
+            self.floatingButton.contentEdgeInsetsFlippedForTrailingImagePosition,
+            NSStringFromUIEdgeInsets(self.floatingButton.contentEdgeInsets));
+      if (self.floatingButton.mode == MDCFloatingButtonModeExtended) {
         [self.floatingButton setImage:[UIImage imageNamed:@"Plus"] forState:UIControlStateNormal];
         [self.floatingButton setTitle:@"Extended" forState:UIControlStateNormal];
       }  else {
--- a/components/Buttons/examples/ButtonsTypicalUse.m
+++ b/components/Buttons/examples/ButtonsTypicalUse.m
@@ -17,6 +17,7 @@
 #import "MaterialButtons.h"
 #import "MaterialTypography.h"
 #import "supplemental/ButtonsTypicalUseSupplemental.h"
+#import "ExampleFloatingButtonThemer.h"
 
 @interface ButtonsTypicalUseViewController ()
 @property(nonatomic, strong) MDCFloatingButton *floatingButton;
@@ -138,6 +139,13 @@
       self.floatingButton.shape += 1;
       if (self.floatingButton.shape > MDCFloatingButtonShapeExtendedTrailingIcon) {
         self.floatingButton.shape = 0;
+        static dispatch_once_t onceToken;
+        dispatch_once(&onceToken, ^{
+          MDCBasicColorScheme *scheme = [[MDCBasicColorScheme alloc] initWithPrimaryColor:UIColor.purpleColor
+                                                                           secondaryColor:UIColor.orangeColor];
+          [ExampleFloatingButtonThemer applyToButton:self.floatingButton withColorScheme:scheme];
+        });
+
       }
       if (self.floatingButton.shape >= MDCFloatingButtonShapeExtendedLeadingIcon) {
         [self.floatingButton setTitle:@"Extended" forState:UIControlStateNormal];
--- a/components/Buttons/examples/ButtonsTypicalUse.m
+++ b/components/Buttons/examples/ButtonsTypicalUse.m
@@ -148,9 +148,9 @@
         }
 
         if (!self.styleChanged) {
-          MDCBasicColorScheme *scheme = [[MDCBasicColorScheme alloc] initWithPrimaryColor:UIColor.purpleColor
-                                                                           secondaryColor:UIColor.orangeColor];
-          [ExampleFloatingButtonThemer applyToButton:self.floatingButton withColorScheme:scheme];
+//          MDCBasicColorScheme *scheme = [[MDCBasicColorScheme alloc] initWithPrimaryColor:UIColor.purpleColor
+//                                                                           secondaryColor:UIColor.orangeColor];
+//          [ExampleFloatingButtonThemer applyToButton:self.floatingButton withColorScheme:scheme];
           self.styleChanged = YES;
         }
       }
@@ -161,8 +161,8 @@
             NSStringFromUIEdgeInsets(self.floatingButton.contentEdgeInsets));
       if (self.floatingButton.mode == MDCFloatingButtonModeExtended) {
         [self.floatingButton setImage:[UIImage imageNamed:@"Plus"] forState:UIControlStateNormal];
-        [self.floatingButton setTitle:@"Extended" forState:UIControlStateNormal];
-      }  else {
+        [self.floatingButton setTitle:@"Default" forState:UIControlStateNormal];
+      } else {
         [self.floatingButton setImage:nil forState:UIControlStateNormal];
         [self.floatingButton setTitle:@"New" forState:UIControlStateNormal];
       }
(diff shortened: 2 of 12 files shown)


components/Buttons/examples/ButtonsContentEdgeInsetsExample.m
```diff
@@ -52,7 +52,8 @@ - (void)viewDidLoad {
   self.flatButton.inkColor = [UIColor colorWithWhite:1.0 alpha:0.1];
   self.flatButton.contentEdgeInsets = UIEdgeInsetsMake(64, 64, 0, 0);
   self.raisedButton.contentEdgeInsets = UIEdgeInsetsMake(0, 0, 64, 64);
-  self.floatingActionButton.contentEdgeInsets = UIEdgeInsetsMake(40, 40, 0, 0);
+  [self.floatingActionButton setContentEdgeInsets:UIEdgeInsetsMake(40, 40, 0, 0)
+                                         forShape:MDCFloatingButtonShapeDefault];
```
romoore: There would be breaking changes by making properties unavailable:
* minimumSize
* maximumSize
* contentEdgeInsets
* hitAreaInsets

The alternative is not much better, however, since "shape" does not have a true "default" value the way UIControlState does.

components/Buttons/src/MDCFloatingButton.h
```diff
@@ -96,6 +118,26 @@ typedef NS_ENUM(NSInteger, MDCFloatingButtonShape) {
 
 - (nullable instancetype)initWithCoder:(nonnull NSCoder *)aDecoder NS_DESIGNATED_INITIALIZER;
 
+- (void)setMinimumSize:(CGSize)minimumSize NS_UNAVAILABLE;
+
+- (void)setMinimumSize:(CGSize)minimumSize forShape:(MDCFloatingButtonShape)shape
+    UI_APPEARANCE_SELECTOR;
```
romoore: Right now, these will "default" to `MDCFloatingButtonShapeDefault`, but that may not be expected behavior. We may prefer to require each shape value to be fully specified.

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -60,79 +75,369 @@ - (instancetype)initWithFrame:(CGRect)frame shape:(MDCFloatingButtonShape)shape
...
 
-- (CGSize)intrinsicContentSize {
-  switch (_shape) {
-    case MDCFloatingButtonShapeDefault:
-      return CGSizeMake([[self class] defaultDimension], [[self class] defaultDimension]);
-    case MDCFloatingButtonShapeMini:
-      return CGSizeMake([[self class] miniDimension], [[self class] miniDimension]);
-    case MDCFloatingButtonShapeLargeIcon:
-      return CGSizeMake([[self class] defaultDimension], [[self class] defaultDimension]);
+//- (CGSize)sizeThatFits:(CGSize)size {
```
romoore: This is an alternate solution that would just add the `imageTitlePadding` value to the size returned by super. However, this has risks by relying on UIButton (or MDCButton) to perform sizing in a compatible manner.

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -60,79 +75,369 @@ - (instancetype)initWithFrame:(CGRect)frame shape:(MDCFloatingButtonShape)shape
...
     case MDCFloatingButtonShapeDefault:
       return CGSizeMake([[self class] defaultDimension], [[self class] defaultDimension]);
     case MDCFloatingButtonShapeMini:
       return CGSizeMake([[self class] miniDimension], [[self class] miniDimension]);
     case MDCFloatingButtonShapeLargeIcon:
       return CGSizeMake([[self class] defaultDimension], [[self class] defaultDimension]);
+    case MDCFloatingButtonShapeExtendedTrailingIcon:
+    case MDCFloatingButtonShapeExtendedLeadingIcon: {
+      const CGSize intrinsicTitleSize
+          = [self.titleLabel sizeThatFits:CGSizeMake(CGFLOAT_MAX, CGFLOAT_MAX)];
```
romoore: Calling `sizeThatFits:` triggers an entire layout pass of the button due to the shadowLayer laying out its sublayers (apparently).

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -60,79 +75,369 @@ - (instancetype)initWithFrame:(CGRect)frame shape:(MDCFloatingButtonShape)shape
...
+  // |    |  |  |  CEI BOT            |   |
+  // +------------------------------------+
+  //
+  // (A) The same spacing on either side of the label.
+  // (SP) The spacing between the image and title
+  // (CEI) Content Edge Insets
+  //
+  // The diagram above assumes an LTR user interface orientation
+  // and a .leadingIcon shape for this button.
+
```
romoore, line 299: The code below is unoptimized/unrefactored. Take it with a grain of salt.

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -60,79 +75,369 @@ - (instancetype)initWithFrame:(CGRect)frame shape:(MDCFloatingButtonShape)shape
...
+  }
+}
+
+- (void)updateShape {
+  BOOL needsLayout = [self updateMinimumSize];
+  needsLayout |= [self updateMaximumSize];
+  [self updateContentEdgeInsets];
+  [self updateHitAreaInsets];
+  if (needsLayout) {
+    [self sizeToFit];
```
romoore, line 557: This is concerning. I found that without a call to `sizeToFit` after UIAppearance applied the min/max sizes, the new sizes were not applied on-screen.  The alternative would be for all clients to do:

```objectivec
// Note that this is NOT |viewWillAppear:|
- (void)viewDidAppear:(BOOL)animated {
  [super viewDidAppear:animated];
  [self.floatingButton sizeToFit];
}
```

components/Buttons/src/MDCFloatingButton.h
```diff
@@ -37,7 +37,15 @@ typedef NS_ENUM(NSInteger, MDCFloatingButtonShape) {
   /**
    A 56-point circular button surrounding a 36-point square icon or short text.
    */
-  MDCFloatingButtonShapeLargeIcon = 2
+  MDCFloatingButtonShapeLargeIcon = 2,
```
randallli: Let us acknowledge that:
enums are much more difficult to deprecate than a mutator class.
It is also much more difficult to expand. 
Is the recommendation to subclass to add shapes?
Is "LargeIcon" really a shape?

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -19,20 +19,31 @@
...
 
-@implementation MDCFloatingButton {
-  MDCFloatingButtonShape _shape;
-}
+@interface MDCFloatingButton ()
+@property(nonatomic, readonly) NSMutableDictionary<NSNumber *, NSValue *> *shapeToHitAreaInsets;
+@property(nonatomic, readonly) NSMutableDictionary<NSNumber *, NSValue *> *shapeToMinimumSize;
+@property(nonatomic, readonly) NSMutableDictionary<NSNumber *, NSValue *> *shapeToMaximumSize;
+@property(nonatomic, readonly) NSMutableDictionary<NSNumber *, NSValue *> *shapeToContentEdgeInsets;
+@property(nonatomic, readonly) NSMutableDictionary<NSNumber *, NSMutableDictionary<NSNumber *, NSNumber *> *> *shapeToStateToElevation;
```
romoore: This is somewhat disconcerting.

components/Buttons/src/MDCFloatingButton.h
```diff
@@ -52,6 +60,20 @@ typedef NS_ENUM(NSInteger, MDCFloatingButtonShape) {
 @interface MDCFloatingButton : MDCButton
 
 /**
+ The shape of this floating button.
+
+ Changing this value will effect how the title and image are positioned.
+ */
+@property(nonatomic, assign) MDCFloatingButtonShape shape;
```
randallli: The requirement to be able to change from Default to Extended opens the API to switch to and from any combination of the other shapes.

components/Buttons/examples/ButtonsContentEdgeInsetsExample.m
randallli: Thanks for acknowledging that we are NS_UNAVAILABLE 5 properties. you missed:
- elevationForState:

components/Buttons/examples/ButtonsTypicalUse.m
```diff
@@ -135,14 +136,26 @@ - (void)viewDidLoad {
...
+          MDCBasicColorScheme *scheme = [[MDCBasicColorScheme alloc] initWithPrimaryColor:UIColor.purpleColor
+                                                                           secondaryColor:UIColor.orangeColor];
+          [ExampleFloatingButtonThemer applyToButton:self.floatingButton withColorScheme:scheme];
+        });
+
+      }
+      if (self.floatingButton.shape >= MDCFloatingButtonShapeExtendedLeadingIcon) {
+        [self.floatingButton setTitle:@"Extended" forState:UIControlStateNormal];
+      } else {
+        [self.floatingButton setTitle:nil forState:UIControlStateNormal];
```
randallli: So when we change shape to not extended we have to nil the title? what happens when we forget to do that?

components/Buttons/src/MDCButton.m
```diff
@@ -676,7 +676,9 @@ - (CGFloat)elevationForState:(UIControlState)state {
 
 - (void)setElevation:(CGFloat)elevation forState:(UIControlState)state {
   _userElevations[@(state)] = @(elevation);
-  self.layer.elevation = [self elevationForState:self.state];
+  if (self.state == state) {
```
randallli, line 679: I think you mean !=

but be careful because we want the default elevation to also be set.

components/Buttons/src/MDCFloatingButton.h
```diff
@@ -96,6 +118,32 @@ typedef NS_ENUM(NSInteger, MDCFloatingButtonShape) {
...
+
+- (void)setContentEdgeInsets:(UIEdgeInsets)contentEdgeInsets forShape:(MDCFloatingButtonShape)shape
+    UI_APPEARANCE_SELECTOR;
+
+- (void)setHitAreaInsets:(UIEdgeInsets)hitAreaInsets NS_UNAVAILABLE;
+
+- (void)setHitAreaInsets:(UIEdgeInsets)insets forShape:(MDCFloatingButtonShape)shape
+    UI_APPEARANCE_SELECTOR;
+
+- (void)setElevation:(CGFloat)elevation forState:(UIControlState)state NS_UNAVAILABLE;
```
randallli: This means we are going to have to overload the MDCButton sets the elevation when state changes to take into account our shape.

components/Buttons/src/MDCFloatingButton.h
```diff
@@ -96,6 +118,32 @@ typedef NS_ENUM(NSInteger, MDCFloatingButtonShape) {
...
+
+- (void)setElevation:(CGFloat)elevation
+            forState:(UIControlState)state
+               shape:(MDCFloatingButtonShape)shape UI_APPEARANCE_SELECTOR;
```
romoore: DRop this concept.

components/Buttons/examples/ButtonsTypicalUse.m
romoore: Maybe we should just hide the title?

components/Buttons/src/MDCFloatingButton.h
ianegordon: +1 to dropping.

setElevation:forStyle:

components/Buttons/src/MDCFloatingButton.h
```diff
@@ -96,6 +118,32 @@ typedef NS_ENUM(NSInteger, MDCFloatingButtonShape) {
...
 - (nullable instancetype)initWithCoder:(nonnull NSCoder *)aDecoder NS_DESIGNATED_INITIALIZER;
 
+- (void)setMinimumSize:(CGSize)minimumSize NS_UNAVAILABLE;
+
+- (void)setMinimumSize:(CGSize)minimumSize forShape:(MDCFloatingButtonShape)shape
+    UI_APPEARANCE_SELECTOR;
+
+- (void)setMaximumSize:(CGSize)maximumSize NS_UNAVAILABLE;
+
+- (void)setMaximumSize:(CGSize)maximumSize forShape:(MDCFloatingButtonShape)shape
```
ianegordon: Let's discuss this.
I'm hoping we can get away without it.

components/Buttons/src/MDCFloatingButton.h
```diff
@@ -37,7 +37,15 @@ typedef NS_ENUM(NSInteger, MDCFloatingButtonShape) {
   /**
    A 56-point circular button surrounding a 36-point square icon or short text.
    */
-  MDCFloatingButtonShapeLargeIcon = 2
+  MDCFloatingButtonShapeLargeIcon = 2,
+  /**
+   An Extended FAB with a 24-point square icon on the leading side of the text.
```
ianegordon: If this supports dynamic type, note how scaling works.  Perhaps something along the lines of:
An extended FAB with a square icon (default 24 x 24) on the leading edge of the text.

ianegordon (commented):
For offline discussion:

Can we get rid of min/max size?
Does max size make sense if these are going to be shown on an iPad?  I would expect the button to respect the size being set by it's parent.

For Mini + Standard is the only difference sizeThatFits: / instrinsicContentSize?

Can we get away with:
@property (assign) mode (expanded / compact)

components/Buttons/examples/ButtonsTypicalUse.m
romoore: I think we should leave it up to clients to add/remove the title along with the shape:
* UIButton ignores the alpha/hidden properties of the titleLabel when performing layout. The imageView is still offset as though the title were visible.
* The title is set on a per-state basis, and the button probably should not be removing all possible states (including the UIControlStateApplication range).
* Compact FABs may use the titleLabel instead of an image in some cases. Some phrases with low character count may be preferable to images.

### Hidden title that contains text (includes frames)
screen shot 2017-11-17 at 9 51 45 am

### FAB with title-only in compact mode
screen shot 2017-11-17 at 10 00 22 am

randallli: sgtm

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -42,104 +55,430 @@ + (CGFloat)miniDimension {
...
-    self.hitAreaInsets = [self defaultHitAreaInsets];
+    [self commonMDCFloatingButtonInit];
+    [self updateType];
   }
   return self;
 }
 
+- (void)commonMDCFloatingButtonInit {
+  NSMutableDictionary *miniTypeMinimumSizeDictionary
+      = [NSMutableDictionary dictionaryWithCapacity:2];
```
randallli: put the = on previous line.

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -42,104 +55,430 @@ + (CGFloat)miniDimension {
...
+  const CGSize miniSizeNormal = CGSizeMake(40, 40);
+  const CGSize defaultSizeNormal = CGSizeMake(56, 56);
+  const CGSize defaultSizeExtendedMinimum = CGSizeMake(132, 48);
+  const CGSize defaultSizeExtendedMaximum = CGSizeMake(328, 0);
+  miniTypeMinimumSizeDictionary[@(MDCFloatingButtonModeNormal)] =
+
```
randallli: remove space

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -42,104 +55,430 @@ + (CGFloat)miniDimension {
...
+  _typetoModeToMinimumSize = [NSMutableDictionary dictionaryWithCapacity:2];
```
randallli: indent

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -42,104 +55,430 @@ + (CGFloat)miniDimension {
...
+  _typetoModeToMinimumSize
+      = [@{
+           @(MDCFloatingButtonTypeMini) :
+             [@{ @(MDCFloatingButtonModeNormal) :
```
randallli: put this in a temp local variable so the dic is not so big.

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -42,104 +55,430 @@ + (CGFloat)miniDimension {
...
+                   [NSValue valueWithCGSize:miniSizeNormal] } mutableCopy],
+           @(MDCFloatingButtonTypeDefault) :
+             [@{ @(MDCFloatingButtonModeNormal) : [NSValue valueWithCGSize:defaultSizeNormal],
```
randallli: temp local variable.

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -42,104 +55,430 @@ + (CGFloat)miniDimension {
...
+      = [NSMutableDictionary dictionaryWithCapacity:2];
+  const CGSize miniSizeNormal = CGSizeMake(40, 40);
+  const CGSize defaultSizeNormal = CGSizeMake(56, 56);
+  const CGSize defaultSizeExtendedMinimum = CGSizeMake(132, 48);
+  const CGSize defaultSizeExtendedMaximum = CGSizeMake(328, 0);
+  miniTypeMinimumSizeDictionary[@(MDCFloatingButtonModeNormal)] =
+
+  _typetoModeToMinimumSize = [NSMutableDictionary dictionaryWithCapacity:2];
+  _typetoModeToMinimumSize
+      = [@{
```
randallli: move equal to previous line.

randallli: space between [ @

randallli (commented):
readability improvements for messy dictionaries.

BigAlInTheHouse:
I'm grateful that through all the endless days and nights. I will have a
word with you all.  FULLVALUEINFOSTUDY.COM

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -42,104 +55,442 @@ + (CGFloat)miniDimension {
...
+           @(MDCFloatingButtonModeNormal) : [NSValue valueWithUIEdgeInsets:UIEdgeInsetsZero]
+           } mutableCopy];
+  NSMutableDictionary *defaultContentEdgeInsets
+      = [@{
+           @(MDCFloatingButtonModeNormal) : [NSValue valueWithUIEdgeInsets:UIEdgeInsetsZero],
+           @(MDCFloatingButtonModeExtended) :
+             [NSValue valueWithUIEdgeInsets:UIEdgeInsetsMake(0, 16, 0, 24)],
+           } mutableCopy];
+  _typeToModeToContentEdgeInsets = [@{ @(MDCFloatingButtonTypeMini) : miniContentEdgeInsets,
```
romoore: More like this section?

randallli: yes

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -42,104 +55,442 @@ + (CGFloat)miniDimension {
...
+- (void)commonMDCFloatingButtonInit {
+  const CGSize miniSizeNormal = CGSizeMake(40, 40);
+  const CGSize defaultSizeNormal = CGSizeMake(56, 56);
+  const CGSize defaultSizeExtendedMinimum = CGSizeMake(132, 48);
+  const CGSize defaultSizeExtendedMaximum = CGSizeMake(328, 0);
+
+
+  _typetoModeToMinimumSize = [NSMutableDictionary dictionaryWithCapacity:2];
+  _typetoModeToMinimumSize
+      = [@{
```
willlarche: I'd love for a comment since this is so unusual looking.

components/Buttons/src/ColorThemer/ExampleFloatingButtonThemer.m
```diff
@@ -0,0 +1,55 @@
+//
```
willlarche: Easy fix for later: license stanzas.

components/Buttons/src/MDCFloatingButton.m
willlarche: Also, this looks like the kind of work that should have a dispatch once on it.

components/Buttons/src/ColorThemer/ExampleFloatingButtonThemer.m
romoore: Done.

components/Buttons/src/MDCFloatingButton.m
romoore: Agreed. Once we settle on the baseline, it should be extracted out into a static bit of code. Instances can then use it to initialize the properties.

components/Buttons/src/MDCButton.m
romoore: ~~No, we only want to update elevation-related properties if the elevation for the current state has changed. Otherwise those properties will be updated on the next state change.~~

Actually, I think it makes sense to check for the explicit fall-through in this case.  I will drop this change from this PR and push it up separately with more context.

romoore:
Closing, but will push up a set of headers and usage code review.

PR #2464, opened 2017-11-16
[AppBar] Check UIAppearance bg color in the a11y mutator

### LINKED ISSUE #2457: [Catalog] Header bg color consistency
ss

chuga:
Also removes the hardcoded styling we have in the app bar examples. This makes Themes work as expected.

Closes #2457

### DIFF
--- a/components/AppBar/examples/AppBarModalPresentationExample.swift
+++ b/components/AppBar/examples/AppBarModalPresentationExample.swift
@@ -28,8 +28,6 @@ class AppBarModalPresentationSwiftExamplePresented: UITableViewController {
 
     self.addChildViewController(appBar.headerViewController)
 
-    let color = UIColor(white: 0.2, alpha:1)
-    appBar.headerViewController.headerView.backgroundColor = color
     let mutator = MDCAppBarTextColorAccessibilityMutator()
     mutator.mutate(appBar)
     self.modalPresentationStyle = .formSheet
@@ -103,8 +101,6 @@ class AppBarModalPresentationSwiftExample: UITableViewController {
 
     self.addChildViewController(appBar.headerViewController)
 
-    let color = UIColor(white: 0.2, alpha:1)
-    appBar.headerViewController.headerView.backgroundColor = color
     let mutator = MDCAppBarTextColorAccessibilityMutator()
     mutator.mutate(appBar)
   }
--- a/components/AppBar/examples/AppBarTypicalUseExample.swift
+++ b/components/AppBar/examples/AppBarTypicalUseExample.swift
@@ -28,8 +28,9 @@ class AppBarTypicalUseSwiftExample: UITableViewController {
 
     self.addChildViewController(appBar.headerViewController)
 
-    let mutator = MDCAppBarTextColorAccessibilityMutator()
-    mutator.mutate(appBar)
+    // Note: You could use MDCAppBarTextColorAccessibilityMutator instead of doing this manually.
+    appBar.navigationBar.tintColor = UIColor.white
+    appBar.navigationBar.titleTextAttributes = [NSForegroundColorAttributeName : UIColor.white]
   }
 
   required init?(coder aDecoder: NSCoder) {
--- a/components/AppBar/examples/AppBarTypicalUseExample.swift
+++ b/components/AppBar/examples/AppBarTypicalUseExample.swift
@@ -19,7 +19,6 @@ import MaterialComponents
 
 class AppBarTypicalUseSwiftExample: UITableViewController {
 
-  // Step 1: Create and initialize an App Bar.
   let appBar = MDCAppBar()
 
   init() {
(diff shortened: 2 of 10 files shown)


components/AppBar/examples/AppBarTypicalUseExample.swift
```diff
@@ -19,21 +19,17 @@ import MaterialComponents
...
 
-    // Step 2: Add the headerViewController as a child.
     self.addChildViewController(appBar.headerViewController)
 
-    let color = UIColor(white: 0.2, alpha:1)
-    appBar.headerViewController.headerView.backgroundColor = color
-    appBar.navigationBar.tintColor = UIColor.white
-    appBar.navigationBar.titleTextAttributes = [NSForegroundColorAttributeName : UIColor.white]
+    let mutator = MDCAppBarTextColorAccessibilityMutator()
+    mutator.mutate(appBar)
```
ianegordon: Can we keep one of these examples as is to demonstrate both with and without the mutator?

ianegordon (approved):

components/AppBar/examples/AppBarTypicalUseExample.swift
chuga: Sure. Any preference?

ianegordon (dismissed):

components/AppBar/src/MDCAppBar.m
```diff
@@ -47,7 +47,13 @@ - (void)mutate:(nonnull MDCAppBar *)appBar {
   UIColor *backgroundColor = appBar.navigationBar.backgroundColor ?:
       appBar.headerViewController.headerView.backgroundColor;
   if (!backgroundColor) {
-    return;
+    // It might be possible that the views haven't received their background colors if we're using
+    // UIAppearance to style them. That's why we check their appearance before giving up.
+    backgroundColor = [[MDCNavigationBar appearance] backgroundColor] ?:
```
romoore, line 52: I have strong reservations about this change. It's possible that the color was explicitly set to `nil`, and that won't be caught by this check.  UIAppearance won't apply the proxied values to an instance if the setter was called.   This may be the case for an AppBar with an imageView as its background (instead of a color), or one that relies on the `backgroundColor` of the container view.

If we **really** want to take this approach, please update the public (header) documentation to make it clear that it will overwrite any instance values with UIAppearance values (if set).  My vote is to update our examples to do mutating in `-viewDidAppear:` and provide that as the best practice.

components/AppBar/examples/AppBarModalPresentationExample.swift
```diff
@@ -28,8 +28,6 @@ class AppBarModalPresentationSwiftExamplePresented: UITableViewController {
 
     self.addChildViewController(appBar.headerViewController)
 
-    let color = UIColor(white: 0.2, alpha:1)
-    appBar.headerViewController.headerView.backgroundColor = color
     let mutator = MDCAppBarTextColorAccessibilityMutator()
     mutator.mutate(appBar)
```
romoore, line 32: If you move the `mutate` call to `-viewDidAppear:`, then any UIAppearance values will be set on the instance. 

If our examples are to demonstrate how to work with AppBars and UIAppearance together, then we should move all of the mutators into `-viewDidAppear`.

romoore (dismissed):
Please consider moving the mutator calls into `-viewDidAppear` in the examples instead of having the mutator check UIAppearance values.

components/AppBar/examples/AppBarModalPresentationExample.swift
ianegordon: Interesting point.  For some reason I had it in my head that viewDidAppear was too late in certain cases, but I'm not sure my reasoning.  Andrés?

chuga: `viewDidAppear` is indeed too late, the wrong colors are visible while the view animates in.

components/AppBar/src/MDCAppBar.m
chuga: `viewDidAppear` is not a realistic option. Other ideas?

samnm: Did you explore running the mutator on the UIAppearance instance?

chuga: No, would that be considered a better approach?

samnm: IMO it would. That way clients using UIAppearance can just use directly after configuring their UIAppearance setup, and clients manually setting values on an MDCAppBar instance can mutate their instance.

chuga: FWIW, if you're setting your background color to nil you should not be calling this mutator as it only works if you do have a background color.

chuga: Thoughts on @samnm's proposal @romoore @ianegordon?

romoore: I think having the client call it explicitly on UIAppearance proxy is better. If it's not already documented, we should ensure they realize they'll have to theme both the FHV and NavBar proxies before calling the mutator on the AppBar proxy.

chuga: Since MDCAppBar is not a view, I can't just use UIAppearance. I'll just drop this PR.

ianegordon: Applying the mutator to the UIAppearance proxy seemed like such a clean solution.  :/
Since MDCAppBar is not a UIView would a possible solution be class properties similar to what we had to do for textfields?

PR #2324, opened 2017-11-01
[Ink] Prototype demo of updated ink animation and behavior

jgunaratne:
Prototype demo of behavior and animation of potential next version of ink. Simplified ink based on new design spec. Will not merge, only for prototype demo.

PR #2476, opened 2017-11-17
[Ink] Prototype demo of updated ink animation and behavior

jgunaratne:
Prototype demo of behavior and animation of potential next version of ink. Simplified ink based on new design spec.


jgunaratne:
Continues work from PR: https://github.com/material-components/material-components-ios/pull/2324

jgunaratne:
Ink animation changes addressed in: https://github.com/material-components/material-components-ios/pull/2488

PR #2484, opened 2017-11-20
[Buttons] Update layer elevation only on change

### LINKED ISSUE #2483: [Buttons] Setting elevation for non-active states causes multiple elevation change animations
### Overview
When setting multiple elevation-for-state values in MDCButton, each one causes an animation to update the elevation rendering.  This can cause visual bugs during animations in components like the Bottom App Bar if values are being set in sequence.

### Reproduction steps
Set the elevation for multiple UIControlState values, starting with the active state

** Expected Result **
Button elevation animates to update only the active state

** Actual Result **
Button elevation animates or all state updates, including inactive states.

### Version number
40.0.3

Any/all

romoore:
When a button's elevation is assigned, it animates a change to the underlying
shadow elevation.  However, this causes animation "jumps" when elevations for
multiple states are being assigned simultaneously. Instead of blindly
assigning the new value to the layer, a check for a change should be performed
instead.

Closes #2483

### DIFF
--- a/components/Buttons/src/MDCButton.m
+++ b/components/Buttons/src/MDCButton.m
@@ -676,7 +676,12 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
 
 - (void)setElevation:(CGFloat)elevation forState:(UIControlState)state {
   _userElevations[@(state)] = @(elevation);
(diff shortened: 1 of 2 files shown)


ianegordon (approved):
LGTM.

Issue #2467, opened 2017-11-17
[Catalog] Text field filled text example does not resize the message box


sigito:
![Filled text example](https://user-images.githubusercontent.com/1129082/32944987-fc23588c-cb91-11e7-8d56-2f82805578ff.gif)

Version: 40.0.0.

willlarche:
Closed in #2489

PR #2489, opened 2017-11-20
[TextFields] Fixes bug: multiline having ambiguous height in the text view

### LINKED ISSUE #2352: [TextFields] Filled multiline height is wrong when empty and minimumLines == 0

willlarche:
Closes #2352


willlarche:
Closes #2467

components/TextFields/src/private/MDCIntrinsicHeightTextView.m
```diff
@@ -0,0 +1,32 @@
...
+
+@implementation MDCIntrinsicHeightTextView
+
+/**
+ When a value in the CGSize of intrinsicContentSize is -1, it's considered undefined. For the
+ MDCMultilineTextField, we want this to always be defined so our layouts are not ambiguous.
+ */
+- (CGSize)intrinsicContentSize {
+  CGSize size = [super intrinsicContentSize];
+  if (size.height == -1) {
```
ianegordon: Is -1 the same as UIViewNoIntrinsicMetric?
If so, prefer UIViewNoIntrinsicMetric.
If not, do we also need special handling of UIViewNoIntrinsicMetric?

components/TextFields/src/MDCMultilineTextField.m
```diff
@@ -88,12 +92,11 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder {
...
-      [self setupTextView];
-    } else if (!_textView) {
-      // It's expected that this will be created by the fundament above. This is a catch.
-      _textView = [[UITextView alloc] initWithFrame:CGRectZero];
-      [self setupTextView];
     }
+    if (!_textView) {
+      _textView = [[MDCIntrinsicHeightTextView alloc] initWithFrame:CGRectZero];
+    }
+    [self setupTextView];
```
ianegordon: Do we want to replace existing UITextViews with IntrinsicHeight versions?
if (!_textView || ![_textView isKindOf:[MDCIntrinsicHeightTextView class]]) {
  _textView = [[MDCIntrinsicHeightTextView alloc] initWithFrame:CGRectZero];
}

ianegordon: setupTextView will overwrite a bunch of properties that may already be encoded.  Is this acceptable?  If so, consider adding a comment that mentioning this and that it is intentional.

components/TextFields/src/MDCMultilineTextField.m
```diff
@@ -685,22 +707,17 @@ - (void)setTextInsetsMode:(MDCTextInputTextInsetsMode)textInsetsMode {
...
 
 - (void)setTextView:(UITextView *)textView {
   if (![_textView isEqual:textView]) {
     [_textView removeFromSuperview];
     _textView = textView;
-    if (textView) {
-      [self setupTextView];
-      [self setupUnderlineConstraints];
-    }
+    [self textViewSafetyCheck];
```
ianegordon: Can we avoid this check here by requiring an MDCIntrinsicHeightTextView as the parameter / property?

components/TextFields/src/MDCMultilineTextField.m
```diff
@@ -233,6 +236,25 @@ - (void)setupTextView {
   self.textView.opaque = NO;
 }
 
+// It's possible we've been passed a text view that's not our special class. We can handle that.
+// It's also something we guarantee to always be there. So, if it's nil, we make a new one.
+- (void)textViewSafetyCheck {
```
ianegordon: Naming is hard.
Consider: 
- (void)promoteToMDCIntrinsicTextView

components/TextFields/src/MDCMultilineTextField.m
```diff
@@ -233,6 +236,25 @@ - (void)setupTextView {
...
+  // We expect users of Interface Builder to put a UITextView in the view. When that happens,
+  // we swap it out for one of our special text views and bring the existing text over.
+  NSString *existingText = _textView.text;
```
ianegordon: Do we care about any other state?

font?
textColor?
textAlignment?
etc.

components/TextFields/src/MDCMultilineTextField.m
```diff
@@ -75,6 +76,9 @@ @implementation MDCMultilineTextField
 - (instancetype)initWithFrame:(CGRect)frame {
   self = [super initWithFrame:frame];
   if (self) {
+    _textView = [[MDCIntrinsicHeightTextView alloc] initWithFrame:CGRectZero];
+    [self setupTextView];
```
ianegordon, line 80: [self setupTextView] is in both init* functions.  Move to commonInit?

components/TextFields/src/MDCMultilineTextField.m
```diff
@@ -685,22 +707,17 @@ - (void)setTextInsetsMode:(MDCTextInputTextInsetsMode)textInsetsMode {
 }
 
 - (UITextView *)textView {
-  if (!_textView) {
-    _textView = [[UITextView alloc] initWithFrame:CGRectZero];
-    [self setupTextView];
-    [self setupUnderlineConstraints];
-  }
+  [self textViewSafetyCheck];
```
ianegordon: The previous implementation lazy-init'd out textView.
Do we no longer need to do this?  (textViewSafetyCheck does not lazy-init)

components/TextFields/src/MDCMultilineTextField.m
```diff
@@ -685,22 +707,17 @@ - (void)setTextInsetsMode:(MDCTextInputTextInsetsMode)textInsetsMode {
...
   return _textView;
 }
 
 - (void)setTextView:(UITextView *)textView {
```
ianegordon: Add a comment explaining the magic here.  Perhaps something along the lines of:
We silently replace the textView with an MDCIntrinsicHeightTextView version because x, y and z.

I can see people getting confused when they assume that the textView they are setting here end up NOT being the textView in the component. :/

ianegordon (dismissed):

components/TextFields/src/MDCMultilineTextField.m
willlarche: This one cannot be moved. It is a dependency for the fundament which is a dependency for the commonInit. I will document that.

components/TextFields/src/MDCMultilineTextField.m
willlarche: It's hard because it also makes sure that it isn't nil.

components/TextFields/src/MDCMultilineTextField.m
willlarche: I was a little torn about those because of all the styling we apply. But I think you're right. The MDCMultilineTextField element is supposed to be as dumb as possible and that should include respecting whatever is set on its junk.

components/TextFields/src/MDCMultilineTextField.m
willlarche: Yes it does.
`  if (!_textView) {
    _textView = [[MDCIntrinsicHeightTextView alloc] initWithFrame:CGRectZero];
`

components/TextFields/src/MDCMultilineTextField.m
willlarche: We could but we were trying to keep that fact private so people just see UITextView in the API instead of a class they have never heard of.

components/TextFields/src/private/MDCIntrinsicHeightTextView.m
willlarche: Correcting.

components/TextFields/src/MDCMultilineTextField.m
willlarche: Done.

components/TextFields/src/MDCMultilineTextField.m
willlarche: Ugh. This isn't gonna work. I'm going to enforce they use my special class and be very verbose with instructions and asserts. This looks terrible!

components/TextFields/src/MDCMultilineTextField.m
willlarche: Added the comment.

willlarche:
Approach of hidden class was creating NUTS code. We'll just have to enforce it. That's not too crazy. PTAL

components/TextFields/src/MDCMultilineTextField.m
ianegordon: SGTM.

ianegordon (approved):

ianegordon:
Thank you @willlarche !

PR #2492, opened 2017-11-20
[TextFields] Placeholder post-floating, color fix.

### LINKED ISSUE #2482: Floating label text fields not reverting placeholder color to it's original after selection.
### Overview
TextField placeholder (with floating label) does not revert to it's original color after being selected and then unselected. 
Before selection:
screen shot 2017-11-17 at 5 11 37 pm
After selection:
screen shot 2017-11-17 at 5 11 22 pm
The after selection screenshot shows that the 2nd text field which has a floating label does not appear the same after selection.  

### Reproduction steps
Run Catalog, tap Text Field, tap Kitchen Sink, tap the second text field (labelled "This is filled and floating"), tap on any other text field to deselect.

### Version number
40.0.3

iPhone 6 with iOS 8.1 and iOS 11.1

willlarche:
Closes #2482

### DIFF
--- a/components/TextFields/src/MDCTextInputControllerDefault.m
+++ b/components/TextFields/src/MDCTextInputControllerDefault.m
@@ -527,12 +527,11 @@ static UITextFieldViewMode _underlineViewModeDefault = UITextFieldViewModeWhileE
 
   // We do this beforehand to flush the layout engine.
   [self.textInput layoutIfNeeded];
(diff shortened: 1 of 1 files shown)


romoore (approved):
Is there a unit test that can cover similar behavior?

willlarche:
As far as I know I can’t unit test becoming first responder.

PR #2515, opened 2017-11-26
[TextFields] Updating encoding and copying.

### LINKED ISSUE #1709: [TextFields] initWithCoder should check for existence of key
If the default value is non-zero we should check for the existence of the key before taking its value.

```
if ([aDecoder containsValueForKey:MDCTrailingViewKey]) {
  _trailingViewMode = (UITextFieldViewMode)[aDecoder decoderIntergerForKey:MDCTrailingViewKey];
}
```

willlarche:
Closes #1709

### DIFF
--- a/components/TextFields/src/private/MDCPaddedLabel.m
+++ b/components/TextFields/src/private/MDCPaddedLabel.m
@@ -16,7 +16,22 @@
 
 #import "MDCPaddedLabel.h"
 
+static NSString *const MDCPaddedLabelHorizontalPaddingKey = @"MDCPaddedLabelHorizontalPaddingKey";
+
 @implementation MDCPaddedLabel
+
+- (instancetype)initWithCoder:(NSCoder *)aDecoder {
+  self = [super initWithCoder:aDecoder];
+  if (self) {
+    _horizontalPadding = (CGFloat)[aDecoder decodeFloatForKey:MDCPaddedLabelHorizontalPaddingKey];
+  }
+  return self;
+}
+
+- (void)encodeWithCoder:(NSCoder *)aCoder {
+  [super encodeWithCoder:aCoder];
+  [aCoder encodeFloat:(float)_horizontalPadding forKey:MDCPaddedLabelHorizontalPaddingKey];
+}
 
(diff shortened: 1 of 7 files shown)


components/TextFields/tests/unit/TextFieldTests.swift
```diff
@@ -153,31 +153,22 @@ class TextFieldTests: XCTestCase {
...
-
     textField.cursorColor = .white
 
     let leadingView = UILabel()
     leadingView.text = "$"
-
     textField.leadingView = leadingView
     textField.leadingViewMode = .unlessEditing
 
+    textField.trailingUnderlineLabel.text = "%@#^#"
```
jgunaratne, line 163: What's going on with this string?

jgunaratne (approved):

components/TextFields/tests/unit/TextFieldTests.swift
willlarche: Just a testing string. NBD.

components/TextFields/src/private/MDCPaddedLabel.m
```diff
@@ -16,8 +16,23 @@
...
+  self = [super initWithCoder:aDecoder];
+  if (self) {
+    _horizontalPadding = (CGFloat)[aDecoder decodeFloatForKey:MDCPaddedLabelHorizontalPaddingKey];
+  }
+  return self;
+}
+
+- (void)encodeWithCoder:(NSCoder *)aCoder {
+  [super encodeWithCoder:aCoder];
+  [aCoder encodeFloat:(float)_horizontalPadding forKey:MDCPaddedLabelHorizontalPaddingKey];
```
ianegordon, line 33: Should we always be encoding this as a double so we don't lose precision on 64-bit platforms?

willlarche: We could! I took inspiration for float casting it from all the other places we usually do CGFloats. If you look around our codebase, there's a lot of `.f`s to avoid the lost of precision warning. Figured this did just like that.

PR #2523, opened 2017-11-27
[ShadowLayer] Animate shadow resizing

samnm:
This PR implicitly animates shadows when the bounds.size of the shadow layer are animated.

This change fixes the animation issue highlighted in the ZShadow demo.

Related to #1287


components/ShadowLayer/src/MDCShadowLayer.m
```diff
@@ -304,25 +334,87 @@ - (void)commonLayoutSublayers {
...
+    // won't know any properties of the animation block until it is commited.
+    MDCPendingAnimation *pendingAnim = [[MDCPendingAnimation alloc] init];
+    pendingAnim.animation = animation;
+
+    return pendingAnim;
+  }
+  return [super actionForKey:key];
+}
+
+- (MDCPendingAnimation *)pendingAnimationWithKeyPath:(NSString *)keyPath fromValue:(id)fromValue {
```
jverkoey: This method is a duplicate?

components/ShadowLayer/src/MDCShadowLayer.m
```diff
@@ -304,25 +334,87 @@ - (void)commonLayoutSublayers {
...
+  _shadowPathIsInvalid = NO;
+}
+
+@end
+
+@implementation MDCAnimatedShapeLayer
+
+- (id<CAAction>)actionForKey:(NSString *)key {
+  if ([key isEqualToString:@"path"] || [key isEqualToString:@"shadowPath"]) {
+    CABasicAnimation *animation = [CABasicAnimation animationWithKeyPath:key];
```
jverkoey: Did you test this using the spring animation APIs? I'm curious whether we should be copying the bounds animation (so that we also spring these properties) or copying its properties (as you're doing now).

jverkoey (changes requested):
Looks good at a first pass, just a few small changes requested..

components/ShadowLayer/src/MDCShadowLayer.m
```diff
@@ -304,25 +334,87 @@ - (void)commonLayoutSublayers {
...
     } else {
       _topShadow.shadowPath = [self defaultShadowPath].CGPath;
     }
   }
+  _shadowPathIsInvalid = NO;
+}
+
+@end
+
+@implementation MDCAnimatedShapeLayer
```
jverkoey: You can remove this intermediary object if you implement CALayerDelegate on `MDCShadowLayer` and you move `actionForKey:`'s logic into `actionForLayer:forKey:`.

components/ShadowLayer/src/MDCShadowLayer.m
samnm: Oops, it was. Removed.

components/ShadowLayer/src/MDCShadowLayer.m
jverkoey: Each of the layers would then set their layer delegate to the MDCShadowLayer instance and you'd avoid having to manage parent/child relationships (you'll still need to store a reference to the shadow layer in the pending animation instance though).

components/ShadowLayer/src/MDCShadowLayer.m
samnm: The existing behavior won't work with springs. I'm testing copying spring animations and it isn't working as expected so I'll have to explore that a bit more. Copying may be better - I don't have a strong preference either way.

components/ShadowLayer/src/MDCShadowLayer.m
samnm: Good to know. I'll give this a go.

components/ShadowLayer/src/MDCShadowLayer.m
jverkoey: May be ok to punt on that for now. This PR is an improvement over the existing behavior as-is.

jverkoey:
Other things to test:

- [x] With springs (may not be working as desired).
- [x] With delay.
- [ ] With user interaction enabled and long durations so that you can change the bounds multiple times while an animation is in-flight (this will test how additive animations/interruptions fare).

components/ShadowLayer/src/MDCShadowLayer.m
```diff
@@ -304,25 +334,77 @@ - (void)commonLayoutSublayers {
...
+    if ([boundsAction isKindOfClass:[CABasicAnimation class]]) {
+      [self configureAnimation:(CABasicAnimation *)boundsAction];
+      [layer addAnimation:self.animation forKey:event];
+    }
+  }
+}
+
+- (void)configureAnimation:(CAAnimation *)template {
+  self.animation.autoreverses = template.autoreverses;
+  self.animation.beginTime = template.beginTime;
```
jverkoey: This line may prove to be problematic if you intend to support delayed animations. Unfortunately, UIKit implements delayed animations using a hidden API on CAAnimation ([beginTimeMode](https://github.com/JaviSoto/iOS10-Runtime-Headers/blob/master/Frameworks/QuartzCore.framework/CAAnimation.h#L15)), which they're able to set to "relative". When you set it, however, it'll get set to "absolute", which will almost always be incorrect.

So here you have to either offset the time to match the layer's time, or copy the animation so that you can inherit the private details of beginTime. The problem with assuming that beginTime is relative however is that sometimes someone might actually have already done the time manipulations for you! In other words: there's no safe way to read beginTime. But! That's more the exception than the rule, so it might be ok in this case to just assume that beginTime will be relative if you don't want to do animation copying.

---

To offset time:

```objc
begimTime = [layer convertTime:CACurrentMediaTime() fromLayer:nil] + beginTime
```

components/ShadowLayer/src/MDCShadowLayer.m
```diff
@@ -304,25 +334,77 @@ - (void)commonLayoutSublayers {
...
+@implementation MDCPendingAnimation
+
+- (void)runActionForKey:(NSString *)event object:(id)anObject arguments:(NSDictionary *)dict {
+  if ([anObject isKindOfClass:[MDCAnimatedShapeLayer class]]) {
+    MDCAnimatedShapeLayer *layer = (MDCAnimatedShapeLayer *)anObject;
+
+    // In order to synchronize our animation with UIKit animations we have to fetch the resizing
+    // animation created by UIKit and copy the configuration to our custom animation.
+    CALayer *animatedLayer = layer.animationParentLayer;
+    CAAnimation *boundsAction = [animatedLayer animationForKey:@"bounds.size"];
```
jverkoey: UIKit will create animations that are additive by default, so if you create your own animation here and attempt to animate it non-additively, there's a decently high likelihood that there are multiple bounds animations associated with the layer if the user tapped quickly. This will be most problematic if user interactions are enabled on the view.

e.g.:

```
bounds.origin=<CABasicAnimation: 0x60400023a020> bounds.size=<CABasicAnimation: 0x60400003e0e0> bounds.origin-2=<CABasicAnimation: 0x60000022f540> bounds.size-2=<CABasicAnimation: 0x60000022f5c0>
```

I don't know of a clean solution here unfortunately, but it's something to consider.

jverkoey: Also this isn't necessarily a new problem in this PR, so I'm not proposing that it need to be fixed here.

jverkoey:
As a completely left-field suggestion, much of the work that's going on here could potentially be alleviated by migrating this API over to a UIView-based set of APIs. If all of the layers were backing a UIView then I think we'd get most of this animation logic for free. Though I do understand that such an approach may introduce more problems than it's solving...so it's a tough call.

jverkoey (approved):
Am going to LGTM this because it's a significant improvement over the existing buggy behavior. Feel free to do the delegate logic refactoring in a follow-up PR if desired.

ianegordon:
LGTM as a (massive) improvement.

In order to repro the spring issue:
Open ZShadow.
Click a square to start the animation.
Click the square again before the animation finished.
Note the system shadow stays with it's parent, our shadow does not.

ianegordon (approved):

components/ShadowLayer/src/MDCShadowLayer.m
samnm: I've changed the implementation to copy the animation instance.

samnm:
OK I'm going to submit this but there are a few issues we need to resolve eventually.

* Doesn't work with Spring animations
* Doesn't work with additive animations (i.e. user interaction)

components/ShadowLayer/src/MDCShadowLayer.m
```diff
@@ -359,33 +367,15 @@ - (void)commonLayoutSublayers {
...
 - (void)runActionForKey:(NSString *)event object:(id)anObject arguments:(NSDictionary *)dict {
-  if ([anObject isKindOfClass:[MDCAnimatedShapeLayer class]]) {
-    MDCAnimatedShapeLayer *layer = (MDCAnimatedShapeLayer *)anObject;
+  if ([anObject isKindOfClass:[CAShapeLayer class]]) {
+    CAShapeLayer *layer = (CAShapeLayer *)anObject;
 
     // In order to synchronize our animation with UIKit animations we have to fetch the resizing
     // animation created by UIKit and copy the configuration to our custom animation.
-    CALayer *animatedLayer = layer.animationParentLayer;
+    CALayer *animatedLayer = (CALayer *)layer.delegate;
```
jverkoey: This is depending on an implementation detail here and won’t be caught if someone changes the delegate in the code above. Can you get access to the parent layer by storing it in the pending animation?

jverkoey (dismissed):

components/ShadowLayer/src/MDCShadowLayer.m
samnm: Good point. Done.

jverkoey (approved):

ianegordon:
Excellent! 🍾

PR #2528, opened 2017-11-28
[Buttons] Add extended FAB layout code

romoore:
Creating (internal for now) the required properties and layout code to support
the extended FAB. Unit tests to confirm the behavior and some minor
adjustments to existing values like default contentEdgeInsets values.

Implements #2495

### DIFF
--- a/components/Buttons/src/MDCFloatingButton.m
+++ b/components/Buttons/src/MDCFloatingButton.m
@@ -31,10 +31,6 @@ static NSString *const MDCFloatingButtonModeKey = @"MDCFloatingButtonModeKey";
 static NSString *const MDCFloatingButtonImageLocationKey = @"MDCFloatingButtonImageLocationKey";
 static NSString *const MDCFloatingButtonImageTitleSpacingKey = @"MDCFloatingButtonImageTitleSpacingKey";
 
-static UIEdgeInsets UIEdgeInsetsFlippedHorizonally(UIEdgeInsets insets) {
-  return UIEdgeInsetsMake(insets.top, insets.right, insets.bottom, insets.left);
-}
-
 // TODO(rsmoore): All properties within here will be made public in a future PR.
 @interface MDCFloatingButton ()
 
@@ -118,10 +114,16 @@ static UIEdgeInsets UIEdgeInsetsFlippedHorizonally(UIEdgeInsets insets) {
     if (@(_shape).integerValue >= 2) {
       _shape = MDCFloatingButtonShapeDefault;
     }
-    _mode = [aDecoder decodeIntegerForKey:MDCFloatingButtonModeKey];
-    _imageLocation = [aDecoder decodeIntegerForKey:MDCFloatingButtonImageLocationKey];
-    _imageTitleSpacing =
-        (CGFloat)[aDecoder decodeDoubleForKey:MDCFloatingButtonImageTitleSpacingKey];
+    if ([aDecoder containsValueForKey:MDCFloatingButtonModeKey]) {
+      _mode = [aDecoder decodeIntegerForKey:MDCFloatingButtonModeKey];
+    }
+    if ([aDecoder containsValueForKey:MDCFloatingButtonImageLocationKey]) {
+      _imageLocation = [aDecoder decodeIntegerForKey:MDCFloatingButtonImageLocationKey];
+    }
+    if ([aDecoder containsValueForKey:MDCFloatingButtonImageTitleSpacingKey]) {
+      _imageTitleSpacing =
+          (CGFloat)[aDecoder decodeDoubleForKey:MDCFloatingButtonImageTitleSpacingKey];
+    }
   }
   return self;
 }
@@ -222,7 +224,7 @@ static UIEdgeInsets UIEdgeInsetsFlippedHorizonally(UIEdgeInsets insets) {
   BOOL isLeadingIcon = self.imageLocation == MDCFloatingButtonImageLocationLeading;
   UIEdgeInsets adjustedLayoutInsets =
       isLeadingIcon ? internalLayoutSpacingInsets
-                    : UIEdgeInsetsFlippedHorizonally(internalLayoutSpacingInsets);
+                    : MDFInsetsFlippedHorizontally(internalLayoutSpacingInsets);
 
   const CGRect insetBounds = UIEdgeInsetsInsetRect(UIEdgeInsetsInsetRect(self.bounds,
                                                                          adjustedLayoutInsets),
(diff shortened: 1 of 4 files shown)


components/Buttons/src/MDCFloatingButton.m
```diff
@@ -71,55 +112,171 @@ - (instancetype)initWithFrame:(CGRect)frame shape:(MDCFloatingButtonShape)shape
 - (instancetype)initWithCoder:(NSCoder *)aDecoder {
   self = [super initWithCoder:aDecoder];
   if (self) {
+    [self commonMDCFloatingButtonInit];
```
samnm, line 111: You are always encoding all properties so no need for commonInit here.

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -19,9 +19,49 @@
...
 static const CGFloat MDCFloatingButtonMiniDimension = 40.0f;
+static const CGFloat MDCFloatingButtonDefaultImageTitleSpacing = 8;
+static const UIEdgeInsets internalLayoutSpacingInsets = (UIEdgeInsets){0, 16, 0, 24};
+
 static NSString *const MDCFloatingButtonShapeKey = @"MDCFloatingButtonShapeKey";
+static NSString *const MDCFloatingButtonModeKey = @"MDCFloatingButtonModeKey";
+static NSString *const MDCFloatingButtonImageLocationKey = @"MDCFloatingButtonImageLocationKey";
+static NSString *const MDCFloatingButtonImageTitleSpacingKey = @"MDCFloatingButtonImageTitleSpacingKey";
+
+static UIEdgeInsets UIEdgeInsetsFlippedHorizonally(UIEdgeInsets insets) {
```
ianegordon: FYI: This also exists as MDFInsetsFlippedHorizontally() in MDFI18n.  Also this will cause an error if our friends in Cupertino ever decide to declare UIEdgeInsetsFlippedHorizonally.  Although they may throw in a 't' 😄

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -19,9 +19,49 @@
...
+  return UIEdgeInsetsMake(insets.top, insets.right, insets.bottom, insets.left);
+}
+
+// TODO(rsmoore): All properties within here will be made public in a future PR.
```
ianegordon: Super-nit: Consider opening an issue and making this "TODO(#3218): All properties within here will be made public in a future PR."

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -71,55 +112,171 @@ - (instancetype)initWithFrame:(CGRect)frame shape:(MDCFloatingButtonShape)shape
...
+  CGPoint titleCenter;
+  CGPoint imageCenter;
+  BOOL isRTL = self.mdf_effectiveUserInterfaceLayoutDirection
+      == UIUserInterfaceLayoutDirectionRightToLeft;
+
+  // isRTL && isLeadingIcon => icon to right of text
+  // !isRTL && !isLeadingIcon => icon to right of text
+  // isRTL && !isLeadingIcon => icon to left of text
+  // !isRTL && isLeadingIcon => icon to left of text
+  if (isRTL == isLeadingIcon) {
```
ianegordon: Consider being a bit more verbose in favor of obvious, and also consider prioritizing the commonly-used path:

// If we are in Left-To-Right and the icon is leading -OR- if we are in Right-To-Left and the icon is trailing
// layout the image to the left of the label.  Otherwise layout the image to the right of the label.
if ( (ifLTR && isLeadingIcon) ||
      (!ifLTR && !isLeadingIcon) {
   // Layout image to the left of the label
   ....
} else {
   // Layout image to the right of the label
   ....
}

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -71,55 +112,171 @@ - (instancetype)initWithFrame:(CGRect)frame shape:(MDCFloatingButtonShape)shape
...
   if (self) {
+    [self commonMDCFloatingButtonInit];
     _shape = [aDecoder decodeIntegerForKey:MDCFloatingButtonShapeKey];
     // Required to migrate any previously-archived FloatingButtons from .largeIcon shape value
     if (@(_shape).integerValue >= 2) {
       _shape = MDCFloatingButtonShapeDefault;
     }
+    _mode = [aDecoder decodeIntegerForKey:MDCFloatingButtonModeKey];
+    _imageLocation = [aDecoder decodeIntegerForKey:MDCFloatingButtonImageLocationKey];
+    _imageTitleSpacing =
```
ianegordon: Consider checking for the existence of the key before decoding _ivars.  Especially any with a default value != 0.

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -71,55 +112,171 @@ - (instancetype)initWithFrame:(CGRect)frame shape:(MDCFloatingButtonShape)shape
...
+    const CGFloat imageCenterX
+        = CGRectGetMaxX(insetBounds) - (imageViewWidth / 2);
+    const CGFloat titleCenterX = CGRectGetMinX(insetBounds) + (titleWidthAvailable / 2);
+    imageCenter = CGPointMake(imageCenterX, boundsCenterY);
+    titleCenter = CGPointMake(titleCenterX, boundsCenterY);
+  } else {
+    const CGFloat imageCenterX
+        = CGRectGetMinX(insetBounds) + (imageViewWidth / 2);
+    const CGFloat titleCenterX
+        = CGRectGetMaxX(insetBounds) - (titleWidthAvailable / 2);
```
samnm: Nit: These assignments don't need to wrap multiple lines.

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -71,55 +112,171 @@ - (instancetype)initWithFrame:(CGRect)frame shape:(MDCFloatingButtonShape)shape
...
+
+  // Position the imageView and titleView
+  //
+  // +------------------------------------+
+  // |    |  |  |  CEI TOP            |   |
+  // |CEI +--+  |       +-----+       |CEI|
+  // | LT ||SP|-- A --|Title|-- A --|RGT|
+  // |    +--+  |       +-----+       |   |
+  // |    |  |  |  CEI BOT            |   |
+  // +------------------------------------+
```
ianegordon, line 215: Nice ascii-work 👍

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -71,55 +112,171 @@ - (instancetype)initWithFrame:(CGRect)frame shape:(MDCFloatingButtonShape)shape
...
 #pragma mark - Subclassing
 
 - (UIEdgeInsets)defaultContentEdgeInsets {
-  switch (_shape) {
-    case MDCFloatingButtonShapeDefault:
-      return UIEdgeInsetsZero;
-    case MDCFloatingButtonShapeMini:
-      return UIEdgeInsetsMake(8, 8, 8, 8);
-  }
+  return UIEdgeInsetsZero;
```
samnm, line 279: Will this change the behavior of any existing clients?

ianegordon (approved):

samnm (approved):
LGTM with a couple of Qs.

components/Buttons/src/MDCFloatingButton.m
samnm: Won't the keys always exist?

components/Buttons/src/MDCFloatingButton.m
romoore: I figured since that would be the next PR on this feature branch, it was OK to just remark for my reviewers. I can open an issue if you like.

components/Buttons/src/MDCFloatingButton.m
romoore: Not from Interface Builder, they wouldn't. And in that case the defaults would be lost. Good catch!

samnm: Ah I didn't think of that.

components/Buttons/src/MDCFloatingButton.m
romoore: I think it would if they're doing any blind arithmetic to supplement these. However, having all edge insets equal is equivalent to having them zero EXCEPT that it would start shrinking images, truncating text, etc.

components/Buttons/src/MDCFloatingButton.m
romoore: I think it makes sense to leave the `commonInit` call here in case the button is unarchived from a storyboard/xib. In that case we explicitly initialize defaults. Although properties are assigned zeroed memory, I think it's safer to use explicit assignments so any future changes to default values can be easily identified with a string search.

screen shot 2017-11-29 at 12 56 01 pm

components/Buttons/src/MDCFloatingButton.m

components/Buttons/src/MDCFloatingButton.m

components/Buttons/src/MDCFloatingButton.m

components/Buttons/src/MDCFloatingButton.m

PR #2537, opened 2017-11-29
[ActivityIndicator] Add a delegate callback for the activity indicator mode animation

jmdetloff:


ianegordon (commented):
This implementation seems completely reasonable, but do we really need this callback?
(There may be an internal reason I'm not aware of making the answer to this question a resounding yes)

activityIndicatorAnimationDidFinish:
activityIndicatorModeAnimationDidFinish:
seem nearly identical.

ianegordon (approved):
LGTM.
Please merge when CI is happy.

PR #2545, opened 2017-11-30
[Ink] Add updated ink to chips, update ink animation when bounds change

jgunaratne:
Add updated ink to chips, update ink animation when bounds change. (Also includes changes to buttons in https://github.com/material-components/material-components-ios/pull/2530)


components/Ink/src/MDCInkView.m
```diff
@@ -56,12 +67,17 @@ - (void)commonMDCInkViewInit {
   self.userInteractionEnabled = NO;
   self.backgroundColor = [UIColor clearColor];
   self.autoresizingMask = UIViewAutoresizingFlexibleHeight | UIViewAutoresizingFlexibleWidth;
+  self.layer.delegate = self;
   self.inkColor = self.defaultInkColor;
   _usesLegacyInkRipple = YES;
+
+  // Use mask layer when the superview has a shadowPath.
+  _maskLayer = [CAShapeLayer layer];
```
samnm: `_maskLayer.delegate = self;`

components/Ink/src/MDCInkView.m
```diff
@@ -219,4 +245,44 @@ - (void)inkLayerAnimationDidEnd:(MDCInkLayer *)inkLayer {
...
 
+#pragma mark - CALayerDelegate
+
+- (id<CAAction>)actionForLayer:(CALayer *)layer forKey:(NSString *)event {
+  if ([event isEqualToString:@"path"] || [event isEqualToString:@"shadowPath"]) {
+
+    // We have to create a pending animation because if we are inside a UIKit animation block we
+    // won't know any properties of the animation block until it is commited.
+    MDCInkPendingAnimation *pendingAnim = [[MDCInkPendingAnimation alloc] init];
+    pendingAnim.animationSourceLayer = self.layer;
```
samnm: In order for this to work with simulator slow animations, you would need change `self.layer` to `self.superview.layer`.  Although that only works if you are animating the bounds of the superview. I'm not sure what the ideal approach would be here.

samnm (dismissed):

components/Ink/src/MDCInkView.m

samnm (approved):

PR #2547, opened 2017-11-30
[Tabs] Allow customization of fonts and text transform

brianjmoore:
`displaysUppercaseTitles` has been put on the road to deprecation and a replacement has been added called `titleTextTransform`. This new property is UIAppearance-compatible and less magical overall. Existing clients can switch at any time, but will not be required to immediately.

Adds two properties to customize the fonts used for selected and unselected tab titles. This allows a bold/regular or regular/thin pair of fonts (for example) to be specified to indicate selection, if desired. Two documented limitations on these:
* They only apply to top-position navigation bars. This keeps the API simple and is consistent with our longer-term direction towards recommending `MDCBottomNavigationBar` instead of bottom tab bars.
* A soft condition of using the API is that you don't set two drastically differently-sized fonts for these states. If you do, layouts may not be optimal. Reasonable designs shouldn't violate this condition.

### DIFF
--- a/components/Tabs/src/MDCTabBar.h
+++ b/components/Tabs/src/MDCTabBar.h
@@ -87,7 +87,7 @@ IB_DESIGNABLE
 
 /**
  Font used for selected item titles.
- By default this uses the MDCTypography button font. Ignored for bottom-position tab bars.
+ By default this uses +[MDCTypography buttonFont]. Ignored for bottom-position tab bars.
 
  Note: Tab sizes are determined based on their unselected state and do not vary based on selection.
  To avoid clipped layouts and other layout issues, the font provided here should have similar
@@ -134,6 +134,7 @@ IB_DESIGNABLE
  The default value is based on the position and is recommended for most applications.
 
  NOTE: This property will be deprecated in a future release. Use `titleTextTransform` instead.
+ https://github.com/material-components/material-components-ios/issues/2552
  */
 @property(nonatomic) IBInspectable BOOL displaysUppercaseTitles;
 
--- a/components/Tabs/src/MDCTabBar.h
+++ b/components/Tabs/src/MDCTabBar.h
@@ -104,13 +104,13 @@ IB_DESIGNABLE
  Padding on the left and right of the tab bar's content for scrollable tabs.
  Default depends on the horizontal size class.
  */
-@property(nonatomic) IBInspectable CGFloat horizontalPadding UI_APPEARANCE_SELECTOR;
+@property(nonatomic) CGFloat horizontalPadding UI_APPEARANCE_SELECTOR;
 
 /**
  Padding on the left and right of each item's content (title or images) for non-justified alignment.
  Default depends on the horizontal size class.
  */
-@property(nonatomic) IBInspectable CGFloat itemHorizontalPadding UI_APPEARANCE_SELECTOR;
+@property(nonatomic) CGFloat itemHorizontalPadding UI_APPEARANCE_SELECTOR;
 
 /**
  Tint color to apply to the tab bar background.
@@ -139,7 +139,7 @@ IB_DESIGNABLE
 
  The default value is MDCTabBarTextTransformAutomatic.
  */
-@property(nonatomic) IBInspectable MDCTabBarTextTransform titleTextTransform UI_APPEARANCE_SELECTOR;
+@property(nonatomic) MDCTabBarTextTransform titleTextTransform UI_APPEARANCE_SELECTOR;
 
 /**
  Template that defines the appearance of the selection indicator.
--- a/components/Tabs/src/MDCTabBar.h
+++ b/components/Tabs/src/MDCTabBar.h
@@ -86,19 +86,16 @@ IB_DESIGNABLE
 @property(nonatomic, nonnull) UIColor *inkColor UI_APPEARANCE_SELECTOR;
 
 /**
- Text attributes used for selected item titles.
- Default uses the regular font from the font loader.
+ Font used for selected item titles.
+ By default uses the regular font from the font loader.
  */
-@property(nonatomic, strong, nonnull)
-    NSDictionary<NSAttributedStringKey, id> *selectedItemTitleTextAttributes UI_APPEARANCE_SELECTOR;
+@property(nonatomic, strong, nonnull) UIFont *selectedItemTitleFont UI_APPEARANCE_SELECTOR;
 
 /**
- Text attributes used for unselected item titles.
- Default uses the regular font from the font loader.
+ Font used for unselected item titles.
+ By default uses the regular font from the font loader.
  */
-@property(nonatomic, strong, nonnull)
-    NSDictionary<NSAttributedStringKey, id> *unselectedItemTitleTextAttributes
-    UI_APPEARANCE_SELECTOR;
+@property(nonatomic, strong, nonnull) UIFont *unselectedItemTitleFont UI_APPEARANCE_SELECTOR;
 
 /**
  Padding on the left and right of the tab bar's content for scrollable tabs.
--- a/components/Tabs/src/MDCTabBar.h
+++ b/components/Tabs/src/MDCTabBar.h
@@ -98,10 +98,10 @@ IB_DESIGNABLE
     UI_APPEARANCE_SELECTOR;
 
 /**
- Spacing between tab bar content (titles or images) for non-justified alignments.
+ Padding on the left and right of each item's content (title or images) for non-justified alignment.
  Default depends on the horizontal size class.
  */
-@property(nonatomic) IBInspectable CGFloat itemSpacing UI_APPEARANCE_SELECTOR;
+@property(nonatomic) IBInspectable CGFloat itemHorizontalPadding UI_APPEARANCE_SELECTOR;
 
 /**
  Tint color to apply to the tab bar background.
--- a/components/Tabs/src/MDCTabBar.h
+++ b/components/Tabs/src/MDCTabBar.h
@@ -85,16 +85,18 @@ IB_DESIGNABLE
 @property(nonatomic, nonnull) UIColor *inkColor UI_APPEARANCE_SELECTOR;
 
 /**
- Display font used for selected item titles.
- Default is the regular font from the font loader.
+ Text attributes used for selected item titles.
+ Default uses the regular font from the font loader.
  */
-@property(nonatomic, strong, null_resettable) UIFont *selectedItemTitleFont UI_APPEARANCE_SELECTOR;
+@property(nonatomic, strong, nonnull)
+    NSDictionary<NSAttributedStringKey, id> *selectedItemTitleTextAttributes UI_APPEARANCE_SELECTOR;
 
 /**
- Display font used for unselected item titles.
- Default is the regular font from the font loader.
+ Text attributes used for unselected item titles.
+ Default uses the regular font from the font loader.
  */
-@property(nonatomic, strong, null_resettable) UIFont *unselectedItemTitleFont
+@property(nonatomic, strong, nonnull)
+    NSDictionary<NSAttributedStringKey, id> *unselectedItemTitleTextAttributes
     UI_APPEARANCE_SELECTOR;
 
 /**
--- a/components/Tabs/src/MDCTabBar.h
+++ b/components/Tabs/src/MDCTabBar.h
@@ -87,13 +87,13 @@ IB_DESIGNABLE
 
 /**
  Font used for selected item titles.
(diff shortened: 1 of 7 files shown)


components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -558,7 +560,13 @@ - (void)updateTransformsAnimated:(BOOL)animated {
 }
 
 - (void)updateTitleFont {
-  _titleLabel.font = _style.titleFont;
+  UIFont *font = nil;
```
ianegordon: super-SUPER-nit: "= nil" is redundant.

ianegordon (approved):
LGTM.
Most nittiest of nits.

components/Tabs/src/MDCTabBar.h
```diff
@@ -84,6 +85,26 @@ IB_DESIGNABLE
...
 
+/**
+ Font used for selected item titles.
+ By default this uses the MDCTypography button font. Ignored for bottom-position tab bars.
+
+ Note: Tab sizes are determined based on their unselected state and do not vary based on selection.
+ To avoid clipped layouts and other layout issues, the font provided here should have similar
+ metrics to `unselectedItemTitleFont`.
+ */
+@property(nonatomic, strong, nonnull) UIFont *selectedItemTitleFont UI_APPEARANCE_SELECTOR;
```
romoore, line 96: Any reason not to proxy [an `NSDictionary *` of text attributes](https://developer.apple.com/library/content/documentation/StringsTextFonts/Conceptual/TextAndWebiPhoneOS/CustomTextProcessing/CustomTextProcessing.html#//apple_ref/doc/uid/TP40009542-CH4-SW24)?  I can't find any UIKit classes proxying a `UIFont *` directly, but plenty that proxy dictionaries for attributed string handling that includes the font.

components/Tabs/src/MDCTabBar.h
```diff
@@ -84,6 +85,26 @@ IB_DESIGNABLE
 /** Ink color for taps on tab bar items. Default: Semi-transparent white. */
 @property(nonatomic, nonnull) UIColor *inkColor UI_APPEARANCE_SELECTOR;
 
+/**
+ Font used for selected item titles.
+ By default this uses the MDCTypography button font. Ignored for bottom-position tab bars.
```
romoore: Super-nit: Consider referencing the class method `+[MDCTypography buttonFont]` instead of using prose.

components/Tabs/src/MDCTabBar.h
```diff
@@ -107,13 +128,22 @@ IB_DESIGNABLE
...
 
 /**
- Indicates if all tab titles should be uppercased for display. If NO, item titles will be
- displayed verbatim.
+ Indicates if all tab titles should be uppercased for display. If NO, item titles will be displayed
+ verbatim.
 
  The default value is based on the position and is recommended for most applications.
+
+ NOTE: This property will be deprecated in a future release. Use `titleTextTransform` instead.
```
romoore, line 136: If we really cannot deprecate in this release cycle, please create a tracking Issue and add a `TODO()` with a link.

components/Tabs/src/MDCTabBarTextTransform.h
```diff
@@ -0,0 +1,29 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+/** Appearance for content within tab bar items. */
+typedef NS_ENUM(NSInteger, MDCTabBarTextTransform) {
+  /** The default text transform is applied based on the bar's position. */
+  MDCTabBarTextTransformAutomatic,
```
romoore: Please assign numeric values for these. It helps avoid future errors with encoding/decoding if someone tries to modify the enumeration.

components/Tabs/src/MDCTabBar.m
```diff
@@ -126,13 +123,14 @@ - (void)commonMDCTabBarInit {
   _barPosition = UIBarPositionAny;
   _hasDefaultItemAppearance = YES;
   _hasDefaultAlignment = YES;
-  _hasDefaultDisplaysUppercaseTitles = YES;
 
   // Set default values
   _alignment = [self computedAlignment];
-  _displaysUppercaseTitles = [self computedDisplaysUppercaseTitles];
+  _titleTextTransform = MDCTabBarTextTransformAutomatic;
```
romoore, line 129: Nit?: Why is this property (and others) not being restored in `initWithCoder:` (or even encoded)?

components/Tabs/src/MDCTabBar.m
```diff
@@ -222,6 +220,22 @@ - (void)setInkColor:(UIColor *)inkColor {
   }
 }
 
+- (void)setUnselectedItemTitleFont:(UIFont *)unselectedItemTitleFont {
+  if ((unselectedItemTitleFont != _unselectedItemTitleFont) &&
```
romoore, line 224: Nit: Please use property getter instead of referencing the backing iVar.  Here and throughout.

components/Tabs/src/MDCTabBar.m
```diff
@@ -256,10 +270,34 @@ - (void)setUnselectedItemTintColor:(UIColor *)unselectedItemTintColor {
   }
 }
 
+- (BOOL)displaysUppercaseTitles {
+  switch (_titleTextTransform) {
```
romoore: Nit: Please use the property accessor.

components/Tabs/src/MDCTabBar.m
```diff
@@ -521,12 +541,39 @@ - (void)updateItemBarStyle {
...
   style.selectionIndicatorColor = self.tintColor;
   style.inkColor = _inkColor;
   style.selectedTitleColor = (_selectedItemTintColor ? _selectedItemTintColor : self.tintColor);
   style.titleColor = _unselectedItemTintColor;
-  style.displaysUppercaseTitles = _displaysUppercaseTitles;
+
+  // Set displaysUppercaseTitles based on titleTextTransform.
+  BOOL displaysUppercaseTitles = NO;
+  switch (_titleTextTransform) {
+    case MDCTabBarTextTransformAutomatic:
```
romoore: Can we use `- (BOOL)displaysUppercaseTitles` from above? This looks like the same logic.

components/Tabs/src/private/MDCItemBarCell.m
romoore: You could also replace this method with

```objectivec
- (void)updateTitleFont {
  _titleLabel.font = self.isSelected ? _style.selectedTitleFont : _style.unselectedTitleFont;
}
```

components/Tabs/src/private/MDCItemBarStyle.h
```diff
@@ -58,8 +58,11 @@
 /** Color of title text when selected. Default is opaque white. */
 @property(nonatomic, strong, nonnull) UIColor *selectedTitleColor;
 
-/** Font used for item titles. */
-@property(nonatomic, nonnull) UIFont *titleFont;
+/** Font used for selected item titles. */
+@property(nonatomic, nonnull) UIFont *selectedTitleFont;
```
romoore: Nit: Please specify `strong` or `weak` retain semantics.

components/Tabs/src/MDCTabBar.m
```diff
@@ -521,12 +541,39 @@ - (void)updateItemBarStyle {
 
   style = [[self class] defaultStyleForPosition:_barPosition itemAppearance:_itemAppearance];
 
+  if ([MDCTabBar isTopTabsForPosition:_barPosition]) {
+    // Top tabs: Use provided fonts.
+    style.selectedTitleFont = _selectedItemTitleFont;
+    style.unselectedTitleFont = _unselectedItemTitleFont;
+  } else {
+    // Bottom navigation: Ignore provided fonts.
+    style.selectedTitleFont = [[MDCTypography fontLoader] regularFontOfSize:12];
```
romoore: If `style.titleFont` is no longer used, should it be deprecated/removed?

romoore (dismissed):
Please see previous review comments. Forgot to select "Request Changes" option.

components/Tabs/src/MDCTabBar.h
brianjmoore: Ian and I discussed this offline and felt that an attributes dictionary would be overkill for this situation and we'd prefer to start simple. We have more of an ability to expand this API in the future than a platform vendor does, so I think this deviation from UIKit makes sense.

components/Tabs/src/MDCTabBar.m
brianjmoore: It has been removed, see MDCItemBarStyle.h change below. No need for deprecations, it's private.

components/Tabs/src/MDCTabBar.h
brianjmoore: I'd prefer to follow the deprecation policy for this one. Created an issue and added a link.

components/Tabs/src/MDCTabBar.h
brianjmoore: Done

components/Tabs/src/MDCTabBar.h
romoore: Ack.

components/Tabs/src/MDCTabBar.m
romoore: ah, sorry, missed that somehow. :(

components/Tabs/src/MDCTabBar.m
brianjmoore: That's a good question. It's not implemented and doing so would be out of scope here: https://github.com/material-components/material-components-ios/issues/2553

romoore: Ack.

components/Tabs/src/MDCTabBar.m
brianjmoore: Sure, but in getters and setters we should still refer to ivars to minimize unexpected side-effects and recursive calls.

romoore: Yes, a setter cannot use a setter call, and a getter cannot use a getter call. I would argue that setters should generally use getters unless there's a specific reason not to.

components/Tabs/src/MDCTabBar.m
brianjmoore: Done

components/Tabs/src/MDCTabBar.m
brianjmoore: Good catch. Done

components/Tabs/src/MDCTabBarTextTransform.h
brianjmoore: Done

components/Tabs/src/private/MDCItemBarStyle.h
brianjmoore: Done

components/Tabs/src/MDCTabBar.m
brianjmoore: In my view the implementation of a setter is explicitly modifying the ivar, so it should use the ivar consistently throughout.

components/Tabs/src/MDCTabBar.m
```diff
@@ -256,10 +270,34 @@ - (void)setUnselectedItemTintColor:(UIColor *)unselectedItemTintColor {
...
+    case MDCTabBarTextTransformUppercase:
+      return YES;
+  }
+}
+
 - (void)setDisplaysUppercaseTitles:(BOOL)displaysUppercaseTitles {
-  _hasDefaultDisplaysUppercaseTitles = NO;
-  _displaysUppercaseTitlesOverride = displaysUppercaseTitles;
-  [self internalSetDisplaysUppercaseTitles:[self computedDisplaysUppercaseTitles]];
+  MDCTabBarTextTransform newTransform = MDCTabBarTextTransformAutomatic;
```
romoore: Nit: The .automatic value can never be used but is assigned initially.  This could be simplified to:

```objectivec
self.titleTextTransform = 
    displaysUppercaseTitles ? MDCTabBarTextTransformUppercase : MDCTabBarTextTransformNone;
```

components/Tabs/src/MDCTabBar.m
romoore: Ack.

romoore (approved):

components/Tabs/src/MDCTabBar.m
brianjmoore: Done

ianegordon:
Customize all the fonts!

PR #2560, opened 2017-12-01
[Ink] Use updated ink for bottom navigation

jgunaratne:
Use updated ink for bottom navigation

Superseded by https://github.com/material-components/material-components-ios/pull/2567


components/Ink/src/MDCInkView.m
```diff
@@ -153,16 +153,6 @@ - (void)startTouchBeganAnimationAtPoint:(CGPoint)point
     inkLayer.inkColor = self.inkColor;
     inkLayer.maxRippleRadius = self.maxRippleRadius;
     inkLayer.animationDelegate = self;
-
-    switch (self.inkStyle) {
```
randallli, line 157: Are we intentionally deleting this? It looks like a accidental deletion given the description of the PR.

randallli (changes requested):

components/Ink/src/MDCInkView.m
jgunaratne: Yes, it's an intentional deletion. inkStyle shouldn't be used with the new version of the ripple since there isn't a distinction between bounded an unbounded ripples.

PR #2563, opened 2017-12-04
[Buttons] Extended FAB

romoore:
This is the introduction of the "extended" or "expanded" form of the Floating Action Button into the Buttons component. 

Implements #2495

PR #2565, opened 2017-12-04
[Buttons] Rename to `imageTitleSpace`

romoore:
The previous name, `imageTitleSpacing` sounds too much like a protocol name,
one that would adjust the space between things.

Amends #2563
Implements #2495

### DIFF
--- a/components/Buttons/src/MDCFloatingButton.m
+++ b/components/Buttons/src/MDCFloatingButton.m
@@ -23,14 +23,14 @@
 
 static const CGFloat MDCFloatingButtonDefaultDimension = 56;
 static const CGFloat MDCFloatingButtonMiniDimension = 40;
-static const CGFloat MDCFloatingButtonDefaultImageTitleSpacing = 8;
+static const CGFloat MDCFloatingButtonDefaultImageTitleSpace = 8;
 static const UIEdgeInsets internalLayoutSpacingInsets = (UIEdgeInsets){0, 16, 0, 24};
 
 static NSString *const MDCFloatingButtonShapeKey = @"MDCFloatingButtonShapeKey";
 static NSString *const MDCFloatingButtonModeKey = @"MDCFloatingButtonModeKey";
 static NSString *const MDCFloatingButtonImageLocationKey = @"MDCFloatingButtonImageLocationKey";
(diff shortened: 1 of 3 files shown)


components/Buttons/src/MDCFloatingButton.m
```diff
@@ -23,14 +23,14 @@
 
 static const CGFloat MDCFloatingButtonDefaultDimension = 56;
 static const CGFloat MDCFloatingButtonMiniDimension = 40;
-static const CGFloat MDCFloatingButtonDefaultImageTitleSpacing = 8;
+static const CGFloat MDCFloatingButtonDefaultimageTitleSpace = 8;
 static const UIEdgeInsets internalLayoutSpacingInsets = (UIEdgeInsets){0, 16, 0, 24};
```
randallli: this one too?

randallli (dismissed):

components/Buttons/src/MDCFloatingButton.m

romoore:
Will merge against develop once the feature branch is merged.

components/Buttons/src/MDCFloatingButton.m
```diff
@@ -23,14 +23,14 @@
 
 static const CGFloat MDCFloatingButtonDefaultDimension = 56;
 static const CGFloat MDCFloatingButtonMiniDimension = 40;
-static const CGFloat MDCFloatingButtonDefaultImageTitleSpacing = 8;
+static const CGFloat MDCFloatingButtonDefaultImageTitleSpace = 8;
 static const UIEdgeInsets internalLayoutSpacingInsets = (UIEdgeInsets){0, 16, 0, 24};
```
randallli, line 27: internalLayoutInsets. "Spacing" provides no new information. Its an inset.

randallli (approved):
feel free to apply suggestion in a different PR.

components/Buttons/src/MDCFloatingButton.m
romoore: Well, they aren't actually used as insets - I'm just reusing that data structure. If it makes more sense, I can keep the values in a non-insets struct.

romoore: Changed in #2578

PR #2578, opened 2017-12-04
[Buttons] Improve FloatingButton documentation

### LINKED ISSUE #2566: [Floating Button] Improve layout documentation
### Overview
The MDCFloatingButton documentation, both public and internal, needs to be improved and clarified.   
* The headers should document the default layout of subviews
* The headers should document how the different insets will affect the layout
* The headers should document how to change sizes for the `.default` and `.mini` shapes
* The internal documentation should provide a bit more into the why/how of the layout code.
* Rename `internalLayoutSpacingInsets`. Either the name is redundant or they shouldn't be insets.

### Version number
40.1.1

### Related issues
Based on PR feedback in #2563

romoore:
* Rename `internalLayoutSpacingInsets` to `internalLayoutInsets`
* Add comments to outline the layout code in the implementation
* Update header documentation to provide more detail about how the new
  setters affect layout.

Closes #2566


components/Buttons/src/MDCFloatingButton.h
```diff
@@ -74,7 +74,16 @@ typedef NS_ENUM(NSInteger, MDCFloatingButtonImageLocation) {
 
 /**
  The mode of the floating button can either be .normal (a circle) or .expanded (a pill-shaped
- rounded rectangle).
+ rounded rectangle). In the @c .normal mode, the button should have either an image or a title,
```
jverkoey, line 77: The "shoulds" here make me nervous - is this something the client needs to be cautious not to do? Can we enforce it somehow?

jverkoey: What happens if both are provided?

components/Buttons/src/MDCFloatingButton.h
```diff
@@ -74,7 +74,16 @@ typedef NS_ENUM(NSInteger, MDCFloatingButtonImageLocation) {
...
+ but not both.  In the @c .expanded mode, the button should have both an image and a title.
+
+ The @c .normal layout is identical to that of UIButton. The content will be centered (or otherwise
+ aligned based on the @c contentHorizontalAlignment and @c contentVerticalAlignment properties. In
+ @c .expanded layout, the image view will be inset from the leading edge (or trailing edge when
+ @c imageLocation is .trailing). The "bounding box" for the title will be inset from the opposite
+ edge and separated from @c imageView by @c imageTitleSpace and the title label will be centered
+ within this box. In @c .expanded mode, the @c contentVerticalAlignment and
+ @c contentHorizontalAlignment properties are ignored.
```
jverkoey, line 86: Great explanation!

jverkoey (approved):
LGTM with minor feedback on guidance wording.

components/Buttons/src/MDCFloatingButton.h
romoore: It's based on Material Design guidelines. I don't think we should force an either-or in this case. It's possible that the client is changing the minimum sizes for `.normal` and can actually fit both. Alternatively, they may be adjusting insets so that the image and text align with one another.

Here's with a standard image + title ("Plus")
fab-image title

Here's one where the insets are adjusted to (poorly) align the image/text. It does violate the design guidelines, but I don't think there's much to be gained by making it hard to achieve.

fab-image title-insets

A more creative button might have the image and text go well together, unicode has a lot of cool glyphs!

fab-image title-glyph

PR #2582, opened 2017-12-04
[TextFields] Base controller class files

willlarche:
This is 95% of what used to be called `-Default` controller.


ianegordon:
Kokoro seems unhappy :/
Unable to find 'private/MDCTextInputControllerBase+Subclassing.h'

ianegordon (approved):

ianegordon:
Testing will fail until entire chain lands.

PR #2597, opened 2017-12-05
[Buttons] Rename .mode as .mdc_expandedMode

romoore:
MDCFloatingButton has a new property to control whether it is an Extended
Floating Action Button (FAB) or a traditional (normal) FAB. The property name
"mode" conflicted with an internal client and in any case does not align with
Apple's recommendations for method/property names in subclasses and
categories.


romoore:
@ianegordon  Naming is hard. Please take a look.

ianegordon (changes requested):
Let's investigate making the change to the internal client.  This seems,.... less than ideal.

romoore:
Internal client resolved the naming conflict.

PR #2601, opened 2017-12-05
[Themes] Add BUILD file for kokoro

romoore:


ajsecord (approved):
Looks like we're only waiting on Travis, and Travis doesn't care about BUILD, so I'm forcing this through.

PR #2480, opened 2017-11-17
[CocoaPods] Added warnings to examples.

randallli:
Turn on warnings from lib for example code.

requires: https://github.com/material-components/material-components-ios/pull/2426 and https://github.com/material-components/material-components-ios/pull/2456

PR #2603, opened 2017-12-05
Fixed method collision build errors

randallli:
multiple methods named 'count' found

Found by turning on warnings via: https://github.com/material-components/material-components-ios/pull/2480


ianegordon (approved):
Generics FTW!
Thank you @randallli !

PR #2619, opened 2017-12-06
[Shapes] Implicitly animate shape path on resizing

samnm:


components/ShadowLayer/src/MDCShadowLayer.h
```diff
@@ -77,3 +77,10 @@
 @property(nonatomic, getter=isShadowMaskEnabled, assign) BOOL shadowMaskEnabled;
 
 @end
+
+/**
+ Subclasses can depend on MDCShadowLayer implementing CALayerDelegate actionForLayer:forKey: in
```
romoore, line 82: Does MDCShadowLayer need to explicitly implement `actionForLayer:forKey:` or does CALayer take care of that?

romoore (approved):

romoore:
Is it possible to backlog an Issue to write a couple of unit tests for the expected behavior?

components/ShadowLayer/src/MDCShadowLayer.h
samnm: As far as I know there is no way to rely on the built-in CAAction creation for headless layers so we explicitly implement `actionForLayer:forKey:`.

romoore: Ack. I missed the method definition in my earlier search. :-S

PR #2621, opened 2017-12-06
Sample of approach where we use the singleton and a delegate to let a …

randallli:
Sample of approach where we use the singleton and a delegate to let a client team choose a custom font for style.

Note: only the Button style was applied and if we go this approach we would apply it to all other styles.


randallli:
UIAppearance is why we are dropping this.

PR #2629, opened 2017-12-07
[Typography] Add standard (non-Dynamic Type) font convenience methods

ianegordon:
Add a method based on mdc_preferredFontWithStyle: that returns a non-scaled font.

### DIFF
--- a/components/Typography/examples/TypographyMaterialStylesViewController.m
+++ b/components/Typography/examples/TypographyMaterialStylesViewController.m
@@ -21,7 +21,8 @@
 @implementation TypographyMaterialStyleViewController {
   NSArray<NSString *> *_strings;
   NSArray<NSString *> *_styleNames;
-  NSArray<UIFont *> *_styleFonts;
+  NSArray<UIFont *> *_styleStandardFonts;
+  NSArray<UIFont *> *_styleDynamicFonts;
 }
 
 - (void)viewDidLoad {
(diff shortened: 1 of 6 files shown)


components/Typography/examples/TypographyMaterialStylesViewController.m
```diff
@@ -44,21 +45,46 @@ - (void)viewDidLoad {
...
+                  @"Display 2 Font (Dynamic Type-enabled)",
                   @"Display 3 Font",
-                  @"Display 4 Font"
+                  @"Display 3 Font (Dynamic Type-enabled)",
+                  @"Display 4 Font",
+                  @"Display 4 Font (Dynamic Type-enabled)"
                   ];
 
-  _styleFonts = @[
+  _styleStandardFonts = @[
```
randallli: why create two separate arrays? just put them interlaced like the names?

components/Typography/examples/TypographyMaterialStylesViewController.m
```diff
@@ -143,7 +169,13 @@ - (UITableViewCell *)tableView:(UITableView *)tableView
                                   reuseIdentifier:@"cell"];
   }
   cell.textLabel.text = _strings[indexPath.section];
-  cell.textLabel.font = _styleFonts[indexPath.row];
+  UIFont *font;
+  if (indexPath.row % 2) {
```
randallli: if we put the font all in the same array we dont have to do any math here. See above.

randallli (approved):

components/Typography/examples/TypographyMaterialStylesViewController.m
ianegordon: Simplification!
Done.

PR #2635, opened 2017-12-07
[TextFields] Updating names of methods in example

willlarche:


romoore (approved):
Is this example broken? I don't see the definitions being updated at the same time.

willlarche:
This was a file I forgot to check in when I updated the main file of the example.

PR #2636, opened 2017-12-07
[TextFields] Updates to controller header.

willlarche:


components/TextFields/src/MDCTextInputController.h
```diff
@@ -82,7 +82,7 @@
 /**
  Default value for disabledColor.
 
- Default is blue.
+ Default is [UIColor lightGrayColor].
```
romoore: Super-duper-nit: `lightGrayColor` is a property. Prefer either

    Default is .lightGrayColor.

or

    Default is UIColor.lightGrayColor.

components/TextFields/src/MDCTextInputController.h
```diff
@@ -253,6 +253,34 @@
...
+ Height of the underline when text field is first responder.
+
+ Default is underlineHeightActiveDefault.
+ */
+@property(nonatomic, assign) CGFloat underlineHeightActive;
+
+/**
+ Default value for underlineHeightActive.
+
+ Default is 2.
```
romoore: Super-nit: Follow Apple's approach of documenting default floats as floating-point literals?

    Default is 2.0.

components/TextFields/src/MDCTextInputController.h
```diff
@@ -253,6 +253,34 @@
...
+ Height of the underline when NOT first responder.
+
+ Default is underlineHeightNormalDefault.
+ */
+@property(nonatomic, assign) CGFloat underlineHeightNormal;
+
+/**
+ Default value for underlineHeightNormal.
+
+ Default is 1.
```
romoore: Super-nit: `Default is 1.0.`

romoore (approved):
Minor nits, feel free to disregard.

components/TextFields/src/MDCTextInputController.h
willlarche: I'll go back thru and do that for all in a nother PR.

romoore: Sounds good to me.

components/TextFields/src/MDCTextInputController.h
willlarche: oh yeah! But what you didn't notice is that all these `Default is [something real]` are false! This is a protocol. I have now updated it to delete them all. (There used to be so few controllers that they weren't a lie. They are lies now.)

PR #2645, opened 2017-12-07
 [Buttons] Corrected imports in examples.

randallli:


components/Buttons/examples/ButtonsSimpleExampleSwiftViewController.swift
```diff
@@ -15,7 +15,8 @@
  */
 
 import Foundation
```
romoore: Doesn't this class require UIKit?

components/Buttons/examples/ButtonsSimpleExampleSwiftViewController.swift
```diff
@@ -15,7 +15,8 @@
...
-import MaterialComponents
+
```
romoore: Should probably add import UIKit

components/Buttons/examples/ButtonsStoryboardAndProgrammatic.swift
```diff
@@ -15,7 +15,8 @@
  */
 
 import Foundation
-import MaterialComponents
+
```
romoore: Also here: import UIKit

jverkoey (approved):
LGTM with @romoore's changes.

PR #2647, opened 2017-12-07
[Dialogs] Add BUILD file and test

romoore:
Adding a single no-op unit test to ensure that Kokoro is actually compiling
the target and that the target settings are correct.

### DIFF
--- a/components/Dialogs/tests/unit/DialogsNoopTest.m
+++ b/components/Dialogs/tests/unit/DialogsNoopTest.m
@@ -24,7 +24,7 @@
 
 @implementation DialogsNoopTest
 
-- (void)simpleBuildTest {
+- (void)testSimpleBuild {
   MDCAlertController *controller = [[MDCAlertController alloc] init];
   XCTAssertNotNil(controller);
 }
--- a/components/Dialogs/tests/unit/DialogsNoopTest.m
+++ b/components/Dialogs/tests/unit/DialogsNoopTest.m
@@ -1,0 +1,32 @@
+/*
+ Copyright 2017-present the Material Components for iOS authors. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


jverkoey (approved):

romoore:
test has wrong method name

components/Dialogs/tests/unit/DialogsNoopTest.m
```diff
@@ -0,0 +1,32 @@
...
+#import "MaterialDialogs.h"
+
+@interface DialogsNoopTest : XCTestCase
+
+@end
+
+@implementation DialogsNoopTest
+
+- (void)testSimpleBuild {
+  MDCAlertController *controller = [[MDCAlertController alloc] init];
```
ianegordon, line 28: init should actually be NS_UNAVAILABLE.
Please replace with:
MDCAlertController *controller = [MDCAlertController alertControllerWithTitle:@"Foo" message:"@"Bar"];

ianegordon (changes requested):

ianegordon:
PS - Marking init NS_UNAVAILABLE in #2654

components/Dialogs/tests/unit/DialogsNoopTest.m
romoore: You want me to change the implementation of MDCAlertController?

romoore:
Canceled Travis. Waiting on #2654 to rebase.

components/Dialogs/tests/unit/DialogsNoopTest.m
ianegordon: We should match UIAlertController's behavior.

[[UIAlertController alloc] init] will create an object.  It will throw an exception when you present it unless you fill in the details, but that is unrelated.

+1 LGTM.

ianegordon (approved):

ianegordon:
Blocking PR was dropped.

PR #2648, opened 2017-12-07
[FeatureHighlight] Add BUILD file and no-op test

romoore:
A simple no-op test that just ensures that Bazel is building/running some of
the code in Feature Highlight.


romoore:
test has wrong method name

ianegordon (approved):

Issue #2650, opened 2017-12-07
[FeatureHighlight] Move Dynamic Type Support to the View


ianegordon:
UIContentSizeCategoryAdjusting is typically implemented on UIView.
Move from MDCFeatureHighlightViewController to MDCFeatureHightView:
@property(nonatomic, readwrite, setter=mdc_setAdjustsFontForContentSizeCategory:)
    BOOL mdc_adjustsFontForContentSizeCategory UI_APPEARANCE_SELECTOR;

mohammadcazig:
The view is not private and not accessible in public interface, are you suggesting to move the logic to the view and have view controller pass the boolean to the view?

romoore:
@ianegordon Could you please clarify what you expect the resulting API to look like?

ianegordon:
Done.

PR #2653, opened 2017-12-07
[FeatureHighlight] Move Dynamic Type support to the view

ianegordon:
This moves the implementation from the view controller to the view.
It also requests non-dynamic type fonts if dynamic type is not enabled.


samnm (approved):

ianegordon:
Closes #2650

PR #2654, opened 2017-12-07
[AlertController] Mark init as unavilable

ianegordon:
We marked initWithCoder and initWithFrame unavailable, but we missed this one.

### DIFF
--- a/components/Dialogs/src/MDCAlertController.h
+++ b/components/Dialogs/src/MDCAlertController.h
@@ -49,6 +49,9 @@
 /** Alert controllers must be created with alertControllerWithTitle:message: */
 - (nullable instancetype)initWithCoder:(nonnull NSCoder *)aDecoder NS_UNAVAILABLE;
 
+/** Alert controllers must be created with alertControllerWithTitle:message: */
+- (nonnull instancetype)init NS_UNAVAILABLE;
+
 /**
  Adds an action to the alert dialog.
 


jverkoey (dismissed):

components/Dialogs/src/MDCAlertController.h
```diff
@@ -49,6 +49,9 @@
 /** Alert controllers must be created with alertControllerWithTitle:message: */
 - (nullable instancetype)initWithCoder:(nonnull NSCoder *)aDecoder NS_UNAVAILABLE;
```
romoore, line 50: This doesn't make sense to me - Why can't I archive my view hierarchy with an alertController being shown?

components/Dialogs/src/MDCAlertController.h
```diff
@@ -49,6 +49,9 @@
...
 
+/** Alert controllers must be created with alertControllerWithTitle:message: */
+- (nonnull instancetype)init NS_UNAVAILABLE;
```
romoore, line 53: I don't understand why this is unavailable. The two properties (title, message) are settable below.

romoore (changes requested):
I don't think this change makes sense:

* `title` and `message` are both settable properties. They can easily be changed after initialization.
* `+alertControllerWithTitle:message:` accepts both parameters to be `nil`. In that case, it's equivalent to calling `-init` and setting the properties later.
* The class factory method is documented explicitly as a convenience initializer. There is no designated initializer on the class (both superclass designated initializers are unavailable).  So a simple alloc+init makes sense.

ianegordon:
Dropping this PR.
This is an easy one, we should match UIAlertController behavior.
Tracking issue #2660

PR #2683, opened 2017-12-11
[FeatureHighlight] Add custom fonts

ianegordon:
Exposes a pair of custom font properties.
These properties integrate with the UIAppearance proxy.


components/FeatureHighlight/src/MDCFeatureHighlightView.h
```diff
@@ -20,7 +20,11 @@
 
 @property(nonatomic, strong, nullable) UIColor *innerHighlightColor UI_APPEARANCE_SELECTOR;
 @property(nonatomic, strong, nullable) UIColor *outerHighlightColor UI_APPEARANCE_SELECTOR;
+
+@property(nonatomic, strong, nullable) UIFont *titleFont UI_APPEARANCE_SELECTOR;
```
randallli, line 24: unit test?

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
```diff
@@ -111,15 +113,15 @@ - (instancetype)initWithFrame:(CGRect)frame {
     _displayMaskLayer.fillColor = [UIColor whiteColor].CGColor;
 
     _titleLabel = [[UILabel alloc] initWithFrame:CGRectZero];
-    _titleLabel.font = [UIFont mdc_standardFontForMaterialTextStyle:MDCFontTextStyleTitle];
+    _titleLabel.font = [UIFont mdc_standardFontForMaterialTextStyle:kTitleTextStyle];
```
randallli: this is suppose to query the singleton.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
```diff
@@ -111,15 +113,15 @@ - (instancetype)initWithFrame:(CGRect)frame {
...
     _titleLabel.textAlignment = NSTextAlignmentNatural;
     _titleLabel.lineBreakMode = NSLineBreakByTruncatingTail;
     _titleLabel.numberOfLines = 0;
     _titleLabel.accessibilityHint = [[self class] dismissAccessibilityHint];
     [self addSubview:_titleLabel];
 
     _bodyLabel = [[UILabel alloc] initWithFrame:CGRectZero];
-    _bodyLabel.font = [UIFont mdc_standardFontForMaterialTextStyle:MDCFontTextStyleSubheadline];
+    _bodyLabel.font = [UIFont mdc_standardFontForMaterialTextStyle:kBodyTextStyle];
```
randallli: suppose to use the singleton

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
```diff
@@ -178,12 +180,70 @@ - (void)setOuterHighlightColor:(UIColor *)outerHighlightColor {
...
+      // If we are automatically adjusting for Dynamic Type resize the font based on the text style
+      _bodyLabel.font =
+          [_bodyFont mdc_fontSizedForMaterialTextStyle:kBodyTextStyle
+                                  scaledForDynamicType:_mdc_adjustsFontForContentSizeCategory];
+    } else {
+      _bodyLabel.font = _bodyFont;
+    }
+  } else {
+    if (_mdc_adjustsFontForContentSizeCategory) {
+      _bodyLabel.font = [UIFont mdc_preferredFontForMaterialTextStyle:kBodyTextStyle];
```
randallli, line 259: we want to use the singleton's value here too

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
```diff
@@ -178,12 +180,70 @@ - (void)setOuterHighlightColor:(UIColor *)outerHighlightColor {
...
+    } else {
+      _bodyLabel.font = [UIFont mdc_standardFontForMaterialTextStyle:kBodyTextStyle];
```
randallli: singleton's value.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
```diff
@@ -178,12 +180,70 @@ - (void)setOuterHighlightColor:(UIColor *)outerHighlightColor {
...
+  [self setNeedsLayout];
+}
+
 - (void)setTitleColor:(UIColor *)titleColor {
   _titleColor = titleColor;
 
   _titleLabel.textColor = titleColor;
 }
 
+- (void)setBodyFont:(UIFont *)bodyFont {
```
randallli, line 239: I think this should read:
```
- (void)setBodyFont:(UIFont *)bodyFont {
  if (bodyFont == nil) {
    _bodyLabel.font = [Typography bodyTextFont];
  } else {
    _bodyLabel.font = bodyFont;
  }
  
  if (_mdc_adjusts...) {
    _bodyLabel.font =
        [_bodyLabel.font mdc_fontSizedForMaterialTextStyle:kBodyTextStyle
  }
}
```

randallli: and that changes to the _mdc_adjust bool and sizeCategoryChanges should just call `[self setBodyFont:self.bodyFont]`

randallli (changes requested):

components/FeatureHighlight/src/MDCFeatureHighlightView.h
ianegordon: Added.

ianegordon:
This component never used the MDCTypography singleton, and I wasn't intending to add support.
In retrospect, having an easy transition for people who are using the singleton will likely make our lives easier.  MDCTypography added where appropriate.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
ianegordon: fontSizedForMaterialTextStyle is intended to scale a custom font.
preferredFontForStyle is intended to eventually migrate to UIContentSizeAdjusting, which is Apple's prescribed API.

Per the Design Doc, properties (set*Font and mdc_setAdjusting) and the sizeCategoryChange handler (and any other font changing method) should call update*Font.

To be clear, this is just a question of whether the code is part of the setter or encapsulated in its own method.  Due to possible KVO side-effects I recommend against implicitly calling complex setters from other methods.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
ianegordon: Done.

ianegordon:
PTAL.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
```diff
@@ -111,15 +113,29 @@ - (instancetype)initWithFrame:(CGRect)frame {
     _displayMaskLayer.fillColor = [UIColor whiteColor].CGColor;
 
     _titleLabel = [[UILabel alloc] initWithFrame:CGRectZero];
-    _titleLabel.font = [UIFont mdc_standardFontForMaterialTextStyle:MDCFontTextStyleTitle];
+    // If we are using the default (system) font loader, retrieve the
+    // font from the UIFont standardFont API.
+    if ([MDCTypography.fontLoader isKindOfClass:[MDCSystemFontLoader class]]) {
```
randallli, line 118: This is not correct anymore since we are going to add a fontloader for each style. Or we are creating a delegate. This is how we globally set a default font faces for all our styles.

The MDCSystemFontLoader should already return the [UIFont mdc_standardFontForMaterialTextStyle:kTitleTextStyle]

concretely we should drop the if clause and just put:
`_titleLabel.font = [MDCTypography titleFont];`

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
randallli: Sorry i had a minor error in my suggestion. I corrected it (inside the if (_mdc_adjusts...){ clause i changed the object calling of the mdc_fontSizedForMaterial.. method)

If we want the updateFont to be in a different method fine but it should source from the label.font not the property that we have created.

components/FeatureHighlight/src/MDCFeatureHighlightView.h
randallli: did you forget to add it to the PR?

randallli (dismissed):
please dismiss my review when you want me to take another look

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
ianegordon: Unfortunately, it won't be orthogonal.

setFont - Comic@22
enable Dynamic Type auto-adjusting : Comic@22 gets scaled to Comic@26
disable Dynamic Type auto-adjusting :
We need the original Comic@22, but we can't get back to there from Comic@26

We need to store the customFont as explicit state.

components/FeatureHighlight/src/MDCFeatureHighlightView.h
ianegordon: UI Test added.

If you are specifically talking about a Unit test for titleFont, it is simply wired to a backing _ivar.

Let's catch up offline if there are still open questions.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
randallli: Yes we talked about that and we decided that it was acceptable to lose size. We would use the "large" or what apple considers default when you turned off the dynamic font.
Since this is developer API action they have a workaround to manually set the precise size again if they need to.

components/FeatureHighlight/src/MDCFeatureHighlightView.h
randallli: oh I was expecting something run by a test. Are you saying you created a manual example to test?

randallli (changes requested):
I think we need changes to MDCTypography to land before this fully works.

The font loader setters per styles or a delegate.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
jverkoey: @randallli it sounds like there might be different assumptions in this discussion. I think what Ian is doing is in line with what we discussed last, so I'll clarify my primary assumption here to make sure it matches yours.

If someone sets a font, it's important that enabling/disabling dynamic type (using the mdc_adjusts bool) does not change the font property.

```swift
// In unit test form:
featureHighlight.font = SomeCustomFont
featureHighlight.mdc_adjusts = true
featureHighlight.mdc_adjusts = false
XCTAssertEqual(featureHighlight.font, SomeCustomFont)
```

It sounds like you feel like the unit test should look like so:

```swift
// In unit test form:
featureHighlight.font = SomeCustomFont
featureHighlight.mdc_adjusts = true
featureHighlight.mdc_adjusts = false
XCTAssertNotEqual(featureHighlight.font, SomeCustomFont)
```

Which I believe we discussed could be confusing to a client (side effects should be avoided!).

If you feel strongly about the alternate implementation @randallli it may be in our best interest to land this change for now and then explore the merits of the alternative approach in a follow-up so that we're not blocking the general body of work here.

ianegordon:
Happy to discuss offline #2689.

Even with the changes you're proposing, we will have the same interface via the customFont properties.  Are we clear to land this (and the other) components based on the API and we can revisit the implementation after Friday?

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
randallli: by copying this logic that will be in [MDCTypography titleFont] you are making it harder for us to pivot if we determine that UIAppearance is not scalable for our needs in GMDC

randallli: by accessing MDCTypography.fontloader you are increasing usage of the very thing you are trying to delete.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
```diff
@@ -111,15 +113,29 @@ - (instancetype)initWithFrame:(CGRect)frame {
...
+      _titleLabel.font = [UIFont mdc_standardFontForMaterialTextStyle:kTitleTextStyle];
+    } else {
+      // There is a custom font loader, retrieve the font from it.
+      _titleLabel.font = [MDCTypography titleFont];
```
randallli, line 122: The logic inside MDCTypography titleFont will be a copy this if else clause.

If as you propose the source of truth should be mdc_standardFontForMaterialTextStyle... The Typography component should call [UIFont mdc_standardFontForMaterialTextStyle:kTitleTextStyle];

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
ianegordon: Do we have evidence the UIAppearance is not scalable?
We already evaluated UIAppearance and decided it was the path forward when we added the color themers.  Happy to revisit that decision, but it might be out of scope for this PR.

As far as utility:

If I want all of my buttons to use GoogleSans:
[MDCButton appearance].titleFont = [UIFont fontNamed:@"Google-Sans"];

Following the pattern of our color themers if we want to turn use Roboto everywhere
MDCFontScheme *scheme = [MDCFontScheme schemeWithFontFamily:@"Roboto"];
[MDCAppBarFontThemer apply:scheme to: [MDCAppBarView appearance]];
[MDCButtonFontThemer apply:scheme to:[MDCButton appearance]];
[MDCFeatureHighlightThemer apply:scheme to:[MDCFeatureHighlightThemer appearance]];
// etc.

Note that this would be the migration path for people using the RobotoFontLoader.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
randallli: I expect this unit test to pass:

Typogragrphy.fontloader = ComicSans
featureHighlight.mdc_adjust = true
XCTAssert(featureHighlight.font.face, comicSans)
featureHighlight.mdc_adjust = false
XCTAssert(featureHighlight.font.face, comicSans)

With the current logic both fail.

randallli (changes requested):

components/FeatureHighlight/src/MDCFeatureHighlightView.h
randallli: I would like a unit test that checks that the backing label's font changes when you set the public api.

Since we are embracing a "I will set a different font then they one you want" (dynamic text) we should unit test that as well.

Also we should test that the default font gets used when we set the font to nil.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
randallli: This does not look like it changed.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
```diff
@@ -111,15 +113,29 @@ - (instancetype)initWithFrame:(CGRect)frame {
...
     _titleLabel.lineBreakMode = NSLineBreakByTruncatingTail;
     _titleLabel.numberOfLines = 0;
     _titleLabel.accessibilityHint = [[self class] dismissAccessibilityHint];
     [self addSubview:_titleLabel];
 
     _bodyLabel = [[UILabel alloc] initWithFrame:CGRectZero];
-    _bodyLabel.font = [UIFont mdc_standardFontForMaterialTextStyle:MDCFontTextStyleSubheadline];
+    // If we are using the default (system) font loader, retrieve the
+    // font from the UIFont standardFont API.
+    if ([MDCTypography.fontLoader isKindOfClass:[MDCSystemFontLoader class]]) {
```
randallli, line 133: The MDCSystemFontLoader class is also kind of an internal implementation detail of MDCTypography.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
```diff
@@ -111,15 +113,29 @@ - (instancetype)initWithFrame:(CGRect)frame {
...
+      _bodyLabel.font = [UIFont mdc_standardFontForMaterialTextStyle:kBodyTextStyle];
+    } else {
+      // There is a custom font loader, retrieve the font from it.
+      _bodyLabel.font = [MDCTypography body1Font];
+    }
     _bodyLabel.shadowColor = nil;
```
randallli, line 139: Just discussed with @jverkoey,
we should add this:
_bodyFont = _bodyLabel.font;

randallli (dismissed):
i have to add a checklist in a couple moments.

ianegordon:
To be reopened in a new PR.

PR #2701, opened 2017-12-13
[FeatureHighlight] Add custom font support

ianegordon:
Add custom font properties for the title and body label.
This fonts are scaled with dynamic type is auto adjusting is set to true.
Unit tests to verify the font set on the property are also set on the label.

### DIFF
--- a/components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
+++ b/components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
@@ -213,8 +213,19 @@ static inline CGPoint CGPointAddedToPoint(CGPoint a, CGPoint b) {
       _titleLabel.font = _titleFont;
     }
   } else {
+    // There is no custom font, so use the default font.
     if (_mdc_adjustsFontForContentSizeCategory) {
-      _titleLabel.font = [UIFont mdc_preferredFontForMaterialTextStyle:kTitleTextStyle];
+      // If we are using the default (system) font loader, retrieve the
+      // font from the UIFont preferredFont API.
+      if ([MDCTypography.fontLoader isKindOfClass:[MDCSystemFontLoader class]]) {
+        _titleLabel.font = [UIFont mdc_preferredFontForMaterialTextStyle:kTitleTextStyle];
+      } else {
+        // There is a custom font loader, retrieve the font and scale it.
+        UIFont *customTypographyFont = [MDCTypography titleFont];
+        _titleLabel.font =
+            [customTypographyFont mdc_fontSizedForMaterialTextStyle:kTitleTextStyle
+                scaledForDynamicType:_mdc_adjustsFontForContentSizeCategory];
+      }
     } else {
       // If we are using the default (system) font loader, retrieve the
       // font from the UIFont standardFont API.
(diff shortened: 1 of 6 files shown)


jverkoey (approved):

components/FeatureHighlight/tests/unit/FeatureHighlightCustomFontTest.m
```diff
@@ -0,0 +1,64 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <XCTest/XCTest.h>
+
+#import "MaterialFeatureHighlight.h"
+#import "../../src/private/MDCFeatureHighlightView+Private.h"
```
randallli: In the past we just exposed the APIs that we wanted instead of importing private headers.
https://github.com/material-components/material-components-ios/blob/develop/components/Slider/tests/unit/SliderTests.m#L28

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
```diff
@@ -111,15 +113,29 @@ - (instancetype)initWithFrame:(CGRect)frame {
     _displayMaskLayer.fillColor = [UIColor whiteColor].CGColor;
 
     _titleLabel = [[UILabel alloc] initWithFrame:CGRectZero];
-    _titleLabel.font = [UIFont mdc_standardFontForMaterialTextStyle:MDCFontTextStyleTitle];
+    // If we are using the default (system) font loader, retrieve the
+    // font from the UIFont standardFont API.
```
randallli: We need additional info:
Since mdc_standardFont... returns a different value than [MDCTypography titleFont] we are jumping through some extra logic to use the newer font sizes in mdc_standardFont... Once [MDCTypograhy titleFont] returns the same value as mdc_standardFont... we can remove some of this logic.

randallli (approved):

components/FeatureHighlight/tests/unit/FeatureHighlightCustomFontTest.m
ianegordon: As it's an implementation detail, I typically wouldn't expose it publicly just so I could test it.  I rank keeping the public API as small as possible more important that exposing more API to make it easier to test.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
ianegordon: Added a link to an issue.  Trying to embed the explanation in a comment seems a bit too much.

components/FeatureHighlight/tests/unit/FeatureHighlightCustomFontTest.m
ianegordon: Done.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
randallli: sgtm

PR #2719, opened 2017-12-13
[Chips] Fix sizeToFit sometimes being too small

samnm:
Due to some intricacies of MDCChipView layout, putting each view's frame through MDCRectAlignToScale would sometimes cause the title to be truncated.

Correctly ensuring chips are fully pixel aligned is nontrivial so I'll create an issue once this PR lands.


romoore (approved):
Is it a bug in the MDCRectAlignToScale implementation or just an artifact of the  Chips layout process?  The expectation of `MDCRectAlignToScale` is that the rect will be the smallest rectangle larger than the argument which is pixel-aligned.

ianegordon:
@romoore Consider opening an issue with your question and assigning it to @samnm since the info may get lost once this PR is merged.

samnm:
@romoore MDCChipView. sizeThatFits isn't aligning rects so it can return a size smaller than required if layoutSubviews does align rects.

Although it *might* be possible that due to floating point inaccuracies sometimes MDCRectAlignToScale unnecessarily expands a rect by a pixel.

PR #2750, opened 2017-12-20
[SnackBar] Increase unit test timeout

romoore:
Swift tests running unit tests can sometimes take longer. Increasing the
timeout for the expectation.


yarneo (approved):
:shipit:

PR #2752, opened 2017-12-20
[KeyboardWatcher] Add no-op test

romoore:


components/private/KeyboardWatcher/BUILD
```diff
@@ -21,4 +24,34 @@ mdc_public_objc_library(
     deps = [
         "//components/private/Application",
     ],
+    sdk_frameworks = [
+        "CoreGraphics",
```
yarneo, line 28: + "UIKit" ?

yarneo (approved):
LGTM % comment

components/private/KeyboardWatcher/BUILD
romoore: Bazel didn't complain, so I didn't add it?

yarneo: Looks like it uses UIKit, and other BUILDs have it there so I would maybe be on the safe side and add it, but upto you, my knowledge of Bazel is limited at this point

romoore: Added the dependency.  Generally, I follow a policy of "if the tool doesn't need it, don't use it. "

yarneo: Taken from https://docs.bazel.build/versions/master/be/objective-c.html

screen shot 2017-12-20 at 12 02 17 pm

I guess it's a NOOP to add it or not. For consistency sake I would add it, if we want to make it cleaner, we should then remove it from all other BUILDs

PR #2760, opened 2017-12-21
[Typography] Add FontThemer & FontScheme to FeatureHighlight

ianegordon:
Add a font scheme and feature highlight font themer.

### DIFF
--- a/components/FeatureHighlight/examples/FeatureHighlightThemerExample.m
+++ b/components/FeatureHighlight/examples/FeatureHighlightThemerExample.m
@@ -24,9 +24,7 @@
 
 @implementation FeatureHighlightThemerExample
 
-- (void)viewDidLoad {
-  [super viewDidLoad];
-
+- (void)applyFeatureHighlightTheming {
   // You would normally set your UIAppearance properties in your AppDelegate in
   // applicationDidFinishLoading.  We are doing it here to keep the revalent
   // code in the sample.
@@ -45,6 +43,12 @@
                             toFeatureHighlightView:[MDCFeatureHighlightView appearance]];
   [MDCFeatureHighlightFontThemer applyFontScheme:fontScheme
                           toFeatureHighlightView:[MDCFeatureHighlightView appearance]];
+}
+
+- (void)viewDidLoad {
+  [super viewDidLoad];
+
+  [self applyFeatureHighlightTheming];
 
   self.view.backgroundColor = [UIColor whiteColor];
 
--- a/components/FeatureHighlight/examples/FeatureHighlightThemerExample.m
+++ b/components/FeatureHighlight/examples/FeatureHighlightThemerExample.m
@@ -1,0 +1,79 @@
+/*
+ Copyright 2017-present the Material Components for iOS authors. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import "MaterialButtons.h"
+#import "MaterialFeatureHighlight.h"
+#import "MaterialThemes.h"
+#import "MaterialTypography.h"
+#import "supplemental/FeatureHighlightExampleSupplemental.h"
+#import "MDCFeatureHighlightColorThemer.h"
+#import "MDCFeatureHighlightFontThemer.h"
+
+@implementation FeatureHighlightThemerExample
+
+- (void)viewDidLoad {
+  [super viewDidLoad];
+
+  // You would normally set your UIAppearance properties in your AppDelegate in
+  // applicationDidFinishLoading.  We are doing it here to keep the revalent
+  // code in the sample.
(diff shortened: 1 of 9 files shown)


ianegordon:
This is an example on how the FontScheme and FontThemer will work with a component.

ianegordon:
screen shot 2017-12-22 at 15 22 37

ianegordon:
This PR is blocked on an open question about whether we should be using the new or old type hierarchy.
I will fill in the appropriate styles once design has made a determination.

components/FeatureHighlight/examples/FeatureHighlightThemerExample.m
```diff
@@ -0,0 +1,79 @@
...
+#import "supplemental/FeatureHighlightExampleSupplemental.h"
+#import "MDCFeatureHighlightColorThemer.h"
+#import "MDCFeatureHighlightFontThemer.h"
+
+@implementation FeatureHighlightThemerExample
+
+- (void)viewDidLoad {
+  [super viewDidLoad];
+
+  // You would normally set your UIAppearance properties in your AppDelegate in
```
randallli: Lets put this inside a method to make it clear what we think should be inside the appDidFinishLoading.

components/FeatureHighlight/examples/FeatureHighlightThemerExample.m
```diff
@@ -0,0 +1,79 @@
...
+  // https://developer.apple.com/documentation/uikit/uiappearance
+  MDCBasicFontScheme *fontScheme = [[MDCBasicFontScheme alloc] init];
+  fontScheme.headline2 = [UIFont fontWithName:@"Zapfino" size:14.0];
+  fontScheme.body2 = [UIFont fontWithName:@"Chalkduster" size:12.0];
+  UIColor *primaryColor = [UIColor purpleColor];
+  MDCBasicColorScheme *colorScheme =
+      [[MDCBasicColorScheme alloc] initWithPrimaryColor:primaryColor];
+
+  [MDCFeatureHighlightColorThemer applyColorScheme:colorScheme
+                            toFeatureHighlightView:[MDCFeatureHighlightView appearance]];
```
randallli, line 43: should this only apply to this view controller? [MDCFeatureHighlightView  appearanceWhenContainedIn:[self class]];

as it stands now it will change the other examples.

components/Themes/src/MDCFontScheme.h
```diff
@@ -0,0 +1,90 @@
...
+#import <UIKit/UIKit.h>
+
+/**
+ A color scheme comprised of set of primary and secondary colors. Material design guidelines
+ recommend using primary and secondary colors with light and dark color variants.
+ */
+@protocol MDCFontScheme <NSObject>
+
+/** The font-face to be used for Display 1. */
+@property(nonatomic, strong, nullable, readonly) UIFont *display1;
```
randallli: I think they should be nonnull so the themer doesn't have to rely on its own defaults.

components/Themes/src/MDCFontScheme.h
```diff
@@ -0,0 +1,90 @@
...
+
+/** The font-face to be used for Display 2. */
+@property(nonatomic, strong, nullable, readonly) UIFont *display2;
```
randallli: By making them a UIFont we are forcing clients to load them all upfront. If we make them a MDFFontDiskLoader pointer then they can be lazy loaded.

randallli: Not all of our components will use the larger display sizes.

randallli (changes requested):

components/Themes/src/MDCFontScheme.h
ianegordon: This is by design. Consider our implementations use customFont = nil as 'default to system'.

So the code for our hypothetical FooThemers becomes:
// Foo has two text elements
// A quote that should be body2
// An author that should be caption
MDCFooComponentThemer
foo.quoteFont = scheme.body2  // If nil this will use the system font
foo.author = scheme.caption      // If nil this will use the system font

components/Themes/src/MDCFontScheme.h
ianegordon: Understood.  We explicitly use UIFont (as opposed to UIFontDescriptor or fontLoader) because we want to know that if UIFont != nil, we will be able to render text with it.
It is totally reasonable that we might have a UseRobotoFontForAllValuesScheme object that uses a fontLoader internally to return a Roboto UIFont for all of the properties.

ianegordon: As far as having multiple (possibly unused) fonts such as the larger display sizes.  The pattern is that you would create a fontScheme once, then apply it to multiple / all of you components, similar to how we (currently) do with the ColorThemers.

components/Themes/src/MDCFontScheme.h
randallli: lets comment that these are setting on component properties that are in spirit null resetable .

components/Themes/src/MDCFontScheme.h
randallli: I realized that lazy loading can be don within the object conforming to the protocol.

randallli (approved):

components/Themes/src/MDCFontScheme.h
ianegordon: Noted.

components/FeatureHighlight/examples/FeatureHighlightThemerExample.m
ianegordon: Done.

ianegordon:
Blocked on +1 from design.

ianegordon:
Pulled in final values from design.

ianegordon:
cla is signed.

PR #2761, opened 2017-12-22
[Examples/Tests] Convert @imports to imports

romoore:
In order to make the internal build systems happier, @import statements should not be used.

### DIFF
--- a/components/Collections/examples/supplemental/CollectionsSectionInsetsExampleSupplemental.h
+++ b/components/Collections/examples/supplemental/CollectionsSectionInsetsExampleSupplemental.h
@@ -14,7 +14,7 @@
  limitations under the License.
  */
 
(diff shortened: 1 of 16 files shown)


ianegordon:
Let's chat on this tomorrow.

randallli (approved):

romoore:
This is necessary for both internal builds, but also for bazel builds on kokoro. In order for us to include building our examples (a very good thing) on Kokoro, we'll have to make this change until bazel supports modules imports.

ianegordon:
Interesting.

ianegordon (commented):
Acknowledged.  Let's make sure we open an issue with the Bazel team if there isn't one already.

ianegordon (approved):

PR #2766, opened 2017-12-22
Add an api to MDCActivityIndicator to allow custom animations 

jmdetloff:
Allows clients to specify animations that should seamlessly blend into the standard rotation cycle of the MDCActivityIndicator

### DIFF
--- a/components/ActivityIndicator/src/MDCActivityIndicator.h
+++ b/components/ActivityIndicator/src/MDCActivityIndicator.h
@@ -16,6 +16,7 @@
 
 #import <UIKit/UIKit.h>
 
+@class MDCActivityIndicatorTransition;
 @protocol MDCActivityIndicatorDelegate;
 
 /**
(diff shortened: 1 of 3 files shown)


jmdetloff:
Hey @ianegordon @randallli Thoughts on this approach? Its a little out there but it does keep from exposing the MDCActivityIndicator internals

ianegordon (changes requested):
Can you point me at the spec / requirement for this?

I suspect if clients want a custom animation they will want it across all of their Activity Indicators.
@property startAnimation UI_APPEARANCE;
@property stopAnimation UI_APPEARANCE;

Add NSCopying to MDCActivityIndicatorAnimation and add to initWithCoder / encodeWithCoder ?

jmdetloff:
This is primarily to enable the refresh control, which animates a refresh arrow spinning and then transforming into the indicator stroke, then back into the refresh icon. I'm not sure this api would usually be appropriate app wide, at least for the usage Im imagining. I figure in most places the normal loading spinner would be what people want. Its only in select cases when clients would want a custom animation, ex a pull to refresh spinner, or maybe a transaction spinner that becomes a checkmark after the transaction completes, ect.

components/ActivityIndicator/src/MDCActivityIndicator.h
```diff
@@ -141,3 +160,34 @@ IB_DESIGNABLE
...
+ or after its standard cycle animation.
+ */
+@interface MDCActivityIndicatorAnimation : NSObject
+
+/**
+ The animations to be performed by MDCActivityIndicator. In this block add CAAnimations to be
+ animated before or after MDCActivityIndicator's cycle animation. MDCActivityIndicator will trigger
+ these animations and call completion after they complete.
+ */
+@property(nonatomic, copy, nonnull) void (^animation)(CGFloat strokeStart, CGFloat strokeEnd);
```
randallli: make a type def for this block def.

components/ActivityIndicator/src/MDCActivityIndicator.m
```diff
@@ -236,9 +242,72 @@ - (void)startAnimating {
...
+  if (!self.window || _backgrounded) {
+    return;
   }
+  
+  [self actuallyStartAnimating];
+}
+
+- (void)startAnimatingWithStartAnimation:(nonnull MDCActivityIndicatorAnimation *)startAnimation
+                         cycleStartIndex:(NSInteger)cycleStartIndex {
+	if (_animating) {
```
randallli: indent

randallli (approved):
lets add an example.

ianegordon:
Got it.
LGTM.

ianegordon (approved):

ianegordon:
Please address Randall's comments and cast the floating point error from Kokoro before merge.

jmdetloff:
Added an example. Any thoughts before merge @randallli?

PR #2778, opened 2018-01-02
[Dialogs] Add designated initializer to MDCAlertController

### LINKED ISSUE #2775: [Dialogs] MDCAlertController has no designated initializer
### Overview
`MDCAlertController` has no designated initializer and has made both superclass designated initializers (`initWithNibName:bundle:` and `initWithCoder:`) unavailable. Calling `init` (the next designated initializer in the hierarchy) results in no initializer calls to `MDCAlertController`.

### Reproduction steps
1. Set a breakpoint in all initializers of `MDCAlertController`
1. Call `[[MDCAlertController alloc] init]`
1. Verify that one of the breakpoints is hit

### Version number
43.1.0

romoore:
If MDCAlertController does not override the superclasses designated
initializers, then its own designated initializer may never get called.
This can lead to problems.  Although the superclass designated
initializers are unavailable in MDCAlertController, they must still have
implementations.

Closes #2775


components/Dialogs/src/MDCAlertController.m
```diff
@@ -124,6 +124,15 @@ - (nonnull instancetype)initWithTitle:(nullable NSString *)title
   return self;
 }
 
+- (nonnull instancetype)initWithNibName:(nullable NSString *)nibNameOrNil
+                                 bundle:(nullable NSBundle *)nibBundleOrNil {
+  return [self initWithTitle:nil message:nil];
+}
+
+- (nullable instancetype)initWithCoder:(nonnull NSCoder *)aDecoder {
```
ajsecord: Does initWithCoder still work with this change?

ajsecord (dismissed):

components/Dialogs/src/MDCAlertController.m
romoore: We can make `initWithCoder:` available to subclasses, but the current design won't allow encoding of the full state of the AlertController. Specifically, the actions array is not possible to encode since its members (`MDCAlertAction *`) have block properties. If we were to switch to a delegate structure, it would be possible to encode the actions.

ianegordon (approved):
LGTM!

PR #2801, opened 2018-01-04
[Dialogs] Explicitly annotate MDCAlertController as not subclassable

### LINKED ISSUE #2787: [Dialogs] Mark MDCAlertController as final / not intended to be subclassed
Like UIAlertController, MDCAlertController is not intended to be subclasses.

Annotate it as final and document this fact.

ianegordon:
MDCAlertController has always been documented as not supporting subclassing, this adds an explicit annotation.

Closes #2787

### DIFF
--- a/components/Dialogs/src/MDCAlertController.h
+++ b/components/Dialogs/src/MDCAlertController.h
@@ -16,6 +16,14 @@
 
 #import <UIKit/UIKit.h>
 
+#ifndef MDC_SUBCLASSING_RESTRICTED
+#if defined(__has_attribute) && __has_attribute(objc_subclassing_restricted)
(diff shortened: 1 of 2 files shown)


romoore (approved):

ianegordon:
Closes #2787

romoore (approved):

ianegordon:
Tested locally on Xcode 9
iPhone Sim @ 10.0
iPhone Sim @ 11.2

ajsecord:
I unfortunately had to revert this change due to multiple broken internal targets. It looks like there are some fairly innocuous subclasses—please take a look at them and plan accordingly.

PR #2805, opened 2018-01-04
[Typography] Suppress pointer warning in tests

romoore:
Internal builds are issuing a warning that `&UIFontWeightMedium` will
never be NULL, although this is incorrect on iOS versions below 8.2.


ajsecord (approved):

ianegordon (approved):
Please merge when CI is happy.

PR #2820, opened 2018-01-08
[Chips] Add custom font support

ianegordon:
Added a titleFont property with a UI_APPEARANCE_SELECTOR annotation.

Added sample demonstrating usage of a custom font.

screen shot 2018-01-08 at 18 53 44


samnm:
I'm not sure we want fonts to be stateful. MDCChipViews could require a resize upon every state change which isn't supported by the supporting Chips code right now.

ianegordon:
PTAL

samnm (approved):

PR #2823, opened 2018-01-09
[Chips] Stateful ink

samnm:


components/Chips/examples/ChipsCustomizedExampleViewController.m
```diff
@@ -31,11 +31,16 @@ + (void)configureChip:(MDCChipView *)chip {
   [chip setTitleColor:customColor forState:UIControlStateNormal];
   [chip setBorderColor:customColor forState:UIControlStateNormal];
   [chip setBorderWidth:2.0 forState:UIControlStateNormal];
+  [chip setInkColor:[customColor colorWithAlphaComponent:0.2] forState:UIControlStateNormal];
```
ajsecord: Cast all of these `0.2` values to CGFloat, they're breaking your 32-bit build.

components/Chips/src/MDCChipView.h
```diff
@@ -211,6 +211,26 @@
 - (void)setElevation:(MDCShadowElevation)elevation forState:(UIControlState)state
     UI_APPEARANCE_SELECTOR;
 
+/*
+ Returns the ink color for a particular control state.
+
+ If no ink color has been set for a given state, the returned value will fall back to the value
+ set for UIControlStateNormal.
```
ajsecord: Please document what the default is for `UIControlStateNormal`.

ajsecord (changes requested):
Please add at least the following unit tests:
- Check that the default is whatever you documented the default to be.
- Set a non-default color for a state and check that it gets returned.
- Set a non-default color for a state, then set it back to `nil`, then check that you get back the color for `Normal`.

ajsecord (approved):

PR #2849, opened 2018-01-17
[ButtonBar] Expose font and padding properties

ianegordon:
Expose the MDCButtonBarButton and its properties so we can explicitly set a custom font.

Add a sample demonstrating the effect.

screen shot 2018-01-17 at 15 31 51

### DIFF
--- a/components/ButtonBar/examples/ButtonBarCustomizedFontExample.m
+++ b/components/ButtonBar/examples/ButtonBarCustomizedFontExample.m
@@ -1,0 +1,117 @@
+/*
+ Copyright 2016-present the Material Components for iOS authors. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+#import "MaterialButtonBar.h"
+
+@interface ButtonBarCustomizedFontExample : UIViewController
+@end
+
+@implementation ButtonBarCustomizedFontExample
+
+- (void)viewDidLoad {
+  [super viewDidLoad];
+
+  MDCButtonBar *buttonBar = [[MDCButtonBar alloc] init];
+
+  // MDCButtonBar ignores the style of UIBarButtonItem.
+  UIBarButtonItemStyle ignored = UIBarButtonItemStyleDone;
+
+  UIBarButtonItem *actionItem =
+      [[UIBarButtonItem alloc] initWithTitle:@"Disable"
+                                       style:ignored
+                                      target:self
+                                      action:@selector(didTapActionButton:)];
+  UIBarButtonItem *secondActionItem =
+      [[UIBarButtonItem alloc] initWithTitle:@"Customization"
+                                       style:ignored
+                                      target:self
+                                      action:@selector(didTapActionButton:)];
+
+  NSArray *items = @[ actionItem, secondActionItem ];
+
+  // Set the title text attributes before assigning to buttonBar.items
+  // because of https://github.com/material-components/material-components-ios/issues/277
+  for (UIBarButtonItem *item in items) {
+    [item setTitleTextAttributes:@{ NSForegroundColorAttributeName : [UIColor greenColor] }
+                        forState:UIControlStateNormal];
+  }
+
+  buttonBar.items = items;
+
+  // MDCButtonBar's sizeThatFits gives a "best-fit" size of the provided items.
+  CGSize size = [buttonBar sizeThatFits:self.view.bounds.size];
+  CGFloat x = (self.view.bounds.size.width - size.width) / 2;
+  CGFloat y = self.view.bounds.size.height / 2 - size.height;
+  buttonBar.frame = CGRectMake(x, y, size.width, size.height);
+  buttonBar.autoresizingMask =
+      (UIViewAutoresizingFlexibleTopMargin | UIViewAutoresizingFlexibleBottomMargin |
+       UIViewAutoresizingFlexibleLeftMargin | UIViewAutoresizingFlexibleRightMargin);
+  [self.view addSubview:buttonBar];
+
+  // Ensure that the controller's view isn't transparent.
+  self.view.backgroundColor = [UIColor colorWithWhite:0.9f alpha:1.0f];
+}
+
+#pragma mark - User actions
+
+- (void)didTapActionButton:(id)sender {
+  NSLog(@"Did tap action item: %@", sender);
+
+  // Disable CUstomization
+  [[MDCButtonBarButton appearance] setTitleFont:nil forState:UIControlStateNormal];
+}
+
+@end
+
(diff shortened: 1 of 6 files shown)


components/ButtonBar/src/MDCButtonBarButton.h
```diff
@@ -0,0 +1,26 @@
+//
+//  MDCButtonBarButton.h
+//  MaterialComponents
+//
+//  Created by Ian Gordon on 1/17/18.
+//
```
samnm: Missing the license comment.

components/ButtonBar/src/MDCButtonBarButton.h
```diff
@@ -0,0 +1,26 @@
...
+
+#import "MaterialButtons.h"
+
+@interface MDCButtonBarButton : MDCFlatButton
+
+// Content padding for the button.
```
samnm: I believe this should be /** */ comment style.

components/ButtonBar/src/private/MDCAppBarButtonBarBuilder.m
```diff
@@ -49,10 +50,10 @@
 @interface MDCButtonBarSandbagView : UIView
 @end
 
-@interface MDCButtonBarButton : MDCFlatButton
+@interface MDCButtonBarButton (MDCAppBarButtonBarBuilder)
 
-// Content padding for the button.
-@property(nonatomic) UIEdgeInsets contentPadding;
+//// Content padding for the button.
+//@property(nonatomic) UIEdgeInsets contentPadding;
```
samnm: Should these commented lines be removed?

samnm (dismissed):

components/ButtonBar/src/MDCButtonBarButton.h
ianegordon: Done.

components/ButtonBar/src/MDCButtonBarButton.h
ianegordon: Done.

components/ButtonBar/src/private/MDCAppBarButtonBarBuilder.m
ianegordon: Done.

ianegordon:
PTAL.

components/ButtonBar/src/private/MDCAppBarButtonBarBuilder.m
```diff
@@ -319,6 +313,10 @@ - (void)layoutSubviews {
   }
 }
 
+- (void)setTitleFont:(nullable UIFont *)font forState:(UIControlState)state {
```
samnm: Do we need this implementation if it only calls super?

samnm (approved):
LGTM, with one final quick question.

components/ButtonBar/src/private/MDCAppBarButtonBarBuilder.m
ianegordon: We do, not sure why, added investigate TODO. #2850

components/ButtonBar/src/MDCButtonBarButton.h
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ The font used by the button's @c title.
+
+ If left unset or reset to nil for a given state, then a default font is used.
+
+ @param font The font.
+ @param state The state.
+ */
+- (void)setTitleFont:(nullable UIFont *)font forState:(UIControlState)state
```
romoore, line 35: You shouldn't need this, it's already declared in MDCButton.h

components/ButtonBar/examples/ButtonBarCustomizedFontExample.m
```diff
@@ -0,0 +1,117 @@
...
+@implementation ButtonBarCustomizedFontExample (GeneralApplicationLogic)
+
+- (id)init {
+  self = [super init];
+  if (self) {
+    self.title = @"Button Bar";
+
+    UIFont *customFont = [UIFont fontWithName:@"American Typewriter" size:10.0f];
+    NSAssert(customFont, @"Unable to instantiate font");
+    [[MDCButtonBarButton appearance] setTitleFont:customFont forState:UIControlStateNormal];
```
romoore: Just so I understand correctly, opening this example will change the font for all ButtonBar buttons throughout the app?

components/ButtonBar/examples/ButtonBarCustomizedFontExample.m
```diff
@@ -0,0 +1,117 @@
...
+
++ (NSArray *)catalogBreadcrumbs {
+  return @[ @"Button Bar", @"Button Bar (Customized)" ];
+}
+
++ (BOOL)catalogIsPrimaryDemo {
+  return NO;
+}
+
++ (NSString *)catalogDescription {
```
romoore: No need for description if it's not the primary.

romoore (approved):

components/ButtonBar/examples/ButtonBarCustomizedFontExample.m
ianegordon: Correct.  However, if you actually click on one of the button bar buttons, it will set the appearance proxy back to nil / default.

components/ButtonBar/src/MDCButtonBarButton.h
ianegordon: Understood.  It is intentionally redeclared to make it obvious that it can be customized.  If we don't think this is necessary I'm happy to remove.
However, when I re-declare, the compiler throws an error unless I re-define.  In this case the implementation just calls super.  I'm not sure why this is the case: #2850

romoore: I suspect this is simply because once an interface is declared, it should be defined. In the case that the superclass removes its declaration, your interface remains the same.  If you prefer to add these into the header strictly for documentation purposes, then I think using a category would be the best way.  It allows redeclaring the API, but would not be enforced by the compiler as long as the category has no implementation.  The drawback would be that the compiler won't notice if the superclass drops its declaration.

components/ButtonBar/examples/ButtonBarCustomizedFontExample.m
```diff
@@ -0,0 +1,117 @@
...
+  self.view.backgroundColor = [UIColor colorWithWhite:0.9f alpha:1.0f];
+}
+
+#pragma mark - User actions
+
+- (void)didTapActionButton:(id)sender {
+  NSLog(@"Did tap action item: %@", sender);
+
+  // Disable CUstomization
+  [[MDCButtonBarButton appearance] setTitleFont:nil forState:UIControlStateNormal];
```
romoore: This won't produce any on-screen change. Can we set the sender's title back to `nil` to match the UIAppearance version? That way the UIAppearance value and on-screen value will be the same.

components/ButtonBar/examples/ButtonBarCustomizedFontExample.m
romoore: Got it. Added a comment above in the action target to recommend changing the instances as well as UIAppearance.

components/ButtonBar/examples/ButtonBarCustomizedFontExample.m
ianegordon: I definitely appreciate the value of this, but the sender is actually a UIBarButtonItem.  The complexity of reseting the fonts currently used would be high.  UIAppearance properties seems like they should really should only be called once, ideally on App startup, so our samples are going to be non-standard.
I've added additional documentation to the sample detailing this.

ianegordon:
Tested locally with Xcode 9 on iPhone X @ 11, iPhone 7+ @ 10.
Skipping Travis.

PR #2866, opened 2018-01-24
[Palettes] Add API to support "walking" the palette colors

romoore:
When a color is chosen from a Palette, it can be helpful to be able to
find darker/lighter colors within the same palette at runtime. For
example, an accessibility mutator can use a darker palette color before
using black.  Even when opacity is used (and adjusted), the resulting
color may not fall within a palette and therefore is less desirable than
one preselected by designers.

This change will introduce two new API methods to MDCPalette:
```objectivec
+ nextDarkerColorInPaletteForColor:(UIColor *)
+ nextLighterColorInPaletteForColor:(UIColor *)
```

Each of these methods will search the predefined palettes for the
provided color, identify that color's position in the palette (e.g.,
tint500 or accent200), and return the next "stepped" color in the
palette.  For example, the next darker color from tint800 is tint900,
and the next lighter color is tint700.  If a next color does not exist
(e.g., darker from tint900), then `nil` is returned.

MDCButtonTitleColorAccessibilityMutator was updated to demonstrate how
the new API can be used.

## Button mutator example
As a demonstration of the behavior, I tested two different buttons that have barely-visible colors set on them.  For each button, I captured the output of the original accessibility mutator and the new mutator.

| Baseline Image       | Original Mutator       | Mutator with "walkable" palette |
|-------------------|----------|-----------|
|flat-tint50-on-grey-baseline | flat-tint50-on-grey-before | flat-tint50-on-grey-after |
|raised-tint900-on-black-baseline|raised-tint900-on-black-before|raised-tint900-on-black-after|

### DIFF
--- a/components/Buttons/src/TitleColorAccessibilityMutator/MDCButtonTitleColorAccessibilityMutator.m
+++ b/components/Buttons/src/TitleColorAccessibilityMutator/MDCButtonTitleColorAccessibilityMutator.m
@@ -19,8 +19,73 @@
 #import <MDFTextAccessibility/MDFTextAccessibility.h>
 #import "MaterialButtons.h"
 #import "MaterialTypography.h"
+#import "MaterialPalettes.h"
 
 @implementation MDCButtonTitleColorAccessibilityMutator
+
++ (UIColor *)passingPaletteColorForColor:(UIColor *)color
+                            onBackground:(UIColor *)background
+                                 options:(MDFTextAccessibilityOptions)options
+                      searchDarkerColors:(BOOL)searchDarkerColors {
+
+  if (searchDarkerColors && [MDCPalette nextDarkerColorInPaletteForColor:color]) {
+    UIColor *nextDarkerColor = [MDCPalette nextDarkerColorInPaletteForColor:color];
+    if ([MDFTextAccessibility textColor:nextDarkerColor
+                passesOnBackgroundColor:background
+                                options:options]) {
+      return nextDarkerColor;
+    }
+    return [[self class] passingPaletteColorForColor:nextDarkerColor
+                                        onBackground:background
+                                             options:options
+                                  searchDarkerColors:YES];
+  }
+  if (!searchDarkerColors && [MDCPalette nextLighterColorInPaletteForColor:color]) {
+    UIColor *nextLighterColor = [MDCPalette nextLighterColorInPaletteForColor:color];
+    if ([MDFTextAccessibility textColor:nextLighterColor
+                passesOnBackgroundColor:background
+                                options:options]) {
+      return nextLighterColor;
+    }
+    return [[self class] passingPaletteColorForColor:nextLighterColor
+                                        onBackground:background
+                                             options:options
+                                  searchDarkerColors:NO];
+  }
+  return nil;
+}
+
++ (UIColor *)passingPaletteColorForColor:(UIColor *)color
+                            onBackground:(UIColor *)background
+                                 options:(MDFTextAccessibilityOptions) options {
+  CGFloat colorContrast = [MDFTextAccessibility contrastRatioForTextColor:color
+                                                        onBackgroundColor:background];
+  if ([MDCPalette nextDarkerColorInPaletteForColor:color]) {
+    UIColor *nextDarkerColor = [MDCPalette nextDarkerColorInPaletteForColor:color];
+    CGFloat darkerContrast =
+        [MDFTextAccessibility contrastRatioForTextColor:nextDarkerColor
+                                      onBackgroundColor:background];
+    if (darkerContrast > colorContrast) {
+      return [[self class] passingPaletteColorForColor:color
+                                          onBackground:background
+                                               options:options
+                                    searchDarkerColors:YES];
+    }
+  }
+  if ([MDCPalette nextLighterColorInPaletteForColor:color]) {
+    UIColor *nextLighterColor = [MDCPalette nextLighterColorInPaletteForColor:color];
+    CGFloat lighterContrast =
+        [MDFTextAccessibility contrastRatioForTextColor:nextLighterColor
+                                      onBackgroundColor:background];
+    if (lighterContrast > colorContrast) {
+      return [[self class] passingPaletteColorForColor:color
+                                          onBackground:background
+                                               options:options
+                                    searchDarkerColors:NO];
+    }
+  }
+  return nil;
+}
 
 + (void)changeTitleColorOfButton:(MDCButton *)button {
   // This ensures title colors will be accessible against the buttons backgrounds.
@@ -41,10 +106,15 @@
       if (![MDFTextAccessibility textColor:existingColor
                    passesOnBackgroundColor:backgroundColor
                                    options:options]) {
-        UIColor *color =
-            [MDFTextAccessibility textColorOnBackgroundColor:backgroundColor
-                                             targetTextAlpha:[MDCTypography buttonFontOpacity]
-                                                     options:options];
+
+        UIColor *color = [[self class] passingPaletteColorForColor:existingColor
+                                                      onBackground:backgroundColor
+                                                           options:options];
+        if (!color) {
+          color = [MDFTextAccessibility textColorOnBackgroundColor:backgroundColor
+                                                   targetTextAlpha:[MDCTypography buttonFontOpacity]
+                                                           options:options];
+        }
         [button setTitleColor:color forState:controlState];
       }
     }
--- a/components/Buttons/src/TitleColorAccessibilityMutator/MDCButtonTitleColorAccessibilityMutator.m
+++ b/components/Buttons/src/TitleColorAccessibilityMutator/MDCButtonTitleColorAccessibilityMutator.m
@@ -23,10 +23,10 @@
 
 @implementation MDCButtonTitleColorAccessibilityMutator
 
-+ (UIColor *)passingPaletteColorForColor:(UIColor *)color
-                            onBackground:(UIColor *)background
-                                 options:(MDFTextAccessibilityOptions)options
-                      searchDarkerColors:(BOOL)searchDarkerColors {
++ (UIColor *)nearestPassingPaletteColorForColor:(UIColor *)color
+                                   onBackground:(UIColor *)background
+                                        options:(MDFTextAccessibilityOptions)options
+                             searchDarkerColors:(BOOL)searchDarkerColors {
 
   if (searchDarkerColors && [MDCPalette nextDarkerColorInPaletteForColor:color]) {
     UIColor *nextDarkerColor = [MDCPalette nextDarkerColorInPaletteForColor:color];
@@ -35,10 +35,10 @@
                                 options:options]) {
       return nextDarkerColor;
     }
-    return [[self class] passingPaletteColorForColor:nextDarkerColor
-                                        onBackground:background
-                                             options:options
-                                  searchDarkerColors:YES];
+    return [[self class] nearestPassingPaletteColorForColor:nextDarkerColor
+                                               onBackground:background
+                                                    options:options
+                                         searchDarkerColors:YES];
   }
   if (!searchDarkerColors && [MDCPalette nextLighterColorInPaletteForColor:color]) {
     UIColor *nextLighterColor = [MDCPalette nextLighterColorInPaletteForColor:color];
@@ -47,17 +47,17 @@
                                 options:options]) {
       return nextLighterColor;
     }
-    return [[self class] passingPaletteColorForColor:nextLighterColor
-                                        onBackground:background
-                                             options:options
-                                  searchDarkerColors:NO];
+    return [[self class] nearestPassingPaletteColorForColor:nextLighterColor
+                                               onBackground:background
+                                                    options:options
+                                         searchDarkerColors:NO];
   }
   return nil;
 }
 
-+ (UIColor *)passingPaletteColorForColor:(UIColor *)color
-                            onBackground:(UIColor *)background
-                                 options:(MDFTextAccessibilityOptions) options {
++ (UIColor *)nearestPassingPaletteColorForColor:(UIColor *)color
+                                   onBackground:(UIColor *)background
+                                        options:(MDFTextAccessibilityOptions) options {
   CGFloat colorContrast = [MDFTextAccessibility contrastRatioForTextColor:color
                                                         onBackgroundColor:background];
   if ([MDCPalette nextDarkerColorInPaletteForColor:color]) {
@@ -66,10 +66,10 @@
         [MDFTextAccessibility contrastRatioForTextColor:nextDarkerColor
                                       onBackgroundColor:background];
     if (darkerContrast > colorContrast) {
-      return [[self class] passingPaletteColorForColor:color
-                                          onBackground:background
-                                               options:options
-                                    searchDarkerColors:YES];
+      return [[self class] nearestPassingPaletteColorForColor:color
+                                                 onBackground:background
+                                                      options:options
+                                           searchDarkerColors:YES];
     }
   }
   if ([MDCPalette nextLighterColorInPaletteForColor:color]) {
@@ -78,10 +78,10 @@
         [MDFTextAccessibility contrastRatioForTextColor:nextLighterColor
                                       onBackgroundColor:background];
     if (lighterContrast > colorContrast) {
-      return [[self class] passingPaletteColorForColor:color
-                                          onBackground:background
-                                               options:options
-                                    searchDarkerColors:NO];
+      return [[self class] nearestPassingPaletteColorForColor:color
+                                                 onBackground:background
+                                                      options:options
+                                           searchDarkerColors:NO];
     }
   }
   return nil;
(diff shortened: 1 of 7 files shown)


components/Buttons/src/TitleColorAccessibilityMutator/MDCButtonTitleColorAccessibilityMutator.m
```diff
@@ -19,9 +19,79 @@
 #import <MDFTextAccessibility/MDFTextAccessibility.h>
 #import "MaterialButtons.h"
 #import "MaterialTypography.h"
+#import "MaterialPalettes.h"
 
 @implementation MDCButtonTitleColorAccessibilityMutator
 
++ (UIColor *)passingPaletteColorForColor:(UIColor *)color
```
willlarche: Said this before but "passing" isn't the best word to me. It's vague. But that's opinion.

willlarche: Also, I don't know if it has a Swift name but it certainly should.

components/Buttons/src/TitleColorAccessibilityMutator/MDCButtonTitleColorAccessibilityMutator.m
```diff
@@ -19,9 +19,79 @@
...
+                            onBackground:(UIColor *)background
+                                 options:(MDFTextAccessibilityOptions)options
+                               tryDarker:(BOOL)tryDarker {
```
willlarche: Is that like "Prefer Darker"?

willlarche: Oh. I see, it's more like "Only Darker".

components/Palettes/src/MDCPalettes.m
```diff
@@ -42,6 +43,34 @@
                          alpha:1];
 }
 
+static void MDCPalettizeColors(MDCPalette *palette) {
```
willlarche: Naming is hard. But I can't think of a better one.

components/Palettes/src/UIColor+MDCPalettable.m
```diff
@@ -0,0 +1,165 @@
...
+#import <CoreGraphics/CoreGraphics.h>
+#import <UIKit/UIKit.h>
+#import <objc/runtime.h>
+
+#import "MDCPalettes.h"
+#import "UIColor+MDCPalettable.h"
+
+@implementation UIColor (MDCPalettable)
+
+static BOOL ColorEqualToColor(UIColor *color, UIColor *otherColor) {
```
willlarche: Put in a comment saying why you need this. I know you ran into weird trouble with numbers but the next person won't know that.

components/Palettes/src/UIColor+MDCPalettable.m
```diff
@@ -0,0 +1,165 @@
...
+  for (size_t i = 0; i < numComponents + 1; ++i) {
+    if (fabs(components[i] - otherComponents[i]) > 0.001) {
+      return NO;
+    }
+  }
+  return YES;
+}
+
++ (MDCPalette *)paletteForColor:(UIColor *)color
+                    darkerColor:(UIColor * __autoreleasing*)darkerColor
```
willlarche: Why you need `__autoreleasing` here? Actually asking. I never use it.

components/Palettes/src/UIColor+MDCPalettable.m
```diff
@@ -0,0 +1,165 @@
...
+          return palette;
+        }
+      }
+    }
+  }
+  return nil;
+}
+
+- (void)mdc_assignPaletteColors {
+  NSValue *flag = objc_getAssociatedObject(self, _cmd);
```
willlarche: I don't love that the object doesn't start with `is` or `has`. I know it's the method signature but maybe it should be a const? Is that possible with associated objects?

components/Palettes/tests/unit/UIColor+MDCPalettableTests.m
```diff
@@ -0,0 +1,55 @@
...
+@implementation UIColor_MDCPalettableTests
+
+- (void)testBluePalette {
+  // Given
+  MDCPalette *bluePalette = MDCPalette.bluePalette;
+  UIColor *tint500 = bluePalette.tint500;
+  UIColor *accent200 = bluePalette.accent200;
+
+  // Then
+  XCTAssertEqual(tint500.mdc_darkerColorInPalette, bluePalette.tint600);
```
willlarche: Ah. So, here's what's bothering me about `darkerColor...` and `lighterColor...` stuff: It's the NEXT darker and NEXT lighter, right? Just darker or just lighter is inspecific.

willlarche (changes requested):
Good work!

willlarche:
And speaking of contrast ratios, what is up with GH?:
screen shot 2018-01-28 at 8 32 11 pm

components/Buttons/src/TitleColorAccessibilityMutator/MDCButtonTitleColorAccessibilityMutator.m
romoore: Agreed. The naming was pulled from the existing MDFTextAccessibility method.  This is just up for review (it won't be merged as-is) so we can debate names and APIs.  I'm hoping for a general LGTM versus OMG, NO!

components/Buttons/src/TitleColorAccessibilityMutator/MDCButtonTitleColorAccessibilityMutator.m
romoore: Yeah, it's a search direction. As the esteemed @ianegordon  likes to say, "naming is hard." 😝

components/Palettes/src/MDCPalettes.m
romoore: Since it's static, we can drop "MDC". PreparePaletteColors(...)?

components/Palettes/src/UIColor+MDCPalettable.m
romoore: **Apprentice**
This [StackOverflow thread talks about why to put the `__autoreleasing` attribute on the method parameters](https://stackoverflow.com/questions/23268576/autoreleasing-in-errornserror-autoreleasing-outerror).  Basically, it's making explicit what ARC will do automatically (for this non-templated parameter).  Since Apple usually makes it explicit for its in-out parameters, I like to do so as well.

**Journeyperson**
A second [StackOverflow thread has a pretty good answer about what happens when a local variable is passed-by-reference to a function as an in-out parameter](https://stackoverflow.com/questions/8862023/in-which-situations-do-we-need-to-write-the-autoreleasing-ownership-qualifier).  

**Master**
This third [StackOverflow thread has a very detailed discussion of what happens to all the objects](https://stackoverflow.com/questions/8814718/handling-pointer-to-pointer-ownership-issues-in-arc).

components/Palettes/tests/unit/UIColor+MDCPalettableTests.m
romoore: Glad I added some tests!

components/Palettes/src/UIColor+MDCPalettable.m
romoore: Done

components/Palettes/src/UIColor+MDCPalettable.m
romoore: Done

components/Palettes/src/MDCPalettes.m
romoore: Ah, I like "LinkPaletteColors" even better!

components/Buttons/src/TitleColorAccessibilityMutator/MDCButtonTitleColorAccessibilityMutator.m
romoore: How about `searchDarkerColors`?

willlarche: Yeah! I like that.

components/Palettes/src/MDCPalettes.m
willlarche: Yes!

components/Palettes/src/UIColor+MDCPalettable.m
willlarche: TY!

willlarche (dismissed):
<3

components/Palettes/src/UIColor+MDCPalettable.h
```diff
@@ -0,0 +1,28 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+#import "MDCPalettes.h"
+
+@protocol MDCPalettable <NSObject>
```
ianegordon: I'm surprised to see us defining a protocol in a class extension.

components/Palettes/src/UIColor+MDCPalettable.h
```diff
@@ -0,0 +1,28 @@
...
+
+@property(nonatomic, strong, nullable) UIColor *mdc_nextDarkerColorInPalette;
+@property(nonatomic, strong, nullable) UIColor *mdc_nextLighterColorInPalette;
+
+@end
+
+@interface UIColor (MDCPalettable) <MDCPalettable>
```
ianegordon: This seems odd.
Can we just add the methods directly on the class?
What happens if self isn't in a palette?

Would this make more sense as methods in MDCPalette?

components/Palettes/src/UIColor+MDCPalettable.h
romoore: I agree and this can easily be removed - this was a quick-and-dirty proof of concept, so I didn't do a lot of clean up along the way.

components/Palettes/src/UIColor+MDCPalettable.h
romoore: Alternative version contained within [MDCPalette](https://github.com/material-components/material-components-ios/pull/2866/commits/b2d9c5937e64ca0adbf7f1a57b2e417ec56ff13a#diff-98bf3a9381b20b3567d58e353c79249eR218).

components/Buttons/src/TitleColorAccessibilityMutator/MDCButtonTitleColorAccessibilityMutator.m
```diff
@@ -19,9 +19,74 @@
 #import <MDFTextAccessibility/MDFTextAccessibility.h>
 #import "MaterialButtons.h"
 #import "MaterialTypography.h"
+#import "MaterialPalettes.h"
 
 @implementation MDCButtonTitleColorAccessibilityMutator
 
++ (UIColor *)passingPaletteColorForColor:(UIColor *)color
```
jverkoey: nearest?

components/Buttons/src/TitleColorAccessibilityMutator/MDCButtonTitleColorAccessibilityMutator.m
jverkoey: nearest?

romoore: `nearestPassingColorForColor:`?

jverkoey: Better, though I'm still not sure what "passing" means in this context :S

romoore: Nor am I, but I leave that definition to MDFTextAccessibility.

components/Palettes/src/MDCPalettes.h
```diff
@@ -215,4 +215,24 @@ CG_EXTERN const MDCPaletteAccent _Nonnull MDCPaletteAccent700Name;
...
 
+/**
+ Returns the next darker color within the Palette and section (tint or accent). If no darker color
+ exists or @c color is not a predefined palette color, then @c nil is returned.
+
+ @param color used as the base color to perform a palette search for the next darker color.
+
+ @returns the next darker color in the palette, if it exists.
+ */
++ (nullable UIColor *)nextDarkerColorInPaletteForColor:(nullable UIColor *)color;
```
ajsecord, line 226: Why are these class methods? The implementation that searches all possible palettes seems problematic and doesn't work with user-defined palettes. Recommend making these normal methods.

ajsecord (changes requested):

components/Palettes/src/MDCPalettes.h
ajsecord: If you absolutely have to use this kind of logic in a situation where you don't already know the right palette to search, then I would recommend breaking out a method that can search a set of palettes for a closest match to a target color. Then the documentation for that method can warn about the performance impact, what happens if there are multiple closest matches, etc.

ianegordon: Naming is hard.

InPalette is implicit as we will be invoking this method on one.

darkerColorThanColor:
colorDarkerThanColor:
darkerColor:

Naming is hard.

romoore: I agree that it only works with predefined methods and the search is undesirable. If we place these methods on a specific palette, how will mutators or views know which palette a color came from?  The original proposal was a category on UIColor, (which could have worked better with custom palettes) but attaching Palette-related data to UIColors was considered undesirable.

Do you think it's better for clients to send palette objects to each view, or did you have another approach in mind?

romoore: For the performance, I wrote a simple test that iterates over several colors and attempts a lookup for lighter/darker colors of each.  I ran it on an iPhone 5c/iOS 10.3.3.

```objectivec
for (int r = 0; r <= 255; r += 8) {
    for (int g = 0; g <= 255; g += 8) {
      for (int b = 0; b <= 255; b += 8) {
        @autoreleasepool {
          UIColor *color = [UIColor colorWithRed:(CGFloat)(r/255.0)
                                           green:(CGFloat)(g/255.0)
                                            blue:(CGFloat)(b/255.0)
                                           alpha:1];
          UIColor *darker = [MDCPalette nextDarkerColorInPaletteForColor:color];
          UIColor *lighter = [MDCPalette nextLighterColorInPaletteForColor:color];
          if (darker || lighter) {
            ++paletteColorsFound;
          }
          ++totalColorsFound;
        }
      }
    }
  }
```

```bash
2018-03-14 15:36:02.900840-0400 MDCCatalog[2874:7883980] Palette colors found 1/32768: 34.688088
2018-03-14 15:36:03.482343-0400 MDCCatalog[2874:7883980] Palette colors found 1/32768: 0.581365
```

This aligns with Time Profiler results on the same device (32.8 seconds):

timeprofiler-iphone5c-32k_colors

The average is 1 ms to determine the darker/lighter colors of an arbitrary color the first time. Subsequent lookups are cached (until the cache is invalidated) and are around 17 µs.

ajsecord: If a color came from a palette, then making code guess the palette after the fact seems a bit silly. If you need a palette in a method, pass the palette into that method.

I would like the following changes to this code:
1. Switch the `nextDarkerColorInPaletteForColor:` and friends to work on an instance of MDCPalette. The user is required to know the palette and we won't be guessing it.
2. If you think guessing a closest known palette from an input color is still valuable, then create a completely separate implementation of that can take an input color and a set of palettes and return an appropriate palette that can be used with `nextDarkerColorInPaletteForColor:`, etc.

romoore: I can make the change to (1), thanks for the feedback.

Regarding (2), I think it's still valuable. Consider a team that builds VCs that are focused on layout, presentation, binding data to the view, etc.  In those cases they make a simple call to an accessibility mutator (or something similar) that modifies the colors in a view.  Instead of the Palette(s) needing to be injected all the way along the view hierarchy, a quick search can be performed to identify a matching color and then mutate using that if needed.

ajsecord: Yep, ok. So if you think 2) is valuable, I would expect an interface like this on MDCPalette:

`+ (??)paletteAndClosestColorFromPalettes:(NSArray<MDCPalette>*)palettes andInputColor:(UIColor *)inputColor`

This would return a pair of `<MDCPalette, MDCPaletteTint>` that indicate the closest palette entry found in any of  `palettes`. (Returning this pair is a bit clunky, but the information coming out is the same.)

The key is to not bake the known set of palettes into this method, because that's just a giant global variable.

You can also provide a _convenience_ method that does search the set of known palettes, but that's clearly documented as a convenience and not the main method.

PR #2871, opened 2018-01-26
[Docs] Adds missing Jazzy yaml files.

shyndman:


ianegordon:
Thank you @shyndman !

ianegordon (approved):

shyndman:
Thank _you_. When will this make its way to stable?

PR #2912, opened 2018-02-07
[CollectionCells] Support images larger than 40x40 in MDCCollectionViewTextCell

### LINKED ISSUE #2911: [CollectionCells] MDCCollectionViewTextCell imageView overlaps text if image is large
The following logic in MDCCollectionViewTextCell tells the UIImageView to size to fit the underlying image:

https://github.com/material-components/material-components-ios/blob/4a011e75a55fa1b0c592c44578a28162ddfa81ce/components/CollectionCells/src/MDCCollectionViewTextCell.m#L161

The logic assumes imageView.image will be at most 40 x 40 pixels in size.

However, if imageView.image is larger than this, the logic causes the image view to extend past the bounds of the cell and overlap the text (which is always laid out with x origin 16 + 40 + 16 = 72 pixels):

https://github.com/material-components/material-components-ios/blob/4a011e75a55fa1b0c592c44578a28162ddfa81ce/components/CollectionCells/src/MDCCollectionViewTextCell.m#L47

bhamiltoncx:
Calling -[UIImageView sizeToFit] will resize the view to fit the image inside.

MDCCollectionViewTextCell currently assumes the image will be 40x40, but
doesn't handle images other than this size.

Currently, if the image is larger than 40x40, the -sizeToFit call on
the image view will cause the view to go outside the bounds of
MDCCollectionViewTextCell and overlap the text.

Closes #2911.

Reviewers: I'd like to add unit tests for this, but I'm not sure how to run MDC tests.
Any suggestions?

### DIFF
--- a/components/CollectionCells/src/MDCCollectionViewTextCell.m
+++ b/components/CollectionCells/src/MDCCollectionViewTextCell.m
@@ -46,6 +46,10 @@ static inline CGFloat CellDefaultDetailTextFontOpacity(void) {
   return [MDCTypography captionFontOpacity];
 }
 
+// Image size.
+static const CGFloat kImageSize = 40;
+// Cell image view padding.
+static const CGFloat kCellImagePaddingLeading = 16;
 // Cell padding top/bottom.
 static const CGFloat kCellTwoLinePaddingTop = 20;
 static const CGFloat kCellTwoLinePaddingBottom = 20;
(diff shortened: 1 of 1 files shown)


ianegordon:
Hi @bhamiltoncx !  Thanks for the PR.  Sorry for the slow response, we were in a code freeze last week.  We'll take a look at this now.

ianegordon:
Before / After
screen shot 2018-02-14 at 12 58 35
screen shot 2018-02-14 at 14 13 51

ianegordon (approved):

ianegordon:
Thank you for the fix @bhamiltoncx 
👍

Issue #2917, opened 2018-02-09
Icons collisions possible 


randallli:
Icons we offer via MDCIcons may collide with  other icons:

internal bug: b/73132837

romoore:
Let's try to add a test to cover the possibility of a nil bundle coming back in [`[MDCIcon bundleNamed]`](https://github.com/material-components/material-components-ios/blob/develop/components/private/Icons/src/MDCIcons.m#L34)

ianegordon:
This is caused by icon de-duplication as part of our internal build process.
It is out of scope of MDC-iOS.

PR #2918, opened 2018-02-09
[MDCIcon] fixed bundle cache call because bundle may not exist

randallli:
The bundle may not be there so we should guard against adding it to the cache since that may crash the any app that uses this code.

Partial fix to https://github.com/material-components/material-components-ios/issues/2917

### DIFF
--- a/components/private/Icons/src/MDCIcons.m
+++ b/components/private/Icons/src/MDCIcons.m
@@ -32,7 +32,9 @@
     NSBundle *baseBundle = [NSBundle bundleForClass:[MDCIcons class]];
     NSString *bundlePath = [baseBundle pathForResource:bundleName ofType:@"bundle"];
     bundle = [NSBundle bundleWithPath:bundlePath];
-    [bundleCache setObject:bundle forKey:bundleName];
+    if (bundle) {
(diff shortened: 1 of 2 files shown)


components/private/Icons/src/MDCIcons.m
```diff
@@ -32,7 +32,9 @@ + (nullable NSBundle *)bundleNamed:(nonnull NSString *)bundleName {
     NSBundle *baseBundle = [NSBundle bundleForClass:[MDCIcons class]];
     NSString *bundlePath = [baseBundle pathForResource:bundleName ofType:@"bundle"];
     bundle = [NSBundle bundleWithPath:bundlePath];
-    [bundleCache setObject:bundle forKey:bundleName];
+    if (bundle) {
```
romoore, line 35: Please add a unit test to cover this case.

romoore (approved):
LGTM, but a unit test (or a fix-now Issue) for the problematic situation would be best.

components/private/Icons/src/MDCIcons.m
randallli: Done.

yarneo (approved):

PR #2920, opened 2018-02-12
[TextFields] Correcting color themer

willlarche:
This color was not supposed to be set.

### DIFF
--- a/components/TextFields/src/ColorThemer/MDCTextFieldColorThemer.m
+++ b/components/TextFields/src/ColorThemer/MDCTextFieldColorThemer.m
@@ -49,14 +49,6 @@
   if ([textInputControllerClass respondsToSelector:@selector(setActiveColorDefault:)]) {
     [textInputControllerClass setActiveColorDefault:colorScheme.primaryColor];
   }
-
-  if ([textInputControllerClass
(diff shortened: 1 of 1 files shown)


ajsecord (approved):

willlarche:
Tested locally to recreate Travis.

components/TextFields/src/ColorThemer/MDCTextFieldColorThemer.m
```diff
@@ -49,14 +49,6 @@ + (void)applyColorScheme:(id<MDCColorScheme>)colorScheme
   if ([textInputControllerClass respondsToSelector:@selector(setActiveColorDefault:)]) {
     [textInputControllerClass setActiveColorDefault:colorScheme.primaryColor];
   }
-
-  if ([textInputControllerClass
```
romoore, line 53: Does the same change need to be made for `floatingPlaceholderNormalColor`? It seems like the default would not change instances that are explicitly themed.

willlarche: Very good catch!

willlarche: Of course, this whole thing makes me wonder if I should be using the secondary color at all.........

Issue #2904, opened 2018-02-06
Some of our classes do not conform to NSSecureCoding ⚠️


stephanemoore:
Source:
https://github.com/material-components/material-components-ios/blob/fdfb54f/components/Themes/src/MDCTonalPalette.h
https://github.com/material-components/material-components-ios/blob/fdfb54f/components/Themes/src/MDCTonalPalette.m

NSSecureCoding documentation:
https://developer.apple.com/documentation/foundation/nssecurecoding?language=objc
http://nshipster.com/nssecurecoding/

ianegordon:
Audit our components to ensure that all are conforming to NSSecureCoding.
Context in the links above.
MDCTonalPalette is an example where we are not.

mohammadcazig:
All tracked and done.

PR #2928, opened 2018-02-13
[MDCTonalPalette] Adding NSSecureCoding. #2904

mohammadcazig:


components/Themes/src/MDCTonalPalette.m
```diff
@@ -59,7 +59,7 @@ - (instancetype)initWithCoder:(NSCoder *)coder {
   self = [super init];
```
romoore, line 59: Make sure to call [super initWithCoder:]

romoore (dismissed):
Please add NSCoding tests for this class.

components/Themes/src/MDCTonalPalette.m
mohammadcazig: Super Class NSObject.

mohammadcazig:
Added some more tests and change the test class name.

ianegordon:
Could you investigate why CI is failing?

Test Case '-[MDCTonalPaletteTests testTonalPaletteEncodingDecoding]' started.
components/Themes/tests/unit/MDCTonalPaletteTests.m:59: error: -[MDCTonalPaletteTests testTonalPaletteEncodingDecoding] : (([decodedTonalPalette.colors isEqualToArray:tonalPalette.colors]) is true) failed
components/Themes/tests/unit/MDCTonalPaletteTests.m:62: error: -[MDCTonalPaletteTests testTonalPaletteEncodingDecoding] : ((decodedTonalPalette.lightColor) equal to (tonalPalette.lightColor)) failed: ("<10388b6f dd7f0000>") is not equal to ("<c0c1896f dd7f0000>")
components/Themes/tests/unit/MDCTonalPaletteTests.m:63: error: -[MDCTonalPaletteTests testTonalPaletteEncodingDecoding] : ((decodedTonalPalette.mainColor) equal to (tonalPalette.mainColor)) failed: ("<d03a8b6f dd7f0000>") is not equal to ("<c0c3896f dd7f0000>")
components/Themes/tests/unit/MDCTonalPaletteTests.m:64: error: -[MDCTonalPaletteTests testTonalPaletteEncodingDecoding] : ((decodedTonalPalette.darkColor) equal to (tonalPalette.darkColor)) failed: ("<503b8b6f dd7f0000>") is not equal to ("<40c4896f dd7f0000>")
Test Case '-[MDCTonalPaletteTests testTonalPaletteEncodingDecoding]' failed (0.005 seconds).
Test Suite 'MDCTonalPaletteTests' failed at 2018-02-13 12:52:28.949.
	 Executed 1 test, with 4 failures (0 unexpected) in 0.005 (0.005) seconds
Test Suite 'unit_tests.xctest' failed at 2018-02-13 12:52:28.949.
	 Executed 1 test, with 4 failures (0 unexpected) in 0.005 (0.005) seconds
Test Suite 'All tests' failed at 2018-02-13 12:52:28.949.
	 Executed 1 test, with 4 failures (0 unexpected) in 0.005 (0.007) seconds

romoore:
This may be an issue on 32-bit simulators where the unarchived UIColor uses single-precision floating point values instead of double-precision.

mohammadcazig:
Investigated the issue. so similar to what Rob said, it seems that archiving and unarchiving happens in single precision and then the comparison on double precision. Based on what I read and tested this seems like an apple internal bug.
https://github.com/material-components/material-components-ios/issues/2931

components/Themes/tests/unit/MDCTonalPaletteTests.m
```diff
@@ -60,6 +60,17 @@ - (void)testTonalPaletteEncodingDecoding {
   XCTAssertEqual(decodedTonalPalette.mainColorIndex, tonalPalette.mainColorIndex);
   XCTAssertEqual(decodedTonalPalette.darkColorIndex, tonalPalette.darkColorIndex);
   XCTAssertEqual(decodedTonalPalette.lightColorIndex, tonalPalette.lightColorIndex);
+
+  // These tests are commented out because of precision bug on apple side when archiving.
+  // Issue #2931.
+  // TODO: Uncomment these tests when this issue is fixed.
```
ianegordon: Super-(SUPER!!!)-nit: I wouldn't bother this time since it will kick off CI again, but consider:
// TODO(#2931): Re-enable tests once Apple's precision bug is fixed
We usually include the issue number in parens after the TODO.

romoore: I would recommend removing the TODO. I don't think Apple's going to change the behavior - we've just worked around it in other tests.

components/Themes/tests/unit/MDCTonalPaletteTests.m
```diff
@@ -60,6 +60,17 @@ - (void)testTonalPaletteEncodingDecoding {
...
+//  XCTAssertTrue([decodedTonalPalette.colors isEqualToArray:tonalPalette.colors]);
```
romoore: Does this fail with XCTAssertEqualObjects?  Another solution is to switch to single-precision color values in the initializers.

mohammadcazig: Done.

components/Themes/tests/unit/MDCTonalPaletteTests.m
mohammadcazig: Done.

ianegordon:
Kokoro is having trouble today, but Travis is back on point.

Issue #2934, opened 2018-02-14
Strong references instead of weak references incidents.


mohammadcazig:
After dealing with implicit self retains in issue: #2933 , there are some cases where we need to use weak references instead of strong so we don't hold on to objects after they are deallocated.
Files that need to be investigated:
MDCBottomAppBarView.m
MDCNavigationBar.m
MDCPageControl.m

mohammadcazig:
For MDCNavigationBar.m I did not find a reason why we handling the case of adding observers from different threads and it seemed safer to leave the code as it is.

mohammadcazig:
I feel like this change does not give us a lot of benefits and it might includes some possible risks. So we decided not to make the change.

PR #2942, opened 2018-02-15
[MDCPageControl][MDCBottomAppBarView] Strong/Weak Self references inside block.

mohammadcazig:
change strong references to weak references so we don't hold on to the components more than we need too.
#2934

### DIFF
--- a/components/BottomAppBar/src/MDCBottomAppBarView.m
+++ b/components/BottomAppBar/src/MDCBottomAppBarView.m
@@ -323,10 +323,11 @@ static const int kMDCButtonAnimationDuration = 200;
   }
   if (animated) {
     [_floatingButton setElevation:1 forState:UIControlStateNormal];
+    __weak typeof(self) weakSelf = self;
     dispatch_after(dispatch_time(DISPATCH_TIME_NOW, kMDCButtonAnimationDuration * NSEC_PER_MSEC),
                    dispatch_get_main_queue(), ^{
-                     [self insertSubview:self.floatingButton atIndex:subViewIndex];
-                     [self.floatingButton setElevation:elevation forState:UIControlStateNormal];
+                     [weakSelf insertSubview:weakSelf.floatingButton atIndex:subViewIndex];
+                     [weakSelf.floatingButton setElevation:elevation forState:UIControlStateNormal];
                    });
   } else {
     [self insertSubview:_floatingButton atIndex:subViewIndex];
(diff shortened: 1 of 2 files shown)


components/BottomAppBar/src/MDCBottomAppBarView.m
```diff
@@ -323,10 +323,11 @@ - (void)setFloatingButtonElevation:(MDCBottomAppBarFloatingButtonElevation)float
   }
   if (animated) {
     [_floatingButton setElevation:1 forState:UIControlStateNormal];
+    __weak typeof(self) weakSelf = self;
     dispatch_after(dispatch_time(DISPATCH_TIME_NOW, kMDCButtonAnimationDuration * NSEC_PER_MSEC),
                    dispatch_get_main_queue(), ^{
-                     [self insertSubview:self.floatingButton atIndex:subViewIndex];
-                     [self.floatingButton setElevation:elevation forState:UIControlStateNormal];
+                     [weakSelf insertSubview:weakSelf.floatingButton atIndex:subViewIndex];
```
romoore, line 329: I think you would want to strongify here.  Calling `-insertSubview:atIndex:` cannot accept a `nil` view.  Even though I believe the right hand side is evaluated first (leaving the receiver `nil`), it still feels like a risk.

components/PageControl/src/MDCPageControl.m
```diff
@@ -171,16 +171,18 @@ - (void)setCurrentPage:(NSInteger)currentPage
       // We are using the delay to increase the time between the end of the extension of the track
       // ahead of the dots movement and the contraction of the track under the dot at the
       // destination.
+      __weak typeof(self) weakSelf = self;
```
romoore, line 174: Can you write a unit test to cover this behavior before/after?

components/PageControl/src/MDCPageControl.m
```diff
@@ -264,16 +266,21 @@ - (void)scrollViewDidScroll:(UIScrollView *)scrollView {
     // If the animation block has a delay it translates to the beginTime of the CAAnimation. We need
     // to ensure that we delay our animation of the page control to keep in sync with the animation
     // of the scrollView.contentOffset.
+    __weak typeof(self) weakSelf = self;
```
romoore, line 269: Please add a unit test for this code as well.

romoore (changes requested):

mohammadcazig:
I feel like this change does not give us a lot of benefits and it might includes some possible risks. So we decided not to make the change.

PR #2956, opened 2018-02-16
[Ink] New bounded ink should ignore maxRippleRadius property

### LINKED ISSUE #329: Bounded ink should ignore the maxRippleRadius property
Right now it's possible to have bounded ink (should fill the view) with a smaller maxRippleRadius and the visual result looks like unbounded ink (a circle). 
1. Create an MDCFlatButton
2. Set .inkMaxRippleRadius to some smaller value.
3. Result: Circular ink, not filling the view.

In MDCInk, just ignore maxRippleRadius if the style is unbounded.

romoore:
For the new Ink, when the view uses `.bounded` for `inkStyle`, it should
ignore the `maxRippleRadius` property. No changes should be made to legacy Ink
since we're migrating away from it and we don't want to make unnecessary
changes to existing clients.

Closes #329


components/Ink/src/MDCInkView.m
```diff
@@ -193,7 +205,8 @@ - (CGFloat)maxRippleRadius {
 }
 
 - (void)setMaxRippleRadius:(CGFloat)radius {
-  if (self.inkLayer.maxRippleRadius != radius) {
+  if (!MDCCGFloatEqual(self.inkLayer.maxRippleRadius, radius)) {
+    _maxRippleRadius = radius;
```
ianegordon: Don't we need to check for inkStyle before we set maxRippleRadius below?
Otherwise what happens if we setInkStyle:Unbounded, then setRadius:12 ?

ianegordon (dismissed):

components/Ink/src/MDCInkView.m
```diff
@@ -18,6 +18,7 @@
 
 #import "private/MDCInkLayer.h"
 #import "private/MDCLegacyInkLayer.h"
+#import "MaterialMath.h"
```
ajsecord: Sort capitals before lowercase.

components/Ink/src/MDCInkView.m
yarneo: Also what happens if `maxRippleRadius` is set to a number initially, but because it is bounded `maxRippleRadius` is 0 (because default is Bounded). Then someone changes the radius to 0 explicitly and makes the ink Unbounded. The number retained in `_maxRippleRadius` will show the previous number and not 0, because it doesn't pass the `MDCCGFloatEqual` conditional above.

romoore: Done.

components/Ink/src/MDCInkView.m
```diff
@@ -193,7 +205,17 @@ - (CGFloat)maxRippleRadius {
 }
 
 - (void)setMaxRippleRadius:(CGFloat)radius {
-  if (self.inkLayer.maxRippleRadius != radius) {
+  _maxRippleRadius = radius;
+  BOOL rippleRadiusChanged = !MDCCGFloatEqual(self.inkLayer.maxRippleRadius, radius);
+  // New Ink will only update inkLayer.maxRippleRadius when using Unbounded ink
+  if (!self.usesLegacyInkRipple) {
+    if (self.inkStyle == MDCInkStyleUnbounded && rippleRadiusChanged) {
```
yarneo: is it more accurate to say here if it isn't Bounded rather than if it's Unbounded? Looks like we want to update maxRippleRadius only when inkStyle is not Bounded?

components/Ink/src/MDCInkView.m
```diff
@@ -193,7 +205,17 @@ - (CGFloat)maxRippleRadius {
 }
 
 - (void)setMaxRippleRadius:(CGFloat)radius {
-  if (self.inkLayer.maxRippleRadius != radius) {
+  _maxRippleRadius = radius;
```
yarneo: aren't we losing the original maxRippleRadius here if we set it to 0 when bounded? so when it is set to unbounded we will get 0 and not the original value

components/Ink/src/MDCInkView.m
romoore: The style is `.unbounded` so I'm trying to work with that terminology. I'm not sure how negating the value of an `NS_ENUM` would make things clearer. Would you prefer to use a `switch` statement instead?

Do you have a suggestion you could put in a snippet?

components/Ink/src/MDCInkView.m
romoore: I'm not sure I understand. The ivar was not previously being used and now it's only used to "restore" the assigned value when switching to `.unbounded` ink.

components/Ink/src/MDCInkView.m
yarneo: From what I understand we want the `maxRippleRadius` to change only in the case where it is not bounded right? Because bounded it needs to be set to 0 no matter what. In that case all other options should allow to have the `maxRippleRadius` to be set, and not only .unbounded. As an example if we add more `inkStyle`s in the future

components/Ink/src/MDCInkView.m
yarneo: Just tested my hypothesis, this seems fine, please ignore.

components/Ink/src/MDCInkView.m
romoore: I've refactored to use a `switch`. Please take a look.

components/Ink/src/MDCInkView.m
```diff
@@ -193,9 +205,27 @@ - (CGFloat)maxRippleRadius {
...
+  // Legacy Ink updates inkLayer.maxRippleRadius regardless of inkStyle
+  if (self.usesLegacyInkRipple) {
     self.inkLayer.maxRippleRadius = radius;
     [self setNeedsLayout];
+  } else {
+    // New Ink Bounded style ignores maxRippleRadius
+    switch(self.inkStyle) {
+      case MDCInkStyleUnbounded:
+        self.inkLayer.maxRippleRadius = radius;
+        [self setNeedsLayout];
```
ianegordon: Why does this require a setNeedsLayout?
If we need one here, do we need one where we set self.inkLayer.maxRippleRadius above?

ianegordon (dismissed):

yarneo (approved):
LGTM % Ian's comment

components/Ink/src/MDCInkView.m
romoore: Done. It's needed for legacy Ink but not new Ink.

ajsecord (commented):
Drive by comment!

components/Ink/src/MDCInkView.m
romoore: Gah! You caught me! 😨 I was trying to sneak it in...

ianegordon (approved):
LGTM, pending a +1 from Travis.

ianegordon:
Would you kick off another TravisCI integration?

romoore:
Travis is failing to find an iOS11/iPhoneX environment:

```
Failed to test MDCUnitTests in catalog/MDCCatalog.xcworkspace.
Build settings from command line:
    GCC_GENERATE_TEST_COVERAGE_FILES = YES
    GCC_INSTRUMENT_PROGRAM_FLOW_ARCS = YES
xcodebuild: error: Unable to find a destination matching the provided destination specifier:
		{ platform:iOS Simulator, OS:11.0, name:iPhone X }
	The requested device could not be found because no available devices matched the request.

```

In the logs, I don't see any iOS 11.x being listed.

yarneo:
did you restart the job via travis, or did you start a new job? i suspect maybe when restarting a job it might cache certain things, like using the wrong xcode (which was the problem during last weekend)

romoore:
I did neither, I clicked "Restart build
screen shot 2018-02-26 at 2 58 17 pm

romoore:
Travis passed, merging.

PR #2962, opened 2018-02-16
[Palettes] Add shared RGBA -> UIColor utility

### LINKED ISSUE #1221: [Palettes] Replace hex to RGB functions w/ dependency to Palettes and its function

romoore:
Creating a single implementation of uint32_t -> UIColor code. We had several
different (but effectively identical) versions of this same code sprinkled
throughout different components. Although this increases dependency on
Palettes, the size of Palettes is only around 30 KiB.

Closes #1221


ajsecord (changes requested):
We try to reduce near-trivial deps between components to avoid a tangled dependency graph. For one-liner/inlined simple functions, just copy-and-paste. Your instinct is correct, but now you know why we do this.

material-admin:
Spoke offline with @romoore and I think that a private colorUtility component would not tie it to Palettes.

randallli:
Spoke offline with @romoore and I think that a private colorUtility component would not tie it to Palettes.

ajsecord:
Please don't! This isn't worth a separate component.

Issue #2975, opened 2018-02-20
[MDCButtonBar] NSSecureCoding support.


mohammadcazig:
We don't respond to the protocol NSSecureCoding since UIView itself do not support NSSecureCoding. But we use secure encoding/decoding wherever possible.

mohammadcazig:
Done.

PR #2976, opened 2018-02-20
[MDCButtonBar] add NSSecureCoding.

mohammadcazig:
#2975

### DIFF
--- a/components/ButtonBar/src/MDCButtonBar.m
+++ b/components/ButtonBar/src/MDCButtonBar.m
@@ -74,7 +74,7 @@ static NSString *const MDCButtonBarButtonLayoutPositionKey = @"MDCButtonBarButto
 
     if ([coder containsValueForKey:MDCButtonBarItemsKey]) {
       // Force going through the setter to ensure KVO is observed for these items
-      self.items = [coder decodeObjectForKey:MDCButtonBarItemsKey];
+      self.items = [coder decodeObjectOfClass:[NSArray class] forKey:MDCButtonBarItemsKey];
     }
   }
   return self;
--- a/components/ButtonBar/src/MDCButtonBar.m
+++ b/components/ButtonBar/src/MDCButtonBar.m
@@ -79,7 +79,7 @@ static NSString *const MDCButtonBarButtonLayoutPositionKey = @"MDCButtonBarButto
       for (id item in items) {
         if (![item isKindOfClass:[UIBarButtonItem class]]) {
           isValid = NO;
-          NSAssert(NO, @"Wrong class time for MDCButtonBar Items when decoding.");
+          NSAssert(NO, @"Wrong class type for MDCButtonBar items when decoding.");
         }
       }
       if (isValid) {
--- a/components/ButtonBar/src/MDCButtonBar.m
+++ b/components/ButtonBar/src/MDCButtonBar.m
@@ -74,7 +74,17 @@ static NSString *const MDCButtonBarButtonLayoutPositionKey = @"MDCButtonBarButto
 
     if ([coder containsValueForKey:MDCButtonBarItemsKey]) {
       // Force going through the setter to ensure KVO is observed for these items
-      self.items = [coder decodeObjectOfClass:[NSArray class] forKey:MDCButtonBarItemsKey];
+      NSArray *items = [coder decodeObjectOfClass:[NSArray class] forKey:MDCButtonBarItemsKey];
+      BOOL isValid = YES;
+      for (id item in items) {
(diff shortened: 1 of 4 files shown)


components/ButtonBar/src/MDCButtonBar.m
```diff
@@ -74,7 +74,7 @@ - (instancetype)initWithCoder:(NSCoder *)coder {
 
     if ([coder containsValueForKey:MDCButtonBarItemsKey]) {
       // Force going through the setter to ensure KVO is observed for these items
-      self.items = [coder decodeObjectForKey:MDCButtonBarItemsKey];
+      self.items = [coder decodeObjectOfClass:[NSArray class] forKey:MDCButtonBarItemsKey];
```
romoore: Please consider encoding/decoding the class information for the array contents.  `NSArray` adopts `NSCoding`, but not necessarily `NSSecureCoding` Something like this:


```objective
-encodeWithCoder:(NSCoder *)coder {
...
  if (self.items) {
    Class *itemClass = [self.items.firstObject class];
    BOOL allObjectsAreSameClass = YES;
    for (NSObject *item in items) {
      if (![item isKindOfClass:itemClass]) {
        allObjectsAreSameClass = NO;
        break;
      }
    }
    if (allObjectsAreSameClass) {
      [coder encodeObject:itemClass forKey:MDCButtonBarItemsClassKey];
    }
    [coder encodeObject:self.items forKey:MDCButtonBarItemsKey];
  }
...
}

  ...
  if ([coder containsValueForKey:MDCButtonBarItemsKey]) {
    Class *itemClass = nil;
    if ([coder containsValueForKey:MDCButtonBarItemsClassKey]) {
      itemClass = [aDecoder decodeObjectOfClass:[Class class] forKey:MDCButtonBarItemsClassKey];
    }
    NSArray *items = [aDecoder decodeObjectOfClass:[NSArray class] forKey:MDCButtonBarItemsClassKey];
    BOOL allObjectsOfSameOrUnknownClass = YES;
    if (itemClass) {
      for (NSObject *item in items) {
        NSAssert([item isKindOfClass:itemClass]);
        if (![item isKindOfClass:itemClass]) {
          allObjectsOfSameClass = NO;
          break;
        }
      }
    } 
    if (allObjectsOfSameOrUnknownClass) {
      self.items = items;
    }
  }
  ...
}
```

romoore (dismissed):
If you're going to add explicit `NSSecureCoding` support, please make sure to add an implementation for `supportsSecureCoding`. If you just want to make the current `NSCoding` slightly more secure, please update the PR comment.  Either one is fine with me (`UIView` does not support `NSSecureCoding`).

components/ButtonBar/src/MDCButtonBar.m
romoore: Also please add unit tests to verify that the initWithCoder and encodeWithCoder work correctly.

mohammadcazig: I added unit tests and also added verification that decoded objects respond to the given protocol (UIBarButtonItem) but I did not dictate a single class. This solution only covers the case where all items are the same class and I don't think adding this code for special case would not be as effective and helpful.

components/ButtonBar/src/MDCButtonBar.m
```diff
@@ -74,7 +74,17 @@ - (instancetype)initWithCoder:(NSCoder *)coder {
 
     if ([coder containsValueForKey:MDCButtonBarItemsKey]) {
       // Force going through the setter to ensure KVO is observed for these items
-      self.items = [coder decodeObjectForKey:MDCButtonBarItemsKey];
+      NSArray *items = [coder decodeObjectOfClass:[NSArray class] forKey:MDCButtonBarItemsKey];
+      BOOL isValid = YES;
+      for (id item in items) {
+        if (![item isKindOfClass:[UIBarButtonItem class]]) {
+          isValid = NO;
+          NSAssert(NO, @"Wrong class time for MDCButtonBar Items when decoding.");
```
romoore: Nit: s/time/type

components/ButtonBar/src/MDCButtonBar.m
romoore: It looks like you're doing [a subclass check](https://github.com/material-components/material-components-ios/pull/2976/commits/f06832c5b2d5fa965ebbf1d81fb25efde5362e7e#diff-8f04e674a4d1b4e8e5877337478ecc48R80) instead of a protocol check. Is this intentional?

romoore (approved):

PR #2984, opened 2018-02-23
[MDCCard] NSSecureCoding support.

mohammadcazig:
#2983 
We don't respond to the protocol NSSecureCoding since UIView itself do not support NSSecureCoding. But we use secure encoding/decoding wherever possible.


yarneo (commented):
Don't you need to also implement `supportsSecureCoding` ?

yarneo (approved):

PR #2989, opened 2018-02-25
[TextFields] Updating placeholder constraints more often

### LINKED ISSUE #2985: [Chips] ChipField misaligns placeholder text
### Overview
When entering chips into a ChipField, the placeholder text is not correctly aligned after the last entered Chip.  This was not a bug in 44.5.0, and appears to be a regression in 44.6.0

Appears to a side-effect of a recent change in Text Fields (#2885).

### Reproduction steps
1. In "ChipsExamplesSupplemental.m", add to `@implementation ChipsInputExampleViewController (Supplemental)`:
    ```objectivec
    + (BOOL)catalogIsPresentable {
      return YES;
    }
    ```
1. Open the example Chips &rarr; Input
1. Type some text and press Return

### Version number
44.6.0

Any/all. Tested on iPhone 5/iOS 8.3, iOS 9.3 (Simulator) and iPhone 4S/iOS 8.3, iOS 9.3 (Simulator)

### LINKED ISSUE #2986: [TextFields] MDCTextField does not update placeholder position when `textInsets` change
### Overview
When the textInsets changes in the superview, MDCTextField does not relayout placeholder text because it only updates its own constraints if the underline position has changed.  This means that if a delegate has adjusted the textInsets (due to inserting custom subviews like Chips), the placeholder position will not get adjusted.

### Reproduction steps
See the reproduction steps for #2985 (Chips) for a ready example in the Catalog.

### Version number
44.6.0 (not present in 44.5.0)

Any/all

### Related issues
Seems to be the root cause of #2985
Seems to have been exposed by #2885

willlarche:
Fixes #2985
Fixes #2986

### DIFF
--- a/components/TextFields/src/private/MDCTextInputCommonFundament.m
+++ b/components/TextFields/src/private/MDCTextInputCommonFundament.m
@@ -527,7 +527,8 @@ static inline UIColor *MDCTextInputUnderlineColor() {
   [self updatePlaceholderAlpha];
   [self.textInput sendSubviewToBack:_borderView];
 
-  if ([self needsUpdateConstraintsForPlaceholderToOverlayViewsPosition]) {
+  if ([self needsUpdateConstraintsForPlaceholderToTextInsets] ||
+      [self needsUpdateConstraintsForPlaceholderToOverlayViewsPosition]) {
     [self.textInput setNeedsUpdateConstraints];
   }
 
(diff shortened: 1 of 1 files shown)


romoore (approved):
Would you please add unit tests for this change:
* What the constraint values should be given different textField (and _textInput) states.
* When `updateConstraints` needs to be called.

willlarche:
#2997 : task for adding unit test.

PR #2993, opened 2018-02-26
[Chips] NSSecureCoding

mohammadcazig:
We don't respond to the protocol NSSecureCoding since UIView itself do not support NSSecureCoding. But we use secure encoding/decoding wherever possible. #2992


ianegordon (approved):

ianegordon:
Kokoro is down.

PR #2998, opened 2018-02-26
[Button|Ink|TextFields|Palettes] Fixed NSNumber conversion static analyzer errors

### LINKED ISSUE #2935: Static Analyzer - Converting NSNumber to primitive
Open MDCCatalog -> MDCDragons
Product Analyze

Multiple warnings:
Converting a pointer value of type 'NSNumber *' to a primitive boolean value; instead, either compare the pointer to nil or call -boolValue.

sigito:
Fixes #2935.

### DIFF
--- a/components/Buttons/src/MDCButton.m
+++ b/components/Buttons/src/MDCButton.m
@@ -686,13 +686,10 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
 
 - (CGFloat)elevationForState:(UIControlState)state {
   NSNumber *elevation = _userElevations[@(state)];
-  if (state != UIControlStateNormal && !elevation) {
+  if (state != UIControlStateNormal && elevation == nil) {
     elevation = _userElevations[@(UIControlStateNormal)];
   }
-  if (elevation) {
-    return (CGFloat)[elevation doubleValue];
-  }
-  return 0;
+  return (CGFloat)elevation.doubleValue;
 }
 
 - (void)setElevation:(CGFloat)elevation forState:(UIControlState)state {
@@ -727,11 +724,7 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
 #pragma mark - Border Width
 
 - (CGFloat)borderWidthForState:(UIControlState)state {
-  NSNumber *borderWidth = _borderWidths[@(state)];
-  if (borderWidth) {
-    return (CGFloat)borderWidth.doubleValue;
-  }
-  return 0;
+  return (CGFloat)_borderWidths[@(state)].doubleValue;
 }
 
 - (void)setBorderWidth:(CGFloat)borderWidth forState:(UIControlState)state {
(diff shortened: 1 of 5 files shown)


components/Buttons/src/MDCButton.m
```diff
@@ -686,13 +686,10 @@ - (void)setBackgroundColor:(UIColor *)backgroundColor forState:(UIControlState)s
...
   NSNumber *elevation = _userElevations[@(state)];
-  if (state != UIControlStateNormal && !elevation) {
+  if (state != UIControlStateNormal && elevation == nil) {
     elevation = _userElevations[@(UIControlStateNormal)];
   }
-  if (elevation) {
-    return (CGFloat)[elevation doubleValue];
-  }
-  return 0;
+  return (CGFloat)elevation.doubleValue;
```
ajsecord: FYI, we usually prefer an explicit value instead of relying on the Objective-C runtime to return zero when the pointer is nil. The reason is twofold: it shows the programmer thought about the case and deliberately chose zero, and if the value to return changes it can be easily updated.

I usually prefer the `?:` form below, but that's up to the individual programmer.

ajsecord (approved):

components/Buttons/src/MDCButton.m
```diff
@@ -727,11 +724,7 @@ - (void)setBorderColor:(UIColor *)borderColor forState:(UIControlState)state {
 #pragma mark - Border Width
 
 - (CGFloat)borderWidthForState:(UIControlState)state {
-  NSNumber *borderWidth = _borderWidths[@(state)];
-  if (borderWidth) {
-    return (CGFloat)borderWidth.doubleValue;
-  }
-  return 0;
+  return (CGFloat)_borderWidths[@(state)].doubleValue;
```
ianegordon: The implementation of borderWidthForState: and elevationForState: are different.
elevation returns valueForState:UIControlStateNormal if valueForState:state doesn't exist.
This one returns 0 and handles in updateBorderWidth.

Please migrate borderWidthForState to the same implementation as elevationForState.

Understood if you want to do this in a different PR.

ianegordon: return (CGFloat)_borderWidth[@(state)].doubleValue doesn't convey to future engineers that undefined values are explicitly intended to be 0.

Please use the Elvis operator:
return _borderWidth[@(state)] ? (CGFloat)_borderWidths[@(state)].doubleValue : (CGFloat)0;

ianegordon (changes requested):

components/Buttons/src/MDCButton.m
sigito: Oh, I did not know that. Thanks.

I used a ternary instead since we want to unbox the value.

components/Buttons/src/MDCButton.m
sigito: Moved logic from `updateBorderWidth`.

Should we apply the same logic for `titleFontForState:`?

sigito: Done. However, I added a `nil` check, since that is exactly what the analyzer complains about.

ianegordon (dismissed):

ianegordon (changes requested):
LGTM.  Fix casting issues in Travis and we can land.
Thanks for the PR!

sigito:
Done.

ianegordon (approved):

ianegordon:
Thank you @sigito !

PR #2999, opened 2018-02-26
[Collections|TextFields|Themes] Removed dead store static analyzer errors

### LINKED ISSUE #2936: Static Analyzer - Dead Store
Open MDCCatalog -> MDCDragons
Product Analyze

Multiple warnings:
Value stored to 'insets' during its initialization is never read

sigito:
Fixes #2936.

### DIFF
--- a/components/TextFields/src/MDCMultilineTextField.m
+++ b/components/TextFields/src/MDCMultilineTextField.m
@@ -481,7 +481,7 @@ static NSString *const MDCMultilineTextFieldTrailingViewModeKey =
 - (CGFloat)trailingViewAlpha {
   // The trailing view has the same behavior as .rightView in UITextField: It has visual precedence
   // over the clear button.
-  CGFloat trailingViewAlpha = self.trailingView.alpha;
+  CGFloat trailingViewAlpha;
   switch (self.trailingViewMode) {
     case UITextFieldViewModeAlways:
       trailingViewAlpha = 1.f;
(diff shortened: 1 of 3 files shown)


components/TextFields/src/MDCMultilineTextField.m
```diff
@@ -481,7 +481,7 @@ - (void)updateTrailingViewAlpha {
 - (CGFloat)trailingViewAlpha {
   // The trailing view has the same behavior as .rightView in UITextField: It has visual precedence
   // over the clear button.
-  CGFloat trailingViewAlpha = self.trailingView.alpha;
+  CGFloat trailingViewAlpha;
```
ajsecord, line 484: This isn't correct... if `self.trailingViewMode` is garbage (enum has an illegal value), then you'll return an uninitialized value, which could be anything (NaN, -22, etc.).

If you want to keep this change, add a default clause to the switch statement and cover debug/release contingencies, like this:

~~~
switch (self.trailingViewMode) {
  ...
  default:
    assert(NO, @"Invalid enumeration value %i.", self.trailingViewMode);
    trailingViewAlpha = self.trailingViewAlpha;
}

ajsecord (dismissed):

components/TextFields/src/MDCMultilineTextField.m
ianegordon: +1 ajsecord's remark, but please use NSAssert() instead of assert().

sigito: Done.

Although, it's not necessarily incorrect atm since current switch statement is exhaustive. `UITextFieldViewMode` has only 4 possible values. However, it does not protect from future additions to this enum or a programmer error initializing to an invalid value.

ianegordon (approved):

PR #3015, opened 2018-02-28
[BottomNavigation] Fix OS version check

romoore:
Instead of relying on a potentially-broken floating point conversion, we
should be using NSProcessInfo to perform OS version checks.


ajsecord (approved):

romoore:
Ignoring kokoro for now. Travis passed.

PR #3017, opened 2018-02-28
[BottomNavigation] Fix memory leak in example

romoore:
The `BottomNavigationTypicalUseExample` has a retain cycle in the
`dispatch_after` block it uses to update the badge value. This causes
example VCs to be accumulated if they are opened repeatedly. A simple
fix is to keep a weak reference in the dispatched block making the next
update call a no-op.

### DIFF
--- a/components/BottomNavigation/examples/BottomNavigationTypicalUseExample.m
+++ b/components/BottomNavigation/examples/BottomNavigationTypicalUseExample.m
@@ -135,7 +135,7 @@
   self.badgeCount++;
   self.bottomNavBar.items[1].badgeValue = [NSNumber numberWithInt:self.badgeCount].stringValue;
 
-  __weak __typeof__(self) weakSelf = self;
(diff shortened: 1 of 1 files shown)


components/BottomNavigation/examples/BottomNavigationTypicalUseExample.m
```diff
@@ -135,8 +135,9 @@ - (void)updateBadgeItemCount {
   self.badgeCount++;
   self.bottomNavBar.items[1].badgeValue = [NSNumber numberWithInt:self.badgeCount].stringValue;
 
+  __weak __typeof__(self) weakSelf = self;
```
ajsecord: Please don't use `__typeof__`, just be explicit with the type.

ajsecord (changes requested):

ajsecord (approved):

romoore:
Skipping kokoro, Travis passed.

Issue #846, opened 2016-11-09
Remove all usage of typeof() and change .xcconfig to disallow GCC extensions


ajsecord:
typeof() is a GCC extension that we don't want to support in the code base.

Internal build settings disallow it by specifying that the compiler should use C99, not C99+GCC extensions, but it seems that MDCCatalog does not do so.

romoore:
Looks like there are no remaining references to `typeof()` in the catalog and no blockers to switching to `-std=c99` in the project file.

ajsecord:
There's still a [few references](https://github.com/material-components/material-components-ios/search?utf8=%E2%9C%93&q=typeof&type=) kicking around...

willlarche:
I thought __typeof was valid.

romoore:
I also thought this issue was to remove `typeof()` (and replace it with) `__typeof__()`?

ajsecord:
There are two closely-related issues, the first is worse than the second.

`typeof()` is a GNU C extension, which isn't something we want to rely on at all. The only reason clang supports it is for backwards compatibility.

`__foo__()` is a standard way of saying "this is completely compiler-dependent, feel free to blow up if you don't happen to understand it." `__typeof__()` a little more honest than `typeof()` (which looks like a real language feature), but has the same problem as `typeof()`.

These features were added for situations [where you don't know the type](https://gcc.gnu.org/onlinedocs/gcc/Typeof.html) (like when writing a macro where you have to declare a temp variable of the _same type_ as an argument. Of course they can be convenient, but it's better to be explicit than save a few characters (especially in shared library code).

willlarche:
OK! So we'll say the best practice is just being explicit about the class?

ajsecord:
Also, extensions are [called out in the style guide](https://engdoc.corp.google.com/eng/doc/objcguide.xml?showone=Nonstandard_Extensions&cl=head#Nonstandard_Extensions).

Yes, please be explicit about the class.

PR #3018, opened 2018-02-28
[FeatureHighlight] Remove use of __typeof__

romoore:
Instead of relying on the compiler to determine the pointer type,
explicit type references should be used.

Related to #3017


ajsecord (approved):

ianegordon:
__weak __typeof__(self) weakSelf = self; is a common pattern both internally and externally.
Is there a specific concern around this?

romoore:
Using `__typeof__` in #3017 was the initial approach, but @ajsecord  requested the use of an explicit type. In the MDC-iOS codebase, this was the only remaining use of any *typeof* variants.

ajsecord:
This has come up before, `typeof` is annoyingly non-standard, and designed for places where you _can't_ know the correct type (e.g. macros). 

https://github.com/material-components/material-components-ios/issues/846#issuecomment-313124727

ianegordon:
SGTM

PR #3033, opened 2018-03-03
Always encode/decode CGFloat as double

### LINKED ISSUE #2606: Always serialize CGFloats as double
When implementing an encode / decoder, always cast CGFloat value to a double before and after encoding.  This will avoid any precision loss on 64-bit platforms.

sigito:
Fixes #2606.


ianegordon:
decodeDoubleForKey: should decode values encoded as floats without complaint, so there are no concerns around decoding values previous encoded as floats.

ianegordon (approved):
Thank you @sigito !

PR #3067, opened 2018-03-08
[Tabs] Don't call UIAppearance in +initialize

### LINKED ISSUE #3066: [Tabs] MDCTabBar should not set UIAppearance in +initialize
### Overview
`MDCTabBar` is setting its defaults via UIAppearance in `+initialize`. Instead, it should set its defaults in `-init` methods and allow clients to use UIAppearance if desired.

### Version number
46.1.1

romoore:
Instead of calling UIAppearance in +initialize, MDCTabBar should be
assigning its default values during initialization.

Closes #3066

### DIFF
--- a/components/Tabs/src/MDCTabBar.m
+++ b/components/Tabs/src/MDCTabBar.m
@@ -108,16 +108,6 @@ static MDCItemBarAlignment MDCItemBarAlignmentForTabBarAlignment(MDCTabBarAlignm
 
 #pragma mark - Initialization
 
-+ (void)initialize {
-  [MDCTabBar appearance].selectedItemTintColor = [UIColor whiteColor];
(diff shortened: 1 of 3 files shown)


components/Tabs/src/MDCTabBar.m
```diff
@@ -193,6 +183,10 @@ - (instancetype)initWithFrame:(CGRect)frame {
 }
 
 - (void)commonMDCTabBarInit {
+  _selectedItemTintColor = [UIColor whiteColor];
+  _unselectedItemTintColor = [UIColor colorWithWhite:1.0f alpha:0.7f];
+  _inkColor = [UIColor colorWithWhite:1.0f alpha:0.7f];
+
```
mohammadcazig, line 189: [nit]: move these to set default value section

mohammadcazig (approved):

PR #3084, opened 2018-03-13
Implicit self

mohammadcazig:
We should not use implicit self inside block.


romoore:
Looks like the branch needs a rebase?

mohammadcazig:
this has some extra commit in it that I revert I am gonna just make another PR.

mohammadcazig:
actually already fixed.

PR #3098, opened 2018-03-15
[FlexibleHeader] Check if contentView is nil before the comparison in hitTest

siyusong:
This PR addresses the problem that caused the rollback: https://github.com/material-components/material-components-ios/pull/3032#issuecomment-373176162

Even though the property is marked as nonnull, there is still a chance that client can still set it to nil, using -setValue:forKey: or Objective-C runtime mechanisms. When it becomes nil, we will be erroneously forwarding the touches that does not belong to the header view to the tracking scroll view.


siyusong:
CC @randallli, @jverkoey

ianegordon:
Hi @siyusong , thanks for the PR!  There may have been some work done there on this side.  Can you rebase and see if a change is still necessary?

siyusong:
Hi @ianegordon - I think the merge conflict is caused by the revert of my original PR in develop branch. I resolved the merge conflict and still want this PR to be reviewed since it will add back the feature I proposed in #3032 in a (hopefully) safer way.

ianegordon:
Gentle ping to @jverkoey to take a look.

jverkoey (approved):

PR #3099, opened 2018-03-15
Release script: checking out stable before pushing to cocopods.

randallli:
Also keeping the error log when the push fails.

### DIFF
--- a/scripts/release
+++ b/scripts/release
@@ -89,7 +89,7 @@ version_for_platform() {
 # Creates a release-candidate branch based off the latest origin/develop
 cut_release() {
   isHotFix=false;
-  
+
   while test $# -gt 0; do
     case "$1" in
       --hotfix)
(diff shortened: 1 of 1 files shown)


scripts/release
```diff
@@ -410,12 +410,13 @@ publish_podspec() {
   echo "Verifying Cocoapod trunk permissions"
   pod trunk me > cocoapods_permissions.log
   if grep -Fq "MaterialComponents" cocoapods_permissions.log; then
-    echo "Running a pod trunk push..."
+    echo "Running a pod trunk push from stable..."
+    git checkout stable
```
ianegordon, line 414: Before pushing, consider checking that this branch is in a clean state:
git status --porcelain

ianegordon (approved):

scripts/release
randallli: we dont do that anywhere else where we switch to the stable branch. (we do that at the beginning of the release though)

Issue #3115, opened 2018-03-19
[BottomNavigation] Bottom Navigation color themer should use greater contrast between colors


romoore:
### Overview
The MDCBottomNavigationBarColorThemer should apply theme colors so that the background, selected, and unselected items have sufficient contrast. 

For example, the MDC catalog default color scheme uses three colors:

```swift
colorScheme = MDCBasicColorScheme(primaryColor: .init(white: 33 / 255.0, alpha: 1),
                                      primaryLightColor: .init(white: 0.7, alpha: 1),
                                      primaryDarkColor: .init(white: 0, alpha: 1))
```

This produces a BottomNav with a contrast ratio that is too low between the unselected items and the background.

![mdc-bottomnav-catalog-theme](https://user-images.githubusercontent.com/1753199/37614194-76129e2a-2b80-11e8-8508-35f9b2a355df.png)

When creating an MDCBasicColorScheme with a named UIColor like `redColor` or `orangeColor` where the brightness is already maximized, then `primaryLightColor` and `primaryColor` are identical, resulting in disappearing selected items.

![mdc-bottomnav-single-red](https://user-images.githubusercontent.com/1753199/37614227-91912720-2b80-11e8-999f-bade27b3a518.png)
![mdc-bottomnav-single-orange](https://user-images.githubusercontent.com/1753199/37614228-919e1318-2b80-11e8-8d61-e216f17477c3.png)

### Version number
48.0.0

romoore:
Something like

```objectivec
+ (void)applyColorScheme:(id<MDCColorScheme>)colorScheme
    toBottomNavigationBar:(MDCBottomNavigationBar *)bottomNavigationBar {
  bottomNavigationBar.selectedItemTintColor = colorScheme.primaryColor;
}
```

Would look like this:

![simulator screen shot - iphone 4s - 2018-03-19 at 14 30 54](https://user-images.githubusercontent.com/1753199/37614858-2bcb8d0c-2b82-11e8-9918-2d503aaf3813.png)

romoore:
This should be obsolete now that we have semantic color themes available.

PR #3116, opened 2018-03-19
[BottomNavigation] Simplify color themer

romoore:
The BottomNavigationBarColorThemer would not work for color schemes generated
from some named UIColors (like .red, .orange). It also generated a poor
contrast ratio when using the MDC Catalog color scheme (dark greys). Instead,
it should use the `primaryColor` for the selected item and leave the other
at the default.  This should generally create a simple, effective set of
colors.

Mitigates #3115

### DIFF
--- a/components/BottomNavigation/src/ColorThemer/MDCBottomNavigationBarColorThemer.m
+++ b/components/BottomNavigation/src/ColorThemer/MDCBottomNavigationBarColorThemer.m
@@ -20,9 +20,7 @@
 
 + (void)applyColorScheme:(id<MDCColorScheme>)colorScheme
     toBottomNavigationBar:(MDCBottomNavigationBar *)bottomNavigationBar {
-  bottomNavigationBar.selectedItemTintColor = colorScheme.primaryLightColor;
-  bottomNavigationBar.unselectedItemTintColor = colorScheme.primaryDarkColor;
-  bottomNavigationBar.barTintColor = colorScheme.secondaryColor;
+  bottomNavigationBar.selectedItemTintColor = colorScheme.primaryColor;
 }
 
 @end
--- a/components/BottomNavigation/tests/unit/BottomNavigationBarColorThemerTests.m
+++ b/components/BottomNavigation/tests/unit/BottomNavigationBarColorThemerTests.m
@@ -1,0 +1,62 @@
+/*
+ Copyright 2018-present the Material Components for iOS authors. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


components/BottomNavigation/src/ColorThemer/MDCBottomNavigationBarColorThemer.m
```diff
@@ -20,9 +20,7 @@ @implementation MDCBottomNavigationBarColorThemer
 
 + (void)applyColorScheme:(id<MDCColorScheme>)colorScheme
     toBottomNavigationBar:(MDCBottomNavigationBar *)bottomNavigationBar {
-  bottomNavigationBar.selectedItemTintColor = colorScheme.primaryLightColor;
-  bottomNavigationBar.unselectedItemTintColor = colorScheme.primaryDarkColor;
-  bottomNavigationBar.backgroundColor = colorScheme.secondaryColor;
+  bottomNavigationBar.selectedItemTintColor = colorScheme.primaryColor;
```
ianegordon: This looks like a change in existing behavior.  Is there an bug?  Are people relying on the existing behavior?

ianegordon (dismissed):

components/BottomNavigation/src/ColorThemer/MDCBottomNavigationBarColorThemer.m
randallli: rob showed this to me offline: if we applied the color scheme to the bottom nav bar it would do weird coloring. so yes there was is a bug. If you only specified the primary color your secondary color would be the same as the primary and this would result in primary on primary background, icon and text becoming invisible.

randallli (approved):

components/BottomNavigation/src/ColorThemer/MDCBottomNavigationBarColorThemer.m
romoore: There are a few problems here:
* All of those properties are optional, but there is no check being performed.
* If the color scheme does not support any of the properties, no theming gets performed at all.
* Other components (AppBar, BottomAppBar, NavigationBar, ButtonBar, Buttons, FlexibleHeader) use `primaryColor` to set the background color.  
   * HeaderStackView uses primary and primaryLight as background colors.
   * FeatureHighlight uses primary and primaryLight for the background colors.
* Other components use colors for foreground elements but not background.
   * TabBar sets the selected/unselected to light/dark, but not the background color. 
   * Slider uses primary and lightPrimary for the track. 
   * PageControl uses light/dark for the indicators. 
   * ProgressView use primary and lightPrimary for track/progress.

This is the only component where we are using the same scheme to set both foreground/accent and background colors.  Given how MDCBasicColorSchemes are automatically generated and how primary/secondary colors are generally used in our components, it seems better for the BottomNavigation to **either** color the background color or the item tints.

ianegordon (approved):
Discussed offline: we agree this is a better implementation but we will test against internal clients to identify any breakages before landing.

ianegordon:
Please test against internal clients and rebase to fix conflict before merge.

romoore:
Automated internal testing found no breaking changes.

romoore:
Merging develop to hopefully get unrelated tests to pass.

PR #3119, opened 2018-03-20
[BottomNavigation] Remove ink for canceled touch

### LINKED ISSUE #2909: [BottomNavigation]  Abandoned ink persists in bottom navigation bar after rotation (b/73060904)
Originally reported by @brianjmoore 

### Overview
The bottom navigation bar can sometimes end up with abandoned ink that sticks around seemingly indefinitely.

bottomnav-ink-bug

### Reproduction steps
1. Build and run the MDC catalog
2. Open the bottom nav example
3. Touch one of the tabs and while holding rotate the device

RESULT: Ink is abandoned, sticks around indefinitely.

### Version number
44.6.0

iPhone 5/iOS 10.3.3

romoore:
When a touch was canceled (e.g., by rotating the device) and the touch event
was canceled, the Bottom Navigation Bar was not canceling the ink animation.

Closes #2909

### DIFF
--- a/components/BottomNavigation/src/MDCBottomNavigationBar.m
+++ b/components/BottomNavigation/src/MDCBottomNavigationBar.m
@@ -389,6 +389,11 @@ static NSString *const kMDCBottomNavigationBarTitleString = @"title";
   [itemView.inkView startTouchEndedAnimationAtPoint:CGPointZero completion:nil];
 }
 
+- (void)didCancelTouchesForButton:(UIButton *)button {
+  MDCBottomNavigationItemView *itemView = (MDCBottomNavigationItemView *)button.superview;
+  [itemView.inkView cancelAllAnimationsAnimated:NO];
+}
+
 #pragma mark - Setters
 
 - (void)setItems:(NSArray<UITabBarItem *> *)items {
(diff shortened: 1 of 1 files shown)


components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -452,6 +452,9 @@ - (void)setItems:(NSArray<UITabBarItem *> *)items {
     [itemView.button addTarget:self
                         action:@selector(didTouchUpOutsideButton:)
               forControlEvents:UIControlEventTouchUpOutside];
+    [itemView.button addTarget:self
+                        action:@selector(didTouchUpOutsideButton:)
+              forControlEvents:UIControlEventTouchCancel];
```
ajsecord: You don't want animation for cancelations.

ajsecord (dismissed):

components/BottomNavigation/src/MDCBottomNavigationBar.m
romoore: Why not? It looks good for the use cases I could think of:

* Tap-and-hold a tab item and then rotate the device (ink dissipates)
* Tap-and-hold a tab item, then swipe down and swipe up to show the Control Panel.

Is there a use case where we are canceling touch events and Ink should not dissipate?

**Animated Cancel**
bottomnav-ink-fix-animated

**Immediate Cancel**
bottomnav-ink-fix-immediate

romoore: Switched to immediately cancel ink after further discussion with @randallli and @yarneo .

randallli: This should work the same way we cancel the ink when a collection view scroll gesture may initially start ink.

ianegordon (approved):
LGTM

PR #3123, opened 2018-03-20
[Chips] Support Dynamic Type and show it in example.

romoore:
**Entry Chips Before**
chips-input-small-before

**Entry Chips After**
chips-input-small-after

**Chips Small (After)**
chips-small-after

**Chips Large (After)**
chips-large-after


yarneo:
ht_tortilla_chip_2_dm_120621_wblog

components/Chips/tests/unit/ChipsTests.m
```diff
@@ -0,0 +1,115 @@
...
+  brightness = MIN(1, MAX(0, brightness - percent));
+
+  return [UIColor colorWithHue:hue saturation:saturation brightness:brightness alpha:alpha];
+}
+
+static inline UIColor *MDCColorLighten(UIColor *color, CGFloat percent) {
+  return MDCColorDarken(color, -percent);
+}
+
+@interface ChipsNoopTest : XCTestCase
```
yarneo: Did you mean to call this class `ChipsTests` instead?

components/Chips/tests/unit/ChipsTests.m
```diff
@@ -0,0 +1,115 @@
...
+
+@end
+
+@implementation ChipsNoopTest
```
yarneo: Did you mean to call this class `ChipsTests` instead?

yarneo (dismissed):

components/Chips/tests/unit/ChipsTests.m
romoore: Yep, good catch. I renamed the file and forgot to rename the class. :P

yarneo:
One open question I wasn't sure about and wanted to know if you had thought about it. Do you think we might need to update the chipView font in `prepareForReuse` in the `MDCChipCollectionViewCell`?

romoore:
If we touch the font in `prepareForReuse` it will disable UIAppearance. I don't think we should be doing that within the component.

yarneo (approved):

PR #3126, opened 2018-03-21
Use safe selector access for color themers

romoore:
Many of our color themers are accessing optional properties without
first verifying that the color scheme responds to the selector.  This
can result in crashes. Instead, each optional property should be checked
for safety before being used.


ajsecord (approved):
Interesting.

PR #3127, opened 2018-03-21
Support Dynamic Type in extensions for iOS >= 10.

Thomas-Redding-G:
[Typography]Support Dynamic Type in extensions for iOS >= 10


ianegordon:
Hi @Thomas-Redding-G , thanks for the PR!
As we still support Xcode 8 we need guards around the @available check.  See:
https://github.com/material-components/material-components-ios/blob/92fecd4cdeff37592d16ddbe556777299e870573/components/Dialogs/src/MDCDialogPresentationController.m#L84

We expect to drop support for Xcode 8 soon, which should simplify this type of code.
Thanks again!

ianegordon (dismissed):
We need the Xcode 8 guards before we can merge.

ianegordon (approved):
Thank you!
We'll merge as soon as CI is happy.

ianegordon:
Kokoro is unresponsive, but TravisCI was successful.

PR #3128, opened 2018-03-21
[Chips] Create a FontThemer

romoore:
Creating a FontThemer to use MDCFontScheme and apply it to an
MDCChipView.

### DIFF
--- a/components/Chips/tests/unit/ChipViewFontThemerTests.m
+++ b/components/Chips/tests/unit/ChipViewFontThemerTests.m
@@ -20,65 +20,6 @@
 #import "MaterialChips.h"
 #import "MDCChipViewFontThemer.h"
 
-@interface FakeFontScheme : NSObject <MDCFontScheme>
-@end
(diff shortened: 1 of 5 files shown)


components/Chips/tests/unit/ChipViewFontThemerTests.m
```diff
@@ -0,0 +1,100 @@
...
+ limitations under the License.
+ */
+
+#import <XCTest/XCTest.h>
+
+#import "MaterialThemes.h"
+#import "MaterialChips.h"
+#import "MDCChipViewFontThemer.h"
+
+@interface FakeFontScheme : NSObject <MDCFontScheme>
```
mohammadcazig: you could use our basic font scheme and set your own values instead of this.

romoore: Done

ianegordon (approved):

PR #3133, opened 2018-03-22
[BottomAppBar, Tabs] Make themer parameters nonnull

romoore:
Themers should not accept nullable pointers.


ianegordon (approved):
Unfortunately this is a breaking change with no migration path.  However, the benefits far outweigh the cost.

PR #3182, opened 2018-03-29
[Dialogs] Fixed regression where in some cases tapping on the buttons in the alert would not do anything

jpoumailloux:
Specifically, we are seeing this behavior (b/77155189 and b/77152040) when a text field is first responder when the alert is presented. When that happens, -[MDCAlertController actionButtonPressed:] (ie. the method expected to be called by UIKit) doesn't get called when the user taps on the buttons.

Per UIKit's documentation, specifying a nil target in -[UIControl addTarget:action:forControlEvents:] is supported ("passing in nil as the target goes up the responder chain"), however it seems to me that it is preferable to be explicit in order to be more robust to the responder chain being not exactly the way it should be.

This is also the previous behavior of MDCAlertController prior to https://github.com/material-components/material-components-ios/commit/b22c8c2650a4a5b21c76cf37b3d1041de639e4e5


romoore (approved):
Seems reasonable to me.

ianegordon (approved):
LGTM!
Will merge once CI is happy.

components/Dialogs/src/private/MDCAlertControllerView+Private.m
```diff
@@ -129,7 +129,7 @@ - (void)addActionButtonTitle:(NSString *)actionTitle selector:(SEL)selector {
   buttonRect.size.height = MAX(buttonRect.size.height, MDCDialogActionButtonHeight);
   buttonRect.size.width = MAX(buttonRect.size.width, MDCDialogActionButtonMinimumWidth);
   actionButton.frame = buttonRect;
-  [actionButton addTarget:nil
```
ianegordon, line 132: We aren't changing any external APIs.

Can we simplify this entire PR to:
[actionButton addTarget:self

ianegordon (changes requested):

components/Dialogs/src/private/MDCAlertControllerView+Private.m
jpoumailloux: We can't do that because self would then refer to the MDCAlertControllerView instance, whereas we want the MDCAlertController (ie. not the view) to be the receiver of the method call. MDCAlertControllerView+Private.h is private and was introduced very recently (13 days ago).

ianegordon: Got it.

ianegordon (approved):

PR #3193, opened 2018-03-29
[NavigationBar] Fix RTL for titleView

ianegordon:
We were applying RTL to titleLabel, but not titleView.

cl/190110908

Imported fix and added test sample for Dragons.

Previous RTL with overlapping views

screen shot 2018-03-29 at 18 44 07

Fixed RTL with proper layout

screen shot 2018-03-29 at 18 36 01

### DIFF
--- a/components/NavigationBar/examples/NavigationBarRTLIcons.m
+++ b/components/NavigationBar/examples/NavigationBarRTLIcons.m
@@ -1,0 +1,134 @@
+/*
+ Copyright 2017-present the Material Components for iOS authors. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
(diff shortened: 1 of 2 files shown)


yarneo:
This looks good, with the caveat that I would of expected the title to not be left aligned to the bar buttons but rather right aligned in this case.

I have copied an example of Apple's Reminders app in hebrew as an example of the text being right aligned with a button on the left.

image1

I think it should be alright though if the layout constraint on that title is to fit all the empty space in the nav bar, because then the RTL language would align to the right. Maybe having the title in Hebrew would make it easier to see the behavior.

ianegordon:
Material and iOS centers these titles.
Often common case where you have a back button.

yarneo:
Gotcha. It does look like it is left aligned rather than center aligned, I opened a bug (https://github.com/material-components/material-components-ios/issues/3202) as this is out of scope for this PR. Thanks

yarneo (approved):

PR #3195, opened 2018-03-30
[Tabs] Fix format string argument types

romoore:
Xcode 9.3 is now complaining that our format strings are accepting the wrong
type for NSIntegers. Explicitly casting to long.


ianegordon (approved):
LGTM, thanks @romoore !

PR #3225, opened 2018-04-04
Revert "[FlexibleHeader] Fix bug where shadow layer's opacity wouldn't be set without a tracking scroll view. (#3201)"

romoore:
This change is resulting in "stacked" shadows when an MDCAppBar is above an
MDCTabBar.  In the particular case of Tabs below an MDCAppBar, the shadow effect is discouraged in designs. We may need to provide client teams with advice on how to integrate MDCAppBar and MDCTabBar in a way that still permits the App Bar to cast a shadow on other (non-scrollable) views in an appropriate way.

This reverts commit d0af3a729b987b8614b2eadc577ce658cd3d0dcc.

Reverts #3201


romoore:
Confirmed with the designer for AppBar that there should be NO shadow (nor even a separation) when Tabs are shown directly below the navigation elements.

Mock of correct behavior:
mdc-appbar-tabs-screenshot

jverkoey (approved):

PR #3229, opened 2018-04-04
Revert "Revert "[FlexibleHeader] Fix bug where shadow layer's opacity wouldn't be set without a tracking scroll view. (#3201)""

romoore:
Reverts material-components/material-components-ios#3225


romoore:
This was supposed to go against release-candidate, not develop. It's not branched from develop.

PR #3247, opened 2018-04-05
[ButtonBar] Allow customizing the inkColor of the buttons

### LINKED ISSUE #1660: AppBar's button ink is the wrong color in MDCCatalog
MDCCatalog has mostly dark colors with a lighter colored ink. The MDCButtonBarButton in the MDCAppBar seems to still have a dark ink applied, so it is barely visible.

Example:
1. Open MDCatalog > App Bar > Demo
1. Tap and hold on the button "Right"
1. Notice that the ink is barely visible as a darker color behind the text.

AlimoRabbani:
Make inkColor for buttons in MDCButtonBar customizable. Also Expose an API on MDCNavigationBar to allow using the custom inkColor in AppBar.

closes #1660


AlimoRabbani:
Will create a new one with clean commit history.

PR #3269, opened 2018-04-06
[FeatureHighlight] Simplifying UpdateTitleFont and UpdateBodyFont.

mohammadcazig:
Cleaning up dynamic type support for feature highlight by defining the default.


jverkoey (approved):
To confirm: will this be a breaking change for clients?

mohammadcazig:
Not a breaking change since there is no change in functionality.

PR #3299, opened 2018-04-09
Forcing Font size now does save the right FontName for every case.

mohammadcazig:
To enforce 20 point size we are using fontWithName:size:.
For some reason even though the printout looks identical comparing the fonts returns false. (Using fontWithSize: did not work for custom fonts - Medium, instead it returned a regular font).

### DIFF
--- a/components/NavigationBar/src/MDCNavigationBar.m
+++ b/components/NavigationBar/src/MDCNavigationBar.m
@@ -264,6 +264,7 @@ static NSString *const MDCNavigationBarTitleAlignmentKey = @"MDCNavigationBarTit
 }
 
 - (void)setTitleFont:(UIFont *)titleFont {
+  // Using "fontWithSize:" did not work for system font medium, instead it returned a regular font.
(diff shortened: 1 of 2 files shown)


components/NavigationBar/src/MDCNavigationBar.m
```diff
@@ -264,7 +264,7 @@ - (void)encodeWithCoder:(NSCoder *)aCoder {
 }
 
 - (void)setTitleFont:(UIFont *)titleFont {
-  _titleFont = [titleFont fontWithSize:kTitleFontSize];
+  _titleFont = [UIFont fontWithName:titleFont.fontName size:kTitleFontSize];
```
ianegordon: Please add  a comment along the lines of:
// -(UIFont)fontWithSize: does not maintain the weight of the original font so we are 
// using +(UIFont)fontWithName:size:

ianegordon (approved):
Please address comment before submit.

components/NavigationBar/src/MDCNavigationBar.m
mohammadcazig: done.

PR #3344, opened 2018-04-12
[Slider] Add tick color API

romoore:
The Slider needs a way to support custom colors for the discrete value "ticks"
within both the filled and unfilled portions of the track.  They should also
be customizable across states (to support disabled colors).

**Before**
slider-tick-ex-before

**After**
slider-tick-ex-after

Partially implements #3137
14f043b


randallli (approved):

jverkoey (approved):
Solid PR.

PR #3357, opened 2018-04-13
[TextFields] Add backgroundColor property

ianegordon:


components/TextFields/src/MDCTextInputControllerBase.h
```diff
@@ -59,6 +59,13 @@ extern const CGFloat MDCTextInputControllerBaseDefaultBorderRadius;
 */
 @interface MDCTextInputControllerBase : NSObject <MDCTextInputControllerFloatingPlaceholder>
 
+/**
+ The color behind the input and label that defines the preferred tap zone.
+
+ Default is borderFillColorDefault.
+ */
+@property(nonatomic, nullable, strong) UIColor *backgroundColor;
```
willlarche: I think this is null_resettable too since it pulls the default if it's nil and the default is null_resettable.

willlarche (dismissed):

willlarche (approved):
Put in a ticket for a unit test when you can.

romoore:
iangordon@ Why was this added? I don't see it used ~anywhere~ in the other subclasses (outlined, filled). Should all of the subclasses be using this property or should the property be moved to `FullWidth`?

PR #3405, opened 2018-04-19
Updating Readme documentation for chips.

mohammadcazig:
Adding step by step instruction how to create all chips types offered in Material Design.

### DIFF
--- a/components/Chips/README.md
+++ b/components/Chips/README.md
@@ -44,24 +44,191 @@ pod install
 
 - - -
 
-## Usage
-
-### Importing
+## Importing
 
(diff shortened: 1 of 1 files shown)


jverkoey (approved):
The structure and code content looks good as a first draft. We'll polish this document up next week as we focus on documentation.

In the end did we not need to enable delays touches for chips tap behaviors?

mohammadcazig:
No we did not need to disable touch delays, it's only for the chip to be responsive faster. other wise collection view first checks to see if touch is scroll. We can add that a note.

PR #3427, opened 2018-04-20
[SnackBar] Extending the timeout for a flaky test.

randallli:
copying the swift test: https://github.com/material-components/material-components-ios/pull/2750


romoore (approved):

randallli:
I landed the other pr which turns the test off.

bhamiltoncx:
Thanks. Looking at the code, this should not take 3 seconds or even 1 second. I think there might be a subtle bug in the implementation.

PR #3444, opened 2018-04-23
[Chips] Reset cell selection state to NO on prepare for reuse.

mohammadcazig:
When reload data we want to make sure we always reset the state. (Other wise animating layout could break the chip selection state).
Step to reproduce the issue in pivotal.
Pivotal issue: https://www.pivotaltracker.com/story/show/157009285


romoore (approved):
Please add the dragon we worked with (as a backlog issue if necessary).

mohammadcazig:
I am gonna update catalog with those examples after this is submitted :)

PR #3450, opened 2018-04-23
Make MDCAppBarContainerViewController set bounds of child view controller

jmdetloff:
The contained child view should fill its container's bounds. @jverkoey as owner, does this approach seem reasonable? This fix is triggered specifically by wanting to use a UITableViewController within MDCAppBarContainerViewController - UITableViewController loads its view with a frame inset equal to the window's top layout guide length. This inset would be removed when pushed to a navigation controller, but isnt currently removed when added to a MDCAppBarContainerViewController first.

### DIFF
--- a/components/AppBar/src/MDCAppBarContainerViewController.m
+++ b/components/AppBar/src/MDCAppBarContainerViewController.m
@@ -50,6 +50,7 @@
 - (void)viewWillLayoutSubviews {
   [super viewWillLayoutSubviews];
   [_appBar.headerViewController updateTopLayoutGuide];
+  self.contentViewController.view.frame = self.view.bounds;
 }
 
 - (BOOL)prefersStatusBarHidden {


jverkoey:
Can you add a screenshot demonstrating the bug that this PR is fixing?

jverkoey (approved):

components/AppBar/src/MDCAppBarContainerViewController.m
```diff
@@ -32,6 +32,10 @@ - (instancetype)initWithContentViewController:(UIViewController *)contentViewCon
 
     _contentViewController = contentViewController;
     [self addChildViewController:contentViewController];
+
+    contentViewController.view.frame = self.view.bounds;
+    contentViewController.view.autoresizingMask =
+        UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
```
jverkoey: As a general rule of thumb it's dangerous to touch a view controller's `.view` inside of a view controller init method.  This logic should be moved to `viewDidLoad` so that we minimize the window of time that a view is loaded for.

jverkoey (dismissed):

jverkoey (approved):

components/AppBar/src/MDCAppBarContainerViewController.m
```diff
@@ -42,6 +42,10 @@ - (void)viewDidLoad {
   [self.view addSubview:self.contentViewController.view];
   [self.contentViewController didMoveToParentViewController:self];
 
+  self.contentViewController.view.frame = self.view.bounds;
+  self.contentViewController.view.autoresizingMask =
+      UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
```
ianegordon: I'm not sure we can override autoresizingMask, that is view-specific.
Otherwise, LGTM.

ianegordon (changes requested):
If it works as expected without modifying the autoresizingMask, LGTM.

components/AppBar/src/MDCAppBarContainerViewController.m
jverkoey: In this case we own the content view so it's reasonable to modify the autoresizingMask accordingly.

ianegordon: It's not our containerView, it's our contentView.  It's given to us by a user.
To put it another way.
I create MyCustomViewController and set it's 
view.autoresizingMask = flexibleHeight | flexibleLeftMargin || flexibleRightMargin
I embed myCustomViewController in a container controller, (UINavigationController, UITabController, MDCAppBarContainerViewController) as a childViewController.
I would *not* expect view.autoresizingMask to be changed.
Make sense?

ianegordon (dismissed):

jmdetloff:
Screenshot of the bug this fixes:

simulator screen shot - iphone x - 2018-04-26 at 14 43 39

The contents of the example are pushed down 20px because the child view controller's view's origin 
is (0, 20) instead of (0, 0)

PR #3498, opened 2018-04-25
[Activity Indicator] Update the minimum radius documented in the header

siyusong:
The actual implementation is changed in https://github.com/material-components/material-components-ios/pull/887, but it seems that the documentation hasn't been updated.


jverkoey (approved):
Thank you!

PR #3590, opened 2018-04-30
[Palettes] Improve example text contrast

romoore:
The text in the Palette examples did not provide sufficient contrast against
the backgrounds. To improve accessibility, the contrast ratio should be at
least 4.5:1.

**Before**
mdc_palettes_before

**After**
mdc_palettes_after


components/Palettes/examples/supplemental/PalettesExampleViewController.swift
```diff
@@ -16,6 +16,16 @@ limitations under the License.
 
 import MaterialComponents
 
+func TextColorFor(backgroundColor: UIColor) -> UIColor {
```
mohammadcazig: document that the default is black.

mohammadcazig (approved):
LGTM, one small comment.

components/Palettes/examples/supplemental/PalettesExampleViewController.swift
romoore: Added docs.

PR #3606, opened 2018-05-01
[TextField] floating placeholder overlaps with border.

mohammadcazig:
Pivotal: https://www.pivotaltracker.com/story/show/157163089
New Color scheme makes the placeholder background transparent so when it floats it overlaps with the border.
Instead of relying on background color we now set the proper border path to achieve the same behavior.
No animation for path change.

screen shot 2018-05-01 at 11 19 50 am

### DIFF
--- a/components/TextFields/src/private/MDCTextInputControllerBase+Subclassing.h
+++ b/components/TextFields/src/private/MDCTextInputControllerBase+Subclassing.h
@@ -31,4 +31,7 @@
 /** Refreshes the layout and style of the placeholder label. Called within updateLayout. */
 - (void)updatePlaceholder;
 
+/** Refreshes the layout and style of the border view. Called within updateLayout. */
+- (BOOL)isPlaceholderUp;
+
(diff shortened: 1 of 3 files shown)


components/TextFields/src/private/MDCTextInputControllerBase+Subclassing.h
```diff
@@ -31,4 +31,7 @@
 /** Refreshes the layout and style of the placeholder label. Called within updateLayout. */
 - (void)updatePlaceholder;
 
+/** Refreshes the layout and style of the border view. Called within updateLayout. */
+- (BOOL)isPlaceholderUp;
```
ianegordon, line 35: This should likely be a readonly property, but we can leave that for another day.

ianegordon (approved):

willlarche:
Since there's no animation, please post a video showing what it DOES look like and the ticket to correct / finish the motion.

PR #3617, opened 2018-05-01
[TextFields] fixing color mapping for floatingNormal and floatingActive Colors.

mohammadcazig:
Both Floating active and normal color should be set in color scheme.

Before:
screen shot 2018-05-02 at 11 16 06 am
screen shot 2018-05-02 at 11 16 23 am

After:
screen shot 2018-05-02 at 11 14 31 am
screen shot 2018-05-02 at 11 14 46 am


randallli (approved):
I would expect unit tests covering this

ianegordon:
Please add a before / after screenshot.

PR #3737, opened 2018-05-04
[Catalog] Shaped Buttons should be all themed using global themes.

mohammadcazig:
Pivotal: https://www.pivotaltracker.com/story/show/157298987
Before:
simulator screen shot - iphone 8 - 2018-05-04 at 15 33 50

After:
screen shot 2018-05-04 at 2 26 07 pm


romoore:
before and after pics, plz!

romoore (approved):

PR #4153, opened 2018-05-10
[Buttons] Fix updateTitles to prevent crash

### LINKED ISSUE #4152: [Buttons] Crash when changing `uppercaseTitle` if more than one title-for-state is set
### Overview
The `uppercaseTitles` property will cause a crash on iOS 10 or below if there is more than one value set in `setTitle:forState:` because it is iterating and mutating the same dictionary.

This does not occur in iOS 11+ nor if there is only a single title-for-state set. There is probably some internal optimization with the enumerator where it doesn't check for modification until the second iteration.

### Reproduction steps

```objectivec
- (void)testUppercaseTitleNoChangedToYes {
  // Given
  NSString *originalTitle = @"some Text";

  // When
  button.uppercaseTitle = NO;
  [button setTitle:originalTitle forState:UIControlStateNormal];
  [button setTitle:originalTitle forState:UIControlStateHighlighted]; // More than one, and it crashes
  [button setTitle:originalTitle forState:UIControlStateDisabled];
  button.uppercaseTitle = YES;

  // Then
  XCTAssertEqualObjects(button.currentTitle,
                        [originalTitle uppercaseStringWithLocale:[NSLocale currentLocale]]);
}
```
### Version number
55.0.1

iPad Mini/iOS 10.3

romoore:
When a button had more than one title-for-state value, changing the property
`-uppercaseTitle` would cause a crash on iOS 10 and below. This was because
the code was both enumerating and mutating the same `_nontransformedTitles`
dictionary.

Closes #4152

### DIFF
--- a/components/Buttons/src/MDCButton.m
+++ b/components/Buttons/src/MDCButton.m
@@ -527,9 +527,10 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
 - (void)updateTitleCase {
   // This calls setTitle or setAttributedTitle for every title value we have stored. In each
   // respective setter the title is upcased if _uppercaseTitle is YES.
-  for (NSNumber *key in _nontransformedTitles.keyEnumerator) {
+  NSDictionary<NSNumber *, NSString *> *nontransformedTitles = [_nontransformedTitles copy];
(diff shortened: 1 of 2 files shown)


jverkoey (approved):
Yay tests!

Issue #2830, opened 2018-01-11
[Snackbar] Snackbar should support leading (left) alignment on tablet


ajsecord:
The Material spec [supports an optional leading (left) alignment for snackbars for tablet](https://material.io/design/components/snackbars.html#placement) (mobile is always full-width).

The change should be fairly trivial to make to [MDCSnackbarManager](https://github.com/material-components/material-components-ios/blob/develop/components/Snackbar/src/MDCSnackbarManager.h).

![components-toasts-tabletdesktop1-left](https://user-images.githubusercontent.com/742162/34850315-532100d0-f6f3-11e7-97e6-bd85cb8ce75a.png)

---
https://www.pivotaltracker.com/story/show/156805088

PR #4163, opened 2018-05-14
[Snackbar] Allow Snackbars to have leading alignment on iPads

andrewoverton:
Snackbars are allowed to have leading alignment on iPads, but until now no implementations have been suggested. Here's an idea for how it could be done using an alignment enum.

If merged this will close issue #2830 and resolve pivotal ticket [156805088](https://www.pivotaltracker.com/n/projects/2150420/stories/156805088).

Here's a screenshot:
simulator screen shot - ipad air 2 - 2018-05-14 at 13 26 01

### DIFF
--- a/components/Snackbar/src/MDCSnackbarAlignment.h
+++ b/components/Snackbar/src/MDCSnackbarAlignment.h
@@ -1,0 +1,28 @@
+/*
+ Copyright 2018-present the Material Components for iOS authors. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+/**
+ States used to configure snackbar alignment.
+ */
+typedef NS_ENUM(NSInteger, MDCSnackbarAlignment) {
+
+  // Snackbar is positioned in the center of the screen. This is the default setting
+  // and only permitted setting for iPhone.
+  MDCSnackbarAlignmentCenter = 0,
+
+  // Snackbar is positioned near the leading margin.
+  MDCSnackbarAlignmentLeading = 1,
+};
--- a/components/Snackbar/src/MDCSnackbarAlignment.h
+++ b/components/Snackbar/src/MDCSnackbarAlignment.h
@@ -14,6 +14,8 @@
  limitations under the License.
  */
 
+#import <Foundation/Foundation.h>
+
 /**
  States used to configure snackbar alignment.
  */
--- a/components/Snackbar/src/MDCSnackbarAlignment.h
+++ b/components/Snackbar/src/MDCSnackbarAlignment.h
@@ -17,12 +17,11 @@
 #import <Foundation/Foundation.h>
 
 /**
- States used to configure snackbar alignment.
+ States used to configure Snackbar alignment.
  */
 typedef NS_ENUM(NSInteger, MDCSnackbarAlignment) {
   /**
-   Snackbar is positioned in the center of the screen. This is the default setting
-   and only permitted setting for iPhone.
+   Snackbar is positioned in the center of the screen.
    */
   MDCSnackbarAlignmentCenter = 0,
 
--- a/components/Snackbar/src/MDCSnackbarAlignment.h
+++ b/components/Snackbar/src/MDCSnackbarAlignment.h
@@ -20,11 +20,14 @@
  States used to configure snackbar alignment.
  */
 typedef NS_ENUM(NSInteger, MDCSnackbarAlignment) {
-
-  // Snackbar is positioned in the center of the screen. This is the default setting
-  // and only permitted setting for iPhone.
+  /**
+   Snackbar is positioned in the center of the screen. This is the default setting
+   and only permitted setting for iPhone.
+   */
   MDCSnackbarAlignmentCenter = 0,
 
-  // Snackbar is positioned near the leading margin.
+  /**
+   Snackbar is positioned near the leading margin.
+   */
   MDCSnackbarAlignmentLeading = 1,
 };
--- a/components/Snackbar/src/MDCSnackbarManager.h
+++ b/components/Snackbar/src/MDCSnackbarManager.h
@@ -44,7 +44,7 @@
 
  @note The setter must be called from the main thread.
  */
-@property (class, nonatomic) MDCSnackbarAlignment alignment;
+@property (class, nonatomic, assign) MDCSnackbarAlignment alignment;
 
 /**
  Shows @c message to the user, in a style consistent with the data contained in @c message.
--- a/components/Snackbar/src/MDCSnackbarManager.h
+++ b/components/Snackbar/src/MDCSnackbarManager.h
@@ -79,13 +79,11 @@
 + (void)setBottomOffset:(CGFloat)offset;
 
 /**
- Determines whether the snackbar is presented in the center of the of the screen or (on iPad only)
- from the leading edge of the screen.
+ Determines the snackbar alignment to the screen.
 
  If called within an animation block, the change will be animated.
 
  @note This method must be called from the main thread.
- @warning On non-iPad devices this setter will be a no-op.
  */
 + (void)setAlignment:(MDCSnackbarAlignment)alignment;
 
--- a/components/Snackbar/src/MDCSnackbarManager.h
+++ b/components/Snackbar/src/MDCSnackbarManager.h
@@ -15,6 +15,7 @@
  */
 
 #import <UIKit/UIKit.h>
+#import "MDCSnackbarAlignment.h"
 
 @class MDCSnackbarMessage;
 @protocol MDCSnackbarSuspensionToken;
@@ -76,6 +77,17 @@
  move and should not be obscured.
  */
 + (void)setBottomOffset:(CGFloat)offset;
+
+/**
+ Determines whether the snackbar is presented in the center of the of the screen or (on iPad only)
+ from the leading edge of the screen.
+
+ If called within an animation block, the change will be animated.
+
+ @note This method must be called from the main thread.
+ @warning On non-iPad devices this setter will be a no-op.
+ */
++ (void)setAlignment:(MDCSnackbarAlignment)alignment;
 
 #pragma mark Suspending
 
--- a/components/Snackbar/src/MDCSnackbarManager.h
+++ b/components/Snackbar/src/MDCSnackbarManager.h
@@ -34,6 +34,19 @@
 @interface MDCSnackbarManager : NSObject
 
 /**
+ Determines the Snackbar alignment to the screen.
+
+ If called within an animation block, the change will be animated.
+
+ @note This setting is only used when both the horizontal and vertical size classes of the presenting
+ window are @c UIUserInterfaceSizeClassRegular. Otherwise @c MDCSnackbarAlignmentCenter
+ will be used.
+
+ @note The setter must be called from the main thread.
+ */
+@property (class, nonatomic) MDCSnackbarAlignment alignment;
+
+/**
  Shows @c message to the user, in a style consistent with the data contained in @c message.
 
  For messages with the same category, the firing of completion blocks has a guaranteed FIFO
@@ -45,7 +58,7 @@
 /**
  MDCSnackbarManager will display the messages in this view.
 
- Call this method to choose where in the view hierarchy snackbar messages will be presented. It is
+ Call this method to choose where in the view hierarchy Snackbar messages will be presented. It is
  only necessary to provide a host view if the default behavior is unable to find one on it's own,
  most commonly when using MDCSnackbarManager inside an application extension. By default, if you use
  MDCSnackbarManager without calling @c setPresentationHostView, the manager will attempt to find a
@@ -77,15 +90,6 @@
  move and should not be obscured.
  */
 + (void)setBottomOffset:(CGFloat)offset;
-
-/**
- Determines the snackbar alignment to the screen.
-
- If called within an animation block, the change will be animated.
-
- @note This method must be called from the main thread.
- */
-+ (void)setAlignment:(MDCSnackbarAlignment)alignment;
 
 #pragma mark Suspending
 
@@ -126,34 +130,34 @@
 #pragma mark Styling
 
 /**
- The color for the background of the snackbar message view.
+ The color for the background of the Snackbar message view.
  */
 @property(class, nonatomic, strong, nullable) UIColor *snackbarMessageViewBackgroundColor;
 
 /**
- The color for the shadow color for the snackbar message view.
+ The color for the shadow color for the Snackbar message view.
  */
 @property(class, nonatomic, strong, nullable) UIColor *snackbarMessageViewShadowColor;
 
 /**
- The color for the message text in the snackbar message view.
+ The color for the message text in the Snackbar message view.
  */
 @property(class, nonatomic, strong, nullable) UIColor *messageTextColor;
 
 /**
- The font for the message text in the snackbar message view.
+ The font for the message text in the Snackbar message view.
  */
 @property(class, nonatomic, strong, nullable) UIFont *messageFont;
 
 /**
- The font for the button text in the snackbar message view.
+ The font for the button text in the Snackbar message view.
  */
 @property(class, nonatomic, strong, nullable) UIFont *buttonFont;
 
 
 /**
  If enabled, modifications of class styling properties will be applied immediately
- to the currently presented snackbar.
+ to the currently presented Snackbar.
 
  Default is set to NO.
  */
@@ -176,7 +180,7 @@
 + (void)setButtonTitleColor:(nullable UIColor *)titleColor forState:(UIControlState)state;
 
 /**
- Indicates whether the snackbar should automatically update its font when the device’s
+ Indicates whether the Snackbar should automatically update its font when the device’s
  UIContentSizeCategory is changed.
 
  This property is modeled after the adjustsFontForContentSizeCategory property in the
@@ -193,7 +197,7 @@
 @end
 
 /**
- A suspension token is returned when messages are suspended by the snackbar manager.
+ A suspension token is returned when messages are suspended by the Snackbar manager.
 
  Messages are suppressed while a suspension token is kept in memory. Messages will resume being
  displayed when the suspension token is released or when the suspension token is passed to
--- a/components/Snackbar/src/MDCSnackbarManager.m
+++ b/components/Snackbar/src/MDCSnackbarManager.m
@@ -534,6 +534,13 @@ static BOOL _shouldApplyStyleChangesToVisibleSnackbars;
   manager.overlayView.bottomOffset = offset;
 }
 
++ (void)setAlignment:(MDCSnackbarAlignment)alignment {
+  NSAssert([NSThread isMainThread], @"setAlignment must be called on main thread.");
+
+  MDCSnackbarManagerInternal *manager = [MDCSnackbarManagerInternal sharedInstance];
+  manager.overlayView.alignment = alignment;
+}
+
 #pragma mark - Suspension
 
 + (id<MDCSnackbarSuspensionToken>)suspendMessagesWithCategory:(NSString *)category {
--- a/components/Snackbar/src/MDCSnackbarManager.m
+++ b/components/Snackbar/src/MDCSnackbarManager.m
@@ -42,7 +42,7 @@ static BOOL UIViewHasFocusedAccessibilityElement(UIView *view) {
 static NSString *const kAllMessagesCategory = @"$$___ALL_MESSAGES___$$";
 
 /**
- The 'actual' snackbar manager which will take care of showing/hiding snackbar messages.
+ The 'actual' Snackbar manager which will take care of showing/hiding Snackbar messages.
  */
 @interface MDCSnackbarManagerInternal : NSObject
 
@@ -60,7 +60,7 @@ static NSString *const kAllMessagesCategory = @"$$___ALL_MESSAGES___$$";
 @property(nonatomic) NSMutableDictionary<NSString *, NSMutableSet<NSUUID *> *> *suspensionTokens;
 
 /**
- The view which will host our snackbar messages.
+ The view which will host our Snackbar messages.
  */
 @property(nonatomic) MDCSnackbarOverlayView *overlayView;
 
@@ -70,7 +70,7 @@ static NSString *const kAllMessagesCategory = @"$$___ALL_MESSAGES___$$";
 @property(nonatomic) UIView *presentationHostView;
 
 /**
- The currently-showing snackbar.
+ The currently-showing Snackbar.
  */
 @property(nonatomic) MDCSnackbarMessageView *currentSnackbar;
 
@@ -130,7 +130,7 @@ static BOOL _shouldApplyStyleChangesToVisibleSnackbars;
 #pragma mark - Message Displaying
 
 /**
- Determines whether or not a message is eligible to be shown based on the snackbar manager's current
+ Determines whether or not a message is eligible to be shown based on the Snackbar manager's current
  configuration.
 
  @note This method should ensure that messages in the same category are not shown out of order.
@@ -208,9 +208,9 @@ static BOOL _shouldApplyStyleChangesToVisibleSnackbars;
   __block BOOL shouldDismiss = YES;
   MDCSnackbarMessageDismissHandler dismissHandler =
       ^(BOOL userInitiated, MDCSnackbarMessageAction *action) {
-        // Because we start a timer to dismiss the snackbar once it is on screen, there exists the
-        // potential to try and dismiss the snackbar view multiple times, say if the user taps on
-        // the snackbar (dismissal one) and then the timer fires (dismissal two). This check ensures
+        // Because we start a timer to dismiss the Snackbar once it is on screen, there exists the
+        // potential to try and dismiss the Snackbar view multiple times, say if the user taps on
+        // the Snackbar (dismissal one) and then the timer fires (dismissal two). This check ensures
         // that the dismissal logic will only fire one time for a given Snackbar view.
         if (shouldDismiss) {
           shouldDismiss = NO;
@@ -252,7 +252,7 @@ static BOOL _shouldApplyStyleChangesToVisibleSnackbars;
                   BOOL hasVoiceOverFocus = UIAccessibilityIsVoiceOverRunning() &&
                                            UIViewHasFocusedAccessibilityElement(strongSnackbarView);
                   if (strongSnackbarView && !hasVoiceOverFocus) {
-                    // Mimic the user tapping on the snackbar.
+                    // Mimic the user tapping on the Snackbar.
                     [strongSnackbarView dismissWithAction:nil userInitiated:NO];
                   }
                 });
@@ -266,7 +266,7 @@ static BOOL _shouldApplyStyleChangesToVisibleSnackbars;
   // Ensure that this method is called on the main thread.
   NSAssert([NSThread isMainThread], @"Method is not called on main thread.");
 
-  // Mark the snackbar as being in the process of dismissal.
+  // Mark the Snackbar as being in the process of dismissal.
   snackbarView.dismissing = YES;
 
   MDCSnackbarMessage *message = snackbarView.message;
@@ -284,7 +284,7 @@ static BOOL _shouldApplyStyleChangesToVisibleSnackbars;
                                        [self deactivateOverlay:self.overlayView];
 
                                        // If VoiceOver had been enabled and the snackbarView was
-                                       // transient, the snackbar was just announced (layout was not
+                                       // transient, the Snackbar was just announced (layout was not
                                        // reported as changed) so there is no need to post a layout
                                        // change here.
                                        if (![self isSnackbarTransient:snackbarView]) {
@@ -294,7 +294,7 @@ static BOOL _shouldApplyStyleChangesToVisibleSnackbars;
 
                                        self.currentSnackbar = nil;
 
-                                       // Now that the snackbar view is offscreen, we can allow more
+                                       // Now that the snackbarView is offscreen, we can allow more
                                        // messages to be shown.
                                        self.showingMessage = NO;
                                        [self showNextMessageIfNecessaryMainThread];
@@ -399,7 +399,7 @@ static BOOL _shouldApplyStyleChangesToVisibleSnackbars;
   // Ensure that this method is called on the main thread.
   NSAssert([NSThread isMainThread], @"Method is not called on main thread.");
 
-  // Make sure that if there is a snackbar on screen, it does not belong to the current category.
+  // Make sure that if there is a Snackbar on screen, it does not belong to the current category.
   if (self.currentSnackbar != nil && !self.currentSnackbar.dismissing) {
     MDCSnackbarMessage *currentMessage = self.currentSnackbar.message;
 
@@ -412,7 +412,7 @@ static BOOL _shouldApplyStyleChangesToVisibleSnackbars;
     }
   }
 
-  // Now that we've ensured that the currently showing snackbar has been taken care of, we can go
+  // Now that we've ensured that the currently showing Snackbar has been taken care of, we can go
   // through pending messages and fire off their completion blocks as we remove them from the
   // queue.
   NSMutableIndexSet *indexesToRemove = [NSMutableIndexSet indexSet];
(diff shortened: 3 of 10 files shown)


yarneo:
Congrats on your first PR!! 🎉🎉🎉🎉

components/Snackbar/src/MDCSnackbarAlignment.h
```diff
@@ -0,0 +1,30 @@
...
+/**
+ States used to configure snackbar alignment.
+ */
+typedef NS_ENUM(NSInteger, MDCSnackbarAlignment) {
+
+  // Snackbar is positioned in the center of the screen. This is the default setting
+  // and only permitted setting for iPhone.
+  MDCSnackbarAlignmentCenter = 0,
+
+  // Snackbar is positioned near the leading margin.
```
jverkoey: Please the comments change here and above to `/** */`.

Context: public APIs require use of `/** */` tags in order to have the docs picked up by our API doc generator.

components/Snackbar/src/MDCSnackbarAlignment.h
```diff
@@ -0,0 +1,30 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+
+/**
+ States used to configure snackbar alignment.
+ */
+typedef NS_ENUM(NSInteger, MDCSnackbarAlignment) {
```
jverkoey: I can't tell at first glance whether this is alignment within the snackbar (e.g. UITextAlignment aligns the text within the label's bounds) or alignment to the screen. `MDCSnackbarScreenAlignment` might help clarify the intended use of this API. What do you think?

components/Snackbar/src/MDCSnackbarManager.h
```diff
@@ -77,6 +78,17 @@
  */
 + (void)setBottomOffset:(CGFloat)offset;
 
+/**
+ Determines whether the snackbar is presented in the center of the of the screen or (on iPad only)
+ from the leading edge of the screen.
```
jverkoey: I fear this documentation will get out of date fairly quickly — can we future-proof it by removing references to the specific values it supports?

E.g. "Determines the snackbar alignment to the screen."

components/Snackbar/src/MDCSnackbarAlignment.h
```diff
@@ -0,0 +1,30 @@
...
+
+  // Snackbar is positioned in the center of the screen. This is the default setting
+  // and only permitted setting for iPhone.
```
jverkoey: I can imagine us wanting to support alternative alignments on large landscape phones in the future, so it may be wise for us to avoid restricting when/how to use the alignments. What do you think?

components/Snackbar/src/private/MDCSnackbarOverlayView.h
```diff
@@ -62,4 +63,15 @@ OBJC_EXTERN NSTimeInterval const MDCSnackbarLegacyTransitionDuration;
...
 @property(nonatomic) CGFloat bottomOffset;
 
+/**
+ Determines whether the snackbar is presented in the center of the of the screen or (on iPad only)
+ the leading edge of the screen.
+
+ If called within an animation block, the change will be animated.
+
+ @note This method must be called from the main thread.
+ @warning On non-iPad devices this setter will be a no-op.
```
jverkoey: Drop this comment - we generally always want properties to store the provided state even if they're not making use of it.

components/Snackbar/src/private/MDCSnackbarOverlayView.m
```diff
@@ -246,7 +246,7 @@ - (void)setSnackbarView:(MDCSnackbarMessageView *)snackbarView {
 
     CGFloat bottomMargin = [self staticBottomMargin];
     CGFloat sideMargin = [self sideMargin];
-    BOOL fullWidth = UI_USER_INTERFACE_IDIOM() != UIUserInterfaceIdiomPad;
+    BOOL isIphone = UI_USER_INTERFACE_IDIOM() == UIUserInterfaceIdiomPhone;
```
jverkoey: Should we be looking at the size traits of the device? E.g. a large iPhone in landscape is almost comparable to an iPad in portrait.

components/Snackbar/src/private/MDCSnackbarOverlayView.m
```diff
@@ -617,6 +627,33 @@ - (void)setBottomOffset:(CGFloat)bottomOffset {
   }
 }
 
+- (void)setAlignment:(MDCSnackbarAlignment)alignment {
+  if (UI_USER_INTERFACE_IDIOM() != UIUserInterfaceIdiomPad) {
+    return;
+  }
```
jverkoey: Drop this early exit - we always want to set the _alignment ivar so that the state isn't lost.

jverkoey (dismissed):
Love the constraint-based solution for layout. I have some small questions about the implementation but overall this is looking like a good direction.

components/Snackbar/src/private/MDCSnackbarOverlayView.m
romoore: Agreed, please consider using horizontal Regular traits instead of iPad. An iPad in 1/3 split-screen multitasking is basically a tall iPhone.

romoore: Or even Regular/Regular if you want to prohibit it in a vertically-compact environment (like a landscape iPhone+).

https://useyourloaf.com/blog/size-classes/

andrewoverton: I agree that size classes are the way to go--I'll look into it!

components/Snackbar/src/MDCSnackbarManager.h
andrewoverton: Oooh, good call 👍

components/Snackbar/src/MDCSnackbarAlignment.h
andrewoverton: I was using `NSTextAlignment` as a guide. `NSTextAlignment` aligns _text_ in a container, like a `UILabel`, for example. Similarly, I figured `MDCSnackbarAlignment` could align a snackbar in a container. I'm open to `MDCSnackbarScreenAlignment` though, definitely.

components/Snackbar/src/MDCSnackbarAlignment.h
andrewoverton: Are you suggesting our implementation allow behavior that goes against the spec? I'm fine with that--I honestly wasn't sure how much I should enforce adherence.. Would you recommend I go so far as to include a _trailing_ alignment case, even though it's a definite no-no, from what I understand? I can see why it might be a wise move for future-proofing.

components/Snackbar/src/MDCSnackbarAlignment.h
jverkoey: Ah, interesting. I just realized that when I look at MDC APIs I tend to subconsciously drop both the MDC and the component name, so I read this API as simply “alignment” and wasn’t sure what was being aligned.

NSTextAlignment on the other hand I read as “text alignment”.

This may just be me though, so I don’t feel super strongly about adding “screen” to the API.

components/Snackbar/src/private/MDCSnackbarOverlayView.m
andrewoverton: I think this makes sense if we actually are okay with people doing things like creating leading aligned snackbars in horizontally compact environments. It sounds like we are 👍

components/Snackbar/src/MDCSnackbarAlignment.h
jverkoey: Generally-speaking we’ve been moving toward increased configurability and decreased opinions baked in to the components themselves.
This is motivated by the fact that configuration is almost always a need, while opinions tend to change from one design system to another. This trend is best exemplified in the themer APIs, which apply a set of opinions on configurable components.

As for adding additional behavior - I think it’s best to keep the scope of this change as-is for now. If someone wants a trailing implementation in the future then we can/will add it then.

components/Snackbar/src/MDCSnackbarAlignment.h
andrewoverton: It wasn't brought up in any of the previous comments--how do we feel about introducing such a big API addition (a whole new header file!) for something that was originally labeled "good first issue"? It seemed a little bold to me, I just couldn't think of a better way haha 🙈

components/Snackbar/src/private/MDCSnackbarOverlayView.m
romoore: I think the guidelines are pretty clear that it's only for "wide layouts". I would not consider a Compact size class to be wide.  An iPhone 4 is probably far too narrow to consider anything other than a full-screen Snackbar.

components/Snackbar/src/MDCSnackbarAlignment.h
jverkoey: Thinking more about this though...I wonder if we do want to encourage thoughtful use of this API on a traits basis? Perhaps by somehow specifying which traits to use a given alignment on?

andrewoverton: Sounds good! Thanks for the insight :D

components/Snackbar/src/private/MDCSnackbarOverlayView.m
andrewoverton: So when someone tries to set a leading alignment in a horizontally compact environment we want to preserve the state (we want the getter to return `MDCSnackbarAlignmentLeading`) but we want the snackbar to _appear_ centered/full-screen. Okie dokie, thanks for the clarification  👍

components/Snackbar/src/MDCSnackbarAlignment.h
andrewoverton: Oh, I just noticed https://github.com/material-components/material-components-ios/pull/4163#discussion_r188414240. Hmmmm.

andrewoverton:
Here's a gif of the size class based solution added in 8ce8cfe. In this case the alignment is leading for both snackbars shown.
may-15-2018 17-33-57

yarneo (commented):
can you please provide screenshots and test out the new material snackbar. You can look at it by pressing the top right nav button in the example you are currently screenshotting. Also please try it out with dynamic type on to see it is working well.

yarneo:
Waiting for design feedback on if landscape on big iPhones is considered a "wide UI". Will follow up.

andrewoverton:
@yarneo thanks for asking an actual designer for clarification, I should've thought to do that 😛 

Here is a gif with dynamic type and the new snackbar:

may-16-2018 10-46-46

components/Snackbar/src/private/MDCSnackbarOverlayView.m
yarneo: We don't want to break appearance by changing the width from full to non-full on mobile devices for landscape. Currently the behavior and per spec is to have it full width, i think we should leave it at that rather than making breaking layout changes.

components/Snackbar/src/private/MDCSnackbarOverlayView.m
```diff
@@ -257,85 +257,79 @@ - (void)setSnackbarView:(MDCSnackbarMessageView *)snackbarView {
...
+                                                                       constant:0];
+        snackbarView.centerConstraint.active = self.alignment == MDCSnackbarAlignmentCenter;
+
+        snackbarView.leadingConstraint  = [NSLayoutConstraint constraintWithItem:snackbarView
+                                                                       attribute:NSLayoutAttributeLeading
+                                                                       relatedBy:NSLayoutRelationEqual
+                                                                          toItem:container
+                                                                       attribute:NSLayoutAttributeLeading
+                                                                      multiplier:1.0
+                                                                        constant:sideMargin];
```
romoore: I like the use of Regular/Regular here, and I think it aligns with the Material Guidelines. Looks like sideMargin is dependent on device idiom (iPad/iPhone). Should we migrate that code to size classes as well? I'm concerned that a split-screen iPad in a Regular(H)/Compact(W) environment will have the wrong margins (24 instead of 8).

components/Snackbar/src/private/MDCSnackbarOverlayView.m
```diff
@@ -257,85 +257,79 @@ - (void)setSnackbarView:(MDCSnackbarMessageView *)snackbarView {
...
+        snackbarView.leadingConstraint.active = self.alignment == MDCSnackbarAlignmentLeading;
+
+        // If not full width, ensure that it doesn't get any larger than our own width.
+        [container
+         addConstraint:[NSLayoutConstraint constraintWithItem:snackbarView
```
romoore: Nit/Indentation: Should be 4 spaces from the previous line.

components/Snackbar/src/private/MDCSnackbarOverlayView.m
```diff
@@ -257,85 +257,79 @@ - (void)setSnackbarView:(MDCSnackbarMessageView *)snackbarView {
...
+                                                    attribute:NSLayoutAttributeWidth
+                                                    relatedBy:NSLayoutRelationLessThanOrEqual
+                                                       toItem:container
+                                                    attribute:NSLayoutAttributeWidth
+                                                   multiplier:1.0
+                                                     constant:-2 * sideMargin]];
+
+        // Also ensure that it doesn't get any smaller than its own minimum width.
+        [container
+         addConstraint:[NSLayoutConstraint constraintWithItem:snackbarView
```
romoore: Nit/Indentation: Same (4 spaces)

components/Snackbar/src/private/MDCSnackbarOverlayView.m
```diff
@@ -257,85 +257,79 @@ - (void)setSnackbarView:(MDCSnackbarMessageView *)snackbarView {
...
+                                                    attribute:NSLayoutAttributeWidth
+                                                    relatedBy:NSLayoutRelationGreaterThanOrEqual
+                                                       toItem:nil
+                                                    attribute:NSLayoutAttributeNotAnAttribute
+                                                   multiplier:1.0
+                                                     constant:[snackbarView minimumWidth]]];
+
+        // Also ensure that it doesn't get any larger than its own maximum width.
+        [container
+         addConstraint:[NSLayoutConstraint constraintWithItem:snackbarView
```
romoore: Nit/Indentation: Same as above.

components/Snackbar/src/private/MDCSnackbarOverlayView.m
```diff
@@ -257,85 +257,79 @@ - (void)setSnackbarView:(MDCSnackbarMessageView *)snackbarView {
...
+                                                   multiplier:1.0
+                                                     constant:[snackbarView maximumWidth]]];
+      } else {
 #if defined(__IPHONE_11_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0)
-          if (@available(iOS 11.0, *)) {
-            leftMargin += self.safeAreaInsets.left;
-            rightMargin += self.safeAreaInsets.right;
-          }
+        if (@available(iOS 11.0, *)) {
+          leftMargin += self.safeAreaInsets.left;
```
romoore: Nit/Suggestion: Consider adding an mdc_safeAreaInsets method similar to [MDCBottomAppBarView](https://github.com/material-components/material-components-ios/blob/develop/components/BottomAppBar/src/MDCBottomAppBarView.m#L251) and [MDCBottomNavigationBar](https://github.com/material-components/material-components-ios/blob/develop/components/BottomNavigation/src/MDCBottomNavigationBar.m#L355).  It will simplify this code (slightly) and allow us to easily migrate in the future.

components/Snackbar/src/private/MDCSnackbarOverlayView.m
```diff
@@ -257,85 +257,79 @@ - (void)setSnackbarView:(MDCSnackbarMessageView *)snackbarView {
...
-        default:
-          NSAssert(NO, @"Unknown size class");
-          break;
+        [container addConstraint:[NSLayoutConstraint constraintWithItem:snackbarView
+                                                              attribute:NSLayoutAttributeLeading
+                                                              relatedBy:NSLayoutRelationEqual
+                                                                 toItem:container
+                                                              attribute:NSLayoutAttributeLeading
+                                                             multiplier:1.0
+                                                               constant:leftMargin]];
```
romoore, line 350: RTL: Please be sure to flip `leftMargin` and `rightMargin` in RTL layouts.  I believe `safeAreaInsets` should always be set correctly for left/right, but the "leading" constraint in an RTL layout should use the `rightMargin` value (since that edge is on the right), not the `leftMargin` value.

components/Snackbar/src/MDCSnackbarAlignment.h
romoore: I wouldn't worry about a new header too much. Since it's used by both SnackbarManager and SnackbarMessageView (and neither publicly references the other), I think it's good to put it in a separate header.

@jverkoey If we want to later change the alignment of the content within the Snackbar, I think we could use something like `MDCSnackbarContentAlignment` to name that value.

components/Snackbar/src/MDCSnackbarAlignment.h
```diff
@@ -0,0 +1,33 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+
+/**
+ States used to configure snackbar alignment.
```
romoore: Super-nit: s/snackbar/Snackbar/

components/Snackbar/src/MDCSnackbarManager.h
```diff
@@ -77,6 +78,15 @@
...
 + (void)setBottomOffset:(CGFloat)offset;
 
+/**
+ Determines the snackbar alignment to the screen.
+
+ If called within an animation block, the change will be animated.
+
+ @note This method must be called from the main thread.
+ */
++ (void)setAlignment:(MDCSnackbarAlignment)alignment;
```
romoore: Please consider making this a property so clients can read the value as well.

components/Snackbar/src/MDCSnackbarManager.h
```diff
@@ -77,6 +78,15 @@
  */
 + (void)setBottomOffset:(CGFloat)offset;
 
+/**
+ Determines the snackbar alignment to the screen.
+
+ If called within an animation block, the change will be animated.
+
+ @note This method must be called from the main thread.
```
romoore: Nit: You can drop the main thread `@note` since this is a UIView, clients are expected to do all UI-related work on the main thread.

components/Snackbar/src/MDCSnackbarAlignment.h
```diff
@@ -0,0 +1,33 @@
...
+ */
+typedef NS_ENUM(NSInteger, MDCSnackbarAlignment) {
+  /**
+   Snackbar is positioned in the center of the screen. This is the default setting
+   and only permitted setting for iPhone.
```
romoore: Nit: Consider putting the default/note about size classes in the docs for the property itself rather than on the enumeration.

```objectivec
/**
 Determines the Snackbar alignment to the screen. If called within an animation block, the change 
 will be animated.

 @note This setting is only used when both the horizontal and vertical size classes of the presenting
       window are @c UIUserInterfaceSizeClassRegular. Otherwise @c MDCSnackbarAlignmentCenter 
       will be used.
 */
```

components/Snackbar/src/MDCSnackbarManager.m
```diff
@@ -534,6 +534,13 @@ + (void)setBottomOffset:(CGFloat)offset {
   manager.overlayView.bottomOffset = offset;
 }
 
++ (void)setAlignment:(MDCSnackbarAlignment)alignment {
+  NSAssert([NSThread isMainThread], @"setAlignment must be called on main thread.");
```
romoore, line 538: Nit: I think this can be removed, clients are expected to perform UI work on the main thread.

components/Snackbar/src/MDCSnackbarMessageView.m
```diff
@@ -924,6 +924,22 @@ - (NSArray *)horizontalButtonLayoutConstraints {
...
+      self.leadingConstraint.active = NO;
+      self.centerConstraint.active = YES;
+      break;
+    case MDCSnackbarAlignmentLeading:
+      self.leadingConstraint.active = YES;
+      self.centerConstraint.active = NO;
+      break;
+    default:
+      NSAssert(NO, @"`alignment` should be a valid MDCSnackbarAlignment value");
+      break;
```
romoore: Please add default behavior here for non-debug releases. I would propose the `MDCSnackbarAlignmentCenter` behavior should be used.

components/Snackbar/src/MDCSnackbarMessageView.m
```diff
@@ -956,7 +972,7 @@ - (CGFloat)contentSafeBottomInset {
   // If a bottom offset has been set to raise the HUD, e.g. above a tab bar, we should ignore
```
romoore: Super-nit: s/HUD/Snackbar/

components/Snackbar/src/private/MDCSnackbarMessageViewInternal.h
```diff
@@ -53,7 +53,19 @@ typedef void (^MDCSnackbarMessageDismissHandler)(BOOL userInitiated,
 /**
  If the snackbar view should be anchored to the bottom of the screen. Default is YES.
  */
-@property(nonatomic) BOOL anchoredToScreenEdge;
+@property(nonatomic) BOOL anchoredToScreenBottom;
+
+/**
+ The layout constraint which determines how far the snackbar is from the leading edge of the screen.
```
romoore: Nit: %s/snackbar/Snackbar/g

(apply change throughout)

components/Snackbar/src/private/MDCSnackbarMessageViewInternal.h
```diff
@@ -53,7 +53,19 @@ typedef void (^MDCSnackbarMessageDismissHandler)(BOOL userInitiated,
...
+ It is active when the alignment of the parent overlay view is MDCAlignmentLeading.
```
romoore: s/MDCAlignmentLeading/MDCSnackbarAlignmentLeading/

components/Snackbar/src/private/MDCSnackbarMessageViewInternal.h
```diff
@@ -53,7 +53,19 @@ typedef void (^MDCSnackbarMessageDismissHandler)(BOOL userInitiated,
...
+ */
+@property(nonatomic, nullable) NSLayoutConstraint *leadingConstraint;
+
+/**
+ The layout constraint used to center the snackbar.
+ It is active when the alignment of the parent overlay view is MDCAlignmentCenter.
```
romoore: s/MDCAlignmentCenter/MDCSnackbarAlignmentCenter/

romoore (changes requested):
Overall looks good to me. A few comments and a handful of nits.  Please also add tests (unit or UI) or file a backlog issue so we can track adding tests.

components/Snackbar/src/MDCSnackbarManager.h
andrewoverton: Since it's a class method that sets a property on the internal snackbar manager would you mind just another method that reads the value from the internal snackbar manager?

components/Snackbar/src/MDCSnackbarManager.h
andrewoverton: Ah, ok, I was just going off what I thought were conventions based off my observations.

components/Snackbar/src/private/MDCSnackbarMessageViewInternal.h
andrewoverton: ~I wouldn't want the diff to get significantly larger so how about I do this just in my changes rather than for all other lowercase mentions of Snackbars?~

components/Snackbar/src/private/MDCSnackbarMessageViewInternal.h
andrewoverton: Good catch!

components/Snackbar/src/private/MDCSnackbarOverlayView.m
andrewoverton: I can do that!

components/Snackbar/src/MDCSnackbarManager.h
jverkoey: For things where there are existing conventions we do prefer staying consistent rather than fixing in-place so as not to increase the scope of PRs. @romoore / @andrewoverton this looks like a good thing to file an issue for to do follow-up cleanup work.

andrewoverton: Alrighty then I will add them back!

components/Snackbar/src/private/MDCSnackbarMessageViewInternal.h
romoore: Agreed, sorry for the confusion (I wrote it poorly).

components/Snackbar/src/MDCSnackbarManager.h
romoore: Maybe a class property would work well?

andrewoverton: Ah, I forgot that class properties were added to Objective-C in 2016, nice

components/Snackbar/src/private/MDCSnackbarMessageViewInternal.h
andrewoverton: I already did it for a lot of occurrences... I can push it up and you can let me know if you want me to re-lowercase them...

romoore: Oh, it's fine. I meant that you didn't need to do unnecessary work. 🤷‍♀️

andrewoverton:
Unless I’m forgetting something (which, let’s be real, is entirely possible!) I’ve addressed all the feedback on this, with one exception: I haven’t written tests. I tried doing a unit test, but ran into problems when I realized I couldn’t mock size classes. I then briefly looked into UI tests, but I guess there are some uncertainties with where we stand on those. My understanding is that Earl Grey v1 is not something we want to move forward with, and XCUITest infrastructure should probably be set up in its own PR. So, I’ve added [a comment](https://github.com/material-components/material-components-ios/issues/3823#issuecomment-390784591) in https://github.com/material-components/material-components-ios/issues/3823 to eventually add test coverage to this.

Here are a few more screenshots, because they can't hurt:

simulator screen shot - ipad air 2 - 2018-05-21 at 17 04 26
simulator screen shot - ipad air 2 - 2018-05-21 at 17 04 29
simulator screen shot - iphone x - 2018-05-21 at 17 01 32
simulator screen shot - iphone x - 2018-05-21 at 17 01 37

components/Snackbar/src/private/MDCSnackbarOverlayView.m
```diff
@@ -246,51 +250,41 @@ - (void)setSnackbarView:(MDCSnackbarMessageView *)snackbarView {
...
-                                                              relatedBy:NSLayoutRelationEqual
-                                                                 toItem:container
-                                                              attribute:NSLayoutAttributeCenterX
-                                                             multiplier:1.0
-                                                               constant:0]];
+      BOOL isRegularWidth =
+          self.traitCollection.horizontalSizeClass == UIUserInterfaceSizeClassRegular;
+      BOOL isRegularHeight =
+          self.traitCollection.verticalSizeClass == UIUserInterfaceSizeClassRegular;
+      if (isRegularWidth && isRegularHeight) {
```
romoore, line 280: If you want to make it easier to unit test, consider refactoring this check into a method that can be (subclassed and) overridden (or partially mocked).

For example, 

```objectivec
- (BOOL)permitsLeadingPresentation {
  // Normally do the trait collection checks here
}

- (void)setSnackbarView:(MDCSnackbarMessageView *)snackbarView {
  ...
  if ([self permitsLeadingPresentation]) {
    ...
  }
}
```

With that, you could hopefully test/verify the behavior of `setSnackbarView:` without requiring actual trait collections to exist.

romoore: The other option is to file an Issue in the backlog and come back to it later.

romoore: Ah, just saw you referenced #3823

components/Snackbar/src/MDCSnackbarManager.h
```diff
@@ -32,6 +33,19 @@
...
+
+ If called within an animation block, the change will be animated.
+
+ @note This setting is only used when both the horizontal and vertical size classes of the presenting
+ window are @c UIUserInterfaceSizeClassRegular. Otherwise @c MDCSnackbarAlignmentCenter
+ will be used.
+
+ @note The setter must be called from the main thread.
+ */
+@property (class, nonatomic) MDCSnackbarAlignment alignment;
```
romoore: Super-nit: Please mark this property as explicitly `assign`.

components/Snackbar/src/private/MDCSnackbarOverlayView.m
```diff
@@ -27,26 +29,26 @@
...
 NSTimeInterval const MDCSnackbarExitTransitionDuration = 0.075f;
 NSTimeInterval const MDCSnackbarLegacyTransitionDuration = 0.5f;
 
-// The scaling starting point for presenting the new snackbar.
+// The scaling starting point for presenting the new Snackbar.
 static const CGFloat MDCSnackbarEnterStartingScale = 0.8f;
 
-// How far from the bottom of the screen should the snackbar be.
+// How far from the bottom of the screen should the Snackbar be.
 static const CGFloat MDCSnackbarBottomMargin_iPhone = 8.f;
```
romoore, line 41: Are these constants still based on the device idiom or should they now reflect the size classes?

components/Snackbar/src/private/MDCSnackbarMessageViewInternal.h
```diff
@@ -51,14 +51,26 @@ typedef void (^MDCSnackbarMessageDismissHandler)(BOOL userInitiated,
...
  */
-@property(nonatomic) BOOL anchoredToScreenEdge;
+@property(nonatomic) BOOL anchoredToScreenBottom;
 
 /**
- Creates a snackbar view to display @c message.
+ The layout constraint which determines how far the Snackbar is from the leading edge of the screen.
+ It is active when the alignment of the parent overlay view is MDCSnackbarAlignmentLeading.
+ */
+@property(nonatomic, nullable) NSLayoutConstraint *leadingConstraint;
```
romoore: Super-nit: Please mark this property and `centerConstraint` as explicitly `strong`.

components/Snackbar/src/private/MDCSnackbarOverlayView.m
```diff
@@ -246,51 +250,41 @@ - (void)setSnackbarView:(MDCSnackbarMessageView *)snackbarView {
...
-                                                                 toItem:container
-                                                              attribute:NSLayoutAttributeCenterX
-                                                             multiplier:1.0
-                                                               constant:0]];
+      BOOL isRegularWidth =
+          self.traitCollection.horizontalSizeClass == UIUserInterfaceSizeClassRegular;
+      BOOL isRegularHeight =
+          self.traitCollection.verticalSizeClass == UIUserInterfaceSizeClassRegular;
+      if (isRegularWidth && isRegularHeight) {
+        snackbarView.centerConstraint =
```
romoore: Style/nit: It feels a little weird to have an externally-available constraint property on the message view object binding it to the presenting container. Would it work to instead track it in the container?

romoore (approved):

components/Snackbar/src/private/MDCSnackbarOverlayView.m
andrewoverton: Sure, I'll do that

components/Snackbar/src/private/MDCSnackbarMessageViewInternal.h
andrewoverton: I moved the constraints like you suggested doing in https://github.com/material-components/material-components-ios/pull/4163#discussion_r189891840 and I'm noticing that a lot of the strong properties surrounding them in their new location aren't explicitly declared strong. I'm torn between following your advice and marking these two constraints explicitly as strong and following Jeff's advice by keeping consistent with what's there 😬

components/Snackbar/src/private/MDCSnackbarOverlayView.m
andrewoverton: I didn't address the bottom margin in this PR--I'll make an issue to make sure that the Snackbar makes no reference to device idioms down the road.

components/Snackbar/src/private/MDCSnackbarMessageViewInternal.h
romoore: I think since it's moving to a class with existing properties, follow the norms there. My previous recommendation was based on my (incorrect?) observation that there seemed to be no previous precedent.

ianegordon (approved):
Whoho!

romoore:
👏 🎆 👏

PR #4175, opened 2018-05-16
[FeatureHighlight] title not attributed.

mohammadcazig:
Rookie mistake in the code. Feature Highlight Title was never set as attributed.


ianegordon:
:/ How is this not a warning?

ianegordon (approved):

PR #4194, opened 2018-05-18
[ShadowElevations] Expose BottomNavigationBar elevation

romoore:
There was a defined elevation for BottomNavigation bar but it was not public.


ianegordon (approved):

ianegordon:
Failure appears flaky.

Issue #4193, opened 2018-05-18
[BottomNavigation] BottomNavigationBar does not cast a shadow


romoore:
## Description
MDCBottomNavigationBar has a shadowElevation but does not seem to cast a shadow on the parent view's background.

## Reproduction steps
1. Open the catalog
1. Open Bottom Navigation &rarr; Start Demo
1. Look very closely at the edge between the Bottom Navigation bar and the background

## Expected behavior
The Bottom Navigation bar has a shadow (8 dps)

## Actual behavior
The Bottom Navigation bar has no shadow

![bn-shadow-none](https://user-images.githubusercontent.com/1753199/40241689-a10b4040-5a89-11e8-8372-4279f9f8e081.png)

## Platform (please complete the following information)
iPhone 4s/iOS 9.0 (Simulator)

romoore:
Closing this. Apparently if you REALLY zoom in, you can see a very slight shadow.

PR #4195, opened 2018-05-18
[BottomNavigation] Fix shadow elevation value

romoore:
The Bottom Navigation should have an elevation of 8 dps instead of 6 based on
the latest design guidance.

**Before**
bottomnav-ex-original

**After**
bottomnav-ex-after

### DIFF
--- a/components/BottomNavigation/BUILD
+++ b/components/BottomNavigation/BUILD
@@ -29,6 +29,7 @@ mdc_public_objc_library(
     ],
     deps = [
         "//components/Ink",
+        "//components/ShadowElevations",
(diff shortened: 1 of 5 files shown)


romoore:
Since this is based off of #4194 , marking as WIP but it should be reviewed.

romoore:
@material-prbot That's from a diffbase, once the parent commit is merged this is a single-component change.

romoore:
The fact that Bottom Navigation Bar has *never* cast a shadow is tracked in #4193

ianegordon (approved):
LGTM once the previous PR has landed.

PR #4196, opened 2018-05-18
[TextFields] Getting rid of paddedLabel Class since it's not needed anymore.

mohammadcazig:


components/TextFields/src/MDCTextInputControllerOutlined.m
```diff
@@ -171,9 +171,9 @@ - (UIBezierPath *)roundedPathFromRect:(CGRect)f
 
   // Draw the path
   [path moveToPoint:CGPointMake(radius + xOffset, yOffset)];
-  [path addLineToPoint:CGPointMake(offset + xOffset - radius, yOffset)];
```
willlarche, line 174: Is this related?

willlarche:
I'd update the PR to not have screenshots since there should be no perceivable difference.

components/TextFields/src/MDCTextInputControllerOutlined.m
mohammadcazig: Yes the padding needed to be accounted for, and instead of that I was using radius the wrong way.

willlarche (approved):

PR #4230, opened 2018-05-22
[Tabs] We should not force viewDidLoad when ViewControllers are set. 

### LINKED ISSUE #4229: [Tabs] MDCTabBarController must not load views of its children when ViewControllers are set.
MDCTabBarController should only load the view of the view controller that is being displayed lazily.
Currently when setting `viewControllers` property all views are loaded.

mohammadcazig:
MDCTabBarController
TabBarController now load the view for it's children and itself when required.
We only load the view for MDCTabBarController when it's added to the view hierarchy or when its explicitly loaded. Same applies to its view controllers, we only load the selected view controller when it's part of the view hierarchy and load the rest lazily as user switch context.

Note that MDCTabBarViewController behavior is currently slightly different than UITabBarViewController.

Calling init in UITabBarViewController forces the UITabBarViewController's view to load, however initWithCoder: through storyboard does not.
Also setting the View Controllers sets the first view controller as selected and loads the first view controllers view as well.

Close #4229


ianegordon:
title edit:
[Tabs] We should *not* force viewDidLoad when ViewControllers are set

components/Tabs/src/MDCTabBarViewController.m
```diff
@@ -165,11 +166,22 @@ - (void)setTabBarHidden:(BOOL)hidden animated:(BOOL)animated {
...
+
+  if (![oldViewControllers isEqual:viewControllers]) {
     // For all view controllers that this is removing, follow UIViewController.h's rules for
     // for removing a child view controller. See the comments in UIViewController.h for more
     // information.
-    for (UIViewController *viewController in _viewControllers) {
+    for (UIViewController *viewController in oldViewControllers) {
       if (![viewControllers containsObject:viewController]) {
         [viewController willMoveToParentViewController:nil];
         if (viewController.viewLoaded) {
```
ianegordon: super-nit: isViewLoaded

components/Tabs/src/MDCTabBarViewController.m
```diff
@@ -165,11 +166,22 @@ - (void)setTabBarHidden:(BOOL)hidden animated:(BOOL)animated {
 }
 
 - (void)setViewControllers:(NSArray<UIViewController *> *)viewControllers {
-  if (![_viewControllers isEqual:viewControllers]) {
+  NSArray<UIViewController *> *oldViewControllers = _viewControllers;
+  _viewControllers = viewControllers;
```
ianegordon: copy so the client can't change this underneath us.

ianegordon (dismissed):

components/Tabs/src/MDCTabBarViewController.m
```diff
@@ -178,36 +190,47 @@ - (void)setViewControllers:(NSArray<UIViewController *> *)viewControllers {
...
 }
 
-- (void)setSelectedViewController:(nullable UIViewController *)selectedViewController {
-  if (_selectedViewController != selectedViewController) {
-    if (selectedViewController) {
-      NSAssert([_viewControllers containsObject:selectedViewController], @"not one of us.");
-    }
-    BOOL animated = NO;
-    UIView *oldView = _selectedViewController.view;
-    _selectedViewController = selectedViewController;
```
ianegordon, line 195: Can we not just move the add childViewController from updateTabBarItems to here?  (And not touch any of the other code)

mohammadcazig: Discussed Offline, This is required to match Apple's behavior.

ianegordon:
Excellent, thank you @mohammadcazig !

ianegordon (approved):

PR #4238, opened 2018-05-22
[Dialogs] Setting Properties on Alert controller will not force the view to load.

### LINKED ISSUE #4239: [Dialogs] View gets loaded when any of the properties are set.
MDCAlertController should only be loaded when the view is explicitly called or presented but not when properties are set on the vc.

mohammadcazig:
Close #4239


ianegordon:
Please return the state of the code regarding title and message back to the original implementation.
91f88a4 is likely a good starting point.

ianegordon (dismissed):
See 91f88a4 for the original implementation.
Specifically look at the weak alertView reference, avoiding properties in the init and checks for existence before calling setters.

ianegordon (approved):

PR #4282, opened 2018-05-25
[TextFields] Post SetText notifications for attributed text

romoore:
TextFields were not always correctly posting notifications when attributed
text was set programmatically. This can cause the placeholder to remain in its
"empty" position and overlap the text.

Also added some examples to demonstrate the behavior and fixed a typo.

### Screenshots
|Style | Before | After |
|--|--|--|
| Filled | single-filled-before | single-filled-after 
| Underlined | single-underline-before | single-underline-after |
| TextArea | textarea-before | textarea-after |

### DIFF
--- a/components/TextFields/examples/TextFieldOutlinedExample.m
+++ b/components/TextFields/examples/TextFieldOutlinedExample.m
@@ -128,8 +128,8 @@
 
   self.messageController =
       [[MDCTextInputControllerOutlinedTextArea alloc] initWithTextInput:textFieldMessage];
-  textFieldMessage.text = @"This is where you could put a multi-line message like an email. It can"
-      "even handle new lines./n";
+  textFieldMessage.text = @"This is where you could put a multi-line message like an email.\nIt can"
+      " even handle new lines.";
   self.messageController.placeholderText = @"Message";
 
   NSDictionary *views = @{
--- a/components/TextFields/src/MDCTextField.m
+++ b/components/TextFields/src/MDCTextField.m
@@ -452,6 +452,17 @@ static const CGFloat MDCTextInputEditingRectRightViewPaddingCorrection = -2.f;
   }
 }
 
+- (void)setAttributedText:(NSAttributedString *)attributedText {
+  [super setAttributedText:attributedText];
+  [_fundament didSetText];
+
+  if (!self.isFirstResponder) {
+    [[NSNotificationCenter defaultCenter]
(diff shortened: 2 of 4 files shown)


andrewoverton (approved):
Setting attributed text the same way we set regular text seems like a good approach to me--LGTM

components/TextFields/examples/TextFieldOutlinedExample.m
```diff
@@ -128,8 +128,8 @@ - (void)viewDidLoad {
 
   self.messageController =
       [[MDCTextInputControllerOutlinedTextArea alloc] initWithTextInput:textFieldMessage];
-  textFieldMessage.text = @"This is where you could put a multi-line message like an email. It can"
-      "even handle new lines./n";
+  textFieldMessage.text = @"This is where you could put a multi-line message like an email.\nIt can"
```
willlarche: I think this conflicts with what I just merged.

willlarche (changes requested):

components/TextFields/src/MDCTextField.m
```diff
@@ -452,6 +452,17 @@ - (void)setText:(NSString *)text {
   }
 }
 
+- (void)setAttributedText:(NSAttributedString *)attributedText {
+  [super setAttributedText:attributedText];
+  [_fundament didSetText];
+
+  if (!self.isFirstResponder) {
```
willlarche, line 459: This is here because typing into the UITextField causes `setText:` to be called. I don't think that will happen with `attributedText` since the user can't input attributed strings.

Please test that and remove if you can, this check for `.isFirstResponder`.

willlarche (changes requested):

components/TextFields/examples/TextFieldOutlinedExample.m
romoore: Yep. Resolved.

components/TextFields/src/MDCTextField.m
romoore: It looks like `-setAttributedText:` is called when the user finishes typing in a text field that previously held attributed text. In that case, it's only called when they did end editing (and `isFirstResponder` is true), but I'm hesitant to remove the check at this point.

willlarche: Deal.

willlarche (approved):

PR #4291, opened 2018-05-25
[Dialogs] AlertController button layout issue

ianegordon:
Code was not handling the case where action buttons hadn't called sizeToFit: resulting in unexpected horizontal or vertical button layouts.

### DIFF
--- a/components/Dialogs/src/private/MDCAlertControllerView+Private.m
+++ b/components/Dialogs/src/private/MDCAlertControllerView+Private.m
@@ -273,7 +273,7 @@ static const CGFloat MDCDialogMessageOpacity = 0.54f;
     size.width = MDCDialogActionsInsets.left + MDCDialogActionsInsets.right;
     for (UIButton *button in self.actionButtons) {
       CGSize buttonSize = [button sizeThatFits:size];
-      size.height += buttonSize.width;
+      size.height += buttonSize.height;
       size.width = MAX(size.width, buttonSize.width);
       if (button != [self.actionButtons lastObject]) {
         size.height += MDCDialogActionsVerticalPadding;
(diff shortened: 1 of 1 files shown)


ianegordon:
Before:

simulator screen shot - iphone 5s - 2018-05-25 at 16 41 28

After:

simulator screen shot - iphone 5s - 2018-05-25 at 16 38 04

components/Dialogs/src/private/MDCAlertControllerView+Private.m
```diff
@@ -271,8 +272,9 @@ - (CGSize)actionButtonsSizeInVericalLayout {
     size.height = MDCDialogActionsInsets.top + MDCDialogActionsInsets.bottom;
     size.width = MDCDialogActionsInsets.left + MDCDialogActionsInsets.right;
     for (UIButton *button in self.actionButtons) {
-      size.height += CGRectGetHeight(button.bounds);
-      size.width = MAX(size.width, CGRectGetWidth(button.bounds));
+      CGSize buttonSize = [button sizeThatFits:size];
+      size.height += buttonSize.width;
```
jverkoey: `+= buttonSize.height`?

jverkoey (approved):
Small potential typo, but otherwise LGTM.

romoore (dismissed):
Can you please add a unit test to cover this bug?

components/Dialogs/src/private/MDCAlertControllerView+Private.m
ianegordon: Done.

ianegordon:
We don't have any UI tests and that is out of scope for this fix. :(
That said, +1, we should definitely have UI tests.

PR #4311, opened 2018-05-29
[Dialogs] Update trackingView's `clipsToBounds` properly

ajsecord:
When presenting a view with rounded corners, trackingView's corners are visible since clipsToBounds is set to NO by default.

Internal change 198049099.

screen shot 2018-05-29 at 4 54 07 pm


ianegordon:
This change will disable any shadowing by the tracking view since the shadow is rendered outside of the view's bounds.
We added support for the tracking view to have a corner radius in #4233 
I'll follow up on the internal request.

PR #4350, opened 2018-06-01
[Slider] Notify VoiceOver of MDCSlider value changes

### LINKED ISSUE #4289: [Slider] voice over should announce value after adjusting the slider
Voice over currently announces the new value after a double tap + swipe up or down to adjust, but not after a double tap + hold + pan to adjust. This is important because for discrete sliders that second method is currently the only method to adjust the slider value.

andrewoverton:
This change makes it so that there is VoiceOver speech when the value of a slider changes. It's adding functionality to discrete sliders that was previously only there for non-discrete sliders. 

This closes https://github.com/material-components/material-components-ios/issues/4289.

Note: As it stands right now, this change also affects non-discrete sliders. It makes it so that if you choose to swipe horizontally instead of using the brief vertical thumb track swipes to change the value you hear the value as you swipe. There is a delay, so there is not an overwhelming barrage of messages. Rather, you drag your thumb, wait 1 second, and hear the new value. You drag it further along, wait another second, and you hear another new value. Honestly I think it's an improvement, but I thought I should make note!

### DIFF
--- a/components/Slider/src/MDCSlider.m
+++ b/components/Slider/src/MDCSlider.m
@@ -488,6 +488,7 @@ static inline UIColor *MDCThumbTrackDefaultColor(void) {
 
 - (void)thumbTrackValueChanged:(__unused MDCThumbTrack *)thumbTrack {
   [self sendActionsForControlEvents:UIControlEventValueChanged];
+  UIAccessibilityPostNotification(UIAccessibilityAnnouncementNotification, self.accessibilityValue);
 }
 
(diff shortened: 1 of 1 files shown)


romoore (approved):
Please verify that while testing at different speech rates the notifications won't be interrupted/overlap.

andrewoverton:
@romoore -- At slow speeds, the new value will interrupt the old value being recited, but I verified that this is the same behavior as with the system UISlider

romoore:
Sounds good enough to me!

PR #4389, opened 2018-06-10
[TextFields] Text Input adjustment.

### LINKED ISSUE #4388: [TextFields] Input text is 1 point too high
before

willlarche:
Closes #4388

Before:
before

After:
after


components/TextFields/src/MDCTextField.m
```diff
@@ -523,6 +523,7 @@ - (CGRect)textRectForBounds:(CGRect)bounds {
                                                     self.placeholderLabel.font.lineHeight) /
                                                 2.f);  // Text field or placeholder
   actualY = textInsets.top - actualY;
+  actualY += 1.f;
```
ianegordon: Super-nit: Consider adding "+ 1." to the previous line.

ianegordon (approved):

components/TextFields/src/MDCTextField.m
willlarche: Done. and made a const.

willlarche:
This will break screenshot tests. They will need to be updated.

PR #4398, opened 2018-06-15
[BottomNavigation] Fix bug when `setItems` does not correctly set items

### LINKED ISSUE #3177: [BottomNavigation] `setItems` does not correctly set `itemViews` (b/77258452)
### Overview
Reproduced from internal issue.

MDCBottomNavigationBar doesn't display views if items is set twice.

It needs self.itemViews = [NSMutableArray array]; on this line
https://github.com/material-components/material-components-ios/blob/develop/components/BottomNavigation/src/MDCBottomNavigationBar.m#L406

In particular, the itemViews grows each time you call this, leaving the old views.  This causes the real itemViews to be scrolled off the right edge.

### Version number
49.0.0

codeman7:
We might want to use [NSMutableArray array]; instead of `removeAllObjects`.

Closes #3177

### DIFF
--- a/components/BottomNavigation/src/MDCBottomNavigationBar.m
+++ b/components/BottomNavigation/src/MDCBottomNavigationBar.m
@@ -408,6 +408,7 @@ static NSString *const kMDCBottomNavigationBarTitleString = @"title";
   if (_items == items) {
     return;
   }
+  
   // Remove existing item views from the bottom navigation so it can be repopulated with new items.
   for (MDCBottomNavigationItemView *itemView in self.itemViews) {
     [itemView removeFromSuperview];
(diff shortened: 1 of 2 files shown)


romoore (changes requested):
Thanks for the fix, @codeman7 . Could you please add unit tests for the change? You may need to redeclare `itemViews` in a category for the tests.

randallli (commented):
test comment

codeman7:
I don't have very much experience with test, I hope this is what your looking for @romoore.

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -408,15 +408,14 @@ - (void)setItems:(NSArray<UITabBarItem *> *)items {
   if (_items == items) {
     return;
   }
-
```
romoore: Nit: Consider restoring this newline and the one ~before~ below?

romoore (approved):
Yep, looks good to me with just a small nit-pick. Thanks!

components/BottomNavigation/tests/unit/BottomNavigationTests.m
```diff
@@ -93,7 +93,8 @@ -(void)testItemReset {
   self.bottomNavBar.items = @[item1, item2, item3];
 
   // Then
-  XCTAssertEqual(self.bottomNavBar.itemViews.count, 3);
+  unsigned int tabsCount = 3;
```
romoore: Please use NSUInteger, since `unsigned long` is used on 64-bit hosts.

ianegordon (approved):
LGTM

Issue #3016, opened 2018-02-28
[BottomNavigation] UIAppearance values not visible until after first tap


romoore:
### Overview
When configuring MDCBottomNavigationBar values (`selectedItemTintColor`, `unselectedItemTintColor`) via UIAppearance, those values are not used until after the first item is selected.

**Initial View Load**
![mdc-bottomnav-before](https://user-images.githubusercontent.com/1753199/36798193-e8383702-1c77-11e8-8cb1-e64e97fae633.png)

**After selecting an item**
![mdc-bottomnav-after](https://user-images.githubusercontent.com/1753199/36798198-eda71c94-1c77-11e8-9bf3-2639b7c72807.png)

### Reproduction steps
 - Detailed steps to reproduce

### Version number
46.0.0

### Operating system and device
iPhone 5/iOS 8.3 (Simulator)

romoore:
Additional report from internal issue (b/77265558):

> When selectedItemTintColor and unselectedItemTintColor are set the MDCBottomNavigationBar sets existing itemViews to the right values, but that assumes items has already been set.  If the tint color is set first then the color does not actually get set.  I believe this is true for font also.
>
> However, setSelectedViewController must be called after all of these or the initial tint is wrong.
>
> So, currently, you have to set items, then set the tints, then select the initial starting item.  Any other order and you don't get the tints.

PR #4401, opened 2018-06-18
[BottomNavigation] Remove UI Appearance selectors. **Breaking change**

codeman7:
Removes all selectors from BottomNavigationBar and its components

This addresses the issue #3016 as we discussed.


romoore:
I can get to this tomorrow, but probably not today.

ianegordon:
I'd prefer to remove all the UIAPPEARANCE annotations at once, possibly in it's own PR / release.  Since we are unable to deprecate these annotations, this will be a major version bump.

ianegordon:
There is another issue that needs to be addressed before we close the issue.  I'm going to close this PR for now, but we should re-open when we remove UIAppearance.

PR #4407, opened 2018-06-19
[List] Initial List PR Feedback

andrewoverton:
I met with @ianegordon last Friday and we took passes through both the APIs and implementations of the list cell classes I’ve been working on.

He suggested I make the following changes to the stuff introduced in https://github.com/material-components/material-components-ios/pull/4382:

- Get rid of the `MDCListBaseCellState ` enum
- Replace state dependent elevation mechanism with an elevation property that works on the underlying shadow layer
- Make the ink spread across the entire screen in the `ManualLayoutListItemCell2.m` example, outside the safe area insets.
- Replace readonly view properties with pass-through accessors for:
  -  UILabels, specifically textColor, font, alignment, and text
  -  UIImageViews, specifically image and cornerRadius
- Remove `MDCTypographyScheme`—use `MDCFontTextStyles` instead
- See if you can avoid setting frames in `preferredLayoutAttributesFittingAttributes:`—do it in `layoutSubviews` instead (I still need to do this.)

It's not complete yet--I'm still working on making it more solid and addressing some iOS 8 bugs, etc. But this is the general idea.

A PR with _just_ the suggested API is coming soon.

### DIFF
--- a/components/List/examples/Manual Layout List 2/ManualLayoutListBaseCell2.h
+++ b/components/List/examples/Manual Layout List 2/ManualLayoutListBaseCell2.h
@@ -20,61 +20,11 @@
 
 @class MDCInkView;
 
-/**
- The cell can be in a Normal, a Highlighted, or a Selected state.
- The Highlighted and Selected states map to the properties of the UICollectionViewCell that
- share the same names.
- */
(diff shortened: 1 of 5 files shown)


andrewoverton:
I'm going to close this because some of the changes it includes are related to things still under discussion in the API review PR.

PR #4408, opened 2018-06-19
[List] Suggested List API

andrewoverton:
Here are the suggested APIs for the two List cell classes in development.

### DIFF
--- a/components/List/examples/supplemental/MDCListBaseCell.h
+++ b/components/List/examples/supplemental/MDCListBaseCell.h
@@ -15,8 +15,6 @@
 
 #import "MaterialShadowElevations.h"
 
-@class MDCInkView;
-
 @interface MDCListBaseCell : UICollectionViewCell
 
 /**
--- a/components/List/examples/supplemental/MDCListBaseCell.h
+++ b/components/List/examples/supplemental/MDCListBaseCell.h
@@ -1,0 +1,27 @@
+/*
+ Copyright 2018-present the Material Components for iOS authors. All Rights Reserved.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+ http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+#import "MaterialShadowElevations.h"
+
+@class MDCInkView;
+
+@interface MDCListBaseCell : UICollectionViewCell
+
+/**
+ The shadow elevation for the cell.
+ */
+@property (nonatomic, assign) MDCShadowElevation elevation;
+
+@end
--- a/components/List/examples/supplemental/MDCListItemCell.h
+++ b/components/List/examples/supplemental/MDCListItemCell.h
@@ -1,0 +1,104 @@
+/*
+ Copyright 2018-present the Material Components for iOS authors. All Rights Reserved.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+ http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+#import "MDCListBaseCell.h"
+
+/**
+ A basic Material List Cell with labels and imageviews.
+ */
+@interface MDCListItemCell : MDCListBaseCell
+
+/**
+ Sets the cell width for the self-sizing cell.
+ Note: The self-sizing is only applied to the height, and the width is set using this method.
+ */
+@property (nonatomic, assign) CGFloat cellWidth;
+
+/**
+ A convenience accessor for the title text.
+ */
+@property (nonatomic, copy, nullable) NSString *titleText;
+
+/**
+ A convenience accessor for the title text color.
+ */
+@property (nonatomic, strong, nullable) UIColor *titleLabelTextColor;
+
+/**
+ A convenience accessor for the title text font.
+ */
+@property (nonatomic, strong, nullable) UIFont *titleLabelFont;
+
+/**
+ A convenience accessor for the title text font.
+ */
+@property (nonatomic, assign) NSTextAlignment titleLabelTextAlignment;
+
+/**
+ A convenience accessor for the detail text.
+ */
+@property (nonatomic, copy, nullable) NSString *detailText;
+
+/**
+ A convenience accessor for the detail text color.
+ */
+@property (nonatomic, strong, nullable) UIColor *detailLabelTextColor;
+
+/**
+ A convenience accessor for the detail text font.
+ */
+@property (nonatomic, strong, nullable) UIFont *detailLabelFont;
+
+/**
+ A convenience accessor for the detail text font.
+ */
+@property (nonatomic, assign) NSTextAlignment detailLabelTextAlignment;
+
+/**
+ The UIImageview used to display the leading image.
+ */
+@property (strong, nonatomic, nonnull, readonly) UIImageView *leadingImageView;
+
+/**
+ A convenience accessor for the leading image.
+ */
+@property (strong, nonatomic, nullable) UIImage *leadingImage;
+
+/**
+ A convenience accessor for the leading image corner radius.
+ */
+@property (nonatomic, assign) UIImage *leadingImageCornerRadius;
+
+/**
+ A convenience accessor for the trailing image.
+ */
+@property (strong, nonatomic, nullable) UIImage *trailingImage;
+
+/**
+ A convenience accessor for the trailing image corner radius.
+ */
+@property (nonatomic, assign) UIImage *trailingImageCornerRadius;
+
+/**
+ Indicates whether the view's contents should automatically update their font when the device’s
+ UIContentSizeCategory changes.
+ This property is modeled after the adjustsFontForContentSizeCategory property in the
+ UIContentSizeCategoryAdjusting protocol added by Apple in iOS 10.0.
+ Default value is NO.
+ */
+@property(nonatomic, readwrite, setter=mdc_setAdjustsFontForContentSizeCategory:)
+BOOL mdc_adjustsFontForContentSizeCategory;
+
+@end
--- a/components/List/examples/supplemental/MDCListItemCell.h
+++ b/components/List/examples/supplemental/MDCListItemCell.h
@@ -21,12 +21,6 @@
 @interface MDCListItemCell : MDCListBaseCell
 
 /**
- Sets the cell width for the self-sizing cell.
- Note: The self-sizing is only applied to the height, and the width is set using this method.
- */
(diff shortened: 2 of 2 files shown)


components/List/examples/supplemental/MDCListBaseCell.h
```diff
@@ -0,0 +1,27 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+#import "MaterialShadowElevations.h"
+
+@class MDCInkView;
```
romoore: Can this forward declaration be dropped? It doesn't look like it's being used.

romoore: Could you please add docs about how you expect this class to be used? 

Related: Based on its location in the file system, it appears to be example code.  Is this actually intended to be a component?

components/List/examples/supplemental/MDCListItemCell.h
```diff
@@ -0,0 +1,104 @@
...
+/**
+ A basic Material List Cell with labels and imageviews.
+ */
+@interface MDCListItemCell : MDCListBaseCell
+
+/**
+ Sets the cell width for the self-sizing cell.
+ Note: The self-sizing is only applied to the height, and the width is set using this method.
+ */
+@property (nonatomic, assign) CGFloat cellWidth;
```
romoore: What happens if this is zero or negative, is the `intrinsicContentSize` of the subviews used to compute it?

components/List/examples/supplemental/MDCListItemCell.h
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+#import <UIKit/UIKit.h>
+
+#import "MDCListBaseCell.h"
+
+/**
+ A basic Material List Cell with labels and imageviews.
+ */
+@interface MDCListItemCell : MDCListBaseCell
```
romoore, line 21: Can you expand the documentation for this class a bit, maybe reference which of the Material Lists row types it supports?

components/List/examples/supplemental/MDCListItemCell.h
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ A convenience accessor for the title text color.
+ */
+@property (nonatomic, strong, nullable) UIColor *titleLabelTextColor;
+
+/**
+ A convenience accessor for the title text font.
+ */
+@property (nonatomic, strong, nullable) UIFont *titleLabelFont;
```
romoore, line 36: Can you please provide documentation of the defaults for these properties? Are the `nullable` properties defaulting to `nil`?

components/List/examples/supplemental/MDCListItemCell.h
```diff
@@ -0,0 +1,104 @@
...
+@property (nonatomic, assign) UIImage *trailingImageCornerRadius;
+
+/**
+ Indicates whether the view's contents should automatically update their font when the device’s
+ UIContentSizeCategory changes.
+ This property is modeled after the adjustsFontForContentSizeCategory property in the
+ UIContentSizeCategoryAdjusting protocol added by Apple in iOS 10.0.
+ Default value is NO.
+ */
+@property(nonatomic, readwrite, setter=mdc_setAdjustsFontForContentSizeCategory:)
```
romoore: Nit: Should this be `nonatomic, assign` instead? I don't think we normally mark properties as `readwrite` unless a superclass defines them as `readonly`.

components/List/examples/supplemental/MDCListItemCell.h
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ A convenience accessor for the trailing image.
+ */
+@property (strong, nonatomic, nullable) UIImage *trailingImage;
+
+/**
+ A convenience accessor for the trailing image corner radius.
+ */
+@property (nonatomic, assign) UIImage *trailingImageCornerRadius;
```
romoore, line 81: Why does an image need a corner radius? Can't clients just set a rounded rectangular image?

components/List/examples/supplemental/MDCListItemCell.h
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ A convenience accessor for the detail text font.
+ */
+@property (nonatomic, assign) NSTextAlignment detailLabelTextAlignment;
+
+/**
+ The UIImageview used to display the leading image.
+ */
+@property (strong, nonatomic, nonnull, readonly) UIImageView *leadingImageView;
```
romoore: I see `leadingImageView` is exposed but not the `titleLabel` nor the `trailingImageView`. Is the leading image view special in some way that we expect clients to use it differently than the other subviews?

romoore (dismissed):

components/List/examples/supplemental/MDCListItemCell.h
andrewoverton: Sure, happy to change. I just copy/pasted this from somewhere else.

components/List/examples/supplemental/MDCListItemCell.h
andrewoverton: This was a request from @ianegordon. He asked that I get rid of the publicly exposed views and replace them with the "pass through" accessors you see here. This was one he mentioned.

You also noticed [here](https://github.com/material-components/material-components-ios/pull/4408#discussion_r197149877) that a UIImageView is still exposed. That was a mistake..

This question of whether or not to expose the views is something that we should all discuss more, because my instinct was actually to expose readonly views rather than have a boatload of pass through properties. I think Ian actually began to reconsider his earlier stance once he realized we'd have to expose pass through properties for all the accessibility related view properties as well. At some point it just becomes too much...

components/List/examples/supplemental/MDCListItemCell.h
andrewoverton: Sure. In this specific case I just looked at UILabel and saw that the default font was nil so I made this nullable.

components/List/examples/supplemental/MDCListBaseCell.h
andrewoverton: Yep, the idea is that eventually List Items will be a component. I wasn't sure what the file system structure conventions were for API reviews.

components/List/examples/supplemental/MDCListItemCell.h
andrewoverton: One thing I was considering was having a private `BOOL` called `hasSetCellWidth` or something like that that would be set to `YES` in the setter for `cellWidth` and set to `NO` in `prepareForReuse`. There'd be an assertion in the layout code that `hasSetCellWidth` had been set to enforce this property's proper usage. This was just an idea I was toying with. Do you have any thoughts?

romoore: I think that seems fragile and prone to lots of frustration by client teams.  The width of the cell may not be known ahead of time as well. I typically expect rows like this to be the full width of the collection view section minus any insets. I would hope a UICollectionViewLayout would take care of that rather than the cell being responsible for determining its size.

components/List/examples/supplemental/MDCListItemCell.h
ianegordon: Apologies if I was not clear.  Based on our earlier discussion re: customizability and accessibility, I believe we will need to expose all of UIViews (the two imageviews and the two labels) instead of just the passthrough properties.

components/List/examples/supplemental/MDCListBaseCell.h
ianegordon: While we will certainly ship with documentation, for this API review, please focus on the functionality and refer to the design doc for context / documentation.

components/List/examples/supplemental/MDCListItemCell.h
ianegordon: Please save documentation until after the design review is done. We don't want to spend time documenting and interface that we end up drastically changing / needing to re-document.

components/List/examples/supplemental/MDCListBaseCell.h
romoore: @andrewoverton  Can you please link to the design doc?

ianegordon (dismissed):

components/List/examples/supplemental/MDCListItemCell.h
romoore: Fair enough, but where does the corner radius come into play? I don't see any such property on `UIImageView`.  Is this being added in iOS 12 and I just missed the update?

components/List/examples/supplemental/MDCListItemCell.h
andrewoverton: Let's assume the width of the cell is the width of the collection view section minus any insets, like you said. In that case, in cellForItemAtIndexPath we'd set the cellWidth of the cell, and it would use that to calculate its height. The cell's height is dependent on the width.

components/List/examples/supplemental/MDCListItemCell.h
andrewoverton: It's on the `CALayer`, of course... I think the idea is just that rounding an image is something people would want to often.

I know the three of us talked about this offline yesterday--I think I'm still in favor of exposing the UIViews 😅

components/List/examples/supplemental/MDCListItemCell.h
romoore: Sorry, could you clarify how the cell's height depends on its width?  Maybe some examples or links to the guidelines?

andrewoverton: The cell overrides `-systemLayoutSizeFittingSize:`. Its implementation uses the various UIView-related properties that have been set and the cell width that is specified in the `cellWidth` property to calculate the frame for every subview. After that a height is calculated considering the final subview frames. The method ultimately returns a CGSize based on this desired width and calculated height. Later, in `layoutSubviews`, subviews are laid out according to the frames calculated before.

andrewoverton:
@romoore, in 06b74b2 I addressed some of your comments. Addressed some small nits. Importantly, I also got rid of the public `cellWidth` property. If it’s private, and we set it [like this](https://github.com/andrewoverton/material-components-ios/blob/01001208817f623c0c0b54eb2d3da05c8a6fc311/components/List/examples/Manual%20Layout%20List%203/MDCListItemCell.m#L139), in `-systemLayoutSizeFittingSize:`, the effect is the same. The drawback of this approach is the client has to reset `estimatedItemSize` whenever the UICollectionView bounds change, because the `targetSize` parameter here is based off that property.

andrewoverton:
Also, @romoore and @ianegordon, where did we land on exposing the UIView properties? Talking about `titleLabel`, `detailLabel`, `leadingImageView`, and `trailingImageView`

romoore:
@andrewoverton Regarding the `estimatedItemSize` and `systemLayoutSizeFittingSize` - how do those play with the [`-itemSize`](https://developer.apple.com/documentation/uikit/uicollectionviewflowlayout/1617711-itemsize?language=objc) property in UICollectionViewFlowLayout and [`collectionView:layout:sizeForItemAtIndexPath:`](https://developer.apple.com/documentation/uikit/uicollectionviewdelegateflowlayout/1617708-collectionview?language=objc) in UICollectionViewDelegateFlowLayout?  Does a more accurate value get passed if the delegate implements the method?

romoore:
Looking at exposing the properties, my vote is to only expose them if we think that subclasses should be allowed to change the layout.  If we want that behavior, then we could expose them in a category `(Subclassing)`, but I'm still in favor of the pass-through properties being the way that clients generally customize the appearance of the cell.

andrewoverton:
My understanding is that if you set itemSize, the items will be that size. If you set estimatedCellSize, the cells' self-sizing capabilities will be used. If neither is set, and sizeForItemAtIndexPath is implemented, then the size returned from that method is used. For this cell I'd recommend using the self-sizing capabilities. It's built with that in mind.

I'm okay with whatever re exposing views

andrewoverton:
I'll just close this until we start talking about it again

PR #4420, opened 2018-06-21
[BottomNavigation] Add example when we explicitly set color

### LINKED ISSUE #4428: [BottomNavigation] Verify order of operations setting colors works as expected.
b/77265558

iOS 11/iPhone X

codeman7:
Add example to dragons for when we explicitly set tint color from buttons for bottom navigation

Closes #4428

### DIFF
--- a/components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
+++ b/components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
@@ -19,7 +19,6 @@ import MaterialComponents
 
 class BottomNavigationExplicitlySetColorExample: UIViewController {
 
-  let appBar = MDCAppBar()
   var colorScheme = MDCSemanticColorScheme()
 
   let bottomNavBar = MDCBottomNavigationBar()
@@ -29,53 +28,11 @@ class BottomNavigationExplicitlySetColorExample: UIViewController {
 
   init() {
     super.init(nibName: nil, bundle: nil)
-    self.title = "Bottom Navigation Set Color (Swift)"
-
-    self.addChildViewController(appBar.headerViewController)
-    let color = UIColor(white: 0.2, alpha: 1)
-    appBar.headerViewController.headerView.backgroundColor = color
-    appBar.navigationBar.tintColor = .white
-    appBar.navigationBar.titleTextAttributes = [NSForegroundColorAttributeName: UIColor.white]
-
-    commonBottomNavigationExampleInit()
   }
 
   @available(*, unavailable)
   required init?(coder aDecoder: NSCoder) {
     super.init(coder: aDecoder)
-  }
-
-  func commonBottomNavigationExampleInit() {
-    view.backgroundColor = .lightGray
-    view.addSubview(bottomNavBar)
-
-    bottomNavBar.titleVisibility = .always
-
-    bottomNavBar.alignment = .centered
-
-    // Add items to the bottom navigation bar.
-    let tabBarItem1 = UITabBarItem(title: "Home", image: UIImage(named: "Home"), tag: 0)
-    let tabBarItem2 =
-      UITabBarItem(title: "Messages", image: UIImage(named: "Email"), tag: 0)
-    let tabBarItem3 =
-      UITabBarItem(title: "Favorites", image: UIImage(named: "Favorite"), tag: 0)
-    bottomNavBar.items = [ tabBarItem1, tabBarItem2, tabBarItem3 ]
-
-    // Select a bottom navigation bar item.
-    bottomNavBar.selectedItem = tabBarItem2;
-
-    // Layout buttons
-    redButton.frame = CGRect(x: view.center.x - 72, y: view.center.y - 72, width: 144, height: 56)
-    redButton.backgroundColor = .red
-    redButton.setTitle("Red theme", for: .normal)
-    redButton.addTarget(self, action: #selector(redTheme), for: .touchUpInside)
-    view.addSubview(redButton)
-
-    blueButton.frame = CGRect(x: view.center.x - 72, y: view.center.y + 16, width: 144, height: 56)
-    blueButton.backgroundColor = .blue
-    blueButton.setTitle("Blue theme", for: .normal)
-    blueButton.addTarget(self, action: #selector(blueTheme), for: .touchUpInside)
-    view.addSubview(blueButton)
   }
 
   func layoutBottomNavBar() {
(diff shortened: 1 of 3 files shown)


components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
```diff
@@ -0,0 +1,145 @@
...
+  var colorScheme = MDCSemanticColorScheme()
+
+  let bottomNavBar = MDCBottomNavigationBar()
+
+  let redButton = MDCButton()
+  let blueButton = MDCButton()
+
+  init() {
+    super.init(nibName: nil, bundle: nil)
+    self.title = "Bottom Navigation Set Color (Swift)"
```
romoore: I believe this will be set automatically by the breadcrumbs, would you please check?

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
```diff
@@ -0,0 +1,145 @@
...
+
+    self.addChildViewController(appBar.headerViewController)
+    let color = UIColor(white: 0.2, alpha: 1)
```
romoore: I think it's better to use the color scheme colors. Preferably in `viewDidLoad` instead of the initializer.

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
```diff
@@ -0,0 +1,145 @@
...
+
+    commonBottomNavigationExampleInit()
+  }
+
+  @available(*, unavailable)
+  required init?(coder aDecoder: NSCoder) {
+    super.init(coder: aDecoder)
+  }
+
+  func commonBottomNavigationExampleInit() {
```
romoore: It looks like most of this can be moved to `viewDidLoad`, is there a reason it has to be called from `init`? If yes, could you please add a comment documenting the reason.

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
```diff
@@ -0,0 +1,145 @@
...
+    view.backgroundColor = .lightGray
+    view.addSubview(bottomNavBar)
+
+    bottomNavBar.titleVisibility = .always
```
romoore: Nit: Let's not use `.always` for `titleVisibility` unless we absolutely must.  Users with impaired color vision may not be able to tell which tab is selected.  Is it in the critical path for this example?

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
```diff
@@ -0,0 +1,145 @@
...
+    bottomNavBar.selectedItem = tabBarItem2;
+
+    // Layout buttons
+    redButton.frame = CGRect(x: view.center.x - 72, y: view.center.y - 72, width: 144, height: 56)
+    redButton.backgroundColor = .red
+    redButton.setTitle("Red theme", for: .normal)
+    redButton.addTarget(self, action: #selector(redTheme), for: .touchUpInside)
+    view.addSubview(redButton)
+
+    blueButton.frame = CGRect(x: view.center.x - 72, y: view.center.y + 16, width: 144, height: 56)
```
romoore: Please consider using `sizeToFit` to support dynamic type.

romoore (dismissed):

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
```diff
@@ -0,0 +1,145 @@
...
+                                                               toBottomNavigation: bottomNavBar)
+    bottomNavBar.backgroundColor = .darkGray
+  }
+
+  override func viewWillAppear(_ animated: Bool) {
+    super.viewWillAppear(animated)
+    self.navigationController?.setNavigationBarHidden(true, animated: animated)
+  }
+
+  @objc func redTheme(_ button: UIButton) {
```
ianegordon, line 106: In order to verify more code paths, please set a new set of items before the colors in this func and please set a new set of items after the colors in the bluefunc below.

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
codeman7: Done

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
codeman7: Done

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
codeman7: Done

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
codeman7: If you don't set the title then the app bar doesn't have a title.

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
codeman7: Done

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
codeman7: Setting up a different example for what we talked about

romoore: Where is the other example?

codeman7: Setting up a different example for when we explicitly reset items and then set their colors or when we set their colors then reset items.

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
```diff
@@ -0,0 +1,130 @@
...
+  override func viewSafeAreaInsetsDidChange() {
+    super.viewSafeAreaInsetsDidChange()
+    layoutBottomNavBar()
+  }
+  #endif
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+
+    view.backgroundColor = .lightGray
```
romoore: Please use `colorScheme.backgroundColor`

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
```diff
@@ -0,0 +1,130 @@
...
+    view.addSubview(bottomNavBar)
+
+    bottomNavBar.alignment = .centered
+
+    // Add items to the bottom navigation bar.
+    let tabBarItem1 = UITabBarItem(title: "Home", image: UIImage(named: "Home"), tag: 0)
+    let tabBarItem2 =
+      UITabBarItem(title: "Messages", image: UIImage(named: "Email"), tag: 0)
```
romoore: Nit: Please make the tags unique.

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
```diff
@@ -0,0 +1,130 @@
...
+    blueButton.sizeToFit()
+    blueButton.frame.origin = CGPoint(x: view.center.x - (blueButton.frame.width / 2),
+                                      y: view.center.y + 16)
+    blueButton.backgroundColor = .blue
+    blueButton.addTarget(self, action: #selector(blueTheme), for: .touchUpInside)
+    view.addSubview(blueButton)
+
+    MDCBottomNavigationBarColorThemer.applySemanticColorScheme(colorScheme,
+                                                               toBottomNavigation: bottomNavBar)
+    bottomNavBar.backgroundColor = .darkGray
```
romoore: This should not be needed, I thought the Bottom Nav's background color is expected to remain clear?

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
```diff
@@ -0,0 +1,130 @@
...
+
+    // Select a bottom navigation bar item.
+    bottomNavBar.selectedItem = tabBarItem2;
+
+    // Layout buttons
+    redButton.setTitle("Red theme", for: .normal)
+    redButton.sizeToFit()
+    redButton.frame.origin = CGPoint(x: view.center.x - (redButton.frame.width / 2),
+                                     y: view.center.y - (redButton.frame.height + 16))
+    redButton.backgroundColor = .red
```
romoore: Nit: Please use material palettes for these colors.  
Suggest red tint400 with black text  blue tint500 with black text.

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
```diff
@@ -0,0 +1,130 @@
...
+    bottomNavBar.backgroundColor = .darkGray
+  }
+
+  override func viewWillAppear(_ animated: Bool) {
+    super.viewWillAppear(animated)
+    self.navigationController?.setNavigationBarHidden(true, animated: animated)
+  }
+
+  @objc func redTheme(_ button: UIButton) {
+    bottomNavBar.selectedItemTintColor = .red
```
romoore: Please consider using material palette colors with theming. It would allow us to test live themer changes instead of just the tint colors.

romoore (dismissed):

ianegordon:
Please address comment before merge.

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
codeman7: That doesn't effect the background for the view. That leaves the view being black.

components/BottomNavigation/examples/BottomNavigationExplicitlySetColorExample.swift
codeman7: Bottom navigation is purple if you keep take this out, this allows us to be able to see the colors change better.

codeman7:
The other example is coming in a different PR.

ianegordon (approved):

PR #4422, opened 2018-06-21
[BottomNavigation] Update ripple color to match spec.

### LINKED ISSUE #4403: [BottomNavigation] Ripple color is wrong when selected and unselected tint colors aren't the same.
When you set `selectedItemTintColor` and `unselectedItemTintColor` are not the same then the ripple stays the same color as originally set, weather that item is selected or not.

Set `selectedItemTintColor` and `unselectedItemTintColor` to be different values.
Go to navigation bar in catalog tap a tab that wasn't originally selected and the ripple is the `unselectedItemTintColor`, then tap the original `selectedItemTintColor` and it's color is still the `selectedItemTintColor`.

I believe when tapping an item it should always be the `selectedItemTintColor` because on a tap it becomes the "selected" item.

bottomnav

iOS 11/iPhone X

codeman7:
Update spec to match what designer said

Closes #4403

### DIFF
--- a/components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
+++ b/components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
@@ -363,8 +363,8 @@ static NSString *const kMDCBottomNavigationItemViewTabString = @"tab";
   if (!self.selected) {
     self.iconImageView.tintColor = self.unselectedItemTintColor;
     self.label.textColor = self.unselectedItemTintColor;
-    CGFloat alpha = MDCBottomNavigationItemViewInkOpacity;
-    self.inkView.inkColor = [self.unselectedItemTintColor colorWithAlphaComponent:alpha];
+    self.inkView.inkColor = [self.selectedItemTintColor
+                             colorWithAlphaComponent:MDCBottomNavigationItemViewInkOpacity];
   }
 }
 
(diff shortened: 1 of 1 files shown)


components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
```diff
@@ -363,8 +363,8 @@ - (void)setUnselectedItemTintColor:(UIColor *)unselectedItemTintColor {
   if (!self.selected) {
     self.iconImageView.tintColor = self.unselectedItemTintColor;
     self.label.textColor = self.unselectedItemTintColor;
-    CGFloat alpha = MDCBottomNavigationItemViewInkOpacity;
-    self.inkView.inkColor = [self.unselectedItemTintColor colorWithAlphaComponent:alpha];
+    self.inkView.inkColor = [self.selectedItemTintColor
+                             colorWithAlphaComponent:MDCBottomNavigationItemViewInkOpacity];
```
ianegordon: layout-nit:
Our coding style typically indents 4 characters when wrapping a line.
Consider:
```
self.inkView.inkColor = 
    [self.selectedItemTintColor colorWithAlphaComponent:MDCBottomNavigationItemViewInkOpacity];
```

ianegordon (changes requested):

ianegordon (approved):
Please address comment before merge.

codeman7:
Should be good now.

PR #4431, opened 2018-06-22
[BottomNavigation] Test cases when we reset the bottom navigation items array

### LINKED ISSUE #4429: [BottomNavigation] Issue with `setItems` call layout of items
Explicitly set items after they have been set, lays out all the items wrong.

iOS 11.4/iPhone X

codeman7:
Added an example to showcase when items are reset to an new array.

Fixed the bug when we reset items and they don't layout correctly.

After a reset, set the selected tab to Null to mirror UITabBar
https://developer.apple.com/documentation/uikit/uitabbar/1623453-selecteditem?language=objc

Added two test to check for selection being Null after a reset and frames not being zero after a reset.

Closes #4429

### DIFF
--- a/components/BottomNavigation/examples/BottomNavigationResetButtons.swift
+++ b/components/BottomNavigation/examples/BottomNavigationResetButtons.swift
@@ -26,9 +26,10 @@ class BottomNavigationResetExample: UIViewController {
   // The tabs that will be reordered later
   let tabBarItem1 = UITabBarItem(title: "Home", image: UIImage(named: "Home"), tag: 0)
   let tabBarItem2 =
-    UITabBarItem(title: "Messages", image: UIImage(named: "Email"), tag: 0)
+    UITabBarItem(title: "Messages", image: UIImage(named: "Email"), tag: 1)
   let tabBarItem3 =
-    UITabBarItem(title: "Favorites", image: UIImage(named: "Favorite"), tag: 0)
+    UITabBarItem(title: "Favorites", image: UIImage(named: "Favorite"), tag: 2)
+  let tabBarItem4 = UITabBarItem(title: "Cake", image: UIImage(named: "Cake"), tag: 3)
 
   let buttonOne = MDCButton()
   let buttonTwo = MDCButton()
@@ -107,7 +108,8 @@ class BottomNavigationResetExample: UIViewController {
   }
 
   @objc func reorderTwo(_ button: UIButton) {
-    bottomNavBar.items = [tabBarItem3, tabBarItem1, tabBarItem2]
+    bottomNavBar.items = [tabBarItem3, tabBarItem1, tabBarItem2, tabBarItem4]
+    bottomNavBar.selectedItem = tabBarItem3
   }
 }
 
--- a/components/BottomNavigation/examples/BottomNavigationResetButtons.swift
+++ b/components/BottomNavigation/examples/BottomNavigationResetButtons.swift
@@ -69,7 +69,7 @@ class BottomNavigationResetExample: UIViewController {
 
   override func viewDidLoad() {
     super.viewDidLoad()
-    view.backgroundColor = MDCPalette.grey.tint100
+    view.backgroundColor = colorScheme.backgroundColor
     view.addSubview(bottomNavBar)
 
     bottomNavBar.alignment = .centered
--- a/components/BottomNavigation/examples/BottomNavigationResetButtons.swift
+++ b/components/BottomNavigation/examples/BottomNavigationResetButtons.swift
@@ -104,12 +104,10 @@ class BottomNavigationResetExample: UIViewController {
 
   @objc func reorderOne(_ button: UIButton) {
     bottomNavBar.items = [tabBarItem2, tabBarItem3, tabBarItem1]
-    
   }
 
   @objc func reorderTwo(_ button: UIButton) {
     bottomNavBar.items = [tabBarItem3, tabBarItem1, tabBarItem2]
-    bottomNavBar.selectedItem = tabBarItem2
   }
 }
 
--- a/components/BottomNavigation/examples/BottomNavigationResetButtons.swift
+++ b/components/BottomNavigation/examples/BottomNavigationResetButtons.swift
@@ -1,0 +1,129 @@
+/*
+ Copyright 2018-present the Material Components for iOS authors. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+import Foundation
+import MaterialComponents
+
+class BottomNavigationResetExample: UIViewController {
+
+  var colorScheme = MDCSemanticColorScheme()
+
+  let bottomNavBar = MDCBottomNavigationBar()
+
+  // The tabs that will be reordered later
+  let tabBarItem1 = UITabBarItem(title: "Home", image: UIImage(named: "Home"), tag: 0)
+  let tabBarItem2 =
+    UITabBarItem(title: "Messages", image: UIImage(named: "Email"), tag: 0)
+  let tabBarItem3 =
+    UITabBarItem(title: "Favorites", image: UIImage(named: "Favorite"), tag: 0)
+
+  let buttonOne = MDCButton()
+  let buttonTwo = MDCButton()
+
+  init() {
+    super.init(nibName: nil, bundle: nil)
+  }
+
+  @available(*, unavailable)
+  required init?(coder aDecoder: NSCoder) {
+    super.init(coder: aDecoder)
+  }
+
+  func layoutBottomNavBar() {
+    let size = bottomNavBar.sizeThatFits(view.bounds.size)
+    let bottomNavBarFrame = CGRect(x: 0,
+                                   y: view.bounds.height - size.height,
+                                   width: size.width,
+                                   height: size.height)
+    bottomNavBar.frame = bottomNavBarFrame
+  }
+
+  override func viewWillLayoutSubviews() {
+    super.viewWillLayoutSubviews()
+    layoutBottomNavBar()
+  }
+
+  #if swift(>=3.2)
+  @available(iOS 11, *)
+  override func viewSafeAreaInsetsDidChange() {
+    super.viewSafeAreaInsetsDidChange()
+    layoutBottomNavBar()
+  }
+  #endif
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+
+    view.backgroundColor = .lightGray
+    view.addSubview(bottomNavBar)
+
+    bottomNavBar.alignment = .centered
+
+    // Add items to the bottom navigation bar.
+    bottomNavBar.items = [ tabBarItem1, tabBarItem2, tabBarItem3 ]
+
+    // Select a bottom navigation bar item.
+    bottomNavBar.selectedItem = tabBarItem2
+
+    // Layout buttons
+    buttonOne.setTitle("Reorder One", for: .normal)
+    buttonOne.sizeToFit()
+    buttonOne.frame.origin = CGPoint(x: view.center.x - (buttonOne.frame.width / 2),
+                                     y: view.center.y - (buttonOne.frame.height + 16))
+    buttonOne.addTarget(self, action: #selector(reorderOne), for: .touchUpInside)
+    view.addSubview(buttonOne)
+
+    buttonTwo.setTitle("Reorder Two", for: .normal)
+    buttonTwo.sizeToFit()
+    buttonTwo.frame.origin = CGPoint(x: view.center.x - (buttonTwo.frame.width / 2),
+                                      y: view.center.y + 16)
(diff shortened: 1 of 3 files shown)


components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -405,7 +405,7 @@ - (void)didCancelTouchesForButton:(UIButton *)button {
 #pragma mark - Setters
 
 - (void)setItems:(NSArray<UITabBarItem *> *)items {
-  if (_items == items) {
+  if ([_items isEqual:items]) {
```
ianegordon: You likely want both so the const and nil case are handled:
if (items == _items || [_items isEqual:items])

ianegordon (approved):

ianegordon:
If you put "WIP" in the PR title, the WIP check will come up red, so people won't merge the PR before you are finished. (WIP = work in progress)

romoore:
As a team member with merge access, nobody will merge your PRs unless you put the label [`PR:merge-when-approved`](https://github.com/material-components/material-components-ios/labels/PR%3Amerge-when-approved).  For contributors without merge access, we tend to merge their PRs if they're unresponsive (and it's ready to go) for a few days.

components/BottomNavigation/examples/BottomNavigationResetButtons.swift
```diff
@@ -0,0 +1,129 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+import Foundation
+import MaterialComponents
+
+class BottomNavigationResetExample: UIViewController {
+
+  var colorScheme = MDCSemanticColorScheme()
```
romoore: Please add a `typographyScheme` property so we can support injecting styles.

components/BottomNavigation/examples/BottomNavigationResetButtons.swift
```diff
@@ -0,0 +1,129 @@
...
+  let tabBarItem2 =
+    UITabBarItem(title: "Messages", image: UIImage(named: "Email"), tag: 1)
+  let tabBarItem3 =
+    UITabBarItem(title: "Favorites", image: UIImage(named: "Favorite"), tag: 2)
+  let tabBarItem4 = UITabBarItem(title: "Cake", image: UIImage(named: "Cake"), tag: 3)
+
+  let buttonOne = MDCButton()
+  let buttonTwo = MDCButton()
+
+  init() {
```
romoore, line 39: Is this required for Swift? Isn't the superclass default behavior of `-init` to call `-initWithNibName:bundle:`?

components/BottomNavigation/examples/BottomNavigationResetButtons.swift
```diff
@@ -0,0 +1,129 @@
...
+  func layoutBottomNavBar() {
+    let size = bottomNavBar.sizeThatFits(view.bounds.size)
+    let bottomNavBarFrame = CGRect(x: 0,
+                                   y: view.bounds.height - size.height,
+                                   width: size.width,
+                                   height: size.height)
+    bottomNavBar.frame = bottomNavBarFrame
+  }
+
+  override func viewWillLayoutSubviews() {
```
romoore, line 57: Can this be accomplished with a resizing mask instead?

components/BottomNavigation/examples/BottomNavigationResetButtons.swift
```diff
@@ -0,0 +1,129 @@
...
+  override func viewSafeAreaInsetsDidChange() {
+    super.viewSafeAreaInsetsDidChange()
+    layoutBottomNavBar()
+  }
+  #endif
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+
+    view.backgroundColor = .lightGray
```
romoore: Please use `colorScheme.backgroundColor`

components/BottomNavigation/examples/BottomNavigationResetButtons.swift
```diff
@@ -0,0 +1,129 @@
...
+    buttonTwo.frame.origin = CGPoint(x: view.center.x - (buttonTwo.frame.width / 2),
+                                      y: view.center.y + 16)
+    buttonTwo.addTarget(self, action: #selector(reorderTwo), for: .touchUpInside)
+    view.addSubview(buttonTwo)
+
+    MDCBottomNavigationBarColorThemer.applySemanticColorScheme(colorScheme,
+                                                               toBottomNavigation: bottomNavBar)
+  }
+
+  override func viewWillAppear(_ animated: Bool) {
```
romoore: Instead of overriding `-viewWillAppear`, can you use `-catalogShouldHideNavigation` and return `true`?

components/BottomNavigation/examples/BottomNavigationResetButtons.swift
```diff
@@ -0,0 +1,129 @@
...
+    self.navigationController?.setNavigationBarHidden(true, animated: animated)
+  }
+
+  @objc func reorderOne(_ button: UIButton) {
+    bottomNavBar.items = [tabBarItem2, tabBarItem3, tabBarItem1]
+  }
+
+  @objc func reorderTwo(_ button: UIButton) {
+    bottomNavBar.items = [tabBarItem3, tabBarItem1, tabBarItem2, tabBarItem4]
+    bottomNavBar.selectedItem = tabBarItem3
```
romoore: Why does this method set the selected item but `reorderOne` does not?

components/BottomNavigation/examples/BottomNavigationResetButtons.swift
```diff
@@ -0,0 +1,129 @@
...
+  }
+}
+
+// MARK: Catalog by convention
+extension BottomNavigationResetExample {
+  class func catalogBreadcrumbs() -> [String] {
+    return ["Bottom Navigation", "Bottom Navigation Reset (Swift)"]
```
romoore: I'm not sure I understand the "reset" part of this. Isn't it a reorder/reassignment example?

components/BottomNavigation/examples/BottomNavigationResetButtons.swift
```diff
@@ -0,0 +1,129 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+import Foundation
+import MaterialComponents
+
+class BottomNavigationResetExample: UIViewController {
```
romoore: Please consider a quick comment describing what this example demonstrates and why.

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -473,7 +473,11 @@ - (void)setItems:(NSArray<UITabBarItem *> *)items {
     [self.itemViews addObject:itemView];
     [self.containerView addSubview:itemView];
   }
+  self.selectedItem = self.items[0];
```
romoore: Would this crash if `items` is set to `nil`?

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -473,7 +473,11 @@ - (void)setItems:(NSArray<UITabBarItem *> *)items {
...
   [self addObserversToTabBarItems];
+  self.selectedItem = NULL;
```
romoore: Should this be `nil` instead of `NULL`?

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -473,7 +473,11 @@ - (void)setItems:(NSArray<UITabBarItem *> *)items {
...
+  [self setNeedsLayout];
+  [self setNeedsDisplay];
```
romoore: Is this required? I don't see a `-drawRect` and I would expect `layoutSubviews` to be sufficient for updating the view.

components/BottomNavigation/tests/unit/BottomNavigationTests.m
```diff
@@ -97,4 +97,17 @@ -(void)testItemReset {
   XCTAssertEqual(self.bottomNavBar.itemViews.count, tabsCount);
 }
 
+-(void)testSelectedItemAfterReset {
```
romoore: Can this check be added to the test above instead of a separate test?

romoore (dismissed):

components/BottomNavigation/examples/BottomNavigationResetButtons.swift
codeman7: Just to showcase when we set a selected item after a reset vs when we just reset.

components/BottomNavigation/examples/BottomNavigationResetButtons.swift
codeman7: This is needed or a crash will occur.

romoore: I don't get a crash when I try:

```objectivec
id bar = [[MDCBottomNavigationBar alloc] init];
```

I also see some of our example VCs don't reimplement `init` and don't crash.  Like [TextFieldOutlinedObjectiveCExample](https://github.com/material-components/material-components-ios/blob/develop/components/TextFields/examples/TextFieldOutlinedExample.m)

Is this a bug in CatalogByConvention?

components/BottomNavigation/examples/BottomNavigationResetButtons.swift
romoore: Ok. Could we perhaps name the method, `reorderItemsAndSetSelected` or something more descriptive?

components/BottomNavigation/examples/BottomNavigationResetButtons.swift
codeman7: I think it's only an issue with Swift

components/BottomNavigation/examples/BottomNavigationResetButtons.swift
codeman7: I don't believe UIViewController responds to this for it's background color.

components/BottomNavigation/examples/BottomNavigationResetButtons.swift
```diff
@@ -0,0 +1,131 @@
...
+  @available(iOS 11, *)
+  override func viewSafeAreaInsetsDidChange() {
+    super.viewSafeAreaInsetsDidChange()
+    layoutBottomNavBar()
+  }
+  #endif
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+    view.backgroundColor = MDCPalette.grey.tint100
```
romoore: Please use the color scheme

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -473,7 +473,9 @@ - (void)setItems:(NSArray<UITabBarItem *> *)items {
     [self.itemViews addObject:itemView];
     [self.containerView addSubview:itemView];
   }
+  self.selectedItem = nil;
```
romoore, line 476: I believe this would be a change in behavior for clients. Does it make sense to check if the previous `selectedItem` is in the current `items` array and only deselect it if it's absent?

components/BottomNavigation/tests/unit/BottomNavigationTests.m
```diff
@@ -97,4 +97,34 @@ -(void)testItemReset {
...
+  UITabBarItem *item3 = [[UITabBarItem alloc] initWithTitle:@"2" image:nil tag:2];
+  [self.bottomNavBar sizeToFit];
+  self.bottomNavBar.frame = CGRectMake(0, 0, 320, 56);
+  [self.bottomNavBar setNeedsLayout];
+
+  // When
+  self.bottomNavBar.items = @[item1, item2];
+  self.bottomNavBar.items = @[item1, item2, item3];
+  [self.bottomNavBar layoutIfNeeded];
+
```
romoore, line 112: Nit: Please add the `Then` block markers here and the other test.

components/BottomNavigation/tests/unit/BottomNavigationTests.m
```diff
@@ -97,4 +97,34 @@ -(void)testItemReset {
...
 }
 
+-(void)testFramesAfterReset {
+  // Given
+  UITabBarItem *item1 = [[UITabBarItem alloc] initWithTitle:@"1" image:nil tag:0];
+  UITabBarItem *item2 = [[UITabBarItem alloc] initWithTitle:@"2" image:nil tag:1];
+  UITabBarItem *item3 = [[UITabBarItem alloc] initWithTitle:@"2" image:nil tag:2];
+  [self.bottomNavBar sizeToFit];
+  self.bottomNavBar.frame = CGRectMake(0, 0, 320, 56);
+  [self.bottomNavBar setNeedsLayout];
```
romoore: Is this necessary?

components/BottomNavigation/examples/BottomNavigationResetButtons.swift
romoore: Does the MDCBottomNavigationBar account for the safe area automatically in `sizeThatFits`?

romoore (changes requested):

components/BottomNavigation/examples/BottomNavigationResetButtons.swift
romoore: Ack.

components/BottomNavigation/src/MDCBottomNavigationBar.m
codeman7: Without this line you get unexpected behavior, where you can't select the item that was previously selected and it doesn't show itself as selected. The default behavior for UITabBar is to reset to nil so that's why we have this.

romoore: Ack.

romoore (approved):
LGTM!

PR #4434, opened 2018-06-24
[TextFields] Making leading and trailing view layout customizeable

willlarche:
Closes the layout part of b/74174981

***Breaks screenshot tests.

Before:
simulator screen shot - iphone x - 2018-06-24 at 17 39 27

After:
simulator screen shot - iphone x - 2018-06-24 at 17 35 45

Also corrects the x of the floating placeholder on a filled text field.
simulator screen shot - iphone x - 2018-06-24 at 17 45 43


components/TextFields/src/MDCTextInput.h
```diff
@@ -205,6 +205,22 @@ typedef NS_ENUM(NSUInteger, MDCTextInputTextInsetsMode) {
 
 @end
 
+@protocol MDCLeadingViewTextInput <MDCTextInput>
```
randallli: We should put some docs for this protocol.

I assume we have a different one because we dont want to implement this for all of our textInput implementations. Once the old ones get deprecated is your long term plan to merge the protocols?

components/TextFields/src/MDCTextField.m
```diff
@@ -254,6 +254,34 @@ - (void)applyCursorColor {
   self.tintColor = self.cursorColor;
 }
 
+#pragma mark - MDCLeadingViewTextInput Implementation
+
+- (void)setLeadingView:(UIView *)leadingView {
```
randallli, line 259: unit test?

components/TextFields/src/MDCTextFieldPositioningDelegate.h
```diff
@@ -44,10 +44,38 @@
  */
 - (CGRect)editingRectForBounds:(CGRect)bounds defaultRect:(CGRect)defaultRect;
 
+/**
+ The area that the leadingView should inhabit when shown.
+
+ @param defaultRect The default value of the leading view rect.
+ */
+- (CGRect)leadingViewRectForBounds:(CGRect)bounds defaultRect:(CGRect)defaultRect;
```
randallli, line 52: these look very unit testable. I'd feel much more comfortable if they were unit tested

randallli (approved):
More docs and some unit tests please.

components/TextFields/src/MDCTextField.m
willlarche: Don't have time for unit tests so I made #4437

components/TextFields/src/MDCTextFieldPositioningDelegate.h
willlarche: #4437

components/TextFields/src/MDCTextInput.h
willlarche: Done.

This is just bringing out from inside the text field and making it possible to do introspection on it. That allows easier layout calculations in the controllers.

This is not implemented on multiline because multiline text fields should not have a leading view.

PR #4436, opened 2018-06-25
[BottomNavigation, Math] Fix `BOOL` types

romoore:
Two files were using C99 `bool` instead of Objective-C `BOOL` types. This
could lead to breakages in the future if the `BOOL` type were redefined by the
Objective-C headers.


codeman7 (approved):
LGTM

ianegordon (approved):

PR #4440, opened 2018-06-26
[List] Suggested Base List Cell API

andrewoverton:
Since there are people interested in getting the base List cell ASAP and the subclass with UILabels and UIImageViews is still under discussion we're exploring releasing the base cell first. So this is the suggested API for that. It's really just a shadow property... The working implementation is [here](https://github.com/andrewoverton/material-components-ios/blob/01001208817f623c0c0b54eb2d3da05c8a6fc311/components/List/examples/Manual%20Layout%20List%203/MDCBaseCollectionViewCell.m) if you want to take a look. It adds an `MDCInkView`, that's pretty much it!

### DIFF
--- a/components/List/src/MDCBaseCell/MDCBaseCell.h
+++ b/components/List/src/MDCBaseCell/MDCBaseCell.h
@@ -1,0 +1,25 @@
+/*
+ Copyright 2018-present the Material Components for iOS authors. All Rights Reserved.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+ http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+#import "MaterialShadowElevations.h"
+
+@interface MDCBaseCell : UICollectionViewCell
+
+/**
+ The shadow elevation for the cell.
+ */
+@property (nonatomic, assign) MDCShadowElevation elevation;
+
+@end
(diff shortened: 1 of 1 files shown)


components/List/src/MDCBaseCell/MDCBaseCell.h
```diff
@@ -0,0 +1,25 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+#import "MaterialShadowElevations.h"
+
+@interface MDCBaseCell : UICollectionViewCell
```
romoore, line 18: Please add a property so clients can set the Ink color for the Ripple.

romoore (dismissed):

andrewoverton:
Here's a (somewhat low quality) gif of a base cell demo that can be found [here](https://github.com/andrewoverton/material-components-ios/tree/293e386d455c8103a26a2fc25f39ad54052e1779/components/List/examples/Manual%20Layout%20List%20Items/MDCBaseCell) if anyone is interested.

jun-29-2018 10-47-12

romoore:
Is it just the gif or does the Ripple not reach the edge of the cell?

andrewoverton:
@romoore it's not just the gif. That's definitely happening for some reason.

romoore:
I suspect the radius may be equal to the cell width, which is too short for a circle inside a rectangle.  I suggest casting some geometry/trig spells to fix it. ;-)

romoore:
```objectivec
#import "MaterialMath.h"

inkRippleRadius = MDCHypot(MDCRectGetWidth(bounds), MDCGetHeight(bounds));
```

Updated: You should probably calculate directly from the point clicked to the farthest edge. That way the ripple animation can correctly compute the animation duration/ripple velocity.

andrewoverton:
Yeah I had briefly tinkered with `maxRippleRadius` and didn't get what I wanted quickly... I need to take another look!

andrewoverton:
The design doc was approved so I'm going to close this and open another with the implementation and a readme etc.

PR #4449, opened 2018-06-28
[TextFields] Support clear button tint color

### LINKED ISSUE #4447: [TextFields] Support clearButton tintColor theming
## Feature Request/Suggestion

TextInputControllers should support customizing the color of their textInput's clearButton's `tintColor` property.

## Additional context

Internal issue: b/79467503

romoore:
Adding support to change the clearButton's tintColor from the TextInputController. This allows more consistent theming across controllers.

Closes #4447 

## Filled

**Before**
screen shot 2018-06-29 at 12 23 12 pm

**After**
screen shot 2018-06-29 at 12 21 51 pm

## Outlined

**Before**
screen shot 2018-06-29 at 12 23 01 pm

**After**
screen shot 2018-06-29 at 12 21 36 pm

### DIFF
--- a/components/TextFields/examples/TextFieldOutlinedExample.m
+++ b/components/TextFields/examples/TextFieldOutlinedExample.m
@@ -50,7 +50,6 @@
                                 toTextInputController:controller];
   [MDCTextFieldTypographyThemer applyTypographyScheme:self.typographyScheme
                                           toTextInput:controller.textInput];
-
 }
 
 - (void)viewDidLoad {
@@ -122,7 +121,6 @@
   self.cityController.placeholderText = @"City";
   [self styleTextInputController:self.cityController];
 
-
   MDCTextField *textFieldState = [[MDCTextField alloc] init];
   textFieldState.translatesAutoresizingMaskIntoConstraints = NO;
 
@@ -134,7 +132,6 @@
   self.stateController.placeholderText = @"State";
   [self styleTextInputController:self.stateController];
 
-
   MDCTextField *textFieldZip = [[MDCTextField alloc] init];
   textFieldZip.translatesAutoresizingMaskIntoConstraints = NO;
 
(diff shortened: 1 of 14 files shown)


components/TextFields/src/MDCTextInputControllerBase.h
```diff
@@ -104,4 +104,15 @@ extern const CGFloat MDCTextInputControllerBaseDefaultBorderRadius;
  */
 @property(nonatomic, assign) NSUInteger minimumLines;
 
+/**
```
willlarche, line 100: I think this should be universal. Put it in `MDCTextInputController`.

willlarche (dismissed):

components/TextFields/src/MDCTextInputControllerBase.h
romoore: We spoke about this offline before I pushed up the PR. You agreed that we should only make the change for Outlined and Filled styles at this point.  The reasoning, as I understood it, is that [Material Guidelines only recommend the Outlined and Filled styles](https://material.io/design/components/text-fields.html#usage) based on internal usability testing.

Has something changed so that we should be actively supporting the other styles with new feature development?

romoore: After discussing, I've filed #4453 as a follow-up to migrate this property up to the root protocol. Due to the amount of testing that should be done across implementing classes, we decided not to add it to this PR.

components/TextFields/examples/TextFieldOutlinedExample.m
```diff
@@ -38,9 +43,26 @@ - (void)dealloc {
...
+                                toTextInputController:controller];
+  [MDCTextFieldTypographyThemer applyTypographyScheme:self.typographyScheme
+                                          toTextInput:controller.textInput];
+}
+
 - (void)viewDidLoad {
   [super viewDidLoad];
-  self.view.backgroundColor = [UIColor whiteColor];
+
+  if (!self.colorScheme) {
```
willlarche, line 58: `viewDidLoad` is guaranteed to be called once. I think you can drop these null checks.

willlarche (approved):
With comment.

components/TextFields/examples/TextFieldOutlinedExample.m
romoore: It's in case the schemes are injected by the Catalog, we don't want to overwrite those with the defaults.

PR #4452, opened 2018-06-29
[TextFields] {BreakingChange} Remove unused `backgroundColor` property

romoore:
Only MDCTextInputControllerFullWidth is using `backgroundColor` in any
meaningful way. For MDCTextInputControllerBase, the property wasn't used
anywhere in the input controller.

Reduces the scope of #3357


andrewoverton (approved):
This LGTM I suppose! @romoore, is there something we need to do when there's a breaking change? I forget...

romoore:
These properties had no effect outside of the FullWidth controller. I suppose we could mark them deprecated THIS release and then delete them NEXT release?

@ianegordon or @randallli might be able to weigh-in here.

willlarche:
I believe this should just go thru without a deprecation. It was dead code in the other styles and isn't leaving the one that's using it. Just making it less public.

randallli (approved):

romoore:
Adding @yarneo as the next release cop.

PR #4454, opened 2018-06-29
[Cards] Add accessibility documentation

galiak11:
Adding accessibility documentation to Cards.
Closes issue: b/78768477.
Also note: "components/Cards/README.md" was generated using the generate_readme script.


components/Cards/docs/accessibility.md
```diff
@@ -0,0 +1,66 @@
...
+
+To help ensure your Cards are accessible to as many users as possible, please be sure to review the following 
+recommendations:
+
+#### Accessibility Labels for Nested Elements
+
+Set an appropriate[`accessibility
+Label`](https://developer.apple.com/documentation/uikit/uiaccessibilityelement/1619577-accessibilitylabel) 
+for meaningful content that is not textual by default, like images. Also make sure to enable the element's 
+[`isAccessible
```
romoore: s/isAccessible Element/isAccessibilityElement/

components/Cards/docs/accessibility.md
```diff
@@ -0,0 +1,66 @@
+### Accessibility Labels
+
+To help ensure your Cards are accessible to as many users as possible, please be sure to review the following 
+recommendations:
+
+#### Accessibility Labels for Nested Elements
+
+Set an appropriate[`accessibility
```
romoore: Please make this match the property name: `s/accessibility Label/accessibilityLabel/`

components/Cards/docs/accessibility.md
```diff
@@ -0,0 +1,66 @@
...
+Label`](https://developer.apple.com/documentation/uikit/uiaccessibilityelement/1619577-accessibilitylabel) 
+for meaningful content that is not textual by default, like images. Also make sure to enable the element's 
+[`isAccessible
+Element`](https://developer.apple.com/documentation/objectivec/nsobject/1615141-isaccessibilityelement) 
+property so it is detectable by assistive technologies.
+
+Swift
+```swift
+myImageView.isAccessibilityElement = true
+myImageView.accessibilityLabel = "Image title or some info"
```
romoore: I think it might be better to provide a concrete example here. I think an MDCButton inside an MDCCard seems more likely than a UIImageView.  If it were only a UIImageView, then I would suggest the card itself be the accessibility element.  Maybe something like this?

```swift
cardDismissButton.accessibilityLabel = "Close"
```

components/Cards/docs/accessibility.md
```diff
@@ -0,0 +1,66 @@
...
+Set an appropriate[`accessibility
+Label`](https://developer.apple.com/documentation/uikit/uiaccessibilityelement/1619577-accessibilitylabel) 
+for meaningful content that is not textual by default, like images. Also make sure to enable the element's 
+[`isAccessible
+Element`](https://developer.apple.com/documentation/objectivec/nsobject/1615141-isaccessibilityelement) 
+property so it is detectable by assistive technologies.
+
+Swift
+```swift
+myImageView.isAccessibilityElement = true
```
romoore: I might be mistaken, but I thought UIImageView has `isAccessibilityElement == true` by default. Does MDCCard disable it somehow that clients will need to restore the property to its default value?

components/Cards/docs/accessibility.md
```diff
@@ -0,0 +1,66 @@
...
+
+Objective-C
+```objc
+myImageView.isAccessibilityElement = YES;
+myImageView.accessibilityLabel = @"Image title or some info";
+```
+
+#### Accessibility Label for Tappable Cards
+
+Tappable or Selectable Cards are often used as items in a collection. In this mode, Cards often expose a single
```
romoore: Are we introducing the terms "Tappable Card" and "Selectable Card"? I don't see them mentioned in the [material guidelines article](https://material.io/design/components/cards.html).

romoore: Nit/style: Let's keep "card" as a common noun (lowercase) to match the capitalization in the Material guidelines documents.

components/Cards/docs/accessibility.md
```diff
@@ -0,0 +1,66 @@
...
+Label`](https://developer.apple.com/documentation/uikit/uiaccessibilityelement/1619577-accessibilitylabel) value 
+for the Card, and make sure to update it each time the Card's content changes. Additionally, make sure to enable 
+the card's [`isAccessible
+Element`](https://developer.apple.com/documentation/objectivec/nsobject/1615141-isaccessibilityelement)
+property so it is detectable by assistive technologies.
+
+Swift
+```swift
+card.isAccessibilityElement = true
+card.accessibilityLabel = "Location \(locationLabel.accessibilityLabel) has
```
romoore: I think an example that doesn't user string concatenation would be better.  Instead, how about a custom label based on the same data that those elements might use instead?  I think we should also avoid plurals as well, since [plurals across languages can be very complex](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LocalizingYourApp/LocalizingYourApp.html#//apple_ref/doc/uid/10000171i-CH5-SW10).

```swift
card.accessibilityLabel = "Location \(userLocation.name) is popular with users who enjoy \(userLocation.popularActivityMatchingUserProfile(userProfile)"
```

components/Cards/docs/accessibility.md
```diff
@@ -0,0 +1,66 @@
...
+
+In addition to aggregating the Card's label information, tappable Cards with actionable nested elements, like 
+buttons, sliders, and so on, should also aggregate the Cards actions into a single list. That is done by following the
+[`UIAccessibilityAction`](https://developer.apple.com/documentation/uikit/accessibility/uiaccessibilityaction) 
+guidelines or by adding  [`UIAccessibilityCustom
+Action`](https://developer.apple.com/documentation/uikit/uiaccessibilitycustomaction) to the Card's 
+[`accessibilityCustom
+Actions`](https://developer.apple.com/documentation/objectivec/nsobject/1615150-accessibilitycustomactions).
+
+For a specific example, check `CardsCollectionTintingExample` and `CardTintExampleCell` in the
```
romoore: I don't see any custom actions in these examples. Can you file an Issue for this and let someone come back to it once those changes have landed?

components/Cards/examples/CardExampleViewController.swift
```diff
@@ -45,6 +45,9 @@ class CardExampleViewController: UIViewController {
     cardScheme.colorScheme = colorScheme
     MDCCardThemer.applyScheme(cardScheme, to: card)
     card.isInteractable = false
+
+    imageView.isAccessibilityElement = true
+    imageView.accessibilityLabel = "Card Image"
```
romoore: What is in the image? Reading this code (as a blind user would experience our interface) I cannot tell what the image contains.  If the image is just decorative, we can hide it from UIAccessibility.

components/Cards/examples/EditReorderCollectionViewController.swift
```diff
@@ -121,6 +121,10 @@ class EditReorderCollectionViewController: UIViewController,
 
     MDCCardThemer.applyScheme(cardScheme, toCardCell: cell)
     cell.isSelectable = (toggle == .edit)
+
+    cell.isAccessibilityElement = true
+    cell.accessibilityLabel = "Cell"
```
romoore, line 126: I think it would be more helpful if the cards themselves had unique content.  Maybe we can name them after their colors (do the colors vary)? Perhaps they could have a text label? What do you think?

romoore (dismissed):

components/Cards/docs/accessibility.md
galiak11: Good catch!

components/Cards/docs/accessibility.md
galiak11: Do you mean to keep "accessibility" and "Label" on the same line? This is done so the line do not exceed allowed length. It prints fine this way when displayed as markup.

components/Cards/docs/accessibility.md
galiak11: I believe buttons generally do not need an accessibility label since they have a meaningful default (the button's title).  Nested elements would need accessibility labels only in cases when they don't have meaningful defaults, like images. Obviously not all images need to be accessible, so this example refers to the case in which it is needed.

components/Cards/docs/accessibility.md
galiak11: Nested elements are completely user driven, so I don't see how MDCCard can disable it. I assume images are not accessible elements by default since there is no sensible default text that can be used for them.

components/Cards/docs/accessibility.md
galiak11: The guidelines refer to "Often cards are one large touch target". The guidelines do not mention selectivity, although MDCCard does implement it (more relevant when a card is an item of a collection). Can we discuss this?

galiak11: Sure

components/Cards/docs/accessibility.md
galiak11: I ended up moving it to a separate PR (https://github.com/material-components/material-components-ios/pull/4472/files).

components/Cards/examples/CardExampleViewController.swift
galiak11: I added some meaningful content label, though that's a little out of context, since the rest of the card is just Lorem Ipsum...

components/Cards/examples/EditReorderCollectionViewController.swift
galiak11: Working on it. Now available in a separate PR: https://github.com/material-components/material-components-ios/pull/4488.

ianegordon (approved):

components/Cards/README.md
```diff
@@ -252,3 +255,92 @@ id<MDCColorScheming> colorScheme = [[MDCSemanticColorScheme alloc] init];
...
+### MDCCard Accessibility
+
+To help ensure your cards are accessible to as many users as possible, please be sure to review the following 
+recommendations:
+
+#### Accessibility for Cards in a Collection
+
+Since assistive technologies visit all cards in a collection in a sequential order, it is often easier to distinguish 
+between elements that belong to different cards, by aggregating all the card's information to a single sentence. 
+This can be done by setting an appropriate [`accessibility
```
randallli: remove the spaces in the properties because they are APIs.

components/Cards/README.md
```diff
@@ -252,3 +255,92 @@ id<MDCColorScheming> colorScheme = [[MDCSemanticColorScheme alloc] init];
...
+Label`](https://developer.apple.com/documentation/uikit/uiaccessibilityelement/1619577-accessibilitylabel) for the 
+card. In addition, the card's [`isAccessiblity
+Element`](https://developer.apple.com/documentation/objectivec/nsobject/1615141-isaccessibilityelement) 
+property should be enabled, so it is treated as a single entity by assistive technologies.
```
randallli: copyedit alternative: Cards are a container element and isAccessibiltElement turns off individually selecting its subelements.

randallli (approved):

ianegordon:
Please consider Randall's comment and merge.
Thank you!

PR #4465, opened 2018-07-02
[TextFields] Migrate textInputClearButtonTintColor to MDCTextInputController, tests

### LINKED ISSUE #4453: [TextFields] Migrate textInputClearButtonTintColor property to MDCTextInputController protocol
## Feature Request/Suggestion
MDCTextInputController (the foundational protocol) should support theming the `textInput`'s `clearButton`'s `tintColor`. After #4449 is merged, only the MDCTextInputControllerBase will support it, but the property should be moved up the type hierarchy.

* Add implementations to all classes that inherit the property
* Unit tests for verifying that `...Default` applies to instance properties
* Unit tests that changing the TextInputController property applies to the textInput
* Unit tests for any themers that make use of the property

Original issue to add property to `...Base` controller: #4447

michdud:
Adding support for theming the clear button tint color in MDCTextInputController. Allows for theming the clear button tint color in legacy, full-width, and underline text fields.

Closes #4453.

### DIFF
--- a/components/TextFields/src/MDCTextInputController.h
+++ b/components/TextFields/src/MDCTextInputController.h
@@ -196,6 +196,16 @@
 @property(nonatomic, nullable, strong) UIView<MDCTextInput> *textInput;
 
 /**
+ The tintColor applied to the textInput's clear button.
+ See @c UIImageView.tintColor for additional details.
+ */
+@property(nonatomic, null_resettable, strong) UIColor *textInputClearButtonTintColor;
+
+/**
+ Default value for @c textInputClearButtonTintColor. */
+@property(class, nonatomic, nullable, strong) UIColor *textInputClearButtonTintColorDefault;
+
+/**
  The font applied to the trailing side underline label.
 
  Default is trailingUnderlineLabelFontDefault.
--- a/components/TextFields/tests/unit/TextFieldControllerClassPropertiesLegacyTests.swift
+++ b/components/TextFields/tests/unit/TextFieldControllerClassPropertiesLegacyTests.swift
@@ -42,6 +42,7 @@ class TextFieldControllerClassPropertiesLegacyTests: XCTestCase {
     MDCTextInputControllerLegacyDefault.floatingPlaceholderNormalColorDefault = nil
     MDCTextInputControllerLegacyDefault.floatingPlaceholderScaleDefault = 0.75
     MDCTextInputControllerLegacyDefault.isFloatingEnabledDefault = true
+    MDCTextInputControllerLegacyDefault.textInputClearButtonTintColorDefault = nil
 
     MDCTextInputControllerLegacyFullWidth.errorColorDefault = nil
     MDCTextInputControllerLegacyFullWidth.inlinePlaceholderColorDefault = nil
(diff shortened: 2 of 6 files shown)


components/TextFields/src/MDCTextInputController.h
```diff
@@ -195,6 +195,16 @@
 /** The text input the controller is affecting. */
 @property(nonatomic, nullable, strong) UIView<MDCTextInput> *textInput;
 
+/**
+ The tintColor applied to the textInput's clear button.
+ See @c UIImageView.tintColor for additional details.
+ */
+@property(nonatomic, null_resettable, strong) UIColor *textInputClearButtonTintColor;
```
romoore, line 202: I think these properties should also be removed from MDCTextInputControllerBase.h so that it's forced to inherit them rather than redeclaring them.

romoore (dismissed):
Thanks, @michdud! Looks good aside from cleaning-up the old header.  Please also add a bit to the PR summary so we keep track of why we think it's good to do this.

components/TextFields/src/MDCTextInputController.h
michdud: Done!

willlarche: I agree.

components/TextFields/tests/unit/TextFieldControllerClassPropertiesLegacyTests.swift
```diff
@@ -356,4 +363,83 @@ class TextFieldControllerClassPropertiesLegacyTests: XCTestCase {
...
+    
+    // Then
+    XCTAssertEqual(textInputLegacyDefault.clearButton.tintColor,
+                   MDCTextInputControllerLegacyDefault.textInputClearButtonTintColorDefault)
+  }
+  
+  func testLegacyDefaultTextInputClearButtonTintColorAppliesToTextField() {
+    // Given
+    let textInputLegacyDefault = MDCTextField()
+    let controllerLegacyDefault = MDCTextInputControllerLegacyDefault(textInput: textInputLegacyDefault)
```
willlarche, line 396: I would love to see all these not with `MDCTextInputControllerLegacy...anything` and instead `MDCTextInputControllerFilled` and `MDCTextInputControllerOutline`.

willlarche (changes requested):

components/TextFields/tests/unit/TextFieldControllerClassPropertiesLegacyTests.swift
michdud: The test cases like these for `MDCTextInputControllerFilled` and `MDCTextInputControllerOutline` are in #4449 (unless you were thinking of different test cases?). Should I remove the legacy tests altogether?

willlarche: Spoke offline. Learned that part is already done!

willlarche (approved):

romoore (approved):
LGTM!

michdud:
I'm happy to merge!

PR #4469, opened 2018-07-02
[Catalog] Make color pickers larger

romoore:
The color picker buttons were not large enough for a 48-point touch target.
Increasing their visual size also improves accessibility for users with
imperfect vision.

## Before

catalog-picker-size-before

## After
| Portrait | Landscape|
|--|--|
| catalog-picker-size-after |  catalog-picker-size-after-landscape |

### DIFF
--- a/catalog/MDCCatalog/MDCThemePickerViewController.swift
+++ b/catalog/MDCCatalog/MDCThemePickerViewController.swift
@@ -157,7 +157,7 @@ class MDCThemePickerViewController: UIViewController, UICollectionViewDataSource
                                           toItem: nil,
                                           attribute: .notAnAttribute,
                                           multiplier: 1,
-                                          constant:numberOfRows * (cellSize + (cellSpacing * 2))))
+                                          constant: numberOfRows * (cellSize + (cellSpacing * 2))))
     view.backgroundColor = .white
   }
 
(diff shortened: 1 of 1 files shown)


catalog/MDCCatalog/MDCThemePickerViewController.swift
```diff
@@ -108,6 +109,10 @@ class MDCThemePickerViewController: UIViewController, UICollectionViewDataSource
     palettesCollectionView.backgroundColor = .white
     palettesCollectionView.contentInset = UIEdgeInsets(top: 0, left: 8, bottom: 0, right: 0)
     view.addSubview(palettesCollectionView)
+    let rowWidth = view.bounds.width - cellSpacing * 2
```
ianegordon, line 129: super-(super!)-nit: Ideally we would use CGRectGetWidth(view.frame) to avoid non-standardized weirdness.

ianegordon (approved):
LGTM.

catalog/MDCCatalog/MDCThemePickerViewController.swift
romoore: Thankfully in Swift, I believe this has been fixed.

ianegordon: Indeed it has!  Disregard,.... and celebrate Swift.

romoore: Here's the API refs.

https://developer.apple.com/documentation/coregraphics/cgrect/1454758-width

yarneo (approved):

romoore:
The cocoapods/podspec job succeeded internally but never reported its progress. Retriggering.

PR #4491, opened 2018-07-09
[TextFields] Make errorText span multiple lines on MDCTextField

andrewoverton:
Before:
simulator screen shot - iphone 8 plus - 2018-07-09 at 13 58 27

After:
simulator screen shot - iphone 8 plus - 2018-07-09 at 13 48 31

Closes https://github.com/material-components/material-components-ios/issues/4476

### DIFF
--- a/components/TextFields/src/private/MDCTextInputCommonFundament.m
+++ b/components/TextFields/src/private/MDCTextInputCommonFundament.m
@@ -317,8 +317,6 @@ static inline UIColor *MDCTextInputUnderlineColor() {
   _trailingUnderlineLabel.opaque = NO;
   [_textInput addSubview:_trailingUnderlineLabel];
 
-  UILayoutPriority defaultPriority = UILayoutPriorityDefaultHigh;
-
   _leadingUnderlineLeading = [NSLayoutConstraint constraintWithItem:_leadingUnderlineLabel
                                                           attribute:NSLayoutAttributeLeading
                                                           relatedBy:NSLayoutRelationEqual
@@ -326,7 +324,7 @@ static inline UIColor *MDCTextInputUnderlineColor() {
                                                           attribute:NSLayoutAttributeLeading
                                                          multiplier:1
                                                            constant:0];
-  _leadingUnderlineLeading.priority = defaultPriority;
+  _leadingUnderlineLeading.priority = UILayoutPriorityDefaultLow;
 
   _trailingUnderlineTrailing = [NSLayoutConstraint constraintWithItem:_trailingUnderlineLabel
                                                             attribute:NSLayoutAttributeTrailing
@@ -335,20 +333,7 @@ static inline UIColor *MDCTextInputUnderlineColor() {
                                                             attribute:NSLayoutAttributeTrailing
                                                            multiplier:1
                                                              constant:0];
-  _trailingUnderlineTrailing.priority = defaultPriority;
-
-  NSLayoutConstraint *labelSpacing =
-      [NSLayoutConstraint constraintWithItem:_leadingUnderlineLabel
-                                   attribute:NSLayoutAttributeTrailing
-                                   relatedBy:NSLayoutRelationEqual
-                                      toItem:_trailingUnderlineLabel
-                                   attribute:NSLayoutAttributeLeading
-                                  multiplier:1
-                                    constant:0];
-  labelSpacing.priority = defaultPriority;
-
-  [NSLayoutConstraint
-      activateConstraints:@[ labelSpacing, _leadingUnderlineLeading, _trailingUnderlineTrailing ]];
+  _trailingUnderlineTrailing.priority = UILayoutPriorityDefaultLow;
 
   NSLayoutConstraint *leadingTop =
   [NSLayoutConstraint constraintWithItem:_leadingUnderlineLabel
@@ -357,8 +342,8 @@ static inline UIColor *MDCTextInputUnderlineColor() {
                                   toItem:_underline
                                attribute:NSLayoutAttributeBottom
                               multiplier:1
-                                constant:5];
-  leadingTop.priority = defaultPriority;
+                                constant:MDCTextInputHalfPadding];
+  leadingTop.priority = UILayoutPriorityDefaultLow;
 
   NSLayoutConstraint *leadingBottom =
       [NSLayoutConstraint constraintWithItem:_textInput
@@ -367,8 +352,8 @@ static inline UIColor *MDCTextInputUnderlineColor() {
                                       toItem:_leadingUnderlineLabel
                                    attribute:NSLayoutAttributeBottom
                                   multiplier:1
-                                    constant:5];
-  leadingBottom.priority = defaultPriority;
+                                    constant:MDCTextInputHalfPadding];
+  leadingBottom.priority = UILayoutPriorityDefaultLow;
 
   NSLayoutConstraint *trailingBottom =
       [NSLayoutConstraint constraintWithItem:_trailingUnderlineLabel
@@ -378,15 +363,15 @@ static inline UIColor *MDCTextInputUnderlineColor() {
                                    attribute:NSLayoutAttributeBottom
                                   multiplier:1
                                     constant:0];
-  trailingBottom.priority = defaultPriority;
+  trailingBottom.priority = UILayoutPriorityDefaultLow;
 
-  [NSLayoutConstraint activateConstraints:@[ leadingTop, leadingBottom, trailingBottom ]];
+  NSArray *constraints = @[ _leadingUnderlineLeading, _trailingUnderlineTrailing, leadingBottom,
+      trailingBottom, leadingTop ];
+  [NSLayoutConstraint activateConstraints:constraints];
 
   // When push comes to shove, the leading label is more likely to expand than the trailing.
   [_leadingUnderlineLabel setContentHuggingPriority:UILayoutPriorityDefaultLow - 1
                                             forAxis:UILayoutConstraintAxisHorizontal];
-  [_leadingUnderlineLabel setContentCompressionResistancePriority:UILayoutPriorityDefaultLow
-                                                          forAxis:UILayoutConstraintAxisHorizontal];
 
   [_trailingUnderlineLabel
       setContentCompressionResistancePriority:UILayoutPriorityRequired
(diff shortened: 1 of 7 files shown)


andrewoverton:
@willlarche, getting rid of the lines that set the `leadingUnderlineLabel`'s constraints' priorities to low caused `testConstraintPriorities` to fail... There's a note above that test case saying "All the constraints created internally by the MDCTextField need to have a rather low priority so they can be overriden by a controller from the outside". Is that still true? Couldn't a controller just set such constraints to be required, rather than high?

willlarche:
The input elements themselves are designed to lay their subviews out with lower priority constraints. That way a controller can always come in and override the constraints to give it a custom layout.

Looks like the most I ever thought that would be is 260. So, if you are changing constraints of the classes `MDCTextField` or `MDCMultilineTextField`, keep them less than 260 and when you make constraints from a controller, I like to start with 750 and go up. But required works too if your logic doesn't need any hierarchy of priority.

andrewoverton:
Ah, you know what @willlarche , I mistakenly believed that constraint priorities were high by default, rather than required. So yep, my earlier comment doesn’t make sense. But what if we just set the constraints to be high here, as I did in 701ee35? That way an external controller would still be able to override them with required constraints, but the internal constraints would be prioritized enough to successfully lay out a multiline leading underline label.

Here is a screenshot with priorities at 260, the value you recommended I not exceed:
simulator screen shot - iphone 8 plus - 2018-07-10 at 10 03 47

Here is a screenshot with priorities at 750/“default high”:
simulator screen shot - iphone 8 plus - 2018-07-10 at 10 03 19

You can also see I set numberOfLines to be 2 instead of 0, as you told me offline that design said it shouldn't exceed two.

ivangodfather:
this will apply also for .helperText ? It has the same issue

willlarche:
@ivangodfather yes it will apply for helper text too!

willlarche (dismissed):
Spoke offline.

ianegordon:
@andrewoverton if this has been addressed offline, please merge.  If not please re-add Will to give approval when he gets back.
Many Thanks!

willlarche:
Then let’s close this. Andrew is working on changes.

andrewoverton:
Ok I'll close while I work on it 👍

andrewoverton:
simulator screen shot - iphone x - 2018-07-17 at 15 10 30

This is a screenshot with my most recent changes @willlarche :D

As we discussed, there is still this bug in the KitchenSink example where the TextFields' bounds change, both in this branch and develop. It is not yet clear whether or not this is a bug in the TextField itself or a consequence of the way this particular ViewController is built.

components/TextFields/src/private/MDCTextInputCommonFundament.m
```diff
@@ -384,14 +388,16 @@ - (void)setupUnderlineLabels {
   _trailingUnderlineLabel.opaque = NO;
   [_textInput addSubview:_trailingUnderlineLabel];
 
+  UILayoutPriority defaultPriority = UILayoutPriorityDefaultHigh;
```
willlarche: We need to find a way to do this with constraints priorities below 260 in the input.

willlarche (dismissed):

components/TextFields/src/private/MDCTextInputCommonFundament.m
andrewoverton: @willlarche, we prefer lower priorities because we need the controller to be able to override the constraints, correct? Can't a controller still override a constraint with a priority of 750 if it needs to?

It just seems to me that these constraints, which are there to keep the labels' contents inside the view bounds, will fail to do so if their priorities are too low and the text grows too long. That's actually what I would _expect_ to happen... 

When would a controller want to overpower these constraints, which seem to me to serve a vital purpose?
Would you be happier with an approach that adds fixed width constraints to the labels? I haven't tried it yet, but it might work as well...

Thank you for your attention/assistance!

apellegrino91:
I have the same problem. How i can fix it?

willlarche:
@apellegrino91 , this is currently in development. You'll have to wait a few weeks for it to be resolved.

andrewoverton:
Here are the latest changes @willlarche:

simulator screen shot - iphone x - 2018-07-23 at 16 29 33

willlarche:
Found a couple bugs doing a quick run on my machine:

1. The trailing label has the wrong vertical placement. It should probably have its top pinned to the leading label if the leading label defines the vertical space.

screen shot 2018-07-24 at 10 24 42 am
screen shot 2018-07-24 at 10 26 13 am


2. There's a lot of new ambiguous layout:

screen shot 2018-07-24 at 10 24 46 am

andrewoverton:
@willlarche I addressed the ambiguous layout warnings and moved the trailing label to be pinned to the underline rather than the textInput bottom

willlarche:
Great! Found another bug.

Text areas are not returning the correct intrinsic content size. The error label's second line is out of bounds. Tested it with all the text inset modes.
simulator screen shot - iphone x - 2018-07-26 at 16 03 23

screen shot 2018-07-26 at 4 03 14 pm

willlarche:
Also, please add the ability to see multi-line errors into the kitchen sink controller and a test or two around the expected height.

andrewoverton:
Omg! 😱I didn't notice that one. I just copied some of the textInset logic from the controller I had been working with to MDCTextInputControllerOutlinedTextArea. Here's a screenshot:
simulator screen shot - iphone x - 2018-07-27 at 13 58 08
That work is in 1541a94.

I made the requested changes to the kitchen sink example in 0a9bfc2.
And I added the test in a700b90. It's a little weird... I think it is doing what it's supposed to do though. It failed before 1541a94 and now it passes.

willlarche:
Oh, be sure to add some long helper text too, for testing. It gets this 2 line feature for free.
simulator screen shot - iphone x - 2018-07-27 at 11 28 07

willlarche:
Found another bug in the custom fonts example:
simulator screen shot - iphone x - 2018-07-27 at 11 47 32
screen shot 2018-07-27 at 11 48 06 am

components/TextFields/tests/unit/TextFieldTests.swift
```diff
@@ -178,19 +178,27 @@ class TextFieldTests: XCTestCase {
     let controller1 = MDCTextInputControllerFilled(textInput: textField1)
     var errorMessage = "Error!"
     controller1.setErrorText(errorMessage, errorAccessibilityValue: nil)
+    textField1.setNeedsLayout()
```
willlarche: Do you have to have this call? Is just `layoutIfNeeded` not enough?

willlarche:
Same in the filled example
simulator screen shot - iphone x - 2018-07-27 at 11 53 48

andrewoverton:
@willlarche , I didn't see that behavior in the custom font or filled example:

simulator screen shot - iphone x - 2018-07-27 at 15 16 27
simulator screen shot - iphone x - 2018-07-27 at 15 12 07

andrewoverton:
So bizarre... The tests were passing locally but failing in CI, so I chose one of the Xcode versions that was failing in CI (9.1), downloaded that version, got tests passing locally, pushed that up, and see now that it failed in CI again :/

jverkoey:
@andrewoverton  This PR is likely a good candidate for doing some smoke testing internally prior to landing in a release.

willlarche:
Also, I just realized we need to make sure you can opt-out of this and have only 1 line. Hopefully that's as easy as changing the number of lines the label can do.

components/TextFields/src/private/MDCTextInputCommonFundament.m
```diff
@@ -344,26 +348,37 @@ - (void)setupUnderlineLabels {
...
 
-  [NSLayoutConstraint activateConstraints:@[ leadingBottom, trailingBottom ]];
+  NSLayoutConstraint *trailingTop =
+  [NSLayoutConstraint constraintWithItem:_underline
+                               attribute:NSLayoutAttributeBottom
+                               relatedBy:NSLayoutRelationEqual
+                                  toItem:_trailingUnderlineLabel
+                               attribute:NSLayoutAttributeTop
+                              multiplier:1
+                                constant:-MDCTextInputHalfPadding];
```
willlarche: I prefer a `-1 * [whatever]` for negating constants. It's easy to miss the hyphen by itself.

andrewoverton:
Through a combination of setting the content compression resistance on the leading underline label AND adding a low priority width constraint I got truncation working with 1 line of text. That's in 350dc1b. I also made the default 1 line of text. That was a judgment call... I can set it back to two though.

willlarche:
Something seems to be inconsistent with the width constraint. This is what I see:

simulator screen shot - iphone 8 plus - 2018-08-14 at 18 26 31

andrewoverton:
@willlarche omg you'd posted a screenshot just like that 18 days ago and I fixed it and I guess I broke it in a commit since then? So finicky 😩Anyway, I assume what you're doing there is looping through the controllers in viewWillAppear and setting error text. I was just able to reproduce it doing that. Wasn't happening if I added the code in viewDidAppear, just viewWillAppear. Also, the layout would get fixed if I made the text field become the first responder. This makes it seem like it's just waiting on a layout pass that's not coming. Moving the setNeedsLayout/layoutSubviews call below the call to updateLayout in da1a6b7 fixes it.

Since it's already happened like 5 times on this PR I wouldn't be surprised if there are some other bugs lurking but I think we should just land this so I can deal with those as they may come. Something about having this thing sit around so long bothers me. Would you be okay with that?

willlarche:
No. If anything it’s proven this PR needs more rigorous testing. It’s far too dangerous at the moment. 

Let’s slow down and make sure everything is perfect.

andrewoverton:
Okay sounds good. Well hopefully da1a6b7 does the trick for now :)

jverkoey:
What's the status of this PR? If it's blocked please close it and re-open once it's ready to pursue again.

andrewoverton:
@willlarche and I talked about it and I think we're going to merge it into a feature branch and merge that into develop _eventually_ 😕

willlarche (approved):

adrianod1as:
Is this (or will be) available in any branch?

romoore:
I believe this was never committed in develop. The work may be prioritized in the future, but at the moment it isn't at the top of our backlog.

PR #4494, opened 2018-07-10
[Codelabs] Add build tests for codelabs

michdud:
Adds build tests for all iOS codelabs.


ianegordon:
Hi @michdud , this is excellent.  But, there appears to be a path issue when I run from the git root directory.

```
$ ./scripts/build_codelabs -v
./scripts/build_codelabs: line 60: cd: external: No such file or directory
```

We may need to base the path on the git root if it is available.
```
git rev-parse --show-toplevel
```

ianegordon (approved):
Thank you @michdud !

PR #4498, opened 2018-07-11
[TextFields] Add accessibility docs for MDCTextField

### LINKED ISSUE #4325: [TextFields] Publish accessibility guidance in readme
Guidance on how clients should handle MDCTextField accessibility should be added to its readme.

andrewoverton:
Adds accessibility docs for MDCTextFields.
There are no code examples because my understanding is that the client doesn't actually have to do anything in most cases to achieve adequate accessibility with VoiceOver.
Closes #4325.

I'm a little concerned by how much was deleted from the component's readme when I ran the generate readme script.

### DIFF
--- a/components/TextFields/README.md
+++ b/components/TextFields/README.md
@@ -20,7 +20,7 @@ Text fields allow users to input text into your app. They are a direct connectio
   <img src="docs/assets/textfields.png" alt="Text Fields" width="375">
 </div>
 
-MDC's text fields come in several styles and have a great range of customization. Google's UX research has determined that Outlined and Filled (aka 'text box') styles perform the best by a large margin. So use  `MDCTextInputControllerFilled`, `MDCTextInputControllerOutlined`, and `MDCTextInputControllerOutlinedTextArea` if you can and set colors and fonts that match your company's branding.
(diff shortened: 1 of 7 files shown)


components/TextFields/README.md
```diff
@@ -150,65 +154,33 @@ This is a multi-line text input. It's subclassed from `UIView` with an embedded
   <li class="icon-list-item icon-list-item">Minimum number of lines</li>
 </ul>
 
-### Text Input Controller Classes: Recommended
```
codeman7: Do we no longer need this?

ianegordon (approved):

ianegordon:
Please consider Cody's question before merging.

components/TextFields/README.md
andrewoverton: Not sure. The generate_readme script got rid of it though... May wait for @jverkoey to get back to get his take on this.

andrewoverton:
As @codeman7 pointed out, after I ran the generate_readme script a bunch got deleted from the component's main readme. @jverkoey thinks this is because the component's readme had been manually edited in the past instead of using the script. So in c83461e what I tried to do is retroactively update the "sub-readmes" used to generate the main one to reflect what had been manually added to the main one.

Issue #3187, opened 2018-03-29
[Chips] "Clear" button in ChipsTypicalUseViewController has no accessibilityHint


romoore:
### Overview
The "Clear" button in the navigation bar of the ChipsTypicalUseViewController has no accessibilityHint value. It's not clear what it would clear.  Perhaps it should also be dimmed if there is no selection to clear.

### Reproduction steps
1. Open Catalog&rarr;Chips&rarr;Demo
1. Use VoiceOver to select the "Clear" Button near the beginning of the screen

**Expected Result**
Clear button is only enabled when it can perform an action.
Selecting it will read the hint, "Unselects all chips"

**Actual Result**
Clear button is always enabled.
It has no `accessibilityHint`.

### Version number
50.0.0

andrewoverton:
Closed by https://github.com/material-components/material-components-ios/pull/4504

PR #4504, opened 2018-07-11
[Chips] Add accessibilityHint to clear button in chips example

andrewoverton:
Adds an `accessibilityHint` to the "Clear" nav bar button in the `ChipsTypicalUseViewController` and makes it enabled only when there are selected items present.
Closes issue #3187


romoore (approved):
Looks good.  Would you please post a screenshot of the newly-disabled "Clear" button in the commit summary?

andrewoverton:
I had been testing with Dragons, and I found that in Catalog the disabled button looked very dark, so I added 81a32e1, which makes it look nice in both.
Here's the disabled state:
simulator screen shot - iphone 8 - 2018-07-11 at 17 15 49
And here's the enabled state:
simulator screen shot - iphone 8 - 2018-07-11 at 17 15 48

PR #4510, opened 2018-07-11
[BottomNavigation] Give MDCBottomNavigationBar a TabBar accessibilityTrait

### LINKED ISSUE #4499: [BottomNavigation] MDCBottomNavigationBar should use  UIAccessibilityTraitTabBar
## Feature Request/Suggestion
MDCBottomNavigationBar should probably have the trait `UIAccessibilityTraitTabBar` in its container.

## Additional context

```objectivec
/*
 Used when a view or accessibility container represents an ordered list of tabs.
 The object with this trait should return NO for isAccessibilityElement.
 */
UIKIT_EXTERN UIAccessibilityTraits UIAccessibilityTraitTabBar NS_AVAILABLE_IOS(10_0);
```

andrewoverton:
Closes #4499

### DIFF
--- a/components/BottomNavigation/src/MDCBottomNavigationBar.m
+++ b/components/BottomNavigation/src/MDCBottomNavigationBar.m
@@ -191,9 +191,11 @@ static NSString *const kMDCBottomNavigationBarOfAnnouncement = @"of";
                                      UIViewAutoresizingFlexibleRightMargin);
   _containerView.clipsToBounds = YES;
   [self addSubview:_containerView];
-  if (@available(iOS 10.0, *)) {
+#if defined(__IPHONE_10_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_10_0)
+  if (&UIAccessibilityTraitTabBar != NULL) {
     _containerView.accessibilityTraits = UIAccessibilityTraitTabBar;
   }
+#endif
   [self setElevation:MDCShadowElevationBottomNavigationBar];
   _itemViews = [NSMutableArray array];
   _itemTitleFont = [UIFont mdc_standardFontForMaterialTextStyle:MDCFontTextStyleCaption];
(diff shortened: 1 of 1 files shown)


ianegordon:
I don't think the guards are quite right.  This appears to be failing on iOS 9 and earlier.

ianegordon (approved):

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -191,6 +191,9 @@ - (void)commonMDCBottomNavigationBarInit {
                                      UIViewAutoresizingFlexibleRightMargin);
   _containerView.clipsToBounds = YES;
   [self addSubview:_containerView];
+#if defined(__IPHONE_10_0) && (__IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_10_0)
+  _containerView.accessibilityTraits = UIAccessibilityTraitTabBar;
```
romoore: I think we should allow this to be used on iOS 10+ today instead of waiting to drop iOS 9 support in Xcode.   Could we switch to something like this?

```objc
if (&UIAccessibilityTraitTabBar != NULL) {
  _containerView.accessibilityTraits = UIAccessibilityTraitTabBar;
}
#endif
```

romoore (dismissed):

romoore:
@ianegordon The guards looked OK for me and nothing was failing. Maybe just a flake in kokoro.

components/BottomNavigation/src/MDCBottomNavigationBar.m
andrewoverton: @romoore I didn't see this comment and just changed it to use `if @available(iOS 10.0, *)` instead. Is that alright?

andrewoverton: Ah, I see why this doesn't work now. I'll use the NULL check approach, thanks @romoore !

PR #4534, opened 2018-07-12
[Slider] Support custom increment and decrement levels

### LINKED ISSUE #4507: [Slider] Support custom increment/decrement levels
Currently we don't allow clients to have custom increment/decrement levels within VoiceOver

## Expected behavior

A client should be able to set these values through a property.

## Actual behavior

We currently have this as a constant.

codeman7:
Closes #4507

### DIFF
--- a/components/Slider/src/MDCSlider.h
+++ b/components/Slider/src/MDCSlider.h
@@ -292,6 +292,20 @@ IB_DESIGNABLE
  */
 @property(nonatomic, assign, getter=isThumbHollowAtStart) BOOL thumbHollowAtStart;
 
+/**
+Tells the accessibility element what amount to increment the value of its content.
+
+ Default increment level is 10%.
+ */
+@property(nonatomic) CGFloat accessibilityIncrementAmount;
+
+/**
+Tells the accessibility element what amount to decrement the value of its content.
+
+ Default decrement level is 10%.
+ */
+@property(nonatomic) CGFloat accessibilityDecrementAmount;
+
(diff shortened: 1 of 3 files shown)


components/Slider/src/MDCSlider.h
```diff
@@ -292,6 +292,20 @@ IB_DESIGNABLE
  */
 @property(nonatomic, assign, getter=isThumbHollowAtStart) BOOL thumbHollowAtStart;
 
+/**
+Tells the accessibility element what amount to increment the value of its content.
+
+ Default increment level is 10%.
+ */
+@property(nonatomic) CGFloat accessibilityIncrementAmount;
```
romoore: I don't see anything like this on UISlider, and I believe users can activate the slider to get smaller adjustments back toward the initial value. Is there a reason we think this API is needed specifically in MDCSlider?

components/Slider/src/MDCSlider.m
```diff
@@ -478,6 +478,24 @@ - (void)accessibilityDecrement {
   }
 }
 
+- (BOOL)accessibilityActivate {
+  CGFloat midPoint = (self.maximumValue - self.minimumValue) / 2.0;
+  CGFloat newValue;
+  CGFloat adjustmentAmount = fabs((self.value - midPoint) / 3.0);
```
romoore: Should we set some minimum adjustment amount? It seems that UISlider won't change when it's less than ~1.5%

romoore (approved):

PR #4535, opened 2018-07-12
[Buttons] Disable FAB animation for VoiceOver

romoore:
When VoiceOver is running and the FAB animations for expanding/collapsing are
triggered, sometimes VoiceOver won't detect the FAB. It can still detect the
FAB if the user "scrubs" their finger to discover it, but it's a big problem
if they don't know they're missing a button.  The effect seems intermittent
and this is the best work-around.


andrewoverton (approved):
LGTM

PR #4542, opened 2018-07-12
[Buttons] Scrub internal doc links

randallli:
[Button] Scrub internal doc links


jverkoey (changes requested):
This PR is pointed against stable. Was that intentional?

randallli:
PR is against stable because of an artifact of our export from internal code. We have instructions to manually change in the UI after the PR is created.

randallli:
rebased and recreated with: https://github.com/material-components/material-components-ios/pull/4559

PR #4558, opened 2018-07-13
[BottomNavigation] Correct misnamed Material guidelines references

randallli:


romoore (approved):
While we're editing these files, please use the term "guidelines" instead of "spec" to refer to the Material guidelines site.

romoore:
Updated the PR title since this doesn't seem to be changing any links.

PR #4564, opened 2018-07-13
[ActivityIndicator] Change override of AccessibilityLabel with setting it to a default.

randallli:
[ActivityIndicator] Change override of AccessibilityLabel with setting it to a default.


romoore (approved):
Do we have a client-facing bug as the reason to make this change? Can you please include the bug link in the commit summary?

randallli:
No this is so we can give good recommendations for accessibility. We want to tell clients to make a descriptive label. If they set one its currently ignored

PR #4583, opened 2018-07-17
CODEOWNERS - Suppress @yarneo on NavigationBar PRs

romoore:
@yarneo does not want to get automatically added to all PRs touching the NavigationBar component, but is still an owner of that component. This will leave @jverkoey as the auto-assigned owner for changes to NavigationBar.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -38,7 +38,7 @@
 /components/LibraryInfo/          @ajsecord
 /components/MaskedTransition/     @jverkoey
 # /components/NavigationBar/        @yarneo 
-components/NavigationBar/         @jverkoey
+/components/NavigationBar/         @jverkoey
 /components/OverlayWindow/        @yarneo
 /components/PageControl/          @randallli
 /components/Palettes/             @ajsecord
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -9,6 +9,13 @@
 # will be requested to review.
 # *.js    @octocat @github/js
 
+# !!!! IMPORTANT !!!!
+# If a component is commented-out with a current contributor/team member
+# it means that the person still "owns" the component but does not want
+# to be automatically added as a reviewer to all PRs touching that component.
```
ianegordon, line 15: I would expect an OWNER to be reviewing all incoming PRs against their components.
This comment is not likely to be visible to people opening PRs.

ianegordon (dismissed):

CODEOWNERS
romoore: This comment is to prevent people from removing commented-out owners. 

We are currently discussing the issue of listed owners reviewing every (possibly trivial) change to their component. If you would like to make that discussion public, perhaps we can open an Issue and debate it there?

randallli (approved):

ianegordon:
We are going to hold off on this change pending some material-automation magic.  We can re-open if necessary.

romoore:
@ianegordon What automation magic?

PR #4617, opened 2018-07-23
[BottomSheet] Replacing typeof(self) by __typeof(self)

jeromelebel:
typeof(self) is not supported by Chromium iOS compiler.

Relates to cl/205664591


jeromelebel:
I signed it!

jverkoey (approved):

PR #4399, opened 2018-06-15
[Chips] Accessibility selected & dimmed states.

galiak11:
Accessibility reads selected Chips as 'selected' and disabled Chips as 'dimmed'.
Closes b/79691548 & b/79692379.

PR #4664, opened 2018-07-31
[Chips] Allow clients to set `accessibilityLabel` on MDCChipView

compositeprimes:
Allow clients to set `accessibilityLabel` on MDCChipView, instead of ignoring any value they set.

This was broken by: #4399


romoore (approved):
Looks good. Please add unit tests for the 3 use cases:

1. Using the explicitly set `accessibilityLabel` value
2. Using `textLabel.accessibilityLabel`
3. Using `textLabel.text`

compositeprimes:
Added unit tests

romoore:
I cleaned up the unit tests a bit, thanks @compositeprimes !

PR #4676, opened 2018-07-31
[ActionSheet] API Review

codeman7:
This is just for a review of the API. A design doc should go out ASAP to follow.


components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,158 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+@class MDCActionSheetAction;
+
+@interface MDCActionSheetController : UIViewController
```
jverkoey: Need docs for the class as well. This should:

- [x] Describe the class' purpose.
- [x] Include a link to the Material spec.
- [x] Include a link to any similar UIKit APIs for reference.
- [x] Describe any notable differences from the UIKit APIs.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,158 @@
...
+ */
++ (nonnull instancetype)actionSheetControllerWithTitle:(nullable NSString *)title;
+
+/**
+ Convenience constructor to create and return a view controller for displaying an alert to the user.
+
+ After creating the alert controller, add actions to the controller by calling -addAction.
+
+ @param title The title of the alert.
+ @param message Descriptive text that summarizes a decision in a sentence of two.
```
jverkoey: [typo]: `or two`.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,158 @@
...
+ @return An initialized MDCActionSheetController object.
+ */
++ (nonnull instancetype)actionSheetControllerWithTitle:(nullable NSString *)title
+                                               message:(nullable NSString *)message;
+
+- (nonnull instancetype)initWithNibName:(NSString *)nibNameOrNil
+                                 bundle:(NSBundle *)nibBundleOrNil;
+
+- (nullable instancetype)initWithCoder:(NSCoder *)aDecoder;
```
jverkoey: I believe you don't have to redefine these APIs because they're part of the parent class, should be able to remove.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ Adds an action to the action sheet.
+
+ Actions are the possible reactions of the user to the presented alert. Actions are added as a
+ list item in the list of the action sheet.
+ Action buttons will be laid out from right to top to bottom depending on the the order they
```
jverkoey: typo: `will be laid out from top to bottom in the order they were added.`

Can drop the "first being top last being bottom" part.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ The actions that the user can take in response to the action sheet.
+
+ The order of the actions in the array matches the order in which they were added to the action sheet.
+ */
+@property (nonatomic, nonnull, readonly) NSArray<MDCActionSheetAction *> *actions;
+
+/**
+ The title of the action sheet controller.
```
jverkoey: What happens if you change this property (and the message property) after the view controller has been presented? Document whether the client can expect the title label to be updated or not.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,158 @@
...
+
+ This property is modeled after the adjustsFontForContentSizeCategory property in the
+ UIConnectSizeCategoryAdjusting protocol added by Apple in iOS 10.0.
+
+ If set to YES, this button will base its text font on MDCFontTextStyleButton.
+
+ Defaults value is NO.
+ */
+@property(nonatomic, readwrite, setter=mdc_setAdjustsFontForContentSizeCategory:)
+BOOL mdc_adjustsFontForContentSizeCategory;
```
jverkoey: [stylenit]: Indent four spaces.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,158 @@
...
+
+@end
+
+/**
+ MDCActionHandler is a block that will be invoked when the action is selected.
+ */
+typedef void (^MDCActionSheetHandler)(MDCActionSheetAction *_Nonnull action);
+
+/**
+  MDCActionSheetAction is passed to MDCActionSheetController to add a button to the action sheet.
```
jverkoey: [stylenit] Here and below, there are two spaces indenting the docs when there should just be one.

jverkoey: "An instance of MDCActionSheetAction is passed..."

The client isn't expected to pass the MDCActionSheetAction class itself.

jverkoey: "to add an action to the action sheet".

It's not necessarily guaranteed that the actions are buttons.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,158 @@
...
+ */
+@interface MDCActionSheetAction : NSObject <NSCopying>
+
+
+/**
+  Action alerts control the list items that will be displayed on the bottom of an action
```
jverkoey: Not sure if this sentence is as-intended? Did you mean something like "Returns an action sheet action populated with the given values"?

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,158 @@
...
+ sheet controller.
+
+ @param title The title of the list item shown on the list
+ @param image The icon of the list item shown on the list
```
jverkoey: [minor] `in the list` rather than `on the list`.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,158 @@
...
+ */
++ (nonnull instancetype)actionWithTitle:(nonnull NSString *)title
+                                  image:(nonnull UIImage *)image
+                                handler:(__nullable MDCActionSheetHandler)handler;
+
+/** Action sheet actions must be created with actionWithTitle:image:handler: */
+- (nonnull instancetype)init NS_UNAVAILABLE;
+
+/**
+ Title of the cell shown on the action sheet.
```
jverkoey: Remove any reference to `cell` in the public docs - this term leaks an implementation detail that could change in the future.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,158 @@
...
+  Action of the cell shown on the action sheet.
+
+  Action sheet actions must have an action taht will be set within
+    actionWithTitle:image:handler: method.
+*/
+@property (nonatomic, nonnull, readonly) MDCActionSheetHandler action;
+
+@end
+
+@protocol MDCActionSheetControllerDelegate <NSObject>
```
jverkoey: Docs for the protocol should:

- Describe that this protocol is used by the MDCActionSheetController to delegate events.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ Called when the user taps the dimmed background or swipes the action sheet off to dismiss
+ the action sheet. This method is not called if the action sheet is dismissed
+ programatically.
```
jverkoey: Is this called if the user dismisses the sheet by selecting an action?

jverkoey:
Looking good! Most feedback is just grammatical/missing docs.

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h
ianegordon: We should match (and document) the behavior of UIAlertController in ActionSheet mode.

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h
ianegordon: I would mark initWithCoder: and initWithNibName:bundle NS_UNAVAILABLE.  I don't expect we would support either.

ianegordon (dismissed):

components/ActionSheet/src/MDCActionSheetController.h
codeman7: We will update the view to reflect the new string.

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

jverkoey:
Spoke offline with @ianegordon: am in agreement that we can mark the initWithNib/initWithCoder APIs as unavailable at this point. We can explore adding support for interface builder as a follow-up in the future.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,153 @@
...
+ Convenience constructor to create and return a view controller for displaying an alert to the user.
+
+ After creating the alert controller, add actions to the controller by calling -addAction.
+
+ @param title The title of the alert.
+ @param message Descriptive text that summarizes a decision in a sentence or two.
+ @return An initialized MDCActionSheetController object.
+ */
++ (nonnull instancetype)actionSheetControllerWithTitle:(nullable NSString *)title
+                                               message:(nullable NSString *)message;
```
ianegordon: Mark as designated initializer.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ Convenience constructor to create and return a view controller for displaying an alert to the user.
+
+ After creating the alert controller, add actions to the controller by calling -addAction.
+
+ @param title The title of the alert.
+ @return An initialized MDCActionSheetController object.
+ */
++ (nonnull instancetype)actionSheetControllerWithTitle:(nullable NSString *)title;
```
ianegordon: This should *not* be a designated initializer, and in the implementation should call:
[self actionSheetControllerWithTitle:title message:nil]

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,153 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+@class MDCActionSheetAction;
+
+@interface MDCActionSheetController : UIViewController
```
ianegordon: Annotate as final to avoid subclasses.
See #2801 for an example of how to do this.

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,175 @@
...
+
+ After creating the alert controller, add actions to the controller by calling -addAction.
+
+ @param title The title of the alert.
+ @return An initialized MDCActionSheetController object.
+ */
++ (nonnull instancetype)actionSheetControllerWithTitle:(nullable NSString *)title;
+
+
+- (nonnull instancetype)init;
```
ianegordon: This is not required in the header, but in the implementation should call:
[self actionSheetControllerWithTitle:nil message:nil];

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,179 @@
...
+#endif
+#endif  // #ifndef MDC_SUBCLASSING_RESTRICTED
+
+@class MDCActionSheetAction;
+
+/**
+ MDCActionSheetController displays an alert message to the user, similar to
+     UIAlertControllerStyleActionSheet.
+
+ A Material Action Sheet consist of a title, message and a list of actions.
```
ajsecord: "consists"

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,179 @@
...
+
+ Learn more at the [Material spec bottom
```
ajsecord: "Learn more about ...", also just use "Material bottom sheet", no need for the extra "spec".

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,179 @@
...
+
+ ### Dependencies
+
+ MDCActionSheetController depends on BottomSheet Component.
+ */
+MDC_SUBCLASSING_RESTRICTED
+@interface MDCActionSheetController : UIViewController
+
+/**
+ Designated constructor to create and return a view controller for displaying an alert to the user.
```
ajsecord: Replaced "constructor" with "initializer"

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,179 @@
...
+
+ @param title The title of the alert.
+ @param message Descriptive text that summarizes a decision in a sentence or two.
+ @return An initialized MDCActionSheetController object.
+ */
++ (nonnull instancetype)actionSheetControllerWithTitle:(nullable NSString *)title
+                                               message:(nullable NSString *)message;
+
+/**
+ Convenience constructor to create and return a view controller for displaying an alert to the user.
```
ajsecord: Replace "constructor" with "initializer"

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,179 @@
...
+/** Action sheet controllers must be created with actionSheetControllerwithTitle:
+   or with actionSheetControllerWithTitle:message:  */
+- (nullable instancetype)initWithCoder:(NSCoder *)aDecoder NS_UNAVAILABLE;
+
+/**
+ Adds an action to the action sheet.
+
+ Actions are the possible reactions of the user to the presented alert. Actions are added as a
+ list item in the list of the action sheet.
+ Action buttons will be laid out from top to bottom depending on the the order they
```
ajsecord: "the the"

ianegordon (approved):
LGTM.

ajsecord (approved):
LGTM with nits

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,179 @@
...
+#else
+#define MDC_SUBCLASSING_RESTRICTED
+#endif
+#endif  // #ifndef MDC_SUBCLASSING_RESTRICTED
+
+@class MDCActionSheetAction;
+
+/**
+ MDCActionSheetController displays an alert message to the user, similar to
+     UIAlertControllerStyleActionSheet.
```
jverkoey: [minorstyle] Remove the extra indentation here.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,179 @@
...
+/**
+ Returns an action sheet action with the populated given values.
+
+ @param title The title of the list item shown in the list
+ @param image The icon of the list item shown in the list
+ @param handler A block to execute when the user selects the action.
+ @return An initialized MDCActionSheetAction object.
+ */
++ (nonnull instancetype)actionWithTitle:(nonnull NSString *)title
+                                  image:(nonnull UIImage *)image
```
andrewoverton: The Apple version has a parameter for style, like "default", "cancel", or "destructive". Should we have our own versions of these?

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,179 @@
...
+
+ After creating the alert controller, add actions to the controller by calling -addAction.
+
+ @param title The title of the alert.
+ @return An initialized MDCActionSheetController object.
+ */
++ (instancetype)actionSheetControllerWithTitle:(nullable NSString *)title;
+
+
+- (instancetype)init;
```
andrewoverton: Does this need to be here?

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,179 @@
...
+
+/** Action sheet controllers must be created with actionSheetControllerWithTitle: or
+ with actionSheetControllerWithTitle:message:  */
+- (nonnull instancetype)initWithNibName:(NSString *)nibNameOrNil
+                                 bundle:(NSBundle *)nibBundleOrNil NS_UNAVAILABLE;
+
+/** Action sheet controllers must be created with actionSheetControllerwithTitle:
```
andrewoverton: _Style nit:_ We should be consistent (at least within a single file) about whether to put newlines after and before the "/**" and "*/" doc delimiter things

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,179 @@
...
+
+ MDCActionSheetController does not support UIPopoverController, instead it will always be presented
+ in a sheet from the bottom.
+
+ ### Dependencies
+
+ MDCActionSheetController depends on BottomSheet Component.
+ */
+MDC_SUBCLASSING_RESTRICTED
+@interface MDCActionSheetController : UIViewController
```
andrewoverton, line 51: Did we deem [text field stuff](https://developer.apple.com/documentation/uikit/uialertcontroller/1620093-addtextfield) overly complex for a first pass? Or are we just not concerned about it for this? Now that I think about it I don't think I've ever seen a text field in an action sheet, just an alert... 🤔

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h
codeman7: I think Apple uses this to render the actions differently, I believe all of our actions would render the same. But, if the team feels differently I'd love know that.

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h
codeman7: I think Action Sheet will just be a list of items.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,182 @@
...
+/**
+ MDCActionSheetActionHandler is a block that will be invoked when the action is selected.
+ */
+typedef void (^MDCActionSheetHandler)(MDCActionSheetAction *_Nonnull action);
+
+/**
+ An instance of MDCActionSheetAction is passed to MDCActionSheetController to add an
+ action to the action sheet.
+ */
+@interface MDCActionSheetAction : NSObject <NSCopying>
```
jverkoey: Consider marking this API as subclassing_restricted as well until shown that we need it otherwise.

jverkoey (approved):
LGTM!

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,182 @@
...
+
+ After creating the alert controller, add actions to the controller by calling -addAction.
+
+ @param title The title of the alert.
+ @param message Descriptive text that summarizes a decision in a sentence or two.
+ @return An initialized MDCActionSheetController object.
+ */
++ (nonnull instancetype)actionSheetControllerWithTitle:(nullable NSString *)title
+                                               message:(nullable NSString *)message;
+
```
romoore, line 64: Should there be a property to allow clients to set the dismiss/collapse button behavior in the header? From the [Material Guidelines article for Bottom Sheets](https://material.io/design/components/sheets-bottom.html#behavior), it seems like sometimes they can be dismissed (X) and other times they are just collapsed back to half (v).

Mock screenshots showing the two header icons side-by-side:
bottomsheet-dismiss-affordance-variety

bottomsheet-modal-dismiss-affordance

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,182 @@
...
+ This property is modeled after the adjustsFontForContentSizeCategory property in the
+ UIConnectSizeCategoryAdjusting protocol added by Apple in iOS 10.0.
+
+ If set to YES, this button will base its text font on MDCFontTextStyleButton.
+
+ Defaults value is NO.
+ */
+@property(nonatomic, readwrite, setter=mdc_setAdjustsFontForContentSizeCategory:)
+    BOOL mdc_adjustsFontForContentSizeCategory;
+
```
romoore, line 134: Should there be APIs for theming? From the [Material Guidelines theming section](https://material.io/design/components/sheets-bottom.html#theming) it looks like color and shape (not yet built on iOS) theming are permitted.

Suggestion:
```objective
@property(nonatomic, strong) UIColor *titleTextColor;
@property(nonatomic, strong) UIFont *titleTextFont;
@property(nonatomic, strong) UIColor *messageTextColor;
@property(nonatomic, strong) UIFont *messageTextFont;
@property(nonatomic, strong) UIColor *actionTextColor;
@property(nonatomic, strong) UIColor *actionTintColor; // for template image tinting
@property(nonatomic, strong) UIColor *inkColor; // Overlay ripple color for touch feedback
```

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,182 @@
...
+ If this is updated after presentation the view will be updated to match the new value.
+ */
+@property (nonatomic, nullable, copy) NSString *message;
+
+/*
+ Indicates whether the button should automatically update its font when the device’s
+ UIContentSizeCategory is changed.
+
+ This property is modeled after the adjustsFontForContentSizeCategory property in the
+ UIConnectSizeCategoryAdjusting protocol added by Apple in iOS 10.0.
```
romoore: Nit/spelling: s/UIConnectSizeCategoryAdjusting/UIContentSizeCategoryAdjusting/

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,182 @@
...
+ Returns an action sheet action with the populated given values.
+
+ @param title The title of the list item shown in the list
+ @param image The icon of the list item shown in the list
+ @param handler A block to execute when the user selects the action.
+ @return An initialized MDCActionSheetAction object.
+ */
++ (nonnull instancetype)actionWithTitle:(nonnull NSString *)title
+                                  image:(nonnull UIImage *)image
+                                handler:(__nullable MDCActionSheetHandler)handler;
```
romoore: What's the idea for a nullable block here? Do we have an expected use case for an action sheet with buttons that do nothing?

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,182 @@
...
+/**
+ Returns an action sheet action with the populated given values.
+
+ @param title The title of the list item shown in the list
+ @param image The icon of the list item shown in the list
+ @param handler A block to execute when the user selects the action.
+ @return An initialized MDCActionSheetAction object.
+ */
++ (nonnull instancetype)actionWithTitle:(nonnull NSString *)title
+                                  image:(nonnull UIImage *)image
```
romoore: Can the image be nullable? Are images required for action items from a design perspective or only recommended? The Design guidelines article seems to indicate that a Menu is a viable alternative on Mobile and the Google App on iOS uses menus without icons.

google_app_menu

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,182 @@
...
+ @param action Will be added to the end of MDCActionSheetController.actions.
+ */
+- (void)addAction:(nonnull MDCActionSheetAction *)action;
+
+/**
+ The actions that the user can take in response to the action sheet.
+
+ The order of the actions in the array matches the order in which they were added to the action sheet.
+ */
+@property (nonatomic, nonnull, readonly) NSArray<MDCActionSheetAction *> *actions;
```
romoore: Nit: Mark this as copy?

components/ActionSheet/src/MDCActionSheetController.h
romoore: I agree with @codeman7 here.  Since modal bottom sheets with action lists aren't expressly intended for confirmation of actions ([like a Dialog is](https://material.io/design/components/dialogs.html#usage)), I don't think we need to mark destructive/cancel actions. I would expect an action like "Delete" would result in a confirmation message or would support undo.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,182 @@
...
+
+ A Material Action Sheet consists of a title, message and a list of actions.
+
+ Learn more about [Material bottom
+ sheet](https://material.io/design/components/sheets-bottom.html)
+ [Material spec list](https://material.io/design/components/lists.html)
+
+ To learn more about
+ [UIAlertController](https://developer.apple.com/documentation/uikit/uialertcontroller)
+ or [UIAlertControllerSytleActionSheet](https://developer.apple.com/documentation/uikit/uialertcontrollerstyle/uialertcontrollerstyleactionsheet)
```
romoore: Nit: s/UIAlertControllerSytleActionSheet/UIAlertControllerStyleActionSheet/

components/ActionSheet/src/MDCActionSheetController.h
romoore: I think that would cause problems since those parameters are currently marked `nonnull` and Swifters could accidentally unwrap some non-optional nils.

components/ActionSheet/src/MDCActionSheetController.h
codeman7: I am going to add all of those API's at a later date.

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h
codeman7: Wouldn't that make a copy of the array and therefore possibly introduce bugs if this array changes and the underlying array doesn't reflect those changes?

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h
codeman7: I don't know why a client would have a button that does nothing but I think it is a possibility. I don't see a benefit in marking it as nonnull.

components/ActionSheet/src/MDCActionSheetController.h
codeman7: Pushing this back to a later date.

components/ActionSheet/src/MDCActionSheetController.h
romoore: When assigning an object from a class cluster that includes mutable variants, I believe the recommended practice is to use `copy` semantics. That prevents a caller from assigning a mutable object and changing it later.

components/ActionSheet/src/MDCActionSheetController.h
romoore: I think the benefit is to help prevent programmer errors. Unless we have a use case, I would think attaching an action that does nothing to an action sheet should be considered incorrect.

codeman7: What about an action that "dismisses" the sheet?

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h
romoore: That should be provided in the sheet's app bar (guidelines excerpt below).  Even if there were an explicit Dismiss/Cancel option, I would expect that is has an action to dismiss the action sheet.  Are you thinking that the ActionSheet would dismiss itself if any list item without an attached handler were activated?

> Full-screen modal bottom sheets contain a dismiss icon that dismisses the sheet entirely from view.
> &mdash; https://material.io/design/components/sheets-bottom.html#modal-bottom-sheet

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,183 @@
...
+ The order of the actions in the array matches the order in which they were added to the action sheet.
+ */
+@property (nonatomic, nonnull, readonly, copy) NSArray<MDCActionSheetAction *> *actions;
+
+/**
+ The title of the action sheet controller.
+
+ If this is updated after presentation the view will be updated to match the new value.
+ */
+@property (nonatomic, nullable, copy) NSString *title;
```
randallli, line 112: If the title and message are nullable why do we have a new designated initializer? 

Either make the title nonnull or allow the other initializers.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,183 @@
...
+
+/**
+ Returns an action sheet action with the populated given values.
+
+ @param title The title of the list item shown in the list
+ @param image The icon of the list item shown in the list
+ @param handler A block to execute when the user selects the action.
+ @return An initialized MDCActionSheetAction object.
+ */
++ (nonnull instancetype)actionWithTitle:(nonnull NSString *)title
```
randallli, line 158: lets mark this designated init.
and make init unavailable. This will make sure the title will always be nonnull.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,183 @@
...
+
+ MDCActionSheetController does not support UIPopoverController, instead it will always be presented
+ in a sheet from the bottom.
+
+ ### Dependencies
+
+ MDCActionSheetController depends on BottomSheet Component.
+ */
+MDC_SUBCLASSING_RESTRICTED
+@interface MDCActionSheetController : UIViewController
```
randallli, line 51: Suggestion:
If we make the public API a protocol we can land this PR without the implementation.

This will make it easier to land smaller PRs and not have to wait until the component is complete before we land anything/everything at once. It also blends nicely with "use more protocols" advice out of our internal conference.

randallli (changes requested):

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,183 @@
...
+
+/**
+ Designated initializer to create and return a view controller for displaying an alert to the user.
+
+ After creating the alert controller, add actions to the controller by calling -addAction.
```
galiak11, line 56: Is this an alert controller or an action sheet controller? or both?
Are we considering combining action sheets and alerts in the same component - similar to UIAlertController?

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -0,0 +1,183 @@
...
+/**
+ The message of the action sheet controller.
+
+ If this is updated after presentation the view will be updated to match the new value.
+ */
+@property (nonatomic, nullable, copy) NSString *message;
+
+/*
+ Indicates whether the button should automatically update its font when the device’s
+ UIContentSizeCategory is changed.
```
galiak11: only button(s)? what about title & message?

components/ActionSheet/src/MDCActionSheetController.h
codeman7: It's just an action sheet controller. I don't think they will ever combine, I think we plan to make them individual components.

PR #4755, opened 2018-08-09
[Buttons] Rename MDCFloatingButton category to MDCButton

codeman7:
In order to take a step towards deprecation of MDCFloatingButton we need to rename this category. This should allow client teams to still use floating buttons animations as before. This change just enables the animations to occur on ANY MDCButton instead of only MDCFloatingButtons.


romoore (commented):
This animation cannot be applied to MDC Buttons in general.

romoore:
This animation is only for Floating Action Buttons (FAB) and not for other Material Button types. It has also only been tested with the MDCFloatingButton layout code, so I'm not sure what effect it would have on an arbitrary button.  

Why do we want to deprecate MDCFloatingButton? Is there an issue or project for that?

PR #4779, opened 2018-08-13
[TextFields] Update placeholder and outline border color when enabling/disabling text fields with outline controllers

### LINKED ISSUE #4678: [TextField] disabled state does not work as described in the documentation
> For usage questions: ask on [Stack  Overflow](https://stackoverflow.com/questions/51626342/mdctextfield-disabled-state-does-not-work-as-described-in-the-documentation).

MDCTextField disabled state does not work as described in the documentation

## Reproduction steps

1. Create MDCTextField 
2. Set disabledColor in MDCTextInputControllerOutlined
3. set textField.enabled = NO

```
@interface MyController ()

@property (nonatomic, strong) MDCTextInputControllerOutlined *textInputController;
@property (nonatomic, strong) MDCTextField *textField;

@end

@implementation MyController

- (void)viewDidLoad {
    [super viewDidLoad];

    self.textField = [[MDCTextField alloc] initWithFrame:CGRectMake(50.f, 50.f, 220.f, 85.f)];
    [self.view addSubview:self.textField];

    self.textInputController = [[MDCTextInputControllerOutlined alloc] initWithTextInput:self.textField];
    self.textInputController.normalColor = [UIColor purpleColor];
    self.textInputController.disabledColor = [UIColor redColor];

    self.textField.enabled = NO;
    self.textField.placeholder = @"placeholder";
    self.textField.text = @"text";
}
```

## Expected behavior

I expect that the color of the disabled state should apply to the placeholder and the border. 
[material documentation](https://material.io/design/components/text-fields.html#outlined-text-field)

## Actual behavior

As a result I see only the dotted line below the color of the disabled state. 

## Platform
 - Devices: All
 - OS: iOS9, iOS10, iOS11

andrewoverton:
Closes #4678.

It made sense to make the visual changes to the textfield desired by the client through the controller, because that's how it works with things like error text, through APIs on the controller like `-setErrorText:errorAccessibilityValue:`. However, `enabled` is a property on the TextField, not the controller, and there isn't _normally_ a line of communication _from_ the text field _to_ the controller. Occasionally there are delegates that allow for this, like `MDCTextFieldPositioningDelegate`. From what I could tell my options were either introduce another one, add a `weak id<MDCTextInputController>` property to `MDCTextField`, or bring KVO of the text field into the controller. I chose KVO...

Also, there was another bug where the underline was visible when a text field with an outline controller was disabled. That's not in the spec. Hiding the underline and not calling `super` in `updateUnderline` was the easiest approach to fixing it. I'm of two minds about it. It feels kind of hacky but it's by far the smallest change.

This gif is of the TextFieldOutlinedExample with the following added to the setup code:
```Objective-C
  self.zipController.normalColor = [UIColor blueColor];
  self.zipController.disabledColor = [UIColor greenColor];
  
  textFieldZip.placeholder = @"placeholder";
  textFieldZip.text = @"text";
  
  dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(2 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
    textFieldZip.enabled = NO;
  });

  dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(4 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
    textFieldZip.enabled = YES;
  });
```

aug-13-2018 17-11-41

### DIFF
--- a/components/TextFields/src/MDCTextField.h
+++ b/components/TextFields/src/MDCTextField.h
@@ -21,6 +21,9 @@
 /** When text is manually set via .text or setText:, this notification fires. */
 extern NSString *_Nonnull const MDCTextFieldTextDidSetTextNotification;
 
+/** When the value of `enabled` changes on the text input, this notification fires. */
+extern NSString *_Nonnull const MDCTextInputDidToggleEnabled;
+
 /**
   Material Design compliant single-line text input.
   https://www.google.com/design/spec/components/text-fields.html#text-fields-single-line-text-field
--- a/components/TextFields/src/MDCTextField.h
+++ b/components/TextFields/src/MDCTextField.h
@@ -22,7 +22,7 @@
 extern NSString *_Nonnull const MDCTextFieldTextDidSetTextNotification;
 
 /** When the value of `enabled` changes on the text input, this notification fires. */
-extern NSString *_Nonnull const MDCTextInputDidToggleEnabled;
+extern NSString *_Nonnull const MDCTextInputDidToggleEnabledNotification;
 
 /**
   Material Design compliant single-line text input.
--- a/components/TextFields/src/MDCMultilineTextField.m
+++ b/components/TextFields/src/MDCMultilineTextField.m
@@ -571,8 +571,9 @@
 - (void)setEnabled:(BOOL)enabled {
   self.fundament.enabled = enabled;
   self.textView.editable = enabled;
-  [[NSNotificationCenter defaultCenter] postNotificationName:MDCTextInputDidToggleEnabled
-                                                      object:self];
+  [[NSNotificationCenter defaultCenter]
+      postNotificationName:MDCTextInputDidToggleEnabledNotification
+                    object:self];
 }
 
 - (void)setExpandsOnOverflow:(BOOL)expandsOnOverflow {
--- a/components/TextFields/src/MDCMultilineTextField.m
+++ b/components/TextFields/src/MDCMultilineTextField.m
@@ -571,6 +571,8 @@
 - (void)setEnabled:(BOOL)enabled {
   self.fundament.enabled = enabled;
   self.textView.editable = enabled;
+  [[NSNotificationCenter defaultCenter] postNotificationName:MDCTextInputDidToggleEnabled
+                                                      object:self];
 }
 
 - (void)setExpandsOnOverflow:(BOOL)expandsOnOverflow {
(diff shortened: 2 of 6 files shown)


willlarche (changes requested):
KVO is a path of last resort and has to be discussed at length with TLs to justify. Luckily there's a built in way for the text inputs to report changes to the controller. In the fundament, `setEnabled:` is already overridden. Put in an `NSNotification` broadcast for `UITextFieldTextDidChangeNotification` and `UITextViewTextDidChangeNotification` just like in `clearButtonDidTouch`.

Considering this is the same logic as in `clearButtonDidTouch`, move it a method in the fundament and call it from `clearButtonDidTouch` and `setEnabled:`.

andrewoverton:
@willlarche I went looking for some kind of UITextField/UITextView notification that I thought might be appropriate but all I saw were TextDidBeginEditing, TextDidEndEditing, and TextDidChange. Since none of those described what was happening, and since our codebase as a whole is pretty KVO friendly compared to others I’ve worked on, I went with KVO. However, I’m happy to observe NSNotifications rather than key-value pairs. Would you be okay with a new NSNotificationName that more accurately describes what’s happening? Like `MDCTextFieldEnabledStateDidChange` or `MDCTextFieldDidToggleEnabled` or something like that?

willlarche:
Yeah I can buy having a dedicated NSNotification for that if you really don't like didChange for that.

andrewoverton:
Cool. Yep, I think it's better because the value of `text` isn't actually changing, the value of `enabled` is. But I'm also totally going off the assumption that Apple doesn't use `TextDidChange` to convey changes in things other than text haha, maybe they do?

willlarche:
I think you're right. I don't think they do.

components/TextFields/src/MDCTextField.h
```diff
@@ -21,6 +21,9 @@
 /** When text is manually set via .text or setText:, this notification fires. */
 extern NSString *_Nonnull const MDCTextFieldTextDidSetTextNotification;
 
+/** When the value of `enabled` changes on the text input, this notification fires. */
+extern NSString *_Nonnull const MDCTextInputDidToggleEnabled;
```
willlarche: The other notifications end in `Notification`. Ask @ianegordon if he'd like to continue that paradigm.

willlarche (approved):
Approved with one comment on naming for the TL to answer.

components/TextFields/src/MDCTextField.h
andrewoverton: Ahhh he can chime in but honestly it should have `Notification` at the end, I'll add it. I might also change `TextInput` to `TextField`. That one was intentional but I don't know...

willlarche: I'd keep "input" because the `enabled` property comes from the `MDCTextInput` protocol. It's in the multiline version too.

PR #4799, opened 2018-08-15
[Snackbar] VoiceOver users can exit the snackbar

### LINKED ISSUE #4798: [Snackbar] Allow VoiceOver and Switch Control users to "exit" the Snackbar view
When a Snackbar Message appears, users of VoiceOver and Switch Control should be able to navigate into and out of the message view. When an action button is available and the focus shifts, users should not become "trapped" within the snackbar message.

romoore:
When a Snackbar message with an action (button) appears, the focus is changed
for UIAccessibility. Instead of forcing the user to take an action (either the
button or the label to dismiss), the user should be able to navigate away from
the Snackbar message and leave it on-screen if they desire.

Clients should be dismissing the Snackbar message programmatically once the
user takes "an action" within the app. UIAccessibility navigation does not
count as an "action" but normal UIVC navigation does.

Closes #4798


yarneo (approved):
sweet!

PR #4804, opened 2018-08-15
[ShadowLayer] Setting the corner radius should recalculate shadowPath

ianegordon:
This mirrors the internal CL:
cl/208469407


romoore (changes requested):
Could we add a dragon for this? Is the bug visible in any of our current examples?

ianegordon:
TODO to add a dragon #4813.
I'd like to land before release.

randallli (approved):

romoore:
Good to know we're allowed to merge with a blocking review.

PR #4849, opened 2018-08-20
[ShadowLayer] Add dragon demonstrating cornerRadius changes aren't rendering

ianegordon:
Added Shadows - Shadows Corner Radius to Dragons to reproduce case affecting internal client.

### DIFF
--- a/components/ShadowLayer/examples/ShadowCornerRadiusExample.m
+++ b/components/ShadowLayer/examples/ShadowCornerRadiusExample.m
@@ -1,5 +1,5 @@
 /*
- Copyright 2016-present the Material Components for iOS authors. All Rights Reserved.
+ Copyright 2018-present the Material Components for iOS authors. All Rights Reserved.
 
  Licensed under the Apache License, Version 2.0 (the "License");
  you may not use this file except in compliance with the License.
--- a/components/ShadowLayer/examples/ShadowCornerRadiusExample.m
+++ b/components/ShadowLayer/examples/ShadowCornerRadiusExample.m
@@ -1,0 +1,122 @@
+/*
+ Copyright 2016-present the Material Components for iOS authors. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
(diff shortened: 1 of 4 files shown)


ianegordon:
This issue was originally reporting internally as:
b/111777852

components/ShadowLayer/examples/ShadowRadiusLabel.h
```diff
@@ -0,0 +1,26 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+#import "MaterialShadowElevations.h"
+
+@interface ShadowRadiusLabel : UILabel
```
codeman7, line 21: Why not make this a UIView, I think UILabels can sometimes render shadows weird.

components/ShadowLayer/examples/ShadowCornerRadiusExample.m
```diff
@@ -0,0 +1,123 @@
...
+- (void)sliderValueChanged:(MDCSlider *)slider {
+  NSInteger points = (NSInteger)round(slider.value);
+  _paper.text = [NSString stringWithFormat:@"%ld pt", (long)points];
+  _paper.cornerRadius = (CGFloat)points;
+  _elevationLabel.text = _paper.text;
+}
+
+@end
+
+@interface ShadowCornerRadiusViewController : UIViewController
```
codeman7, line 93: I thought we were supposed to add color/typography themers to all examples?

codeman7 (approved):

components/ShadowLayer/examples/ShadowRadiusLabel.h
ianegordon: I want to explicitly show the feedback of label value changing with the contrast of the cornerRadius not changing.

components/ShadowLayer/examples/ShadowCornerRadiusExample.m
ianegordon: Catalog examples should be themed.
Dragons may or may not be.  For testing and developing certain things it may make sense to avoid the theming subsystems.

PR #4850, opened 2018-08-20
[TextFields] Don't set non-floating placeholder color to active color while editing

### LINKED ISSUE #4844: [TextFields] Ability to set text color of inline placeholder in MDCTextInputController while editing
This is an internal issue. If you are a Googler, please visit [b/112707843](http://b/112707843) for more details.

andrewoverton:
Closes #4844.

Previously, non-floating placeholder text color changed to `activeColor` while editing. An internal client sensed something incorrect about this, and Phil confirmed that it should stay set the value of `inlinePlaceholderColor`.

Before:
simulator screen shot - iphone x - 2018-08-20 at 15 56 25

After:
simulator screen shot - iphone x - 2018-08-20 at 15 56 00

### DIFF
--- a/components/TextFields/src/MDCTextInputControllerBase.m
+++ b/components/TextFields/src/MDCTextInputControllerBase.m
@@ -437,7 +437,7 @@ static UITextFieldViewMode _underlineViewModeDefault = UITextFieldViewModeWhileE
                            ? self.errorColor
                            : nonErrorColor;
   } else {
-    placeholderColor = self.textInput.isEditing ? self.activeColor : self.inlinePlaceholderColor;
+    placeholderColor = self.inlinePlaceholderColor;
   }
   if (!self.textInput.isEnabled) {
     placeholderColor = self.disabledColor;


romoore:
Does this change the floating placeholder color in the Outlined text controller?

andrewoverton:
@romoore This change is in the else block following an `if ([self isPlaceholderUp])`, so it won't affect TextFields with floating placeholders. I've only seen TextFields with outline controllers have floating placeholders. I'm not actually sure if they can have non-floating placeholders, but if they can, this change will affect them.

romoore:
Thanks @andrewoverton . I thought `isPlaceholderUp` is only checking for its current state, not its capability. Good to know this doesn't affect controllers that support floating. 👍

andrewoverton:
I think `isPlaceholderUp ` _does_ refer to the state rather than the capability.. I'm not sure where the capability comes from, but I can look into it.

willlarche (approved):
With much thanks!

PR #4867, opened 2018-08-21
[Math] added MDCCeilScaled and MDCFloorScaled (#1917)

### LINKED ISSUE #1917: [Math] Implement MDCCeilScaled()
Throughout Text Fields there are many instances of the following:

CGFloat scale = UIScreen.mainScreen.scale;
CGFloat placeholderEstimatedHeight = 
    MDCCeil(self.textInput.placeholderLabel.font.lineHeight * scale) / scale;

We should consider if we want to pull this into MDCMath.

Name and interface TBD.

fluxminds:
closes https://github.com/material-components/material-components-ios/issues/1917

### DIFF
--- a/components/private/Math/src/MDCMath.h
+++ b/components/private/Math/src/MDCMath.h
@@ -131,6 +131,22 @@ static inline CGFloat MDCSqrt(CGFloat value) {
 #endif
 }
 
+static inline CGFloat MDCCeilScaled(CGFloat value, CGFloat scale) {
+  if (MDCCGFloatEqual(scale, 0)) {
+    return 0.0f;
+  }
+
+  return MDCCeil(value * scale) / scale;
+}
+
+static inline CGFloat MDCFloorScaled(CGFloat value, CGFloat scale) {
+  if (MDCCGFloatEqual(scale, 0)) {
+    return 0.0f;
+  }
+
+  return MDCFloor(value * scale) / scale;
+}
+
 /**
  Expand `rect' to the smallest standardized rect containing it with pixel-aligned origin and size.
  If @c scale is zero, then a scale of 1 will be used instead.
--- a/components/private/Math/src/MDCMath.h
+++ b/components/private/Math/src/MDCMath.h
@@ -131,6 +131,14 @@ static inline CGFloat MDCSqrt(CGFloat value) {
 #endif
 }
 
+/**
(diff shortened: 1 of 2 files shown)


components/private/Math/tests/unit/MDCMathTests.m
```diff
@@ -23,6 +23,32 @@ @interface MDCMathTests : XCTestCase
 
 @implementation MDCMathTests
 
+#pragma mark - MDCCeilScaled and MDCFloorScaled
+
+- (void)testMDCFloorScaled {
```
romoore, line 28: Please add more tests and structure them in the Given-When-Then style that we try to encourage in general. We prefer to have 1 test for each test condition whenever possible.

components/private/Math/tests/unit/MDCMathTests.m
```diff
@@ -23,6 +23,32 @@ @interface MDCMathTests : XCTestCase
...
+  CGFloat inputNumber = 1.3f;
+
+  CGFloat scale1 = 2.0f;
+  CGFloat expectedOutputNumber1 = 1.0f;
+  XCTAssertEqual(MDCFloorScaled(inputNumber, scale1), expectedOutputNumber1);
```
romoore: Please use XCTAssertEqualWithAccuracy(...) when comparing floating point values.

components/private/Math/src/MDCMath.h
```diff
@@ -131,6 +131,22 @@ static inline CGFloat MDCSqrt(CGFloat value) {
 #endif
 }
 
+static inline CGFloat MDCCeilScaled(CGFloat value, CGFloat scale) {
```
romoore: Please add documentation here and in `MDCFloorScaled` to describe how these work, particularly with different values of `scale`.

romoore (approved):

fluxminds:
@romoore Thanks for the review. Comments have been addressed in 911f1233ba1bddafad0b5f4388bb414b3bcf2347

components/private/Math/tests/unit/MDCMathTests.m
romoore: Please add tests for negative values as well.

Overall looks good!

romoore (approved):

PR #4871, opened 2018-08-21
[Codelabs] Update build tests for codelabs

michdud:
Allows building from branch, tag, or commit. Defaults to build only 104 complete and 111 complete.

### DIFF
--- a/scripts/build_codelabs
+++ b/scripts/build_codelabs
@@ -40,6 +40,7 @@ function is_xcpretty_available() {
 function check_arguments() {
   if [ -z $1 ]; then
     echo "Argument must be provided to flag."
+    echo "$(basename $0) -h for help"
     exit 1
   fi
 }
@@ -50,7 +51,8 @@ function check_arguments() {
 function check_reference() {
   reference=$(git ls-remote https://github.com/material-components/material-components-ios.git $1 | wc -l)
   if [ $reference -eq 0 ]; then
-    echo "Invalid reference name."
+    echo "Unable to find $1 on GitHub."
+    echo "$(basename $0) -h for help"
     exit 1
   fi
 }
@@ -61,6 +63,7 @@ function check_reference() {
 function check_commit() {
   if [ ! $(git cat-file commit $1) ]; then
     echo "Invalid commit SHA."
+    echo "$(basename $0) -h for help"
     exit 1
   fi
 }
@@ -88,6 +91,7 @@ while [ $# -gt 0 ]; do
       check_reference $1
       if [ $has_source -eq 1 ]; then
         echo "Only one of --remote-branch/--commit/--tag allowed."
+        echo "$(basename $0) -h for help"
         exit -1
       fi
       export TEST_BRANCH=$1
@@ -100,6 +104,7 @@ while [ $# -gt 0 ]; do
       check_commit $1
       if [ $has_source -eq 1 ];then
         echo "Only one of --remote-branch/--commit/--tag allowed."
+        echo "$(basename $0) -h for help"
         exit -1
       fi
       export TEST_COMMIT=$1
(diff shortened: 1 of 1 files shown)


scripts/build_codelabs
```diff
@@ -32,24 +34,126 @@ function is_xcpretty_available() {
...
+      shift
+      ;;
+    -h|--help)
+      echo
+      echo "Builds MDC iOS Codelabs."
+      echo
+      echo "The default branch of MDC iOS that is used for testing is your current branch."
+      echo "Your current branch is: $DEFAULT_BRANCH"
+      echo
+      echo "The default codelabs built are 104 Complete and 111 Complete. You can build all with -a."
```
ianegordon: Consider:
By default the Swift and Objective-C Google I/O 2018 Codelab 104 and Codelab 111 targets will be built.
You can build all codelabs by specifying --all.

scripts/build_codelabs
```diff
@@ -32,24 +34,126 @@ function is_xcpretty_available() {
...
+  fi
+}
+
+# Checks that the argument is a valid reference.
+#
+# Exits if the reference is not valid.
+function check_reference() {
+  reference=$(git ls-remote https://github.com/material-components/material-components-ios.git $1 | wc -l)
+  if [ $reference -eq 0 ]; then
+    echo "Invalid reference name."
```
ianegordon: Consider:
"Unable to find $1 on GitHub"

scripts/build_codelabs
```diff
@@ -32,24 +34,126 @@ function is_xcpretty_available() {
...
+      ;;
+    -a|--all)
+      all_codelabs=1
+      shift
+      ;;
+    -h|--help)
+      echo
+      echo "Builds MDC iOS Codelabs."
+      echo
+      echo "The default branch of MDC iOS that is used for testing is your current branch."
```
ianegordon: Consider:
By default the codelabs are tested against the remote branch tracking your current branch.

scripts/build_codelabs
```diff
@@ -32,24 +34,126 @@ function is_xcpretty_available() {
...
+      echo "Your current branch is: $DEFAULT_BRANCH"
+      echo
+      echo "The default codelabs built are 104 Complete and 111 Complete. You can build all with -a."
+      echo
+      echo "Options:"
+      echo "-v,  --verbose              Verbose logging."
+      echo "-rb, --remote-branch        Remote branch of MDC-iOS with which to test codelabs."
+      echo "-c,  --commit               Commit SHA of MDC-iOS with which to test codelabs."
+      echo "-t,  --tag                  Tag of MDC-iOS with which to test codelabs."
+      echo "-a,  --all                  Build all the codelabs."
```
ianegordon, line 146: Super-SUPER-nit: Add -h, --help to this menu

ianegordon (approved):
Consider adding something along the following to every error message.
echo "$(basename($0)) -h for help"

I tried running this from a non-tracked branch and it said "Invalid reference name." which wasn't super obvious.

This is great!

scripts/build_codelabs
michdud: Right now, the script points to the master branch of the codelabs (I can't see an I/O branch there). Should I refer to the codelabs as the Google I/O 2018 codelabs?

ianegordon: Good point. 
Could you add tag to the current master called googleio2018 at head? (Outside of this PR)
That way we can continue to point at master, but we'll have a reference if we need it.
I think referring to them as the 'codelabs' is fine in the general case, in this specific case I would urge details.  My thinking is that if something goes wrong, the person will end up coming through this path (--help !) and will gain some insight.

michdud:
I'm ready to merge!

PR #4898, opened 2018-08-26
[Contributing] Update Alpha doc

codeman7:
Update alpha_components.md so that future contributors don't get stuck on imports. The import statements can't follow the standard convention if they are in the MaterialComponentsAlpha phase then a component will need to import `MaterialComponentsAlpha` instead of `MaterialComponents`.


jverkoey (commented):
Why’d we close this PR?

codeman7:
I didn't think it fixed my issue but I guess I had to update the podspec in the unit test as well.

jverkoey (approved):

PR #4901, opened 2018-08-27
[Slider] Allow for different thumb radiuses for different states

### LINKED ISSUE #4883: [Slider] Support different Thumb radius for UIControlState
## Feature Request/Suggestion

The MDCSlider thumb should support changing size (radius) for each UIControlState value.

For example:
* Disabled: 2 points
* Normal: 6 points
* Selected: 10 points

---
#### Internal data
- Associated internal bug: [b/117176821](http://b/117176821)

codeman7:
This change allows clients to set different thumb radiuses for all different `UIControlState`s. This will match the existing API for trackTickColor, trackFillColor, trackBackgroundColor as well as some other properties.

Closes #4883

### DIFF
--- a/components/Slider/src/MDCSlider.h
+++ b/components/Slider/src/MDCSlider.h
@@ -194,6 +194,7 @@ IB_DESIGNABLE
  The radius of the cursor (thumb).
 
  Default value is 6 points.
+ @note Has no effect if @c statefulAPIEnabled is @c YES.
  */
 @property(nonatomic, assign) CGFloat thumbRadius UI_APPEARANCE_SELECTOR;
 
--- a/components/Slider/src/MDCSlider.h
+++ b/components/Slider/src/MDCSlider.h
@@ -191,14 +191,6 @@ IB_DESIGNABLE
 @property(nonatomic, strong, nullable) UIColor *inkColor;
 
 /**
- The radius of the cursor (thumb).
-
- Default value is 6 points.
- @note Has no effect if @c statefulAPIEnabled is @c YES.
- */
-@property(nonatomic, assign) CGFloat thumbRadius UI_APPEARANCE_SELECTOR;
-
-/**
  The elevation of the cursor (thumb).
 
  Default value is MDCElevationNone.
@@ -316,6 +308,15 @@ IB_DESIGNABLE
 @property(nonatomic, assign, getter=isThumbHollowAtStart) BOOL thumbHollowAtStart;
 
 #pragma mark - To be deprecated
+
+/**
+ The radius of the cursor (thumb).
+
+ Default value is 6 points.
+ @note This API is planned for deprecation. Use @c setThumbRadius:forState: instead.
+ @note Has no effect if @c statefulAPIEnabled is @c YES.
+ */
+@property(nonatomic, assign) CGFloat thumbRadius UI_APPEARANCE_SELECTOR;
 
 /**
  The color of the cursor (thumb) and track while the slider is disabled.
--- a/components/Slider/src/MDCSlider.h
+++ b/components/Slider/src/MDCSlider.h
@@ -171,8 +171,9 @@ IB_DESIGNABLE
 
  @param thumbRadius The radius of the thumb.
  @param state The state of the slider.
+ @note If a negative value is passed the absolute value will be used.
  */
-- (void)setThumbRadius:(CGFloat)thumbRadius forState:(UIControlState)state;
+- (void)setThumbRadius:(float)thumbRadius forState:(UIControlState)state;
 
 /**
  Returns the thumb radius associated with the specified state.
@@ -181,7 +182,7 @@ IB_DESIGNABLE
  @returns The thumb radius for the specified state. If no radius has been set for the specific
           state, this method returns the radius associated with the @c UIControlStateNormal state.
  */
-- (CGFloat)thumbRadiusForState:(UIControlState)state;
+- (float)thumbRadiusForState:(UIControlState)state;
 
 /**
  The color of the Ink ripple.
--- a/components/Slider/src/MDCSlider.h
+++ b/components/Slider/src/MDCSlider.h
@@ -173,7 +173,7 @@ IB_DESIGNABLE
  @param state The state of the slider.
  @note If a negative value is passed the absolute value will be used.
  */
-- (void)setThumbRadius:(float)thumbRadius forState:(UIControlState)state;
+- (void)setThumbRadius:(double)thumbRadius forState:(UIControlState)state;
 
 /**
  Returns the thumb radius associated with the specified state.
@@ -182,7 +182,7 @@ IB_DESIGNABLE
  @returns The thumb radius for the specified state. If no radius has been set for the specific
           state, this method returns the radius associated with the @c UIControlStateNormal state.
  */
-- (float)thumbRadiusForState:(UIControlState)state;
+- (double)thumbRadiusForState:(UIControlState)state;
 
 /**
  The color of the Ink ripple.
--- a/components/Slider/src/MDCSlider.h
+++ b/components/Slider/src/MDCSlider.h
@@ -162,28 +162,6 @@ IB_DESIGNABLE
 - (nullable UIColor *)backgroundTrackTickColorForState:(UIControlState)state;
 
 /**
- Sets the thumb radius to use for the specified state.
-
- In general, if a property is not specified for a state, the default is to use the @c
- UIControlStateNormal value. If the @c UIControlStateNormal value is not set, then the property
- defaults to a default value. Therefore, at a minimum, you should set the value for the normal
- state.
-
- @param thumbRadius The radius of the thumb.
- @param state The state of the slider.
- */
-- (void)setThumbRadius:(CGFloat)thumbRadius forState:(UIControlState)state;
-
-/**
- Returns the thumb radius associated with the specified state.
-
- @params state The state that uses the thumb radius.
- @returns The thumb radius for the specified state. If no radius has been set for the specific
-          state, this method returns the radius associated with the @c UIControlStateNormal state.
- */
-- (CGFloat)thumbRadiusForState:(UIControlState)state;
-
-/**
  The color of the Ink ripple.
 
  Defaults to transparent blue.
--- a/components/Slider/src/MDCSlider.h
+++ b/components/Slider/src/MDCSlider.h
@@ -162,6 +162,28 @@ IB_DESIGNABLE
 - (nullable UIColor *)backgroundTrackTickColorForState:(UIControlState)state;
 
 /**
+ Sets the thumb radius to use for the specified state.
+
+ In general, if a property is not specified for a state, the default is to use the @c
+ UIControlStateNormal value. If the @c UIControlStateNormal value is not set, then the property
+ defaults to a default value. Therefore, at a minimum, you should set the value for the normal
+ state.
+
+ @param thumbRadius The radius of the thumb.
+ @param state The state of the slider.
+ */
+- (void)setThumbRadius:(CGFloat)thumbRadius forState:(UIControlState)state;
+
+/**
+ Returns the thumb radius associated with the specified state.
+
+ @params state The state that uses the thumb radius.
+ @returns The thumb radius for the specified state. If no radius has been set for the specific
+          state, this method returns the radius associated with the @c UIControlStateNormal state.
+ */
+- (CGFloat)thumbRadiusForState:(UIControlState)state;
+
+/**
  The color of the Ink ripple.
 
  Defaults to transparent blue.
--- a/components/Slider/src/MDCSlider.m
+++ b/components/Slider/src/MDCSlider.m
@@ -226,11 +226,10 @@ static inline UIColor *MDCThumbTrackDefaultColor(void) {
   if ([_thumbRadiusesForState objectForKey:@(state)]) {
     return [_thumbRadiusesForState[@(state)] floatValue];
   }
-  CGFloat radius = kSliderDefaultThumbRadius;
-  if (state != UIControlStateNormal) {
-    radius = [_thumbRadiusesForState[@(UIControlStateNormal)] floatValue];
+  if ([_thumbRadiusesForState objectForKey:@(UIControlStateNormal)]) {
+    return [_thumbRadiusesForState[@(UIControlStateNormal)] floatValue];
   }
-  return radius;
+  return kSliderDefaultThumbRadius;
 }
 
 - (void)updateColorsForState {
@@ -267,6 +266,8 @@ static inline UIColor *MDCThumbTrackDefaultColor(void) {
   if (self.statefulAPIEnabled) {
     return;
   }
+  float radius = (float)thumbRadius;
+  _thumbRadiusesForState[@(UIControlStateNormal)] = [NSNumber numberWithFloat:radius];
   _thumbTrack.thumbRadius = thumbRadius;
 }
 
--- a/components/Slider/src/MDCSlider.m
+++ b/components/Slider/src/MDCSlider.m
@@ -106,7 +106,7 @@ static inline UIColor *MDCThumbTrackDefaultColor(void) {
   _backgroundTickColorsForState[@(UIControlStateNormal)] = UIColor.blackColor;
   _thumbRadiusesForState = [@{} mutableCopy];
   _thumbRadiusesForState[@(UIControlStateNormal)] =
-      [NSNumber numberWithDouble:kSliderDefaultThumbRadius];
+      [NSNumber numberWithFloat:kSliderDefaultThumbRadius];
   [self addSubview:_thumbTrack];
 }
 
@@ -215,7 +215,8 @@ static inline UIColor *MDCThumbTrackDefaultColor(void) {
 
 - (void)setThumbRadius:(CGFloat)thumbRadius forState:(UIControlState)state {
   NSAssert(thumbRadius >= 0, @"Thumb radius must be a positive value");
-  _thumbRadiusesForState[@(state)] = [NSNumber numberWithDouble:thumbRadius];
+  float radius = (float)thumbRadius;
+  _thumbRadiusesForState[@(state)] = [NSNumber numberWithFloat:radius];
   if (self.state == state) {
     [self updateThumbForState];
   }
@@ -223,11 +224,11 @@ static inline UIColor *MDCThumbTrackDefaultColor(void) {
 
 - (CGFloat)thumbRadiusForState:(UIControlState)state {
   if ([_thumbRadiusesForState objectForKey:@(state)]) {
-    return [_thumbRadiusesForState[@(state)] doubleValue];
+    return [_thumbRadiusesForState[@(state)] floatValue];
   }
   CGFloat radius = kSliderDefaultThumbRadius;
   if (state != UIControlStateNormal) {
-    radius = [_thumbRadiusesForState[@(UIControlStateNormal)] doubleValue];
+    radius = [_thumbRadiusesForState[@(UIControlStateNormal)] floatValue];
   }
   return radius;
 }
--- a/components/Slider/src/MDCSlider.m
+++ b/components/Slider/src/MDCSlider.m
@@ -255,6 +255,7 @@ static inline UIColor *MDCThumbTrackDefaultColor(void) {
 - (void)updateThumbForState {
   if (!self.isStatefulAPIEnabled) {
     _thumbTrack.thumbRadius = kSliderDefaultThumbRadius;
+    return;
   }
 
   _thumbTrack.thumbRadius = [self thumbRadiusForState:self.state];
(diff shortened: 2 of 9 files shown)


components/Slider/src/MDCSlider.m
```diff
@@ -209,6 +213,24 @@ - (UIColor *)backgroundTrackTickColorForState:(UIControlState)state {
   return color;
 }
 
+- (void)setThumbRadius:(CGFloat)thumbRadius forState:(UIControlState)state {
+  _thumbRadiusesForState[@(state)] = [NSNumber numberWithFloat:thumbRadius];
```
romoore: Should we NSAssert if they assign a negative value?

components/Slider/src/MDCSlider.m
```diff
@@ -209,6 +213,24 @@ - (UIColor *)backgroundTrackTickColorForState:(UIControlState)state {
...
+  if (self.state == state) {
+    [self updateThumbForState];
+  }
+}
+
+- (CGFloat)thumbRadiusForState:(UIControlState)state {
+  if ([_thumbRadiusesForState objectForKey:@(state)]) {
+    return [_thumbRadiusesForState[@(state)] floatValue];
+  }
+  CGFloat radius = 0.f;
```
romoore: Can we use the default value instead of 0? I don't think we want clients to be "hiding" the thumb by making it too small to render.

components/Slider/tests/unit/SliderTests.m
```diff
@@ -913,6 +913,42 @@ - (void)testThumbElevationDefault {
   XCTAssertEqual(self.slider.thumbElevation, MDCShadowElevationNone);
 }
 
+- (void)testThumbRadiusForState {
```
romoore, line 916: Please test the default/fallback behavior as well.

components/Slider/src/MDCSlider.h
```diff
@@ -161,6 +161,28 @@ IB_DESIGNABLE
...
+
+ In general, if a property is not specified for a state, the default is to use the @c
+ UIControlStateNormal value. If the @c UIControlStateNormal value is not set, then the property
+ defaults to a default value. Therefore, at a minimum, you should set the value for the normal
+ state.
+
+ @param thumbRadius The radius of the thumb.
+ @param state The state of the slider.
+ */
+- (void)setThumbRadius:(CGFloat)thumbRadius forState:(UIControlState)state;
```
romoore: What is happening to `-thumbRadius`?  Does it just get assigned to the `.normal` state?

components/Slider/src/MDCSlider.m
```diff
@@ -229,6 +251,14 @@ - (void)updateColorsForState {
   _thumbTrack.trackOffTickColor = [self backgroundTrackTickColorForState:self.state];
 }
 
+- (void)updateThumbForState {
+  if (!self.isStatefulAPIEnabled) {
+    return;
+  }
```
romoore, line 256: I feel as though if the stateful API is disabled, we should restore the ThumbTrack's thumbRadius to the `-thumbRadius` property value.  What do you think?

components/Slider/tests/unit/SliderTests.m
romoore, line 916: Please add a test that the legacy behavior is not affected when the API is off.

romoore (dismissed):

components/Slider/src/MDCSlider.m
codeman7: Done

components/Slider/src/MDCSlider.m
codeman7: Done

components/Slider/src/MDCSlider.h
codeman7: Yes

components/Slider/src/MDCSlider.m
codeman7: Done

components/Slider/tests/unit/SliderTests.m
codeman7: Done

codeman7: I believe this is the same as the comment above.

components/Slider/src/MDCSlider.h
romoore: Please document the behavior in the header. Similar updates to the documentation was added for the other properties affected by the "stateful API" flag.

components/Slider/src/MDCSlider.m
```diff
@@ -229,6 +252,14 @@ - (void)updateColorsForState {
   _thumbTrack.trackOffTickColor = [self backgroundTrackTickColorForState:self.state];
 }
 
+- (void)updateThumbForState {
+  if (!self.isStatefulAPIEnabled) {
+    _thumbTrack.thumbRadius = kSliderDefaultThumbRadius;
+  }
```
romoore: Won't this be overridden by the stateful API value below?

components/Slider/tests/unit/SliderTests.m
romoore: I don't see the test added. Can you please clarify what you mean by "the same comment above"? Do you mean "Done"?

screen shot 2018-08-27 at 3 40 21 pm

romoore (dismissed):

components/Slider/tests/unit/SliderTests.m
codeman7: What's the difference between the legacy behavior and fallback behavior?

components/Slider/src/MDCSlider.m
codeman7: Forgot to add the return statement.

components/Slider/src/MDCSlider.h
codeman7: I added documentation in the header for `thumbRadius` property.

components/Slider/tests/unit/SliderTests.m
romoore: Sorry, by "legacy behavior" I meant when `.statefulAPIEnabled = NO;`.  Fallback is that each of the UIControl states will use the value for `.normal` if no explicit value is configured for that state.

components/Slider/src/MDCSlider.h
```diff
@@ -172,6 +194,7 @@ IB_DESIGNABLE
  The radius of the cursor (thumb).
 
  Default value is 6 points.
+ @note Has no effect if @c statefulAPIEnabled is @c YES.
```
romoore: Can we discourage clients from using this? I hope we are planning to deprecate this property in the future. 

Would you be willing to follow the pattern adopted for `trackBackgroundColor`, `color`, and others and document the upcoming deprecation as well as the recommended alternative (the new `thumbRadius:forState:` APIs)?

components/Slider/tests/unit/SliderTests.m
```diff
@@ -913,6 +913,64 @@ - (void)testThumbElevationDefault {
...
+  // Given
+  self.slider.statefulAPIEnabled = YES;
+
+  // When
+  CGFloat defaultThumbRadius = self.slider.thumbRadius;
+  CGFloat thumbRadius = 100.f;
+  self.slider.thumbRadius = thumbRadius;
+
+  XCTAssertEqual(self.slider.thumbRadius, defaultThumbRadius);
+
```
romoore: Nit: Please remove newline.

components/Slider/tests/unit/SliderTests.m
```diff
@@ -913,6 +913,64 @@ - (void)testThumbElevationDefault {
...
+- (void)testThumbRadiusStatefulAPI {
+  // Given
+  self.slider.statefulAPIEnabled = YES;
+
+  // When
+  CGFloat defaultThumbRadius = self.slider.thumbRadius;
+  CGFloat thumbRadius = 100.f;
+  self.slider.thumbRadius = thumbRadius;
+
```
romoore, line 960: Nit: Missing "Then" comment.

components/Slider/tests/unit/SliderTests.m
```diff
@@ -913,6 +913,64 @@ - (void)testThumbElevationDefault {
...
+  self.slider.highlighted = YES;
+  XCTAssertEqual(self.slider.thumbRadius, highlighted);
+
+  self.slider.enabled = YES;
+  self.slider.selected = YES;
+  self.slider.highlighted = NO;
+  XCTAssertEqual(self.slider.thumbRadius, selected);
+}
+
+- (void)testThumbRadiusStatefulAPI {
```
romoore: Nit/suggestion: `testThumbRadiusHasNoEffectWhenStatefulAPIIsEnabled`

components/Slider/tests/unit/SliderTests.m
romoore: I don't see the test for the fallback behavior.  Could you please add a test similar to those in cl/194783040 where each UIControlState is checked?

components/Slider/tests/unit/SliderTests.m
codeman7: Done

components/Slider/tests/unit/SliderTests.m
codeman7: Done

components/Slider/tests/unit/SliderTests.m
codeman7: Done

components/Slider/src/MDCSlider.h
codeman7: Done

components/Slider/tests/unit/SliderTests.m
codeman7: Done

jverkoey:
Is this PR stuck? Either mark as needs changes, label it as blocked, land it, or close it please.

codeman7:
It's just waiting review.

components/Slider/src/MDCSlider.h
```diff
@@ -162,18 +162,33 @@ IB_DESIGNABLE
...
+ In general, if a property is not specified for a state, the default is to use the @c
+ UIControlStateNormal value. If the @c UIControlStateNormal value is not set, then the property
+ defaults to a default value. Therefore, at a minimum, you should set the value for the normal
+ state.
+
+ @param thumbRadius The radius of the thumb.
+ @param state The state of the slider.
  */
-@property(nonatomic, strong, nullable) UIColor *inkColor;
+- (void)setThumbRadius:(CGFloat)thumbRadius forState:(UIControlState)state;
```
romoore: Can we please document what happens with negative or zero values?

components/Slider/src/MDCSlider.m
```diff
@@ -209,6 +213,25 @@ - (UIColor *)backgroundTrackTickColorForState:(UIControlState)state {
   return color;
 }
 
+- (void)setThumbRadius:(CGFloat)thumbRadius forState:(UIControlState)state {
+  NSAssert(thumbRadius >= 0, @"Thumb radius must be a positive value");
```
romoore: Please add code to handle this case in non-debug code.

components/Slider/src/MDCSlider.m
```diff
@@ -209,6 +213,25 @@ - (UIColor *)backgroundTrackTickColorForState:(UIControlState)state {
...
+  float radius = (float)thumbRadius;
+  _thumbRadiusesForState[@(state)] = [NSNumber numberWithFloat:radius];
+  if (self.state == state) {
```
romoore: Please update the thumb radius regardless of the current state. Consider what happens if it is currently `.highlighted` and the value for `.normal` was just updated. If we're relying on the fallback behavior for `.highlighted`, then we should see the change immediately.

components/Slider/src/MDCSlider.m
```diff
@@ -209,6 +213,25 @@ - (UIColor *)backgroundTrackTickColorForState:(UIControlState)state {
...
+    [self updateThumbForState];
+  }
+}
+
+- (CGFloat)thumbRadiusForState:(UIControlState)state {
+  if ([_thumbRadiusesForState objectForKey:@(state)]) {
```
romoore: Please use a local value (`NSNumber`) for the return value so we don't need to perform a dictionary lookup twice.

components/Slider/src/MDCSlider.m
```diff
@@ -209,6 +213,25 @@ - (UIColor *)backgroundTrackTickColorForState:(UIControlState)state {
   return color;
 }
 
+- (void)setThumbRadius:(CGFloat)thumbRadius forState:(UIControlState)state {
+  NSAssert(thumbRadius >= 0, @"Thumb radius must be a positive value");
+  float radius = (float)thumbRadius;
```
romoore: Let's keep this as a CGFloat or change the API to be `float` type. I'd prefer that we retain the full value provided by clients.

components/Slider/src/MDCSlider.m
```diff
@@ -209,6 +213,25 @@ - (UIColor *)backgroundTrackTickColorForState:(UIControlState)state {
...
+  _thumbRadiusesForState[@(state)] = [NSNumber numberWithFloat:radius];
+  if (self.state == state) {
+    [self updateThumbForState];
+  }
+}
+
+- (CGFloat)thumbRadiusForState:(UIControlState)state {
+  if ([_thumbRadiusesForState objectForKey:@(state)]) {
+    return [_thumbRadiusesForState[@(state)] floatValue];
```
romoore: Please use `doubleValue` and cast to CGFloat.

components/Slider/src/MDCSlider.m
```diff
@@ -229,9 +252,22 @@ - (void)updateColorsForState {
...
+}
+
 #pragma mark - ThumbTrack passthrough methods
 
 - (void)setThumbRadius:(CGFloat)thumbRadius {
+  if (self.statefulAPIEnabled) {
+    return;
+  }
+  float radius = (float)thumbRadius;
+  _thumbRadiusesForState[@(UIControlStateNormal)] = [NSNumber numberWithFloat:radius];
```
romoore: Let's make this a `numberWithDouble:`

components/Slider/tests/unit/SliderTests.m
```diff
@@ -913,6 +913,80 @@ - (void)testThumbElevationDefault {
   XCTAssertEqual(self.slider.thumbElevation, MDCShadowElevationNone);
 }
 
+- (void)testThumbRadiusForState {
+  // Given
+  self.slider.statefulAPIEnabled = YES;
+  CGFloat normal = 12.f;
```
romoore: Nit/naming: Let's name these `normalRadius`, `disabledRadius`, *etc.*

components/Slider/tests/unit/SliderTests.m
```diff
@@ -913,6 +913,80 @@ - (void)testThumbElevationDefault {
...
+  self.slider.highlighted = YES;
+  XCTAssertEqual(self.slider.thumbRadius, highlighted);
+
+  self.slider.enabled = YES;
+  self.slider.selected = YES;
+  self.slider.highlighted = NO;
+  XCTAssertEqual(self.slider.thumbRadius, selected);
+}
+
+- (void)testThumbRadiusHasNoEffectWhenStatefulAPIIsEnabled {
```
romoore: If I turn on stateful API, set the radius, then turn off the stateful API, should it be ignored? For the reverse (stateful off, setRadius, stateful on), updating the radius will affect the stateful API.

components/Slider/tests/unit/SliderTests.m
```diff
@@ -913,6 +913,80 @@ - (void)testThumbElevationDefault {
...
+  XCTAssertEqual(self.slider.thumbRadius, defaultThumbRadius);
+}
+
+- (void)testThumbRadiusDefaultWhenStatefulAPIIsDisabled {
+  // Given
+  self.slider.statefulAPIEnabled = NO;
+
+  // When
+  CGFloat thumbRadius = 40.f;
+  [self.slider setThumbRadius:thumbRadius];
```
romoore: Please use dot notation for property assignments `self.slider.thumbRadius = thumbRadius;`

romoore (dismissed):
Please add the "fallback" unit tests we talked about for the new stateful API.

components/Slider/src/MDCSlider.m
codeman7: Have instead decided to just always use the absolute value.

components/Slider/src/MDCSlider.h
codeman7: Have documented that if a client uses a negative value then the absolute value will be used.

components/Slider/src/MDCSlider.m
codeman7: Going to use the absolute value.

components/Slider/src/MDCSlider.m
codeman7: Done

components/Slider/src/MDCSlider.m
codeman7: Updated the API to use a `float` instead of a `double` or `CGFloat`.

components/Slider/src/MDCSlider.m
codeman7: Updated the API to use `float`.

components/Slider/tests/unit/SliderTests.m
codeman7: Done

components/Slider/tests/unit/SliderTests.m
codeman7: Done

components/Slider/src/MDCSlider.m
codeman7: Done

components/Slider/src/MDCSlider.m
codeman7: Done

components/Slider/tests/unit/SliderTests.m
codeman7: No, if you set the radius when stateful API is on then that is the thumb radius. If you turn it off later how would we know what it was before setting it? We would have to store a value inside the implementation `nonStatefulThumbRadius` we don't do that for any other properties and that seems *hacky*.

codeman7:
The only test we discussed was added `testThumbRadiusDefaultWhenStatefulAPIIsDisabled`.

components/Slider/src/MDCSlider.m
```diff
@@ -229,9 +250,22 @@ - (void)updateColorsForState {
...
+}
+
 #pragma mark - ThumbTrack passthrough methods
 
 - (void)setThumbRadius:(CGFloat)thumbRadius {
+  if (self.statefulAPIEnabled) {
+    return;
+  }
+  double radius = (double)thumbRadius;
+  _thumbRadiusesForState[@(UIControlStateNormal)] = [NSNumber numberWithDouble:radius];
```
romoore, line 268: I don't think that using the old API should affect the new API.  What do you think?  Probably easiest to just assign the `_thumbTrack.thumbRadius`  if the new API is disabled.

components/Slider/tests/unit/SliderTests.m
```diff
@@ -913,6 +913,96 @@ - (void)testThumbElevationDefault {
...
+
+  NSUInteger maxState = UIControlStateNormal | UIControlStateSelected | UIControlStateDisabled |
+  UIControlStateHighlighted;
+
+  for (NSUInteger state = 0; state <= maxState; ++state) {
+    XCTAssertEqual([self.slider thumbRadiusForState:state], thumbRadius);
+  }
+}
+
+- (void)testThumbRadius {
```
romoore, line 997: Please add a tests for negative and zero values.

components/Slider/tests/unit/SliderTests.m
romoore: Done with `-testThumbRadiusDefaultWhenStatefulAPIEnabled`

components/Slider/src/MDCSlider.h
```diff
@@ -162,18 +162,34 @@ IB_DESIGNABLE
...
+ UIControlStateNormal value. If the @c UIControlStateNormal value is not set, then the property
+ defaults to a default value. Therefore, at a minimum, you should set the value for the normal
+ state.
+
+ @param thumbRadius The radius of the thumb.
+ @param state The state of the slider.
+ @note If a negative value is passed the absolute value will be used.
  */
-@property(nonatomic, strong, nullable) UIColor *inkColor;
+- (void)setThumbRadius:(double)thumbRadius forState:(UIControlState)state;
```
romoore, line 176: Any reason why this cannot be `CGFloat` at the API? It would match the legacy API's type and using explicit `double` types in UI APIs is pretty unusual on iOS.

components/Slider/src/MDCSlider.m
romoore: Given what you needed to do (affect the new API from the old API), I think this might have been a mistake. I was thinking that an iVar could be used, but it may just be better to go back to the original version.  When we spoke offline about why a client would be turning the API on and off, I think you made a good point.  We can probably expect clients to only go from the legacy to the new API and not back again.

romoore (changes requested):

PR #4917, opened 2018-08-28
[Dialogs] Add accessibilityIdentifier to Actions.

### LINKED ISSUE #4916: [Dialogs] Support accessibilityIdentifier for MDCAlertAction
## Feature Request/Suggestion

For clients to support UI testing, an accessibilityIdentifier is useful on the Buttons in an MDCAlertController's view. MDCAlertAction should expose `accessibilityIdentifier` to support this use case.

romoore:
To improve support for UI testing, `accessibilityIdentifier` values
should be attached to the action buttons of MDCAlertController views.
Instead of exposing the views and the buttons, an
`accessibilityIdentifier` property can be assigned to the MDCAlertAction
directly using the UIAccessibilityIdentification protocol.

Closes #4916

### DIFF
--- a/components/Dialogs/tests/unit/MDCAlertControllerTests.m
+++ b/components/Dialogs/tests/unit/MDCAlertControllerTests.m
@@ -243,7 +243,8 @@ static NSString *const MDCAlertControllerSubclassValueKey = @"MDCAlertController
 
 - (void)testAccessibilityIdentifiersAppliesToAlertControllerViewButtons {
   // Given
-  MDCAlertController *alertController = [MDCAlertController alertControllerWithTitle:@"Title" message:@"message"];
+  MDCAlertController *alertController = [MDCAlertController alertControllerWithTitle:@"Title"
+                                                                             message:@"message"];
   MDCAlertAction *action1 = [MDCAlertAction actionWithTitle:@"button1" handler:nil];
   action1.accessibilityIdentifier = @"1";
   MDCAlertAction *action2 = [MDCAlertAction actionWithTitle:@"buttonA" handler:nil];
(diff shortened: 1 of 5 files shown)


components/Dialogs/tests/unit/MDCAlertControllerTests.m
```diff
@@ -237,4 +241,38 @@ - (void)testTheViewIsNotLoadedWhenPropertiesAreSet {
...
+  // Force the view to load
+  if (@available(iOS 9.0, *)) {
+    [alertController loadViewIfNeeded];
+  } else {
+    (void)alertController.view;
+  }
+
+  // Then
+  NSArray<UIButton *> *buttons = alertController.alertView.actionButtons;
+  XCTAssertEqual(buttons.count, 2U);
```
codeman7, line 266: What is the `2U`?

romoore: 2 as an unsigned value.

romoore:
If this gets approved, please keep rerunning kokoro until we no longer fail from "503" errors.

components/Dialogs/tests/unit/MDCAlertControllerTests.m
codeman7: So 2 or -2?

codeman7 (approved):

jverkoey (approved):
Solid turnaround @romoore. Thank you!

components/Dialogs/tests/unit/MDCAlertControllerTests.m
romoore: Not quite. Since `count` is an NSUInteger, when comparing it to another value (a constant in this case) the comparison should be of the same type. Some build systems will complain if a signed value and unsigned value are compared, so it's typically preferred to just explicitly mark that this is the unsigned value 2 and not the signed value 2.

randallli (approved):

PR #4919, opened 2018-08-28
[ActionSheet] Add scrim accessibility properties.

### LINKED ISSUE #4902: [ActionSheet] Support VO features for scrim.
Add the same VO support for the scrim that BottomSheet has.

codeman7:
Before this commit the scrim doesn't have any of the properties needed for accessibility, this will add the functionality that will allow clients to set accessibility properties on the scrim.

Closes #4902

### DIFF
--- a/components/ActionSheet/src/MDCActionSheetController.h
+++ b/components/ActionSheet/src/MDCActionSheetController.h
@@ -142,7 +142,8 @@ __attribute__((objc_subclassing_restricted))
 
 @property(nonatomic, nonnull, strong) UIColor *backgroundColor;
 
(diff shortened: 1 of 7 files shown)


components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -140,6 +142,8 @@ __attribute__((objc_subclassing_restricted))
 
 @property(nonatomic, nonnull, strong) UIColor *backgroundColor;
 
+@property(nonatomic, strong, nonnull) MDCBottomSheetTransitionController *transitionController;
```
ianegordon: Can this be readonly?

ianegordon (approved):
LGTM, but please consider comment before merge.

PR #4921, opened 2018-08-28
[ShadowLayer] Mark shadowPath invalid on cornerRadius change

ianegordon:
Note the cornerRadius changes are not getting reflected in the before version.

| Before | After |
|--|--|
|shadowradius-broken|shadowradius-fixed|

### DIFF
--- a/components/ShadowLayer/src/MDCShadowLayer.m
+++ b/components/ShadowLayer/src/MDCShadowLayer.m
@@ -16,6 +16,8 @@
 
 #import "MDCShadowLayer.h"
 
+#import "MDCMath.h"
+
 static const CGFloat kShadowElevationDialog = 24.0;
 static const float kKeyShadowOpacity = 0.26f;
 static const float kAmbientShadowOpacity = 0.08f;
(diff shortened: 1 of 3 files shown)


components/ShadowLayer/src/MDCShadowLayer.m
```diff
@@ -299,6 +301,17 @@ - (void)setElevation:(CGFloat)elevation {
   _bottomShadow.shadowOpacity = shadowMetrics.bottomShadowOpacity;
 }
 
+- (void)setCornerRadius:(CGFloat)cornerRadius {
+  BOOL radiusChanged = !MDCCGFloatEqual(self.cornerRadius, cornerRadius);
```
codeman7, line 305: Question: In a setter shouldn't we use the ivar?

codeman7 (approved):

yarneo (approved):
Let's just make sure we have a tracking of the work needed (if needed) for MDCShapedShadowLayer as well.

components/ShadowLayer/src/MDCShadowLayer.m
romoore: ~Yes we should.~ Ah, no. Only in the getter should we avoid accessing the getter. Sorry!

romoore: Drive-by comment: Let's please add a unit test for this method. To make sure there isn't an infinite loop. :)

ianegordon: Yes, typically we prefer to use the ivar in init* methods and getters / setters.
However, in this case, cornerRadius is exposed by CALayer, our super class, so we don't have direct access to the internal data structure storing cornerRadius.

ianegordon:
Note the cornerRadius changes are not getting reflected in the before version.

Before:
shadowradius-broken

After:
shadowradius-fixed

components/ShadowLayer/src/MDCShadowLayer.m
romoore: Maybe next time it can make things clearer to use `super.cornerRadius`?

jverkoey:
@ianegordon You'll be OOO for the next bit. Either land or close this PR please.

jverkoey:
Closing this PR for now. @ianegordon please re-open if you still intend to pursue this.

PR #4926, opened 2018-08-28
[ActionSheet] Update header to support RTL

codeman7:
This fix updates the labels width in the header to allow BiDi.

### DIFF
--- a/components/ActionSheet/src/private/MDCActionSheetHeaderView.m
+++ b/components/ActionSheet/src/private/MDCActionSheetHeaderView.m
@@ -64,6 +64,7 @@ static const CGFloat MiddlePadding = 8.f;
   CGSize size = CGRectInfinite.size;
   size.width = CGRectGetWidth(self.bounds);
   CGRect labelFrame = [self frameWithSafeAreaInsets:self.bounds];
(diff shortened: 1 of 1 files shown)


components/ActionSheet/src/private/MDCActionSheetHeaderView.m
```diff
@@ -68,10 +68,10 @@ - (void)layoutSubviews {
   CGSize titleSize = [_titleLabel sizeThatFits:labelFrame.size];
   CGSize messageSize = [_messageLabel sizeThatFits:labelFrame.size];
   CGRect titleFrame = CGRectMake(LeadingPadding + labelFrame.origin.x, TopStandardPadding,
-                                 titleSize.width, titleSize.height);
+                                 labelFrame.size.width, titleSize.height);
```
ianegordon, line 72: Not just this line, but a few places in this method.
We cannot directly access the data of a CGRect unless it has been run through CGRectStandardize().  
Either do so, or use CGRectGetHeight(), CGRectMinX(), etc.

ianegordon (dismissed):

ianegordon (approved):

PR #4940, opened 2018-08-29
Release candidate

randallli:
todo: fill this out


romoore (dismissed):
Let's not revert this change. The internal BUILD files just need to import the strings resources the way we did for Bazel when this landed.

components/Snackbar/tests/unit/MDCSnackbarMessageViewTests.m
```diff
@@ -73,15 +73,6 @@ - (void)testAccessibilityLabelSetFromSnackbarMessageProperty {
                         self.message.accessibilityLabel);
 }
 
-- (void)testAccessibilityHintDefaultIsNotNil {
```
romoore: Here is where I meant to put my comment....

PR #4946, opened 2018-08-29
[ActionSheet] Add example with too many options to fit on screen

codeman7:
This PR adds an example so showcase behavior when there are too many options on to present them on screen initially.

### DIFF
--- a/components/ActionSheet/examples/ActionSheetSwiftExample.swift
+++ b/components/ActionSheet/examples/ActionSheetSwiftExample.swift
@@ -23,7 +23,7 @@ class ActionSheetSwiftExample: UIViewController {
   var typographyScheme = MDCTypographyScheme()
   let tableView = UITableView()
   enum ActionSheetExampleType {
-    case typical, title, message, noIcons, titleAndMessage, dynamicType, delayed
+    case typical, title, message, noIcons, titleAndMessage, dynamicType, delayed, thirtyOptions
   }
   typealias ExamplesTuple = (label: String, type: ActionSheetExampleType)
   let data: [ExamplesTuple] = [
@@ -33,7 +33,8 @@ class ActionSheetSwiftExample: UIViewController {
     ("No Icons", .noIcons),
     ("With Title and Message", .titleAndMessage),
     ("Dynamic Type Enabled", .dynamicType),
-    ("Delayed", .delayed)
+    ("Delayed", .delayed),
+    ("Thirty options", .thirtyOptions)
   ]
   let cellIdentifier = "BaseCell"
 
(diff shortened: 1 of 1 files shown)


jverkoey:
Screenshots please.

components/ActionSheet/examples/ActionSheetSwiftExample.swift
```diff
@@ -204,4 +207,16 @@ extension ActionSheetSwiftExample {
...
   }
+
+  static func thirtyOptions() -> MDCActionSheetController {
+    let actionSheet = MDCActionSheetController(title: "Action sheet", message: messageString)
+    for i in 1...30 {
+      let action = MDCActionSheetAction(title: "Action \(i)",
+                                        image: UIImage(named: "Home")!,
+                                        handler: nil)
+      actionSheet.addAction(action)
+      }
```
jverkoey: [stylenit] Indentation.

jverkoey (approved):

components/ActionSheet/examples/ActionSheetSwiftExample.swift
```diff
@@ -33,7 +33,8 @@ class ActionSheetSwiftExample: UIViewController {
     ("No Icons", .noIcons),
     ("With Title and Message", .titleAndMessage),
     ("Dynamic Type Enabled", .dynamicType),
-    ("Delayed", .delayed)
+    ("Delayed", .delayed),
+    ("Thirty options", .thirtyOptions)
```
yarneo: nit: capitalize options

components/ActionSheet/examples/ActionSheetSwiftExample.swift
```diff
@@ -204,4 +207,16 @@ extension ActionSheetSwiftExample {
     actionSheet.addAction(actionThree)
     return actionSheet
   }
+
+  static func thirtyOptions() -> MDCActionSheetController {
+    let actionSheet = MDCActionSheetController(title: "Action sheet", message: messageString)
+    for i in 1...30 {
+      let action = MDCActionSheetAction(title: "Action \(i)",
+                                        image: UIImage(named: "Home")!,
```
yarneo: instead of force unwrapping check outside the for loop if this image exists

yarneo: `guard` or `if let`

codeman7: If we do a check outside the loop then it will force us to have two `for` loops, instead I removed the `!` because the image can be `nil`.

components/ActionSheet/examples/ActionSheetSwiftExample.swift
codeman7: Done

components/ActionSheet/examples/ActionSheetSwiftExample.swift
codeman7: Done

codeman7:
simulator screen shot - iphone x - 2018-08-29 at 11 06 20
simulator screen shot - iphone x - 2018-08-29 at 11 06 26

PR #4984, opened 2018-08-31
[ActionSheet] Add properties to allow for theming

### LINKED ISSUE #4903: [ActionSheet] Expose properties to enable theming

codeman7:
This PR adds properties to the components to allow for theming.

Closes #4903

### DIFF
--- a/components/ActionSheet/src/MDCActionSheetController.h
+++ b/components/ActionSheet/src/MDCActionSheetController.h
@@ -176,11 +176,6 @@ __attribute__((objc_subclassing_restricted))
  */
 @property(nonatomic, strong, nullable) UIColor *actionTintColor;
 
-/**
- The rendering mode that will be applied to the image.
- */
-@property(nonatomic, nonnull) UIImageRenderingMode *imageRenderingMode;
-
 @property(nonatomic, strong, readonly, nonnull)
     MDCBottomSheetTransitionController *transitionController;
 
--- a/components/ActionSheet/src/MDCActionSheetController.h
+++ b/components/ActionSheet/src/MDCActionSheetController.h
@@ -162,7 +162,7 @@ __attribute__((objc_subclassing_restricted))
 @property(nonatomic, strong, nullable) UIColor *titleTextColor;
 
 /**
- The color applied to the message of the actio sheet controller.
+ The color applied to the message of the action sheet controller.
  */
 @property(nonatomic, strong, nullable) UIColor *messageTextColor;
 
--- a/components/ActionSheet/src/MDCActionSheetController.h
+++ b/components/ActionSheet/src/MDCActionSheetController.h
@@ -140,7 +140,28 @@ __attribute__((objc_subclassing_restricted))
  */
 @property(nonatomic, nullable, strong) UIFont *actionFont;
 
+/**
+ The color applied to the sheet view of the action sheet controller.
+ */
 @property(nonatomic, nonnull, strong) UIColor *backgroundColor;
+
+/**
+ The color applied that will be applied to message if there is a message. If there is no message
+ then this color will be applied to the title.
+ */
+@property(nonatomic, strong, nullable) UIColor *primaryTextColor;
+/**
+ The color applied to the title if there is both a title and a message.
+ */
+@property(nonatomic, strong, nullable) UIColor *secondaryTextColor;
+/**
+ The color applied to the item list labels of Alert Controller.
+ */
+@property(nonatomic, strong, nullable) UIColor *actionColor;
+/**
+ The color applied to the icon items of Alert Controller.
+ */
+@property(nonatomic, strong, nullable) UIColor *imageColor;
 
 @property(nonatomic, strong, readonly, nonnull)
     MDCBottomSheetTransitionController *transitionController;
--- a/components/ActionSheet/src/MDCActionSheetController.h
+++ b/components/ActionSheet/src/MDCActionSheetController.h
@@ -146,17 +146,6 @@ __attribute__((objc_subclassing_restricted))
 @property(nonatomic, nonnull, strong) UIColor *backgroundColor;
 
 /**
- The color applied that will be applied to message if there is a message. If there is no message
- then this color will be applied to the title.
- */
-@property(nonatomic, strong, nullable) UIColor *primaryTextColor;
-
-/**
- The color applied to the title if there is both a title and a message.
- */
-@property(nonatomic, strong, nullable) UIColor *secondaryTextColor;
-
-/**
  The color applied to the title of the action sheet controller.
  */
 @property(nonatomic, strong, nullable) UIColor *titleTextColor;
--- a/components/ActionSheet/src/MDCActionSheetController.h
+++ b/components/ActionSheet/src/MDCActionSheetController.h
@@ -222,7 +222,7 @@ typedef void (^MDCActionSheetHandler)(MDCActionSheetAction *_Nonnull action);
  Action sheet actions must have a title that will be set within actionWithTitle:image:handler:
  method.
  */
-@property(nonatomic, nonnull, readonly) NSString *title;
+@property(nonatomic, nonnull, copy, readonly) NSString *title;
 
 /**
  Image of the list item shown on the action sheet.
--- a/components/ActionSheet/src/MDCActionSheetController.h
+++ b/components/ActionSheet/src/MDCActionSheetController.h
@@ -150,18 +150,36 @@ __attribute__((objc_subclassing_restricted))
  then this color will be applied to the title.
  */
 @property(nonatomic, strong, nullable) UIColor *primaryTextColor;
+
 /**
  The color applied to the title if there is both a title and a message.
  */
 @property(nonatomic, strong, nullable) UIColor *secondaryTextColor;
+
 /**
- The color applied to the item list labels of Alert Controller.
+ The color applied to the title of the action sheet controller.
  */
-@property(nonatomic, strong, nullable) UIColor *actionColor;
+@property(nonatomic, strong, nullable) UIColor *titleTextColor;
+
 /**
- The color applied to the icon items of Alert Controller.
+ The color applied to the message of the actio sheet controller.
  */
-@property(nonatomic, strong, nullable) UIColor *imageColor;
+@property(nonatomic, strong, nullable) UIColor *messageTextColor;
+
+/**
+ The color applied to the item list labels of action sheet controller.
+ */
+@property(nonatomic, strong, nullable) UIColor *actionTextColor;
+
+/**
+ The color applied to the icon items of action sheet controller.
+ */
+@property(nonatomic, strong, nullable) UIColor *actionTintColor;
+
+/**
+ The rendering mode that will be applied to the image.
+ */
+@property(nonatomic, nonnull) UIImageRenderingMode *imageRenderingMode;
 
 @property(nonatomic, strong, readonly, nonnull)
     MDCBottomSheetTransitionController *transitionController;
@@ -209,7 +227,7 @@ typedef void (^MDCActionSheetHandler)(MDCActionSheetAction *_Nonnull action);
  Action sheet actions must have a title that will be set within actionWithTitle:image:handler:
  method.
  */
-@property (nonatomic, nonnull, readonly) NSString *title;
+@property(nonatomic, nonnull, readonly) NSString *title;
 
 /**
  Image of the list item shown on the action sheet.
@@ -217,7 +235,7 @@ typedef void (^MDCActionSheetHandler)(MDCActionSheetAction *_Nonnull action);
  Action sheet actions must have an image that will be set within actionWithTitle:image:handler:
  method.
 */
-@property (nonatomic, nullable, readonly) UIImage *image;
+@property(nonatomic, nullable, readonly) UIImage *image;
 
 /**
  The @c accessibilityIdentifier for the view associated with this action.
--- a/components/ActionSheet/src/MDCActionSheetController.m
+++ b/components/ActionSheet/src/MDCActionSheetController.m
@@ -317,22 +317,6 @@ static NSString *const ReuseIdentifier = @"BaseCell";
   return self.view.backgroundColor;
 }
 
-- (void)setPrimaryTextColor:(UIColor *)primaryTextColor {
-  _header.primaryColor = primaryTextColor;
-}
-
-- (UIColor *)primaryTextColor {
-  return _header.primaryColor;
-}
-
-- (void)setSecondaryTextColor:(UIColor *)secondaryTextColor {
-  _header.secondaryColor = secondaryTextColor;
-}
-
-- (UIColor *)secondaryTextColor {
-  return _header.secondaryColor;
-}
-
 - (void)setTitleTextColor:(UIColor *)titleTextColor {
   _header.titleTextColor = titleTextColor;
 }
--- a/components/ActionSheet/src/MDCActionSheetController.m
+++ b/components/ActionSheet/src/MDCActionSheetController.m
@@ -265,7 +265,6 @@ static NSString *const ReuseIdentifier = @"BaseCell";
   cell.actionFont = self.actionFont;
   cell.actionTextColor = self.actionTextColor;
   cell.contentView.tintColor = self.actionTintColor;
-  cell.imageRenderingMode = self.imageRenderingMode;
   cell.accessibilityIdentifier = action.accessibilityIdentifier;
   return cell;
 }
@@ -345,15 +344,6 @@ static NSString *const ReuseIdentifier = @"BaseCell";
 - (UIColor *)messageTextColor {
   return _header.messageTextColor;
 }
-
-/*- (void)setImageRenderingMode:(UIImageRenderingMode *)imageRenderingMode {
-  //_imageRenderingMode = imageRenderingMode;
-  [self.view setNeedsLayout];
-}
-
-- (UIImageRenderingMode *)imageRenderingMode {
-  return self.imageRenderingMode;
-}*/
 
 #pragma mark - Dynamic Type
 
--- a/components/ActionSheet/src/MDCActionSheetController.m
+++ b/components/ActionSheet/src/MDCActionSheetController.m
@@ -73,7 +73,6 @@ static NSString *const ReuseIdentifier = @"BaseCell";
 }
 
 @synthesize mdc_adjustsFontForContentSizeCategory = _mdc_adjustsFontForContentSizeCategory;
-@synthesize backgroundColor = _backgroundColor;
 
 + (instancetype)actionSheetControllerWithTitle:(NSString *)title message:(NSString *)message {
   return [[MDCActionSheetController alloc] initWithTitle:title message:message];
@@ -306,13 +305,8 @@ static NSString *const ReuseIdentifier = @"BaseCell";
 
 - (void)setBackgroundColor:(UIColor *)backgroundColor {
   _backgroundColor = backgroundColor;
-  self.view.backgroundColor = backgroundColor;
   _tableView.backgroundColor = backgroundColor;
   _header.backgroundColor = backgroundColor;
-}
-
-- (UIColor *)backgroundColor {
-  return _backgroundColor;
 }
 
 - (void)setPrimaryTextColor:(UIColor *)primaryTextColor {
--- a/components/ActionSheet/src/MDCActionSheetController.m
+++ b/components/ActionSheet/src/MDCActionSheetController.m
@@ -73,6 +73,7 @@ static NSString *const ReuseIdentifier = @"BaseCell";
 }
 
 @synthesize mdc_adjustsFontForContentSizeCategory = _mdc_adjustsFontForContentSizeCategory;
+@synthesize backgroundColor = _backgroundColor;
 
 + (instancetype)actionSheetControllerWithTitle:(NSString *)title message:(NSString *)message {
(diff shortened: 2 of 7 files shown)


components/ActionSheet/src/private/MDCActionSheetItemTableViewCell.m
```diff
@@ -148,14 +148,13 @@ - (void)layoutSubviews {
...
-
-  _imageView.image = _itemAction.image;
+  _imageView.image = [_itemAction.image imageWithRenderingMode:UIImageRenderingModeAlwaysTemplate];
 }
 
 - (void)setAction:(MDCActionSheetAction *)action {
   _itemAction = [action copy];
   _textLabel.text = _itemAction.title;
-  _imageView.image = _itemAction.image;
+  _imageView.image = [_itemAction.image imageWithRenderingMode:UIImageRenderingModeAlwaysTemplate];
```
romoore: Let's allow clients choose their rendering mode. They may have a branded image that should retain its full color palette.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -140,8 +140,29 @@ __attribute__((objc_subclassing_restricted))
...
+/**
+ The color applied to the sheet view of the action sheet controller.
+ */
 @property(nonatomic, nonnull, strong) UIColor *backgroundColor;
 
+/**
+ The color applied that will be applied to message if there is a message. If there is no message
+ then this color will be applied to the title.
+ */
+@property(nonatomic, strong, nullable) UIColor *primaryTextColor;
```
romoore: Please add unit tests for all of these APIs.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -140,8 +140,29 @@ __attribute__((objc_subclassing_restricted))
...
+ */
+@property(nonatomic, strong, nullable) UIColor *secondaryTextColor;
+/**
+ The color applied to the item list labels of Alert Controller.
+ */
+@property(nonatomic, strong, nullable) UIColor *actionColor;
+/**
+ The color applied to the icon items of Alert Controller.
+ */
+@property(nonatomic, strong, nullable) UIColor *imageColor;
```
romoore: Is this a tint color? If it is, let's keep `tint` in the name somewhere so clients know what it will do. Tint has a very specific application to images in UIKit.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -140,8 +140,29 @@ __attribute__((objc_subclassing_restricted))
...
+ */
+@property(nonatomic, strong, nullable) UIColor *primaryTextColor;
+/**
+ The color applied to the title if there is both a title and a message.
+ */
+@property(nonatomic, strong, nullable) UIColor *secondaryTextColor;
+/**
+ The color applied to the item list labels of Alert Controller.
+ */
+@property(nonatomic, strong, nullable) UIColor *actionColor;
```
romoore: Is this a text color? Maybe we can call it `actionTextColor`?

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -140,8 +140,29 @@ __attribute__((objc_subclassing_restricted))
  */
 @property(nonatomic, nullable, strong) UIFont *actionFont;
 
+/**
+ The color applied to the sheet view of the action sheet controller.
+ */
 @property(nonatomic, nonnull, strong) UIColor *backgroundColor;
 
+/**
+ The color applied that will be applied to message if there is a message. If there is no message
```
romoore: This seems unnecessarily binding.  I think it's simpler if we provide simple properties mapping onto the view elements.  

If we have `titleTextColor` and `messageTextColor`, clients can choose how to apply the colors.  They will know if they have a title (and message) and can pick colors that way.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -140,8 +140,29 @@ __attribute__((objc_subclassing_restricted))
...
+ then this color will be applied to the title.
+ */
+@property(nonatomic, strong, nullable) UIColor *primaryTextColor;
+/**
```
romoore: Nit: Please add an extra line between the previous property and the start of comments.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -140,8 +140,29 @@ __attribute__((objc_subclassing_restricted))
...
+ The color applied to the title if there is both a title and a message.
+ */
+@property(nonatomic, strong, nullable) UIColor *secondaryTextColor;
+/**
+ The color applied to the item list labels of Alert Controller.
```
romoore: Shouldn't this be "Action Sheet" or something like that?

romoore (dismissed):
Please add before/after screenshots.

components/ActionSheet/src/MDCActionSheetController.h
codeman7: https://github.com/material-components/material-components-ios/pull/4993

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.h

codeman7:
| Before | After |
| -------|------|
|simulator screen shot - iphone x - 2018-09-04 at 21 48 21|simulator screen shot - iphone x - 2018-09-04 at 21 47 10|

components/ActionSheet/src/private/MDCActionSheetItemTableViewCell.m
codeman7: I don't believe that will be needed, if it is then I will add it at a later date. This will have to be on the action items and if that is the case then tint color, actionTextColor, actionFont should all be on the action. As for now I believe maintaining this API I have proposed is the best solution.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -150,18 +150,31 @@ __attribute__((objc_subclassing_restricted))
  then this color will be applied to the title.
  */
 @property(nonatomic, strong, nullable) UIColor *primaryTextColor;
```
romoore: is `primaryTextColor` still required? It seems like it's better to have the new properties that assign values directly to the UI elements.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -188,15 +222,15 @@ typedef void (^MDCActionSheetHandler)(MDCActionSheetAction *_Nonnull action);
  Action sheet actions must have a title that will be set within actionWithTitle:image:handler:
  method.
  */
-@property (nonatomic, nonnull, readonly) NSString *title;
+@property(nonatomic, nonnull, readonly) NSString *title;
```
romoore: Since this is an `NSString`, please have it use `copy` semantics instead of `strong` (the default).

romoore: It should also be marked `nullable` since I see the constructor being described in the docs actually accepts `nullable` parameters. 

>  \+ (nonnull instancetype)actionSheetControllerWithTitle:(nullable NSString *)title;

components/ActionSheet/src/MDCActionSheetController.m
```diff
@@ -114,6 +114,7 @@ - (instancetype)initWithTitle:(NSString *)title message:(NSString *)message {
     _header.title = [title copy];
     _header.message = [message copy];
     self.backgroundColor = [UIColor whiteColor];
+    self.actionTintColor = [UIColor blackColor];
```
romoore: Please don't use property setters in initializers. Directly access the autosynthesized ivar instead.  Please also use dot syntax for properties (on UIColor in this case).

```objc
_actionTintColor = UIColor.blackColor;
```

components/ActionSheet/src/MDCActionSheetController.m
```diff
@@ -114,6 +114,7 @@ - (instancetype)initWithTitle:(NSString *)title message:(NSString *)message {
     _header.title = [title copy];
     _header.message = [message copy];
     self.backgroundColor = [UIColor whiteColor];
```
romoore: Please don't call `self.backgroundColor` in the initializer. I see it's a custom setter and the object isn't finished being initialized yet.

components/ActionSheet/src/MDCActionSheetController.h
romoore: Please remove `primaryTextColor` and `secondaryTextColor`

components/ActionSheet/src/private/MDCActionSheetHeaderView.h
```diff
@@ -34,4 +34,12 @@
 
 @property (nonatomic, strong, nonnull) UIFont *messageFont;
 
+@property(nonatomic, strong, nullable) UIColor *primaryColor;
```
romoore: Please remove `primaryColor` and `secondaryColor`.

romoore (dismissed):

components/ActionSheet/src/private/MDCActionSheetHeaderView.h
codeman7: Primary color is required for a themer.

components/ActionSheet/src/MDCActionSheetController.h
codeman7: Primary text color is required for the themer.

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.m

components/ActionSheet/src/MDCActionSheetController.m

components/ActionSheet/src/private/MDCActionSheetHeaderView.h
romoore: I believe the themer can simply set the titleTextColor and messageTextColor, correct? Why are separate properties required just for a themer?

components/ActionSheet/src/private/MDCActionSheetItemTableViewCell.m
romoore: Do you mean you will remove it at a later date? I'm proposing that the UIImage provided be the one directly applied and that we don't create a second image with a new rendering mode. This matches the behavior of UIKit and MDC components behave.

components/ActionSheet/src/private/MDCActionSheetHeaderView.m
```diff
@@ -201,4 +209,34 @@ - (void)mdc_setAdjustsFontForContentSizeCategory:(BOOL)adjusts {
...
+
+- (void)setMessageTextColor:(UIColor *)messageTextColor {
+  _messageLabel.textColor = messageTextColor;
+}
+
+- (UIColor *)messageTextColor {
+  return _messageLabel.textColor;
+}
+
+- (BOOL)labelExist:(UILabel *)label {
```
romoore: This method seems to be unused. Can it be removed?

components/ActionSheet/src/MDCActionSheetController.h
romoore: That PR has been closed. Are you planning to add the tests here to verify the behavior before merging?

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -140,8 +140,42 @@ __attribute__((objc_subclassing_restricted))
...
+
+/**
+ The color applied to the item list labels of action sheet controller.
+ */
+@property(nonatomic, strong, nullable) UIColor *actionTextColor;
+
+/**
+ The color applied to the icon items of action sheet controller.
+ */
+@property(nonatomic, strong, nullable) UIColor *actionTintColor;
```
romoore, line 166: Why not use `UIView.tintColor`? That would match what our other components do. Are we using that for some other part of the UI?

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -140,8 +140,42 @@ __attribute__((objc_subclassing_restricted))
...
+ */
+@property(nonatomic, strong, nullable) UIColor *secondaryTextColor;
+
+/**
+ The color applied to the title of the action sheet controller.
+ */
+@property(nonatomic, strong, nullable) UIColor *titleTextColor;
+
+/**
+ The color applied to the message of the actio sheet controller.
```
romoore: Nit: s/actio/action/

components/ActionSheet/src/MDCActionSheetController.m
```diff
@@ -113,7 +114,8 @@ - (instancetype)initWithTitle:(NSString *)title message:(NSString *)message {
     _header = [[MDCActionSheetHeaderView alloc] initWithFrame:CGRectZero];
     _header.title = [title copy];
     _header.message = [message copy];
-    self.backgroundColor = [UIColor whiteColor];
+    _backgroundColor = UIColor.whiteColor;
```
romoore, line 117: Do you need to set the `backgroundColor` values for `view`, `header`, and `tableView`?

components/ActionSheet/src/MDCActionSheetController.m
```diff
@@ -73,6 +73,7 @@ @implementation MDCActionSheetController {
 }
 
 @synthesize mdc_adjustsFontForContentSizeCategory = _mdc_adjustsFontForContentSizeCategory;
+@synthesize backgroundColor = _backgroundColor;
```
romoore, line 76: This doesn't seem to be necessary. Previously you were using `self.view.backgroundColor` for the getter, and the setter assigned `backgroundColor` to `self.view`, `tableView`, and `backgroundColor`

components/ActionSheet/src/MDCActionSheetController.m
```diff
@@ -301,13 +305,46 @@ - (UIFont *)messageFont {
...
 - (void)setBackgroundColor:(UIColor *)backgroundColor {
+  _backgroundColor = backgroundColor;
   self.view.backgroundColor = backgroundColor;
   _tableView.backgroundColor = backgroundColor;
   _header.backgroundColor = backgroundColor;
 }
 
 - (UIColor *)backgroundColor {
-  return self.view.backgroundColor;
+  return _backgroundColor;
```
romoore, line 317: I think it makes sense to leave this as a pass-through to `self.view.backgroundColor`.

romoore (dismissed):

components/ActionSheet/src/private/MDCActionSheetHeaderView.h
codeman7: If there is only one label then that will be `onSurfaceColor` with `0.6` alpha. If there is both then the title will be `0.87` alpha. It depends if there is both labels or not and if a client applies a themer and then sets a label the theme won't be applied. `titleTextColor` and `messageTextColor` are not required, `primary` and `secondary` are.

components/ActionSheet/src/MDCActionSheetController.h
codeman7: I will reopen that PR once this one lands.

components/ActionSheet/src/private/MDCActionSheetHeaderView.m
codeman7: Done

components/ActionSheet/src/MDCActionSheetController.h

components/ActionSheet/src/MDCActionSheetController.m
codeman7: Done

components/ActionSheet/src/MDCActionSheetController.m
codeman7: This has been removed.

components/ActionSheet/src/MDCActionSheetController.h
codeman7: The tint color is being applied to the image. A client may need an image that is red and an label that is blue.

components/ActionSheet/src/MDCActionSheetController.m
codeman7: Removed `self.view.backgroundColor`.

codeman7:
Test can be tracked at https://github.com/material-components/material-components-ios/issues/5039

components/ActionSheet/src/MDCActionSheetController.h
romoore: I thought UILabel doesn't apply `tintColor` to the text color? At least I've never been able to get it to work.  Are you finding that `tintColor` is being applied to the label's text color in some circumstances?

components/ActionSheet/src/MDCActionSheetController.m
romoore: I'm confused by this change. It looks like you're no longer setting the view's `backgroundColor` property, is that correct? Should these views initialize without a background color? Do `_header` and `_tableView` no longer require background colors either?

components/ActionSheet/src/MDCActionSheetController.m
```diff
@@ -301,13 +304,41 @@ - (UIFont *)messageFont {
 }
 
 - (void)setBackgroundColor:(UIColor *)backgroundColor {
-  self.view.backgroundColor = backgroundColor;
+  _backgroundColor = backgroundColor;
   _tableView.backgroundColor = backgroundColor;
   _header.backgroundColor = backgroundColor;
 }
 
-- (UIColor *)backgroundColor {
```
romoore: Why was this removed? It seems that for clients, the value of `- [controller backgroundColor]` will no longer be affected by `- [controller setBackgroundColor:]` or am I missing something?

romoore (dismissed):
I'm most concerned about the conflicting behavior with `primaryTextColor` and `titleTextColor` and the apparent removal of support for `backgroundColor`. I would also appreciate replies to the unresolved comment threads from earlier reviews.

components/ActionSheet/src/MDCActionSheetController.h
codeman7: I can remove `titleTextColor` and `messageTextColor` because those change based on each other. I cannot remove `primaryTextColor` and `secondaryTextColor` because those are needed for the color themer. I understand that you requested `titleTextColor` and `messageTextColor` but those changed based off of the `title` and `message` property. I believe that a better API is `primaryTextColor` and `secondaryTextColor`.

components/ActionSheet/src/MDCActionSheetController.h
codeman7: I think you may be looking at the wrong initializer. The initializer for this `title` property is `(nonnull instancetype)actionWithTitle:(nonnull NSString *)title image:(nullable UIImage *)image handler:(__nullable MDCActionSheetHandler)handler;` you are referencing the initializer for `MDCActionSheetController`, this property is in `MDCActionSheetAction`.

components/ActionSheet/src/MDCActionSheetController.h
codeman7: If you are referring to the implementation why don't I apply `tintColor`, I think there is a disconnect because that is what is currently being used. If you are asking why the _API_ isn't `@property(nonatomic, strong, nullable) UIColor *tintColor` I feel that that can be confusing because clients may wonder what all this `tint` will be applied to. If you are referencing the `UIKit` equivalent `UIAlertController` with style `ActionSheet` doesn't have a `tintColor`. For now, I feel this is a better `API`.

components/ActionSheet/src/MDCActionSheetController.m
codeman7: I am now applying background color to the `view`, `_header`, and `_tableView`.

components/ActionSheet/src/MDCActionSheetController.m
codeman7: I added it back, if this is added then `@synthesize backgroundColor` is required, that is now added as well..

components/ActionSheet/src/private/MDCActionSheetItemTableViewCell.m
codeman7: If you don't set the `renderingMode` then the image will not be the correct color when you apply `tintColor` I can remove this line but in doing so it doesn't allow clients to set a `tintColor` on the images.

romoore: Clients can set the rendering mode. This matches what our other components do. The `tintColor` applied will be the one from the UIImageView, either inherited or explicitly applied.

codeman7:
I have removed `primaryTextColor` and `secondaryTextColor`.

components/ActionSheet/src/MDCActionSheetController.h
jverkoey: There's some nuance here to ensure we consider:

1. If a client sets the tintColor of the action sheet, this will be propagated to any subviews that respond to tintColor. @codeman7 If we decide to go with this explicit actionTintColor API we need to make sure it plays nicely with the existing tintColor behavior. Have you explored that already?
2. I like the idea of having an explicit API that very clearly affects one thing, but 1. makes me a bit nervous that we might collide with expectations about how tintColor is propagating to our component.

jverkoey:
This may be a good candidate PR to split apart along the lines of tint-based APIs and non-tint-based APIs because the questions around the tint-based APIs feel like they need a bit more exploration.

Issue #4964, opened 2018-08-30
[Dialogs] Expose corner radius property


galiak11:
Exposing the corner radius property in MDCAlertController.

This requires followup work on correctly applying shadows to the dialogs with rounded corners, which will be implemented in a followup PR.

Issue: # - 4964
Issue: b/113257098.

PR #4988, opened 2018-08-31
[Dialogs] Expose corner radius property

galiak11:
Exposing a corner radius property on MDCAlertController.

This requires followup work on correctly applying a shadow to rounded corner dialogs, which will be implemented in a followup PR.

Issue: #4964
Issue: b/113257098


romoore:
@galiak11 Can you please fix-up the commit summary? It seems to be just a copy-paste of the Issue, which is self-referential.

components/Dialogs/src/MDCAlertController.h
```diff
@@ -88,6 +88,9 @@
 /** The color applied to the button ink effect of Alert Controller.*/
 @property(nonatomic, strong, nullable) UIColor *buttonInkColor;
 
+/** The corner radius of Alert Controller. Default radius is CALayer's default radius (0.0).*/
+@property(nonatomic, assign) CGFloat cornerRadius;
```
romoore: Please add unit tests for the new API.

components/Dialogs/src/MDCAlertController.h
```diff
@@ -88,6 +88,9 @@
 /** The color applied to the button ink effect of Alert Controller.*/
 @property(nonatomic, strong, nullable) UIColor *buttonInkColor;
 
+/** The corner radius of Alert Controller. Default radius is CALayer's default radius (0.0).*/
```
romoore: Please format these comments. 

Suggested:
```objc
/** 
 Pass-through property for the @c MDCAlertController's view's CALayer property.

 @seealso CALayer for the default value and behavior of cornerRadius.
 */
```

romoore (changes requested):
Please add before/after images (or gifs) for this change. Please also consider adding/updating a Dragon example for this API.

galiak11:
@romoore Can you please elaborate regarding the "Can you please fix-up the commit summary?" comment?  Current commit pretty much describes the work that was done in this PR - which is exposing the corner radius.
Also, in regards to the before/after images - there are none since this PR doesn't introduce any UI changes. An upcoming PR updating the examples will have most likely those.

components/Dialogs/src/MDCAlertController.m
```diff
@@ -212,6 +212,19 @@ - (void)setButtonTitleColor:(UIColor *)buttonColor {
...
 
+- (void)setCornerRadius:(CGFloat)cornerRadius {
+  _cornerRadius = cornerRadius;
+  if (self.alertView) {
+    self.alertView.cornerRadius = cornerRadius;
+  }
+  // make sure to pass through the new shape to the dialog's tracking & shadow layers
+  MDCDialogPresentationController *dialogPresentationController =
+      self.mdc_dialogPresentationController;
+  if (dialogPresentationController) {
```
romoore: Please remove the `nil` check since `nil` safely receives all messages.

http://google.github.io/styleguide/objcguide.html#nil-checks

components/Dialogs/src/MDCAlertController.m
```diff
@@ -212,6 +212,19 @@ - (void)setButtonTitleColor:(UIColor *)buttonColor {
   }
 }
 
+- (void)setCornerRadius:(CGFloat)cornerRadius {
+  _cornerRadius = cornerRadius;
+  if (self.alertView) {
```
romoore: Please remove the `nil` check since `nil` safely receives all messages.

http://google.github.io/styleguide/objcguide.html#nil-checks

components/Dialogs/src/MDCAlertControllerView.h
```diff
@@ -26,6 +26,8 @@
 @property(nonatomic, strong, nullable) UIColor *buttonColor UI_APPEARANCE_SELECTOR;
 @property(nonatomic, strong, nullable) UIColor *buttonInkColor UI_APPEARANCE_SELECTOR;
 
+@property(nonatomic, assign) CGFloat cornerRadius UI_APPEARANCE_SELECTOR;
```
romoore: Do we want to extend UIAppearance support? Since this property is settable on the controller, I don't think it's absolutely necessary for styling.  

cc/ @jverkoey since @ianegordon  is OOO this week.

components/Dialogs/tests/unit/MDCAlertControllerTests.m
```diff
@@ -273,4 +273,21 @@ - (void)testAccessibilityIdentifiersAppliesToAlertControllerViewButtons {
   XCTAssertEqualObjects(button2.accessibilityIdentifier, @"A");
 }
 
+- (void)testCornerRadius {
```
romoore: Please add a test to verify that the default radius is 0 since it is documented in the public header.

components/Dialogs/tests/unit/MDCAlertControllerTests.m
```diff
@@ -273,4 +273,21 @@ - (void)testAccessibilityIdentifiersAppliesToAlertControllerViewButtons {
...
+  // Given
+  CGFloat cornerRadius = 36.0;
```
romoore: Please explicitly cast `36.0` to (CGFloat) or we are likely to get errors when compiling for 32-bit architectures. Implicit loss of precision is configured as an error for MDC-iOS.

Alternatively, you can just use the integral literal `36` since that can implicitly convert to single- and double-precision floats without loss of precision.

components/Dialogs/tests/unit/MDCAlertControllerTests.m
```diff
@@ -273,4 +273,21 @@ - (void)testAccessibilityIdentifiersAppliesToAlertControllerViewButtons {
...
+                                                                   message:@"message"];
+  [alert addAction:[MDCAlertAction actionWithTitle:@"action1" handler:nil]];
+  [alert addAction:[MDCAlertAction actionWithTitle:@"action2" handler:nil]];
+
+  // When
+  alert.cornerRadius = cornerRadius;
+
+  // Then
+  MDCAlertControllerView *view = (MDCAlertControllerView *)alert.view;
+  XCTAssertEqual(view.layer.cornerRadius, cornerRadius);
```
romoore: Please use `XCTAssertEqualWithAccuracy` when comparing floating point numbers.

components/Dialogs/src/private/MDCAlertControllerView+Private.m
```diff
@@ -255,6 +255,15 @@ - (void)setButtonInkColor:(UIColor *)color {
   }
 }
 
+- (CGFloat)cornerRadius {
+  return self.layer.cornerRadius;
+}
+
+- (void)setCornerRadius:(CGFloat)cornerRadius {
+  self.layer.cornerRadius = cornerRadius;
```
romoore: Suggestion/optimization: Please use `MDCCGFloatEqual(CGFloat, CGFloat)` to check if the new and old values are different. If they are the same, let's return early and avoid calling `setNeedsLayout`.

components/Dialogs/src/MDCAlertController.h
```diff
@@ -88,6 +88,9 @@
 /** The color applied to the button ink effect of Alert Controller.*/
 @property(nonatomic, strong, nullable) UIColor *buttonInkColor;
 
+/** The corner radius applied to the Alert Controller shape. Default to 0 (no round corners) */
```
romoore: Should this be "Alert Controller **view**" instead of "shape"?  It doesn't look like the property is being applied to any Shaping scheme objects.

romoore (approved):

romoore:
Good to know there are no before/after changes.

Regarding the commit summary, it seems identical to the issue being referenced. I suspect that either the Issue or this PR have the wrong description.

components/Dialogs/src/MDCAlertControllerView.h
galiak11: Just note that corner radius is also extended on MDCCard. Are dialogs different?

romoore: When that Cards API was written, I believe we still supported UIAppearance. The commit [history shows that `cornerRadius` was added on February 1, 2018](https://github.com/material-components/material-components-ios/blame/develop/components/Cards/src/MDCCard.h#L27).   UIAppearance [support was dropped](https://github.com/material-components/material-components-ios/pull/3212) [somewhere in mid-March](https://www.pivotaltracker.com/n/projects/2150420/stories/155533126/comments/187416781).

components/Dialogs/src/private/MDCAlertControllerView+Private.m
```diff
@@ -19,6 +19,7 @@
 
 #import "MaterialButtons.h"
 #import "MaterialTypography.h"
```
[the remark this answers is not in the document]
romoore: Order of imports: MaterialMath before MaterialTypography.

components/Dialogs/src/MDCAlertControllerView.h
galiak11: Okay. Should we remove the UI_APPEARANCE_SELECTOR from all other MDCAlertControllerView properties, like titleFont, titleColor, etc. ?

romoore: I think that's out of scope for this PR, and in general we're not removing anything unless it causes problems with another feature/bug fix. I understand we're just not adding it to any new APIs.

/cc @jverkoey to confirm our UIAppearance support policy.

Issue #4945, opened 2018-08-29
[ActionSheet] Open up to correct height


codeman7:
An ActionSheet should open up to either:
MAX: Half the screen height minus half a action item height plus safe area
or 
MIN: Height of content plus safe area
Whatever is smaller of the two.

## Reproduction steps

1. Go to Dragons
2. Tap on ActionSheet
3. Open up thirty options

## Expected behavior

Sheet should open up to half the screen height minus half an action item height.

## Actual behavior

Sheet opens up to full screen 

## Platform (please complete the following information)

 - Device: All
 - OS: All

PR #5013, opened 2018-09-05
[BottomSheet] Allow for custom height when initially opened

codeman7:
This will allow clients and other components that to open to a custom height instead of half the screen height or the content height.

Part of #4945


codeman7:
| Before | After |
| ------- | ----- |
|simulator screen shot - iphone 8 - 2018-09-05 at 13 03 38|simulator screen shot - iphone 8 - 2018-09-05 at 13 04 42|

PR #5021, opened 2018-09-05
[BottomNavigation] Adding functionality to MDCBottomNavigationBar to be able to access the tappable view associated with a tab item

### LINKED ISSUE #4824: BottomNavigation should provide a way to get to the viewItems in order to connect to feature highlight
## Feature Request/Suggestion
How do I use FeatureHighlight on a MDCBottomNavigationBar?

I need the internal itemView, which is not exposed.

Can we add a method 
- (UIView *)viewForItem:(UITabBarItem *)item;

## Additional context

more info in internal yaqs/6188300342919168

andrewplai:
This will allow MDCBottomNavigationBar to work with popover / tooltip libraries which require an anchor view for positioning (i.e, if we want to display a tooltip over one of the tabs).

Closes #4824


components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -320,6 +320,15 @@ - (UIEdgeInsets)mdc_safeAreaInsets {
   return insets;
 }
 
+- (UIView *)viewForItem:(UITabBarItem *)item {
+  NSUInteger itemIndex = [_items indexOfObject:item];
```
romoore: Please add an explicit check for `NSNotFound`

romoore (changes requested):
Thank you for the PR! This looks like a great improvement to BottomNavigation.  The only thing missing is some unit tests.  Please add tests to exercise the new method, especially the error handling (when a UITabBarItem is not found).

Not a problem as far as I can tell, but there seems to be something a bit odd with your PR. It appears this might be a very large merge commit that includes your proposed changes. Feel free to push up a clean PR (or force-push on this branch) if you need to untangle things. :)

andrewplai:
Sounds good!   And I did notice that this PR did have a bunch of files from a merge on it as well, so I've gone ahead and created a new [PR](https://github.com/material-components/material-components-ios/pull/5061) for this to keep things cleaner.

PR #5023, opened 2018-09-06
[BottomSheet] Allow for custom height when initially opened

codeman7:
This will allow clients and other components that to open to a custom height instead of half the screen height or the content height.

**Test are WIP** Everything else you can review. **Test are WIP**

Part of #4945

| Before | After |
| ------- | ----- |
|simulator screen shot - iphone 8 - 2018-09-05 at 13 03 38|simulator screen shot - iphone 8 - 2018-09-05 at 13 04 42|

### DIFF
--- a/components/BottomSheet/src/MDCBottomSheetPresentationController.h
+++ b/components/BottomSheet/src/MDCBottomSheetPresentationController.h
@@ -66,8 +66,12 @@
 
 /**
  This is used to set a custom height on the sheet view.
+
+ @note This will override setting the preferredContentSize if a positive value is passed.
+ @note If a non-positive value is passed then the sheet will open up to either half the screen
+ height or the size of the contentViewController whatever value is smaller.
  */
-@property(nonatomic) CGFloat preferredSheetHeight;
+@property(nonatomic, assign) CGFloat preferredSheetHeight;
 
 /**
  If @c YES, then the dimmed scrim view will act as an accessibility element for dismissing the
--- a/components/BottomSheet/src/MDCBottomSheetPresentationController.h
+++ b/components/BottomSheet/src/MDCBottomSheetPresentationController.h
@@ -65,6 +65,11 @@
 @property(nonatomic, assign) BOOL dismissOnBackgroundTap;
 
 /**
+ This is used to set a custom height on the sheet view.
+ */
+@property(nonatomic) CGFloat preferredSheetHeight;
+
+/**
  If @c YES, then the dimmed scrim view will act as an accessibility element for dismissing the
  bottom sheet.
 
--- a/components/BottomSheet/src/MDCBottomSheetPresentationController.m
+++ b/components/BottomSheet/src/MDCBottomSheetPresentationController.m
@@ -169,10 +169,10 @@ static UIScrollView *MDCBottomSheetGetPrimaryScrollView(UIViewController *viewCo
 
 - (void)updatePreferredSheetHeight {
   CGFloat preferredContentHeight;
-  if (self.preferredSheetHeight == 0.f) {
-    preferredContentHeight = self.presentedViewController.preferredContentSize.height;
-  } else {
+  if (self.preferredSheetHeight > 0.f) {
     preferredContentHeight = self.preferredSheetHeight;
+  } else {
+    preferredContentHeight = self.presentedViewController.preferredContentSize.height;
   }
 
   // If |preferredSheetHeight| has not been specified, use half of the current height.
--- a/components/BottomSheet/src/MDCBottomSheetPresentationController.m
+++ b/components/BottomSheet/src/MDCBottomSheetPresentationController.m
@@ -168,7 +168,12 @@ static UIScrollView *MDCBottomSheetGetPrimaryScrollView(UIViewController *viewCo
 }
 
 - (void)updatePreferredSheetHeight {
-  CGFloat preferredContentHeight = self.presentedViewController.preferredContentSize.height;
+  CGFloat preferredContentHeight;
+  if (self.preferredSheetHeight == 0.f) {
+    preferredContentHeight = self.presentedViewController.preferredContentSize.height;
+  } else {
+    preferredContentHeight = self.preferredSheetHeight;
+  }
 
   // If |preferredSheetHeight| has not been specified, use half of the current height.
   if (MDCCGFloatEqual(preferredContentHeight, 0)) {
@@ -232,6 +237,11 @@ static UIScrollView *MDCBottomSheetGetPrimaryScrollView(UIViewController *viewCo
 
 - (UIAccessibilityTraits)scrimAccessibilityTraits {
   return _scrimAccessibilityTraits;
+}
+
+- (void)setPreferredSheetHeight:(CGFloat)preferredSheetHeight {
+  _preferredSheetHeight = preferredSheetHeight;
+  [self updatePreferredSheetHeight];
 }
 
 #pragma mark - MDCSheetContainerViewDelegate
--- a/components/BottomSheet/src/MDCBottomSheetTransitionController.h
+++ b/components/BottomSheet/src/MDCBottomSheetTransitionController.h
@@ -46,9 +46,14 @@
 @property(nonatomic, assign) BOOL dismissOnBackgroundTap;
 
 /**
- The height of the content view.
+ This is used to set a custom height on the sheet view. This is can be used to set the initial
+ height when the ViewController is presented.
+
+ @note This will override setting the preferredContentSize if a positive value is passed.
+ @note If a non-positive value is passed then the sheet will open up to either half the screen
+ height or the size of the contentViewController whatever value is smaller.
  */
-@property(nonatomic) CGFloat preferredSheetHeight;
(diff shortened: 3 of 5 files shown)


romoore:
In the screenshots, what are the before/after comparing?  Was there some functionality that was broken previously that you can show now works correctly? Should we demonstrate setting a custom height position that differs from the content height?

components/BottomSheet/src/MDCBottomSheetPresentationController.m
```diff
@@ -168,7 +168,12 @@ - (void)viewWillTransitionToSize:(CGSize)size
 }
 
 - (void)updatePreferredSheetHeight {
-  CGFloat preferredContentHeight = self.presentedViewController.preferredContentSize.height;
+  CGFloat preferredContentHeight;
+  if (self.preferredSheetHeight == 0.f) {
```
romoore: Please consider checking if the height is less than or equal to zero. I don't think we should present a view with a negative `preferredSheetHeight`.

components/BottomSheet/src/MDCBottomSheetPresentationController.h
```diff
@@ -64,6 +64,11 @@
  */
 @property(nonatomic, assign) BOOL dismissOnBackgroundTap;
 
+/**
+ This is used to set a custom height on the sheet view.
```
romoore, line 68: Please expand on this a bit to document how it's used in conjunction with `preferredContentSize`. Specifically, it seems to override that value and is not (for instance) used as a "backup" value if the VC gives an invalid value `preferredContentSize`.

Please also document that it will be ignored if it's non-positive.

components/BottomSheet/src/MDCBottomSheetTransitionController.h
```diff
@@ -45,6 +45,11 @@
  */
 @property(nonatomic, assign) BOOL dismissOnBackgroundTap;
 
+/**
+ The height of the content view.
```
romoore: Please expand on this documentation as well. I believe this is a pass-through property to the presentation controller and so (if set) will affect the initial height of the sheet when presented.

romoore (dismissed):
Please add unit tests for the new API.

components/BottomSheet/src/MDCBottomSheetPresentationController.m
codeman7: Done.

components/BottomSheet/src/MDCBottomSheetPresentationController.h
codeman7: Done

components/BottomSheet/src/MDCBottomSheetTransitionController.h
codeman7: Done

codeman7:
simulator screen shot - iphone x - 2018-09-07 at 09 49 43

If you set the `preferredSheetHeight` to `50`. With these changes this can only be done by changing the implementation of `MDCBottomSheet`.

codeman7:
This is changing the `presentationController` and `transitionController`. I don't see a way to write a test that would not require making a new class that uses these for presentation and transitioning. Within that class there would have to be a way to set a property i.e. `preferredSheetHeight` and test against that. In order for `MDCBottomSheet` to use these new properties, within `viewDidLayoutSubviews` it will have to set the desired value on `self.mdc_presentationController.preferredSheetHeight`.  If a property is added to MDCBottomSheet to enable this behavior then a test would be an appropriate step but at this stage I disagree. I may not have the depth of knowledge of bottom sheet required to write this test, if that is the case then I will gladly learn how this can be done.

romoore:
Since MDCBottomSheetPresentationController isn't marked with subclassing restricted, you should be able to write a Fake subclass that returns a value for `presentedViewController` that will meet the needs of these tests.

It doesn't look like your change is touching `viewDidLayoutSubviews`, though there may be some dependency that I can't see. I believe you should be able to write/update tests for the following use cases without any dependency on a host application:

* Test the behavior of `-[MDCBottomSheetPresentationController updatePreferredSheetHeight]` to test each of the affected code paths.
* Checking that setting `- MDCBottomSheetTransitionController.preferredSheetHeight` will change the value in the associated MDCBottomSheetPresentationController instance.

romoore (changes requested):
Since MDCBottomSheetPresentationController isn't marked with subclassing restricted, you should be able to write a Fake subclass that returns a value for presentedViewController that will meet the needs of these tests.

It doesn't look like your change is touching `viewDidLayoutSubviews`, though there may be some dependency that I can't see so I could be wrong. I believe you should be able to write/update tests for the following use cases without any dependency on a host application:

* Test the behavior of -`[MDCBottomSheetPresentationController updatePreferredSheetHeight]` to test each of the affected code paths.
* Checking that setting `- MDCBottomSheetTransitionController.preferredSheetHeight` will change the value in the associated MDCBottomSheetPresentationController instance.

codeman7:
Please don't review the test, yet.

codeman7:
Don't see a way to write test.

romoore:
What problem are you encountering?

components/BottomSheet/src/MDCBottomSheetPresentationController.m
```diff
@@ -168,7 +168,12 @@ - (void)viewWillTransitionToSize:(CGSize)size
 }
 
 - (void)updatePreferredSheetHeight {
-  CGFloat preferredContentHeight = self.presentedViewController.preferredContentSize.height;
+  CGFloat preferredContentHeight;
+  if (self.preferredSheetHeight > 0.f) {
+    preferredContentHeight = self.preferredSheetHeight;
+  } else {
+    preferredContentHeight = self.presentedViewController.preferredContentSize.height;
+  }
```
jverkoey, line 176: Can you help me better understand when someone would use `preferredSheetHeight` instead of adjusting `self.presentedViewController.preferredContentSize`?

I ask because this code path implies that they're semantically similar and I'm curious if we can remove the need for an extra API.

jverkoey:
@romoore Looking at the code I think @codeman7 will have a hard time getting the `_sheetView` initialization code path to execute without significantly refactoring this change.

https://github.com/codeman7/material-components-ios/blob/aca4509e46ca31a12e7058df2bb34eaf1a6a12f6/components/BottomSheet/src/MDCBottomSheetPresentationController.m#L75-L103

I believe that such a refactoring represents a separate and probably reasonable idea to improve our testing, but that this particular feature as-is can't be addressed without significant work there. This is an example where filing a task to address that issue separately will be beneficial. Whether we address that testing refactoring before or after this PR, I'll leave up to @codeman7, but it is a good improvement to make so that we can improve the testability of this component.

@romoore it sounds like you have some thoughts on how we might be able to improve the testability of this component - can you capture them in a separate issue so that @codeman7 has a good place to start from?

components/BottomSheet/src/MDCBottomSheetPresentationController.m
codeman7: `preferredContentSize` is an Apple API so we don't know exactly what all it does. When you set the `preferredContentSize` it also set's the `frame` and `bounds` of the `viewController`. This becomes an issue on screen size changes. Within ActionSheet I initially set the `preferredContentSize` but then you can't get the size of the screen if you rotate or the application window changes. According to design a list within a bottom sheet (ActionSheet) should be half the screen size minus half a cell height. For example on an iPhone X (812dp) it would be 406dp - 28dp (56dp cell) so 378dp. If you set the `preferredContentSize` you cannot handle rotations unless you recreate `setNeedsLayout` logic in a `setNeedsPreferredContentSize` model.

codeman7:
You can create a fake subclass but you still can't `present`. You can get the `presentedView` which is the `_sheetView` but you cannot `present` in a unit test. I'm sure there is a way but I am not aware of one. Additionally I have a feeling it would have to be a _hack_ on UIKit and if that changes at a later date then this test could become flake or fail even if the behavior is correct.

romoore:
I'm afraid I still don't understand which part of the tests are undesirable. To provide better clarity on what I was thinking, I pushed up #5084 which include some tests using fakes that partly exercise the new `updatePreferredSheetHeight` method.  I believe adding the remaining code flows should be straightforward.

Is there something about using fakes to create the internal conditions of these tests that are undesirable? I agree that it's a bit gnarly at the moment, but as @jverkoey  pointed out, the classes could be refactored a bit to make these things easier to test.

codeman7:
If it works and you feel sufficient tests then I am all for it. Feel free to merge #5084 I think my lack of `obj-c` knowledge blocked me from understanding what you desired. If your happy with #5084 I have no problem with it, I just need the functionality.

romoore:
Gentle ping to @jverkoey to revisit why we might not want the tests in this PR.

jverkoey:
I'm concerned that the proposed testing implementation is not clear enough to stand on its own. It's important that @codeman7 understand the implementation of the proposed test because his understanding is a proxy for someone else's understanding in the future.

I'd like for us to land this functionality and evaluate what we can do to improve the testability of this component as a separate track of work such that more than one person understands the implementation details of the tests.

PR #5024, opened 2018-09-06
[ActionSheet] Open to the correct height

codeman7:
This will for bleeding if there are too many options to show on screen at once.

| Before | After |
| ------- | ------- |
|simulator screen shot - iphone 8 - 2018-09-05 at 15 30 55|simulator screen shot - iphone 8 - 2018-09-05 at 15 28 49|

Depends on #5023 
Part of #4945

### DIFF
--- a/components/ActionSheet/src/MDCActionSheetController.m
+++ b/components/ActionSheet/src/MDCActionSheetController.m
@@ -141,6 +141,7 @@ static NSString *const ReuseIdentifier = @"BaseCell";
 - (void)viewWillLayoutSubviews {
   [super viewWillLayoutSubviews];
 
+  [self correctSheetHeight];
   CGSize size = [_header sizeThatFits:CGRectStandardize(self.view.bounds).size];
   _header.frame = CGRectMake(0, 0, self.view.bounds.size.width, size.height);
   UIEdgeInsets insets = UIEdgeInsetsMake(_header.frame.size.height, 0, 0, 0);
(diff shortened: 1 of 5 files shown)


romoore:
Please clean up the commit summary. I'm not sure what the part about bleeding is referring to.

romoore:
In the "After" screenshot, it looks like the sheet is no longer expanding to half the screen height.  Isn't the default to open one of:
* the preferredContentSize of the VC
* half the height of the screen

components/ActionSheet/src/MDCActionSheetController.m
```diff
@@ -153,6 +154,25 @@ - (void)viewWillLayoutSubviews {
...
 
+- (void)correctSheetHeight {
+  // If there are too many options to fit on half of the screen then show as many options as
+  // possible minus half a cell, to allow for bleeding and signal to the user that the sheet is
+  // scrollable content.
+  if (_tableView.contentSize.height > (CGRectGetHeight(self.view.bounds) / 2)) {
+    CGFloat maxHeight = CGRectGetHeight(self.view.bounds) / 2;
+    CGFloat headerHeight = [_header sizeThatFits:CGRectStandardize(self.view.bounds).size].height;
+    CGFloat cellHeight = _tableView.contentSize.height / (CGFloat)_actions.count;
+    int amountOfCellsToShow = (int)(maxHeight - headerHeight) / (int)cellHeight;
```
romoore: Please use `NSInteger` instead of `int`

components/BottomSheet/src/MDCBottomSheetPresentationController.h
```diff
@@ -64,6 +64,11 @@
  */
 @property(nonatomic, assign) BOOL dismissOnBackgroundTap;
 
+/**
+ This is used to set a custom height on the sheet view.
+ */
+@property(nonatomic) CGFloat preferredSheetHeight;
```
romoore, line 70: Please mark this `nonatomic, assign` to match the file style.

romoore (dismissed):
Looks like this might have gotten a bit mangled with #5023.

Please add unit tests for the new behavior.

components/ActionSheet/src/MDCActionSheetController.m

components/BottomSheet/src/MDCBottomSheetPresentationController.h
codeman7: Done

codeman7:
My understanding after speaking to design is, list with scrollable content within a bottom sheet should open up to half of the screen height minus half of a list item or to the list height. The initial height should be the smaller of those two values. Half of the height of the view is correct. But, if there are more options that will fit on half of the height (of the view) and the list height is a factor of the _super_ view's height then a user may not be aware it is scrollable content. This new behavior is correct and makes for a better user experience in my opinion. The term _bleeding_  was told to be by design, my understanding is the _list item_ bleeds onto the view.

codeman7:
Test will be added for this behavior in a follow up PR after this lands. I don't want to block this behavior on test. The test can be tracked [here](https://github.com/material-components/material-components-ios/issues/5038).

components/ActionSheet/src/MDCActionSheetController.m
```diff
@@ -153,6 +154,25 @@ - (void)viewWillLayoutSubviews {
   _tableView.contentOffset = CGPointMake(0, -size.height);
 }
 
+- (void)correctSheetHeight {
```
romoore, line 157: If you refactor this method to return the correct sheet height as a `CGFloat` instead of reassigning it as a side-effect, it should be much easier to test.  I'm just taking a quick glance, but it seems you could get away with adding  few actions so that the tableView's `contentSize` is non-zero and setting `self.view.bounds` to something reasonable.

It looks like that is enough to trigger the new layout code and allow you to verify the behavior.

If you also wanted to test changes to `viewWillLayoutSubviews`, calling `layoutIfNeeded` after setting `self.view.frame` (or calling `setNeedsLayout`) should be enough to force a layout pass.

romoore (changes requested):
I think it's worth adding the unit tests to this change. The change itself seems primarily focused on a single method (with a side-effect in `viewWillLayoutSubviews`).  I've provided more thoughts in the inline comment, but I believe the test could be written without significant hurdles. Having the test committed with the change would help to verify the behavior is as-expected under a range of starting conditions.

codeman7:
Don't see a path forward for #5023 so closing this.

PR #5028, opened 2018-09-06
[MaskedTransition] Update Swift example to use MDCFloatingButton

codeman7:
Previously we didn't use MDCFloatingButton or any themers in our example, this adds that.

| Before | After |
| ------- | ------- |
|simulator screen shot - iphone x - 2018-09-06 at 15 32 16|simulator screen shot - iphone x - 2018-09-06 at 15 28 36|


romoore:
Are those buttons out of the Safe Area? They look like they're bumping right up against the edge of the safe area - can we add a little padding?

components/MaskedTransition/examples/MaskedTransitionTypicalUse.swift
```diff
@@ -169,3 +180,17 @@ private class ModalViewController: UIViewController {
     dismiss(animated: true)
   }
 }
+
+extension MaskedTransitionTypicalUseSwiftExample {
+  // MARK: - CatalogByConvention
+  class func catalogMetadata() -> [String: Any] {
+    return [
+      "breadcrumbs" : [ "Masked Transition", "Masked Transition (Swift)" ],
```
romoore, line 194: are there no string constants for these?

jverkoey:
Awesome, thank you! I expect this highlights https://github.com/material-components/material-components-ios/issues/2219 as a bug even better now.

components/MaskedTransition/examples/MaskedTransitionTypicalUse.swift
codeman7: Not that I am aware of.

codeman7:
Addressed the safe area
simulator screen shot - iphone x - 2018-09-06 at 22 46 47

yarneo (approved):

PR #5061, opened 2018-09-07
[BottomNavigation] Expose UIView for a given UITabBarItem.

andrewplai:
This will allow MDCBottomNavigationBar to work with popover / tooltip libraries which require an anchor view for positioning (i.e, if we want to display a tooltip over one of the tabs).

Closes [#4824](https://github.com/material-components/material-components-ios/issues/4824)

### DIFF
--- a/components/BottomNavigation/tests/unit/BottomNavigationTests.m
+++ b/components/BottomNavigation/tests/unit/BottomNavigationTests.m
@@ -163,19 +163,35 @@
   XCTAssert(self.bottomNavBar.itemViews.lastObject.label.isHidden);
 }
 
--(void)testViewForItem {
+-(void)testViewForItemFound {
+  // Given
+  UITabBarItem *item1 = [[UITabBarItem alloc] initWithTitle:@"1" image:nil tag:0];
+  UITabBarItem *item2 = [[UITabBarItem alloc] initWithTitle:@"2" image:nil tag:0];
+
+  // When
+  self.bottomNavBar.items = @[item1, item2];
+
+  // Then
+  MDCBottomNavigationItemView *viewForItem1 =
+      (MDCBottomNavigationItemView *)[self.bottomNavBar viewForItem:item1];
+  MDCBottomNavigationItemView *viewForItem2 =
+      (MDCBottomNavigationItemView *)[self.bottomNavBar viewForItem:item2];
+  XCTAssertNotEqual(viewForItem1, viewForItem2);
+  XCTAssertTrue([self.bottomNavBar.itemViews containsObject:viewForItem1], @"BottomNavBar.itemViews did not contain the view (%@) returned for UITabBarItem (%@)", viewForItem1, item1);
+  XCTAssertTrue([self.bottomNavBar.itemViews containsObject:viewForItem2], @"BottomNavBar.itemViews did not contain the view (%@) returned for UITabBarItem (%@)", viewForItem2, item2);
+}
+
+-(void)testViewForItemNotFound {
+  // Given
   UITabBarItem *item1 = [[UITabBarItem alloc] initWithTitle:@"1" image:nil tag:0];
   UITabBarItem *item2 = [[UITabBarItem alloc] initWithTitle:@"2" image:nil tag:0];
   UITabBarItem *item3 = [[UITabBarItem alloc] initWithTitle:@"3" image:nil tag:0];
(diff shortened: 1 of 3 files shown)


components/BottomNavigation/tests/unit/BottomNavigationTests.m
```diff
@@ -163,4 +163,19 @@ -(void)testTitleVisibility {
   XCTAssert(self.bottomNavBar.itemViews.lastObject.label.isHidden);
 }
 
+-(void)testViewForItem {
```
romoore: Please split this into a test found items (at the correct index) and another for an item that is not found.

components/BottomNavigation/tests/unit/BottomNavigationTests.m
```diff
@@ -163,4 +163,19 @@ -(void)testTitleVisibility {
...
+  UITabBarItem *item1 = [[UITabBarItem alloc] initWithTitle:@"1" image:nil tag:0];
```
romoore: Please mark each section of your test with comments for the "Given" (set-up), "When" (condition being tested), and "Then" (assertions) portions of your test.  See `testSelectedItemAfterReset` for an example in this file.

components/BottomNavigation/tests/unit/BottomNavigationTests.m
```diff
@@ -163,4 +163,19 @@ -(void)testTitleVisibility {
...
+  UITabBarItem *item2 = [[UITabBarItem alloc] initWithTitle:@"2" image:nil tag:0];
+  UITabBarItem *item3 = [[UITabBarItem alloc] initWithTitle:@"3" image:nil tag:0];
+  self.bottomNavBar.items = @[item1, item2, item3];
+
+  XCTAssertEqual([self.bottomNavBar viewForItem:item1], self.bottomNavBar.itemViews[0]);
```
romoore: Let's change these asserts to not require the ordering of `itemViews`. Instead, let's just assert that the returned object is an element of the array and that for two items they are not the same view.  Something like:

```objc
UIView *viewForItem1 = [self.bottomNavBar viewForItem:item1];
UIView *viewForItem2 = [self.bottomNavBar viewForItem:item2];

// Then
XCTAssertNotEqual(viewForItem1, viewForItem2);
XCTAssertTrue([self.bottomNavBar.itemViews containsObject:viewForItem1], @"BottomNavBar.itemViews did not contain the view (%@) returned for UITabBarItem (%@)", viewForItem1, item1);
XCTAssertTrue([self.bottomNavBar.itemViews containsObject:viewForItem2], @"BottomNavBar.itemViews did not contain the view (%@) returned for UITabBarItem (%@)", viewForItem2, item2);

```

romoore (approved):
Looks good overall. Some feedback on the tests. Thanks for the PR!

components/BottomNavigation/tests/unit/BottomNavigationTests.m
andrewplai: Sounds good.

components/BottomNavigation/tests/unit/BottomNavigationTests.m
andrewplai: Sounds good.

components/BottomNavigation/tests/unit/BottomNavigationTests.m
andrewplai: Sounds good.

andrewplai:
Thanks for the feedback, and I've gone ahead and made the updates to the tests [here](https://github.com/material-components/material-components-ios/pull/5061/commits/5e1bf7a2815036a195382be8beb38bed2c79f74c).

romoore (approved):
LGTM!

romoore:
I've marked this as merge-when-approved since I'll be offline in a few minutes. Anyone else with write access will be able to merge it once it passes CI builds.

components/BottomNavigation/tests/unit/BottomNavigationTests.m
```diff
@@ -163,4 +163,35 @@ -(void)testTitleVisibility {
   XCTAssert(self.bottomNavBar.itemViews.lastObject.label.isHidden);
 }
 
+-(void)testViewForItemFound {
```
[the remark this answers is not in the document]
romoore: A little late to the party, @material-automation !

Issue #4970, opened 2018-08-30
FlexibleHeaderView should respect contentInsetAdjustmentBehavior of scrollView


wenyuzhang666:
Currently FlexibleHeaderView doesn't take contentInsetAdjustmentBehavior into consideration, so it still adjusts the contentInset if scrollView doesn't want it to be adjusted.

A related bug issue is https://github.com/material-components/material-components-ios/issues/3716

wenyuzhang666:
Fixed in https://github.com/material-components/material-components-ios/pull/5073. Closing.

PR #5073, opened 2018-09-10
[FlexibleHeaderView] Make FexibleHeaderView respect contentInsetAdjustmentBehavior (#4970)

fluxminds:
This fix is compatible for iOS 11 and above. For pre iOS11 version, FlexibleHeaderView still insets the contentInset. To fix it, we need to use a deprecated API automaticallyAdjustsScrollViewInsets from a view controller, which is not supported by the current FlexibleHeaderView architecture. Suggestions are welcomed.

Before the Change:
pc2

After the Change:
pc1

### DIFF
--- a/components/FlexibleHeader/src/MDCFlexibleHeaderView.m
+++ b/components/FlexibleHeader/src/MDCFlexibleHeaderView.m
@@ -635,6 +635,13 @@ static inline MDCFlexibleHeaderShiftBehavior ShiftBehaviorForCurrentAppContext(
                       && [self trackingScrollViewIsWebKit])) {
     return 0;
   }
+  if (@available(iOS 11.0, *)) {
+    // Don't adjust the contentInset if scrollView's behavior doesn't want it.
+    // Compatible to iOS 11 and above
+    if (scrollView.contentInsetAdjustmentBehavior == UIScrollViewContentInsetAdjustmentNever) {
+      return 0;
+    }
+  }
 
   MDCFlexibleHeaderScrollViewInfo *info = [_trackedScrollViews objectForKey:scrollView];
   if (!info) {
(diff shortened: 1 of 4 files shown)


components/FlexibleHeader/src/MDCFlexibleHeaderView.m
```diff
@@ -635,6 +635,13 @@ - (CGFloat)fhv_enforceInsetsForScrollView:(UIScrollView *)scrollView {
                       && [self trackingScrollViewIsWebKit])) {
     return 0;
   }
+  if (@available(iOS 11.0, *)) {
+    // Don't adjust the contentInset if scrollView's behavior doesn't want it.
+    // Compatible to iOS 11 and above
+    if (scrollView.contentInsetAdjustmentBehavior == UIScrollViewContentInsetAdjustmentNever) {
+      return 0;
+    }
+  }
```
jverkoey: This looks like a risky behavioral change that we'll want to gate behind a runtime flag.

Please add a public BOOL property that is NO by default and, when enabled, enables this new code path. This will let us gradually migrate clients to the new behavior so that we can resolve any dependency on the prior behavior.

jverkoey (changes requested):

components/FlexibleHeader/src/MDCFlexibleHeaderView.m
fluxminds: Sure. What do you think would be a suggested name for this property? I am not sure if there is a convention of the flag names for migration. Thanks!

jverkoey: No formal conventions exist yet, though that's probably a good idea for us to explore as a team decision. Filed https://github.com/material-components/material-components-ios/issues/5154 to track.

Some possible suggestions:

- disableContentInsetAdjustmentWhenContentInsetAdjustmentBehaviorIsNever
- respectsContentInsetAdjustmentNever

fluxminds: Thanks for the suggestion and creating the ticket to track. I will use the first one for now.

jverkoey (approved):

jverkoey:
No longer a risky change.

Issue #5128, opened 2018-09-14
[TextFields] Rethink RTL in TextFields


andrewoverton:
The RTL logic in TextFields is inconsistently applied and hard to follow. Refactors should be considered.

andrewoverton:
I'm actually going to close this because it's a duplicate of https://github.com/material-components/material-components-ios/issues/4820 ("Textfield does not work correctly on RTL direction")

PR #5129, opened 2018-09-14
[TextFields] Fix outline text field clear button layout in RTL

andrewoverton:
I previously issued https://github.com/material-components/material-components-ios/pull/4974, but it only fixed the issue for filled text fields.

I also created https://github.com/material-components/material-components-ios/issues/5128 to refactor RTL logic in TextFields, because it exists some places and not others, and where it does exist it could probably be made less confusing.

Here's are some screenshots:
RTL outlined:
rtl_outlined
RTL filled:
rtl_filled
LTR outlined:
ltr_outlined
LTR filled:
ltr_filled

### DIFF
--- a/components/TextFields/src/MDCTextField.m
+++ b/components/TextFields/src/MDCTextField.m
@@ -482,11 +482,7 @@ static const CGFloat MDCTextInputTextRectYCorrection = 1.f;
 
   // Standard textRect calculation
   UIEdgeInsets textInsets = self.textInsets;
-  if (self.mdf_effectiveUserInterfaceLayoutDirection == UIUserInterfaceLayoutDirectionRightToLeft) {
-    textRect.origin.x += textInsets.right;
-  } else {
-    textRect.origin.x += textInsets.left;
-  }
+  textRect.origin.x += textInsets.left;
   textRect.size.width -= textInsets.left + textInsets.right;
 
   // Adjustments for .leftView, .rightView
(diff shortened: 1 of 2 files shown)


yarneo:
do you mind in your screenshots showing RTL with an RTL language displayed?

components/TextFields/src/MDCTextField.m
```diff
@@ -534,7 +530,7 @@ - (CGRect)textRectForBounds:(CGRect)bounds {
     // If there is a rightView, the clearButton will not be shown.
   } else {
     CGFloat clearButtonWidth = CGRectGetWidth(self.clearButton.bounds);
-    clearButtonWidth += 2 * MDCTextInputClearButtonImageBuiltInPadding;
+    clearButtonWidth += 2 * (MDCTextInputClearButtonImageBuiltInPadding * -1);
```
yarneo: nit: please stay consistent with your `* -1` placement here and below

andrewoverton: Oh no... It seems pretty broken with Hebrew, for example. Even on develop. Here's Hebrew on develop:

simulator screen shot - iphone x - 2018-09-14 at 11 31 45

That's as far left the cursor goes :(

yarneo: how does it look like in RTL though? :) this is the LTR layout right?

yarneo: also, yeah, this is a problem. probably should be fixed in a follow up

andrewoverton: Same deal.. Messed up:
simulator screen shot - iphone x - 2018-09-14 at 11 37 01
FWIW I know @willlarche has verified with an Apple engineer at WWDC that there are bugs with cursor location/textRect on UITextField in certain RTL situations.

willlarche: It's true. We could do a workaround tho where the text insets are enlarged laterally.

andrewoverton: So @yarneo, there are still clearly issues with RTL in text fields in general, but what do you think about merging this as a step in the right direction? One that also happens to address the immediate problem described in the issue.

yarneo: sgtm

yarneo (approved):

PR #5144, opened 2018-09-17
[TextFields] Send .editingChanged control event on clear button touch

### LINKED ISSUE #3150: [TextFields] .editingChanged not fired on clear button.
MDCTextField doesn't fire `UIControlEvents.editingChanged` when user tapped its clear button while UITextField does.

```
// UITextField
textField.addTarget(self, action: #selector(onTextFieldEditingChanged(_:)), for: .editingChanged)

// MDCTextField
materialTextField.addTarget(self, action: #selector(onTextFieldEditingChanged(_:)), for: .editingChanged)

@objc
func onTextFieldEditingChanged(_ textField: UITextField) {
    debugPrint(textField.text ?? "")
}
```

We can confirm this difference simply with the code below.

I think this behavior should be adjusted.

andrewoverton:
These changes make it so that when the user taps the clear button on an MDCTextField the .editingChanged UIControlEvent is sent.

Closes #3150.

### DIFF
--- a/components/TextFields/src/private/MDCTextInputCommonFundament.m
+++ b/components/TextFields/src/private/MDCTextInputCommonFundament.m
@@ -596,9 +596,11 @@ static inline UIColor *MDCTextInputUnderlineColor() {
       [[NSNotificationCenter defaultCenter] postNotificationName:UITextViewTextDidChangeNotification
                                                           object:textField.textView];
     } else if ([self.textInput isKindOfClass:[UITextField class]]) {
+      UITextField *textField = (UITextField *)self.textInput;
       [[NSNotificationCenter defaultCenter]
           postNotificationName:UITextFieldTextDidChangeNotification
                         object:self.textInput];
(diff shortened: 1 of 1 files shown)


jverkoey (approved):
This looks like a behavioral change. I'd probably mark this PS as a risky change - what do you think?

andrewoverton:
@jverkoey, I added the label, but I don't think it'll cause any problems. It's making the behavior more closely aligned with UITextField. I wouldn't expect someone to count on it _not_ being like UITextField in this regard. Hmmm.

jverkoey:
Yep - that intuition's fair.

I think this is slightly risky due to the nature of clients depending on our code behaving in ways that are technically incorrect. There's a chance this falls into that category because we might now be invoking client code via `UIControlEventEditingChanged` events in ways that we previously were not. Should be low risk, but still potentially risky!

andrewoverton:
True... I guess let's see what happens :/ 

MJ

PR #5159, opened 2018-09-18
[Dialogs] Customize tint color of alert title icon

galiak11:
Update AlertController API to customize the alert title, including: title alignment, title icon & titleIcon tint color.

#5174
[b/116100167](http://b/116100167)
b/113257098
b/113257097

### DIFF
--- a/components/Dialogs/src/private/MDCAlertControllerView+Private.m
+++ b/components/Dialogs/src/private/MDCAlertControllerView+Private.m
@@ -54,6 +54,7 @@ static const CGFloat MDCDialogMessageOpacity = 0.54f;
 }
 
 @dynamic titleAlignment;
+@dynamic titleIcon;
 
 - (instancetype)initWithFrame:(CGRect)frame {
   self = [super initWithFrame:frame];
@@ -190,12 +191,19 @@ static const CGFloat MDCDialogMessageOpacity = 0.54f;
   } else if (self.titleIconImageView == nil) {
     self.titleIconImageView = [[UIImageView alloc] initWithImage:titleIcon];
     self.titleIconImageView.contentMode = UIViewContentModeScaleAspectFit;
+    [self tintTitleIcon];
     [self.contentScrollView addSubview:self.titleIconImageView];
   } else {
     self.titleIconImageView.image = titleIcon;
   }
 
+  self.titleIconImageView.tintColor = self.titleIconTintColor;
   [self.titleIconImageView sizeToFit];
+}
+
+- (void)setTitleIconTintColor:(UIColor *)titleIconTintColor {
+  _titleIconTintColor = titleIconTintColor;
+  self.titleIconImageView.tintColor = titleIconTintColor;
 }
 
 - (NSString *)message {
--- a/components/Dialogs/src/private/MDCAlertControllerView+Private.m
+++ b/components/Dialogs/src/private/MDCAlertControllerView+Private.m
@@ -183,15 +183,17 @@ static const CGFloat MDCDialogMessageOpacity = 0.54f;
   self.titleLabel.textAlignment = titleAlignment;
 }
 
+- (UIImage *)titleIcon {
+  return self.titleIconImageView.image;
+}
+
 - (void)setTitleIcon:(UIImage *)titleIcon {
-  _titleIcon = titleIcon;
   if (titleIcon == nil) {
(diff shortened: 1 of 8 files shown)


components/Dialogs/src/private/MDCAlertControllerView+Private.m
```diff
@@ -194,6 +195,19 @@ - (void)setTitleIcon:(UIImage *)titleIcon {
...
 }
 
+- (void)setTitleIconTintColor:(UIColor *)titleIconTintColor {
+  _titleIconTintColor = titleIconTintColor;
+  [self tintTitleIcon];
+}
+
+- (void)tintTitleIcon {
+  if (self.titleIconTintColor != nil && self.titleIcon != nil) {
+    UIImage *image = [self.titleIcon imageWithRenderingMode:UIImageRenderingModeAlwaysTemplate];
```
ianegordon: Typically I would not expect us to re-initialize this image.  I would expect the client to give us an image with the proper rendering mode.
Based on that, can we eliminate this method completely?
In setTitleIcon: and setTitleIconTintColor: instead of [self tintTitleIcon], just call self.titleIconImageView.tint = tint.

ianegordon (dismissed):

components/ActionSheet/tests/unit/MDCActionSheetTableCellTest.m
```diff
@@ -139,7 +139,7 @@ - (void)testSetActionFont {
   // When
   self.actionSheet.actionFont = actionFont;
   NSArray *cells = [MDCActionSheetTestHelper getCellsFromActionSheet:self.actionSheet];
-  XCTAssertNotEqual(cells.count, 0);
+  XCTAssertNotEqual(cells.count, 0.0f);
```
ianegordon: We should not be comparing an NSUInteger with a floating point value.

This seems unrelated to dialogs, did a stray commit get pulled in?

components/Dialogs/src/private/MDCAlertControllerView+Private.m
```diff
@@ -198,6 +202,11 @@ - (void)setTitleIcon:(UIImage *)titleIcon {
   [self.titleIconImageView sizeToFit];
 }
 
+- (void)setTitleIconTintColor:(UIColor *)titleIconTintColor {
+  _titleIconTintColor = titleIconTintColor;
+  self.titleIconImageView.tintColor = titleIconTintColor;
```
ianegordon, line 208: Since self.titleIconImageView doesn't always exist, it might be worth wrapping this in an if to explicitly communicate this with engineers.
if (self.titleIconImageView) {
  self.titleIconImageView.tintColor = titleIconTintColor;
}

components/Dialogs/src/private/MDCAlertControllerView+Private.m
```diff
@@ -198,6 +202,11 @@ - (void)setTitleIcon:(UIImage *)titleIcon {
   [self.titleIconImageView sizeToFit];
 }
 
+- (void)setTitleIconTintColor:(UIColor *)titleIconTintColor {
+  _titleIconTintColor = titleIconTintColor;
```
ianegordon, line 207: You will need to apply _titleIconTintColor to newly created titleIconImageView.
Add this in setTitleIcon:

ianegordon (approved):
Please resolve issues and get successful kokoro builds before merge.

components/ActionSheet/tests/unit/MDCActionSheetTableCellTest.m
galiak11: It was suppose to be an unsigned int. either way - this is already fixed in develop, so im removing this from the commit.

components/Dialogs/src/private/MDCAlertControllerView+Private.m
galiak11: Good catch. Done.

components/Dialogs/src/private/MDCAlertControllerView+Private.m
galiak11: Done.

galiak11:
Updating AlertController API to customize the tintColor of the alert's titleIcon.

Includes tests and examples.
issue: #5174
issue: Update MDC API to customize MDCAlertController title🐛

PR #5164, opened 2018-09-18
[Dialogs] Customize alert title alignment

galiak11:
Set title alignment for AlertController.

#5174
Update MDC API to customize MDCAlertController title🐛
b/116100167
b/113257097

image

### DIFF
--- a/components/Dialogs/src/MDCAlertController.m
+++ b/components/Dialogs/src/MDCAlertController.m
@@ -214,16 +214,22 @@ static NSString *const kMaterialDialogsBundle = @"MaterialDialogs.bundle";
 
 - (void)setTitleAlignment:(NSTextAlignment)titleAlignment {
   _titleAlignment = titleAlignment;
-  self.alertView.titleAlignment = titleAlignment;
+  if (self.alertView) {
+    self.alertView.titleAlignment = titleAlignment;
+  }
 }
 
 - (void)setCornerRadius:(CGFloat)cornerRadius {
   _cornerRadius = cornerRadius;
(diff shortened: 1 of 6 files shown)


components/Dialogs/src/private/MDCAlertControllerView+Private.m
```diff
@@ -171,6 +171,11 @@ - (void)setTitleColor:(UIColor *)titleColor {
   _titleLabel.textColor = titleColor;
 }
 
+- (void)setTitleAlignment:(NSTextAlignment)titleAlignment {
+  _titleAlignment = titleAlignment;
+  _titleLabel.textAlignment = titleAlignment;
```
ianegordon: Can we wire the titleAlignment getter/setter directly to _titleLabel.textAlignment and get rid of the _titleAligment ivar?

ianegordon (changes requested):

components/Dialogs/src/MDCAlertController.m
```diff
@@ -212,6 +212,11 @@ - (void)setButtonTitleColor:(UIColor *)buttonColor {
   }
 }
 
+- (void)setTitleAlignment:(NSTextAlignment)titleAlignment {
+  _titleAlignment = titleAlignment;
+  self.alertView.titleAlignment = titleAlignment;
```
ianegordon: It is not required, but consider adding if (self.alertView) to be consistent with the other properties and add context for the next engineer looking at this code.

ianegordon (approved):

PR #5167, opened 2018-09-19
[Dialogs] Customize alert title icon

galiak11:
Adding a title icon property to MDCAlertController.

The title icon is an image that is shown above the dialog title, and is automatically aligned with the title, supporting all NSTextAlignment alignment options (center, natural, left, right, etc).

Provided images are scaled to fit 24pt.

#5174
Update MDC API to customize MDCAlertController title🐛
b/116100167
b/113257097

Center Alignment (24dp icon):
image

Left Alignment (24dp icon):
image

Right Alignment (48dp icon):
image


components/Dialogs/src/private/MDCAlertControllerView+Private.m
```diff
@@ -28,6 +28,9 @@
 
 static const UIEdgeInsets MDCDialogContentInsets = {24.0, 24.0, 24.0, 24.0};
 static const CGFloat MDCDialogContentVerticalPadding = 20.0;
+static const CGFloat MDCDialogTitleIconVerticalPadding = 12.0;
+
+static const CGFloat MDCDialogTitleIconSize = 24.0;
```
ianegordon: Should this size vary based on device screen or dynamic type?

components/Dialogs/src/private/MDCAlertControllerView+Private.m
```diff
@@ -176,6 +179,21 @@ - (void)setTitleAlignment:(NSTextAlignment)titleAlignment {
   _titleLabel.textAlignment = titleAlignment;
 }
 
+- (void)setTitleIcon:(UIImage *)titleIcon {
+  _titleIcon = titleIcon;
+  if (titleIcon == nil) {
+    self.titleIconImageView = nil;
```
ianegordon, line 189: self.contentScrollView will still have a reference to this.

ianegordon (changes requested):

components/Dialogs/examples/DialogsAlertCustomizationViewController.swift
```diff
@@ -83,19 +88,72 @@ class DialogsAlertCustomizationViewController: MDCCollectionViewController {
...
+    self.present(alert, animated: true, completion: nil)
+  }
+
+  func didTapCenteredTitleWithIcon() {
+    let alert = createMDCAlertController(title: "Center Aligned Title")
+    alert.titleIcon = sampleIcon()
     alert.titleAlignment = .center // todo: theme with themer when available
+
+    MDCAlertControllerThemer.applyScheme(alertScheme, to: alert)
+    self.present(alert, animated: true, completion: nil)
```
codeman7: _Tip_: Have all the `didTapXYZ` functions return the _alert_ then just call
`self.present(alert, animated: true, completion: nil)` once.

components/Dialogs/examples/DialogsAlertCustomizationViewController.swift
```diff
@@ -83,19 +88,72 @@ class DialogsAlertCustomizationViewController: MDCCollectionViewController {
...
+
+    let alert = createMDCAlertController(title: nil)
+    alert.titleIcon = sampleIcon()
+
     MDCAlertControllerThemer.applyScheme(alertScheme, to: alert)
     self.present(alert, animated: true, completion: nil)
   }
 
-  private func createMDCAlertController(title: String) -> MDCAlertController {
+  private func createMDCAlertController(title: String?) -> MDCAlertController {
```
codeman7, line 158: _Tip 1_: Change the function signature to 
`createMDCAlertController(title: String?, iconStandardSize: Bool = true)`

Then in the `createMDCAlertController` function call `sampleIcon()` with respective value.

_Tip 2_: This can be marked `class` if you think it may be subclassed or `static` it doesn't rely on `self` at all.

codeman7 (approved):

components/Dialogs/examples/DialogsAlertCustomizationViewController.swift
galiak11: Not all examples are going to have an icon, and there are 2 kinds of icons, which will make the interface unnecessarily complicated.

components/Dialogs/src/private/MDCAlertControllerView+Private.m
galiak11: I removed the size limitation, relying instead on the size of the user's icon-image.  Should we enforce any limitations or perform any checks for the image size?

components/Dialogs/src/private/MDCAlertControllerView+Private.m
galiak11: Done.

components/Dialogs/examples/DialogsAlertCustomizationViewController.swift
galiak11: Done

components/Dialogs/src/private/MDCAlertControllerView+Private.m
```diff
@@ -422,24 +451,54 @@ - (void)layoutSubviews {
...
   CGRect messageFrame = CGRectMake(MDCDialogContentInsets.left,
                                    CGRectGetMaxY(titleFrame) + contentInternalVerticalPadding,
                                    messageSize.width, messageSize.height);
 
   self.titleLabel.frame = titleFrame;
   self.messageLabel.frame = messageFrame;
 
+  if (self.titleIconImageView != nil) {
+    // match the titleIcon alignment to the title alignment
+    CGFloat titleIconPosition = titleFrame.origin.x;
```
ianegordon, line 488: We can't access CGRect struct data directly.  Use CGRectAccessor() methods.

components/Dialogs/src/private/MDCAlertControllerView+Private.m
```diff
@@ -422,24 +451,54 @@ - (void)layoutSubviews {
...
+      CGRectMake(MDCDialogContentInsets.left, titleTop, titleSize.width, titleSize.height);
   CGRect messageFrame = CGRectMake(MDCDialogContentInsets.left,
                                    CGRectGetMaxY(titleFrame) + contentInternalVerticalPadding,
                                    messageSize.width, messageSize.height);
 
   self.titleLabel.frame = titleFrame;
   self.messageLabel.frame = messageFrame;
 
+  if (self.titleIconImageView != nil) {
+    // match the titleIcon alignment to the title alignment
```
ianegordon, line 487: This seems strange to me, don't we always want to center the icon?

ianegordon (approved):

components/Dialogs/src/private/MDCAlertControllerView+Private.m
galiak11: Done.

components/Dialogs/src/private/MDCAlertControllerView+Private.m
galiak11: There are actually no specifications for this scenario. The requirements only cover 2 modes - (1) a centered icon with a centered title, or (2) a left/natural aligned title with no icon.   All other options are available in MDC because the 2 properties (icon and title-alignment) are independent of each other.  
Having the same alignment for icon & title seems to look better (see screenshots in the main comment).

ianegordon:
@galiak11 Please rebase and merge.

PR #5182, opened 2018-09-20
[TextFields] Remove extension check in example

codeman7:
Previously we checked against `iOSApplicationExtension` and `iOS` but if the `iOS` version is 9.0 or above then the ApplicationExtension will be as well. This will remove a lot of warnings in our kokoro logs when trying to track down errors.

| Before | After |
| ------ | ------ |
|simulator screen shot - iphone xs max - 2018-09-20 at 08 30 07|simulator screen shot - iphone xs max - 2018-09-20 at 08 29 32|
|simulator screen shot - iphone xs max - 2018-09-20 at 08 29 32|simulator screen shot - iphone xs max - 2018-09-20 at 08 30 07|


jverkoey (approved):
What's an example of one of the warnings this resolves?

codeman7:
screen shot 2018-09-20 at 9 57 05 am

This removes a lot of that exact warning in kokoro error logs.

PR #5242, opened 2018-09-25
[BottomNavigation] Fix swift example

codeman7:
### Context
In working on #4042 I discovered that Bottom Navigation (Swift) example isn't working as intended
### The problem
No elements were on screen
### The fix
Call the common init in the initializer

### Screenshots
| Before | After |
| ----- | ----- |
|simulator screen shot - iphone xs max - 2018-09-25 at 15 18 57|simulator screen shot - iphone xs max - 2018-09-25 at 15 14 18|

### DIFF
--- a/components/BottomNavigation/examples/BottomNavigationTypicalUseExample.swift
+++ b/components/BottomNavigation/examples/BottomNavigationTypicalUseExample.swift
@@ -25,6 +25,7 @@ class BottomNavigationTypicalUseSwiftExample: UIViewController {
 
   init() {
     super.init(nibName: nil, bundle: nil)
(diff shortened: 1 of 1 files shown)


ianegordon (approved):
I was initially concerned that we are no longer reacting to safeAreainsets changing, but I would expect setNeedsLayout to be invoked if viewSafeAreaInsetsDidChange.
If this matches your understanding, LGTM.

codeman7:
Simply adding the `commonInit` would result in the below screenshot.
simulator screen shot - iphone x - 2018-09-26 at 09 17 47

PR #5244, opened 2018-09-25
[NavigationDrawer] Use CGRect accessor functions

andrewoverton:
Follow up to https://github.com/material-components/material-components-ios/pull/5179 that replaces `frame.size.height` with `CGRectGetHeight(...)`, etc.


ianegordon (approved):
LGTM.
Consider clang formats before merge.

PR #5248, opened 2018-09-26
[ActionSheet] Update MDCActionSheetHelperTest

codeman7:
### Context
Kokoro doesn't check against frameworks as rigorous as internal test

### Problem
MDCActionSheetHelperTest method colorsToTest used CIColor but CoreImage was never imported.

### The fix
Import CoreImage framework and add it in the BUILD file.

### DIFF
--- a/components/ActionSheet/tests/unit/MDCActionSheetTestHelper.m
+++ b/components/ActionSheet/tests/unit/MDCActionSheetTestHelper.m
@@ -14,6 +14,8 @@
 
 #import "MDCActionSheetTestHelper.h"
 
+#import <CoreImage/CoreImage.h>
+
 #import "../../src/private/MDCActionSheetHeaderView.h"
 #import "../../src/private/MDCActionSheetItemTableViewCell.h"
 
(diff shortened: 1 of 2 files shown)


components/ActionSheet/tests/unit/MDCActionSheetTestHelper.m
```diff
@@ -24,10 +24,8 @@ + (NSArray *)colorsToTest {
   UIColor *hsbColor = [UIColor colorWithHue:0.8f saturation:0.8f brightness:0.8f alpha:0.8f];
   UIColor *blackWithAlpha = [UIColor.greenColor colorWithAlphaComponent:0.8f];
   UIColor *black = UIColor.blackColor;
-  CIColor *ciColor = [[CIColor alloc] initWithColor:UIColor.blackColor];
```
randallli: Oh is it because we are using CoreImage?
If so then we can just add the dep to the build target and import it.

codeman7: I can add the import

codeman7: I am missing `#import <CoreImage/CoreImage.h>`

randallli: yes cool thanks.

randallli (approved):
Fix the description to reflect the actual changes. (its describes an earlier iteration)

PR #5296, opened 2018-10-02
[Dialogs] Add shadow elevation

codeman7:
### Context
Clients may need to set custom elevation levels for their custom dialogs
### The problem
Currently we don't expose any of the necessary properties to do this.
### The fix
Expose those properties so clients can have a custom elevation for their Dialog.

### Additionally
Test and Examples have been added.

### Screenshots
| Default | Custom |
| - | - |
|simulator screen shot - iphone x - 2018-10-02 at 15 40 25|simulator screen shot - iphone x - 2018-10-02 at 15 40 21|


ianegordon (approved):
Thank you @codeman7 !

PR #5300, opened 2018-10-02
[TextFields] Fix the error accessibility label

yilei:
The value was not respected previously.


andrewoverton (approved):
Thank you for the PR! `errorAccessibilityValue` not being used was definitely a regression.

I did a fair amount of testing with this change and it seems to have not broken anything. LGTM!

PR #5326, opened 2018-10-04
[MDCSlider] Allow setting exclusiveTouch on Sliders.

### LINKED ISSUE #5236: Allow setting exclusiveTouch on Sliders
## Feature Request/Suggestion

Allow setting exclusiveTouch on Sliders. https://developer.apple.com/documentation/uikit/uiview/1622453-exclusivetouch

## Additional context

Exclusivetouch allows the developer to stop multitouch interactions within a window that are undesirable.

---
#### Internal data
- Associated internal bug: [b/117134562](http://b/117134562)

gabebear:
reopening GH-5237

Exclusivetouch allows the developer to stop multitouch interactions within a window that are undesirable. https://developer.apple.com/documentation/uikit/uiview/1622453-exclusivetouch

closes #5236

### DIFF
--- a/components/private/ThumbTrack/src/MDCThumbView.m
+++ b/components/private/ThumbTrack/src/MDCThumbView.m
@@ -25,8 +25,6 @@
 
 @implementation MDCThumbView
 
-static const CGFloat kMinTouchSize = 48;
-
 + (Class)layerClass {
   return [MDCShadowLayer class];
 }
@@ -73,18 +71,6 @@ static const CGFloat kMinTouchSize = 48;
   [super layoutSubviews];
   self.layer.shadowPath =
       [UIBezierPath bezierPathWithRoundedRect:self.bounds cornerRadius:_cornerRadius].CGPath;
-}
-
-- (BOOL)pointInside:(CGPoint)point withEvent:(__unused UIEvent *)event {
-  CGFloat dx = MIN(0, _cornerRadius - kMinTouchSize / 2);
-  // Converts point to presentation layer coordinate system so gesture will land on the right visual
-  // position. Assuming superview is not animated.
-  if (self.layer.presentationLayer) {
-    point = [(CALayer *)self.layer.presentationLayer convertPoint:point
-                                                        fromLayer:self.layer.modelLayer];
-  }
-  CGRect rect = CGRectInset(self.bounds, dx, dx);
-  return CGRectContainsPoint(rect, point);
 }
 
 - (void)setIcon:(nullable UIImage *)icon {
--- a/components/Slider/src/MDCSlider.m
+++ b/components/Slider/src/MDCSlider.m
@@ -385,6 +385,11 @@ static inline UIColor *MDCThumbTrackDefaultColor(void) {
 
 #pragma mark - UIView methods
 
+- (void)setExclusiveTouch:(BOOL)exclusiveTouch {
+  [super setExclusiveTouch:exclusiveTouch];
+  _thumbTrack.exclusiveTouch = exclusiveTouch;
+}
+
 - (CGSize)intrinsicContentSize {
   return CGSizeMake(kSliderDefaultWidth, kSliderFrameHeight);
 }


components/private/ThumbTrack/src/MDCThumbView.m
```diff
@@ -75,18 +73,6 @@ - (void)layoutSubviews {
       [UIBezierPath bezierPathWithRoundedRect:self.bounds cornerRadius:_cornerRadius].CGPath;
 }
 
-- (BOOL)pointInside:(CGPoint)point withEvent:(__unused UIEvent *)event {
```
randallli, line 78: Removing this seems unrelated to the PR's description. 
This logic ensures that the thumbnail is accessible. Please elaborate why we are removing it.

randallli (dismissed):

components/private/ThumbTrack/src/MDCThumbView.m
gabebear: MDCThumbTrack is the superview of MDCThumbView which handles all the touch events and has it's hit area expanded. 

Having this subview have an expanded hit area that isn't handling the events makes the exclusive logic fail. Without this change setting exclusiveTouch on MDCSlider will make the material slider not respond when other controls are active, but won't stop other controls from responding when the MDCSlider is active. You can see this if you try the "MDCSlider and UISlider Compared" example with just the other code.

jverkoey: Dialing in @randallli's feedback here with a question: with this logic removed, is the tap target for the thumbtrack still at least 48pts along both the horizontal and vertical axis? This is an important metric we use to evaluate whether a component is accessible.

gabebear: Yes, MDCThumbTrack still expands the tap and drag targets to 48px. https://github.com/material-components/material-components-ios/blob/develop/components/private/ThumbTrack/src/MDCThumbTrack.m#L34

jverkoey: It looks like you're linking to the code that you're deleting here - was that intentional?

gabebear: No, the code in MDCThumbTrack.m that I linked to is nearly identical to the code I'm deleting in MDCThumbView.m. I was meaning to point out that the logic isn't being lost.

The code in MDCThumbTrack.m actually expands the hit area.

jverkoey: Got it! Thank you for clarifying.

jverkoey:
Adding @romoore for additional insight on the ThumbTrack component.

components/private/ThumbTrack/src/MDCThumbView.m
romoore: I can't find a difference between including this code and removing it.  I'd prefer to make the smallest change necessary to achieve the Issue's goal to minimize any unintended side effects.

Can you please provide steps to reproduce the problematic scenario you're describing? Here's what I attempted:

1. Configured the first Slider in SliderCollectionViewController to `exclusiveTouch = YES`
1. Opened the main Demo of MDCSlider (SliderCollectionViewController)
1. Used multi-point touch to tap and touch+drag the first and second Sliders. Alternated between having the `exclusiveTouch` slider receive the first and second touches.

**Expected result**
Removing the code in MDCThumbView would cause a difference in behavior.

**Actual result**
I observed no difference in behavior. Here is what I found in my testing:
* If multi-point single tap was used, the Slider with the first touch responded.
* If multi-point touch-and-drag was used, the Slider with `exclusiveTouch = NO` responded.

Testing of this PR with and without the changes to MDCThumbView

|Without Changes to MDCThumbView|With Changes to MDCThumbView|
|---|---|
|slider-thumbview-nochange|slider-thumbview-change|

romoore (dismissed):

gabebear:
The problem can't easily be replicated in the simulator.

Set exclusiveTouch on the MDCSlider and start sliding one slider and then use another finger to try to interact with the other control.

If you start using the MDCSlider first, the UISlider still works without removing the code.
withoutremoval

With removing the code you can start using either slider and the other will respect the exclusiveTouch. 
withremoval

jverkoey:
@romoore + @randallli - your follow up is needed here.

romoore (approved):

romoore:
@gabebear Thanks for explaining and the very awesome gifs!

PR #5341, opened 2018-10-07
[List] Fix MDCSelfSizingStereoCell not using the right text color and not resetting all label properties on re-use

jpoumailloux:
There were two issues:
1) the -titleLabel and -detailLabel UILabels were not getting their -textColor set, despite the helper methods -defaultTitleLabelTextColor and -defaultDetailLabelTextColor defined in the file.
2) within -prepareForReuse, the -textColor and -numberOfLines properties on the -titleLabel and -detailLabel UILabels weren't getting reset.


andrewoverton (approved):
Thank you for the PR!

PR #5342, opened 2018-10-08
[Chips] Add support for third party keyboards

codeman7:
### Context
A client found that MDCChipFields were acting differently when used with a third party keyboard (GBoard).
### The problem
When a user adds a chip and then taps the "delete" key on the keyboard the chip would delete but the placeholder would not move to the correct place. 
### The fix
Adding a call to `clearTextInput` within the `removeChips` function updates the placeholder to act the way it should. This is the only call in `createNewChipFromInput` that has any effect on the placeholder, so this behavior is the same as when we add the chip.
### Related Bugs
b/116498542

### Tests
Since this is a third party keyboard causing the error, I don't think test are necessary for this PR.

### Videos
| Before | After |
| - | - |
| beforeupdate|chipsafterdelete|

### DIFF
--- a/components/Chips/src/MDCChipField.m
+++ b/components/Chips/src/MDCChipField.m
@@ -16,6 +16,7 @@
 
 #import <MDFInternationalization/MDFInternationalization.h>
 
+#import "../../TextFields/src/private/MDCTextInputCommonFundament.h"
 #import "MaterialMath.h"
 #import "MaterialTextFields.h"
 
@@ -55,6 +56,14 @@ const UIEdgeInsets MDCChipFieldDefaultContentEdgeInsets = {
 
 @property(nonatomic, weak) id<MDCChipTextFieldDelegate> deletionDelegate;
 
+@end
+
+@interface MDCTextField (Fundament)
+@property(nonatomic, strong) MDCTextInputCommonFundament *fundament;
+@end
+
+@interface MDCTextInputCommonFundament (Placeholder)
+- (void)updatePlaceholderPosition;
 @end
 
 @implementation MDCChipTextField
@@ -330,7 +339,7 @@ const UIEdgeInsets MDCChipFieldDefaultContentEdgeInsets = {
     [self.delegate chipField:self didRemoveChip:chip];
   }
   [self notifyDelegateIfSizeNeedsToBeUpdated];
-  [self clearTextInput];
+  [self.textField.fundament updatePlaceholderPosition];
   [self invalidateIntrinsicContentSize];
   [self setNeedsLayout];
 }
--- a/components/Chips/src/MDCChipField.m
+++ b/components/Chips/src/MDCChipField.m
@@ -58,11 +58,11 @@ const UIEdgeInsets MDCChipFieldDefaultContentEdgeInsets = {
 
 @end
 
-@interface MDCTextField (Fundament)
+@interface MDCTextField (MDCChipFieldTestingVisability)
 @property(nonatomic, strong) MDCTextInputCommonFundament *fundament;
 @end
 
-@interface MDCTextInputCommonFundament (Placeholder)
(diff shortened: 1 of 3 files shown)


romoore:
can we follow the approach being taken in #5343 where we update the placeholder position instead of changing the text?

components/Chips/src/MDCChipField.m
```diff
@@ -57,6 +58,14 @@ @interface MDCChipTextField : MDCTextField
 
 @end
 
+@interface MDCTextField (Fundament)
```
romoore: Please name the category to something prefixed with MDCChips or MDCChipField to avoid naming conflicts.

Suggestion: @interface MDCTextField (MDCChipFieldVisibility)

components/Chips/src/MDCChipField.m
```diff
@@ -57,6 +58,14 @@ @interface MDCChipTextField : MDCTextField
...
+@property(nonatomic, strong) MDCTextInputCommonFundament *fundament;
+@end
+
+@interface MDCTextInputCommonFundament (Placeholder)
```
romoore: Please name the category to something prefixed with MDCChips or MDCChipField to avoid naming conflicts.

Suggestion: @interface MDCTextField (MDCChipFieldVisibility)

romoore (dismissed):
Please add a unit test to verify the updated placeholder origin position after deleting a chip.

components/Chips/src/MDCChipField.m
```diff
@@ -57,6 +58,14 @@ @interface MDCChipTextField : MDCTextField
 
 @end
 
+@interface MDCTextField (MDCChipFieldTestingVisability)
```
romoore: Nit/spelling: s/Visability/Visibility/

components/Chips/src/MDCChipField.m
```diff
@@ -57,6 +58,14 @@ @interface MDCChipTextField : MDCTextField
...
+@property(nonatomic, strong) MDCTextInputCommonFundament *fundament;
+@end
+
+@interface MDCTextInputCommonFundament (MDCChipFieldTestingVisability)
```
romoore: Nit/spelling: s/Visability/Visibility/

components/Chips/tests/unit/ChipsTests.m
```diff
@@ -294,6 +294,39 @@ - (void)testPointInsideWithCustomHitAreaInsets {
                         withEvent:nil]);
 }
 
+- (void)testRemoveChipsManually {
```
romoore, line 297: I think this test would be less implementation-specific if you only verify that the X origin of the placeholder when the Chip is remove is less than when a Chip is present.  Instead of verifying that adding and removing a Chip results in exactly the same X position.

components/Chips/tests/unit/ChipsTests.m
```diff
@@ -294,6 +294,39 @@ - (void)testPointInsideWithCustomHitAreaInsets {
...
+  MDCChipField *field = [[MDCChipField alloc] init];
+  field.frame = CGRectMake(0, 0, 200, 50);
+  field.textField.text = @"Test";
+  field.textField.placeholder = @"Test";
+  [field setNeedsLayout];
+  [field layoutIfNeeded];
+  CGFloat initialPlaceholderOriginX =
+      CGRectStandardize(field.textField.placeholderLabel.frame).origin.x;
+
+  // When
```
romoore, line 306: Each test should have a single "When" and "Then" section. I believe that getting the position of the initial X position and the "with chip" X position are both part of the "When" or "Given" phases of the test.

Consider this suggestion:

```
// Given
Prepare chipfield
Add a chip, determine the location of placeholder

// When
Remove the chip

// Then
Get the position of the placeholder
Verify that the placeholder X-position is less than the initial value (with the Chip)
```

romoore (dismissed):

romoore (approved):

PR #5343, opened 2018-10-08
[Chips] Fix placeholder for manually adding chips

codeman7:
### Context
A client is adding chips using the `addChip` function with chips they have created. 
### The problem
When you call the `addChip` function with your custom chip the placeholder wouldn't update its position.
### The fix
Add a call to `clearTextInput` within the `addChip` function, this matches what is done in `createNewChipFromInput`.
### Related bugs
b/116495610

### Alternatives
We could expose the `clearTextInput` method and allow clients to call that right after they call `addChip`
or
We could ask clients to update the placeholder within the MDCTextField in MDCChipField
Neither of which I felt was as clean as this approach.

### Videos

| Before | After |
| - | - | 
|develop|chips-add|

**_Note_** In video I am tapping the "add chip" button.

### Snippet to reproduce
```diff
+ ChipsInputExampleViewController.m
```

```objectivec

- (void)viewDidLoad {
  UIButton *_button = [[UIButton alloc] init];
  _button.backgroundColor = self.colorScheme.onBackgroundColor;
  [_button setTitle:@"Add chip" forState:UIControlStateNormal];
  [_button sizeToFit];
  _button.center = self.view.center;
  [_button addTarget:self action:@selector(addChipX) forControlEvents:UIControlEventTouchUpInside];
  [self.view addSubview:_button];
}
- (void)addChipX {
  MDCChipView *chip = [[MDCChipView alloc] init];
  chip.titleLabel.text = @"Chips";
  MDCChipViewScheme *scheme = [[MDCChipViewScheme alloc] init];
  scheme.colorScheme = self.colorScheme;
  scheme.typographyScheme = self.typographyScheme;
  scheme.shapeScheme = self.shapeScheme;
  [MDCChipViewThemer applyScheme:scheme toChipView:chip];

  [_chipField addChip:chip];
}
```

### DIFF
--- a/components/Chips/src/MDCChipField.m
+++ b/components/Chips/src/MDCChipField.m
@@ -58,11 +58,11 @@ const UIEdgeInsets MDCChipFieldDefaultContentEdgeInsets = {
 
 @end
 
-@interface MDCTextField (MDCChipFieldTestingVisibility)
+@interface MDCTextField (MDCChipFieldTestingVisability)
 @property(nonatomic, strong) MDCTextInputCommonFundament *fundament;
 @end
 
-@interface MDCTextInputCommonFundament (MDCChipFieldTestingVisibility)
+@interface MDCTextInputCommonFundament (MDCChipFieldTestingVisability)
 - (void)updatePlaceholderPosition;
 @end
 
--- a/components/Chips/src/MDCChipField.m
+++ b/components/Chips/src/MDCChipField.m
@@ -58,11 +58,11 @@ const UIEdgeInsets MDCChipFieldDefaultContentEdgeInsets = {
 
 @end
 
-@interface MDCTextField (MDCChipFieldTestingVisability)
+@interface MDCTextField (MDCChipFieldTestingVisibility)
 @property(nonatomic, strong) MDCTextInputCommonFundament *fundament;
 @end
 
-@interface MDCTextInputCommonFundament (MDCChipFieldTestingVisability)
+@interface MDCTextInputCommonFundament (MDCChipFieldTestingVisibility)
 - (void)updatePlaceholderPosition;
 @end
 
--- a/components/Chips/src/MDCChipField.m
+++ b/components/Chips/src/MDCChipField.m
@@ -19,7 +19,6 @@
 #import "../../TextFields/src/private/MDCTextInputCommonFundament.h"
 #import "MaterialMath.h"
 #import "MaterialTextFields.h"
-#import "../../TextFields/src/private/MDCTextInputCommonFundament.h"
 
 static NSString *const MDCChipFieldTextFieldKey = @"textField";
 static NSString *const MDCChipFieldDelegateKey = @"delegate";
--- a/components/Chips/src/MDCChipField.m
+++ b/components/Chips/src/MDCChipField.m
@@ -327,6 +327,7 @@ const UIEdgeInsets MDCChipFieldDefaultContentEdgeInsets = {
   if ([self.delegate respondsToSelector:@selector(chipField:didAddChip:)]) {
     [self.delegate chipField:self didAddChip:chip];
   }
+  [self clearTextInput];
   [self notifyDelegateIfSizeNeedsToBeUpdated];
   [self invalidateIntrinsicContentSize];
   [self setNeedsLayout];
--- a/components/Chips/src/MDCChipField.m
+++ b/components/Chips/src/MDCChipField.m
@@ -19,6 +19,7 @@
 #import "../../TextFields/src/private/MDCTextInputCommonFundament.h"
 #import "MaterialMath.h"
 #import "MaterialTextFields.h"
+#import "../../TextFields/src/private/MDCTextInputCommonFundament.h"
 
 static NSString *const MDCChipFieldTextFieldKey = @"textField";
 static NSString *const MDCChipFieldDelegateKey = @"delegate";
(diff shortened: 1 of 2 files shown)


romoore (dismissed):
This looks like a layout bug in MDCChipField and I'm wary of introducing a breaking change like this (removing any user-entered text) when a chip is added programmatically. 

I suspect what actually needs to happen is an explicit call to `- [MDCTextField updatePlaceholderPosition]` at some point. Is there a way we can trigger a relayout of the MDCTextInput within the MDCChipField without removing the user's text?

components/Chips/src/MDCChipField.m
```diff
@@ -318,6 +327,7 @@ - (void)addChip:(MDCChipView *)chip {
   if ([self.delegate respondsToSelector:@selector(chipField:didAddChip:)]) {
     [self.delegate chipField:self didAddChip:chip];
   }
+  [self.textField.fundament updatePlaceholderPosition];
```
romoore, line 330: If this works, I think it's better than the previous approach of clearing the text input. Can you please file a backlog Issue to provide a public (private?) API for this behavior so we can call it without a unique category within Chips?

components/Chips/tests/unit/ChipsTests.m
```diff
@@ -294,6 +294,27 @@ - (void)testPointInsideWithCustomHitAreaInsets {
...
+  [fakeField setNeedsLayout];
+  [fakeField layoutIfNeeded];
+  [fakeField addChip:fakeChip];
+  [fakeField layoutIfNeeded];
+
+  // Then
+  CGRect placeholderFrame = CGRectStandardize(fakeField.textField.placeholderLabel.frame);
+  // MDCTextField's adds 8dp padding to the text
+  CGFloat expectedXPosition = CGRectGetWidth(fakeChip.frame) + 8.f;
+  XCTAssertEqual(placeholderFrame.origin.x, expectedXPosition);
```
romoore: Please use XCTAssertEqualsWithAccuracy() when comparing CGFloat values.

components/Chips/tests/unit/ChipsTests.m
```diff
@@ -294,6 +294,27 @@ - (void)testPointInsideWithCustomHitAreaInsets {
                         withEvent:nil]);
 }
 
+- (void)testAddChipsManuallyPlaceholderCorrectPosition {
```
romoore, line 321: Tip: Instead of checking the exact position of the placeholder text, consider testing that the addition of a Chip will increase the X-value of the placeholder's position.

Something like:

```
// When
CGFloat placeholderInitialXPosition = fakeField.placeholder.frame.origin.x;
[fakeField addChip:fakeChip];
[fakeField layoutIfNeeded];

// Then
CGFloat placeholderFinalXPosition = fakeField.placeholder.frame.origin.x;
XCTAssertGreaterThan(placeholderFinalXPosition, placeholderInitialXPosition);
```

What do you think?

components/Chips/src/MDCChipField.m
```diff
@@ -57,6 +58,14 @@ @interface MDCChipTextField : MDCTextField
 
 @end
 
+@interface MDCTextField (Fundament)
```
romoore: Please name the category to something prefixed with `MDCChips` or `MDCChipField` to avoid naming conflicts.

Suggestion: `@interface MDCTextField (MDCChipFieldVisibility)`

components/Chips/src/MDCChipField.m
```diff
@@ -57,6 +58,14 @@ @interface MDCChipTextField : MDCTextField
...
+@property(nonatomic, strong) MDCTextInputCommonFundament *fundament;
+@end
+
+@interface MDCTextInputCommonFundament (Placeholder)
```
romoore: Please name the category to something prefixed with MDCChips or MDCChipField to avoid naming conflicts.

Suggestion: @interface MDCTextField (MDCChipFieldVisibility)

romoore (approved):

components/Chips/src/MDCChipField.m
codeman7: Done

components/Chips/tests/unit/ChipsTests.m
codeman7: Done

components/Chips/tests/unit/ChipsTests.m
codeman7: Isn't necessary because using XCTAssertGreaterThan.

components/Chips/src/MDCChipField.m
codeman7: Done

components/Chips/src/MDCChipField.m
codeman7: Done

components/Chips/src/MDCChipField.m
```diff
@@ -57,6 +58,14 @@ @interface MDCChipTextField : MDCTextField
 
 @end
 
+@interface MDCTextField (MDCChipFieldTestingVisability)
```
romoore: Nit/spelling: s/Visability/Visibility/

components/Chips/src/MDCChipField.m
```diff
@@ -57,6 +58,14 @@ @interface MDCChipTextField : MDCTextField
...
+@property(nonatomic, strong) MDCTextInputCommonFundament *fundament;
+@end
+
+@interface MDCTextInputCommonFundament (MDCChipFieldTestingVisability)
```
romoore: Nit/spelling: s/Visability/Visibility/

romoore (approved):
Two spelling nits for the category name, otherwise looks good to me. 👍

PR #5357, opened 2018-10-08
[List] Generalize List themers

### LINKED ISSUE #5356: [List] Internal Issue #117213598
This internal issue [b/117213598](http://b/117213598) wasn't picked up by Material Mirror because it's not filed under MDC, but there is public work associated with it.

andrewoverton:
These changes generalize List themers so that for the List component there is one themer per subsystem, each with one method per List Item class.

Closes #5356


codeman7 (approved):

codeman7:
_Tip:_ We talked about updating MDCListTypographyThemer to `#import MaterialList.h` instead of `#import MDCSelfSizingStereoCell` you may want to do that in a follow up PR.

andrewoverton:
Thanks @codeman7 !

PR #5362, opened 2018-10-08
[Dialogs] add scrimColor to MDCAlertController

galiak11:
Adding the scrimColor property to MDCAlertController, which controls the background color when the alert is presented by a MDCDialogPresentationController. Also, fix MDCAlertColorThemer to use the new property when theming the dialog, and update relevant tests and examples.

Issues: b/116845327, b/117173678.

### DIFF
--- a/components/Dialogs/src/MDCAlertController.m
+++ b/components/Dialogs/src/MDCAlertController.m
@@ -236,11 +236,7 @@ static NSString *const kMaterialDialogsBundle = @"MaterialDialogs.bundle";
 
 - (void)setScrimColor:(UIColor *)scrimColor {
   _scrimColor = scrimColor;
-  MDCDialogPresentationController *dialogPresentationController =
-      self.mdc_dialogPresentationController;
-  if (dialogPresentationController) {
-    dialogPresentationController.scrimColor = scrimColor;
-  }
+  self.mdc_dialogPresentationController.scrimColor = scrimColor;
 }
 
 - (void)setCornerRadius:(CGFloat)cornerRadius {
@@ -248,21 +244,12 @@ static NSString *const kMaterialDialogsBundle = @"MaterialDialogs.bundle";
   if (self.alertView) {
     self.alertView.cornerRadius = cornerRadius;
   }
-  // make sure to pass the new shape through to the dialog's tracking/shadow layer
-  MDCDialogPresentationController *dialogPresentationController =
-      self.mdc_dialogPresentationController;
-  if (dialogPresentationController) {
-    dialogPresentationController.dialogCornerRadius = cornerRadius;
-  }
+  self.mdc_dialogPresentationController.dialogCornerRadius = cornerRadius;
 }
 
 - (void)setElevation:(CGFloat)elevation {
   _elevation = elevation;
-  MDCDialogPresentationController *dialogPresentationController =
-      self.mdc_dialogPresentationController;
-  if (dialogPresentationController) {
-    dialogPresentationController.dialogElevation = elevation;
-  }
+  self.mdc_dialogPresentationController.dialogElevation = elevation;
 }
 
 - (void)mdc_setAdjustsFontForContentSizeCategory:(BOOL)adjusts {
--- a/components/Dialogs/src/MDCAlertController.m
+++ b/components/Dialogs/src/MDCAlertController.m
@@ -234,6 +234,16 @@ static NSString *const kMaterialDialogsBundle = @"MaterialDialogs.bundle";
   }
 }
 
+- (void)setScrimColor:(UIColor *)scrimColor {
+  _scrimColor = scrimColor;
+  // make sure to pass the new shape through to the dialog's tracking/shadow layer
+  MDCDialogPresentationController *dialogPresentationController =
+      self.mdc_dialogPresentationController;
+  if (dialogPresentationController) {
+    dialogPresentationController.scrimColor = scrimColor;
+  }
+}
+
 - (void)setCornerRadius:(CGFloat)cornerRadius {
   _cornerRadius = cornerRadius;
   if (self.alertView) {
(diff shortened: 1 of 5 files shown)


components/Dialogs/src/MDCAlertController.m
```diff
@@ -234,6 +234,16 @@ - (void)setTitleIconTintColor:(UIColor *)titleIconTintColor {
   }
 }
 
+- (void)setScrimColor:(UIColor *)scrimColor {
+  _scrimColor = scrimColor;
+  // make sure to pass the new shape through to the dialog's tracking/shadow layer
```
codeman7: _tip:_ I think this comment needs to be updated

components/Dialogs/tests/unit/MDCAlertControllerCustomizationTests.m
```diff
@@ -96,16 +96,33 @@ - (void)testApplyingScrimColorToPresentationController {
...
+  XCTAssertEqualObjects(presentationController.scrimColor, scrimColor);
 }
 
+- (void)testApplyingScrimColorToAlert {
+  // Given
+  UIColor *scrimColor = [UIColor.blueColor colorWithAlphaComponent:0.3];
+  MDCDialogPresentationController *presentationController = alert.mdc_dialogPresentationController;
+  if (presentationController == nil) {
+    return;  // don't fail the test if mdc_dialogPresentationController is nil
+  }
```
codeman7: _tip:_ Do we need these lines?

components/Dialogs/tests/unit/MDCAlertControllerCustomizationTests.m
```diff
@@ -96,16 +96,33 @@ - (void)testApplyingScrimColorToPresentationController {
...
+
+  // When
+  alert.scrimColor = scrimColor;
+
+  // Then
+  XCTAssertEqualObjects(presentationController.scrimColor, scrimColor);
```
codeman7: _tip:_ Aren't we testing the same thing as the test above? Should we test `XCTAssertEqualObjects(alert.scrimColor, scrimColor);`?

codeman7 (approved):

components/Dialogs/tests/unit/MDCAlertControllerCustomizationTests.m
galiak11: Which lines are you referring to? 
There are space lines between the Given/When/Then sections in all tests.

components/Dialogs/tests/unit/MDCAlertControllerCustomizationTests.m
galiak11: Using the same check doesn't mean the tests are the same.
The "When" section (which defines what's being tested) is different in each test: one tests the presentation controller API, while the other tests the alert controller API.

components/Dialogs/src/MDCAlertController.m
galiak11: Done.

components/Dialogs/tests/unit/MDCAlertControllerCustomizationTests.m
codeman7: Yeah I didn't see the difference in the "When".

components/Dialogs/tests/unit/MDCAlertControllerCustomizationTests.m
codeman7: I believe the return makes it so we can silently fail the test.

galiak11: Or silently succeed...

Any other suggestions? Can we mark this as resolved?

components/Dialogs/tests/unit/MDCAlertControllerCustomizationTests.m
galiak11: can we mark this as resolved?

components/Dialogs/src/MDCAlertController.m
```diff
@@ -234,6 +234,15 @@ - (void)setTitleIconTintColor:(UIColor *)titleIconTintColor {
   }
 }
 
+- (void)setScrimColor:(UIColor *)scrimColor {
+  _scrimColor = scrimColor;
+  MDCDialogPresentationController *dialogPresentationController =
+      self.mdc_dialogPresentationController;
+  if (dialogPresentationController) {
```
romoore: Nit: Please remove the `nil` check, [they shouldn't be used in Objective-C](http://google.github.io/styleguide/objcguide.html#nil-checks) unless it would cause a crash.

components/Dialogs/tests/unit/MDCAlertControllerCustomizationTests.m
```diff
@@ -20,43 +20,44 @@
...
 
-MDCAlertController *alert;
-MDCAlertControllerView *alertView;
-
 - (void)setUp {
   [super setUp];
-  alert = [MDCAlertController alertControllerWithTitle:@"Title" message:@"Message"];
-  alertView = (MDCAlertControllerView *)alert.view;
+  self.alert = [MDCAlertController alertControllerWithTitle:@"Title" message:@"Message"];
+  self.alertView = (MDCAlertControllerView *)self.alert.view;
```
romoore, line 35: Tip: Use `- [MDCAlertController.alertView]` to safely access the `MDCAlertControllerView` property. If you don't want to expose it in a testing category, please perform a type check `-isKindOfClass:` first.

romoore (approved):
Looks good. Thanks for removing the global variables!

components/Dialogs/tests/unit/MDCAlertControllerCustomizationTests.m
galiak11: Added the type check.

PR #5378, opened 2018-10-09
[List] Add global themer for List component

### LINKED ISSUE #5375: [List] Create a themer that applies schemes for all relevant subsystems
This internal issue [b/117489647](http://b/117489647) wasn't picked up by Material Mirror because it's not filed under MDC, but there is public work associated with it.

andrewoverton:
Closes #5375

Adds a global themer for Lists. The relevant subsystems are typography and color.

### DIFF
--- a/components/List/src/ListThemer/MDCListScheme.h
+++ b/components/List/src/ListThemer/MDCListScheme.h
@@ -1,0 +1,54 @@
+// Copyright 2018-present the Material Components for iOS authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import "MaterialColorScheme.h"
+#import "MaterialTypographyScheme.h"
+
+#import <Foundation/Foundation.h>
+
+/**
+ Objects conforming to MDCListScheming provide themers for each subsystem relevant to List Items.
+ */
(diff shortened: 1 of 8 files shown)


components/List/src/ListThemer/MDCListScheme.h
```diff
@@ -0,0 +1,54 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import "MaterialColorScheme.h"
+#import "MaterialTypographyScheme.h"
+
+#import <Foundation/Foundation.h>
+
+/**
+ Objects conforming to MDCListScheming provide themers for each subsystem relevant to List Items.
```
codeman7, line 21: _Tip:_  MDCListScheming represents the design parameters for a MDCList.
 An instance of this protocol can be applied to an instance of MDCList using the
 MDCListThemer APIs.

components/List/src/ListThemer/MDCListScheme.h
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ The typography scheme to be applied to a list cell.
+ */
+@property(nonnull, readonly, nonatomic) id<MDCTypographyScheming> typographyScheme;
+
+@end
+
+/**
+ MDCListScheme is a concrete implementation of MDCListScheming.
```
codeman7, line 38: _Tip:_ 
 An MDCListScheme is a mutable representation of the design parameters for an MDCList.

components/List/tests/unit/MDCListThemerTest.m
```diff
@@ -0,0 +1,89 @@
...
+// limitations under the License.
+
+#import "MaterialList+ColorThemer.h"
+#import "MaterialList+ListThemer.h"
+#import "MaterialList+TypographyThemer.h"
+#import "MaterialList.h"
+
+#import <XCTest/XCTest.h>
+
+#ifndef CGFLOAT_EPSILON
```
codeman7, line 22: _Tip:_ I think you can get rid of this an `#import MaterialMath.h` and use the `MDCCGFloatEqual` function

codeman7 (approved):
Maybe add some test for the MDCListScheme class?

PR #5388, opened 2018-10-10
[ShadowLayer] Subclass UIView in corner radius example

codeman7:
### Context
In working on #4848 I noticed that the example uses an UILabel which may have a different backing layer (possibly CATextLayer) than UIView. This change makes it easier for another team member to address this issue.
### The fix
Subclass UIView which I believe is the most common use case for using a custom shadow layer in this example.
### Related issue
#4848
### Videos
| Before | After |
| - | - |
|develop|shadowafter|


jverkoey (approved):

jverkoey (commented):
Is the label no longer updating for the shadow elevation?

codeman7:
Good catch I'll fix that right now.

PR #5390, opened 2018-10-10
[ShadowLayer] Corner radius

codeman7:
<h3>

```diff
- Warning - This has been attempted 3 times
```
</h3>

### Context
Clients may want to set the corner radius of a UIView that they have overridden the layerClass on to use MDCShadowLayer. Currently if the shadow layer's corner radius changes those changes are not rendered (see Shadow Corner Radius example in MDCDragons). This has been attempted in [this PR](https://github.com/material-components/material-components-ios/pull/4921) and [this PR](https://github.com/material-components/material-components-ios/pull/5224) both had to be rolled back. This is a similar change to #4921 except this change does NOT mark `shadowPathIsInvalid` or call `setNeedsLayout`. This differs from #5224  in that it doesn't set the `shadowPath`'s for the sublayers or set `shadowPathIsInvalid`.

### The problem
Corner radius changes aren't rendered when using MDCShadowLayer.

### The fix
Set the corner radius on the layer and sublayers and update the mask for the sublayers.

### Alternatives
@andrewoverton Suggest
```objectivec
-(void)setCornerRadius:(CGFloat)cornerRadius {

  [self layoutSublayers];
}
```
### Videos
| Before | After |
| - | - |
|shadowafter|shadowcorner|

### DIFF
--- a/components/ShadowLayer/src/MDCShadowLayer.m
+++ b/components/ShadowLayer/src/MDCShadowLayer.m
@@ -203,6 +203,19 @@ static const float kAmbientShadowOpacity = 0.08f;
   return [UIBezierPath bezierPathWithRect:self.bounds];
 }
 
+- (void)setCornerRadius:(CGFloat)cornerRadius {
+  super.cornerRadius = cornerRadius;
+
+  _topShadow.cornerRadius = cornerRadius;
+  _bottomShadow.cornerRadius = cornerRadius;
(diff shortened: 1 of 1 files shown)


ianegordon:
As you noted, this is our fourth attempt at this.
Would you put together an internal test CL and run a global presubmit?

Issue #5396, opened 2018-10-11
[Dialogs] Retain cycle between MDCAlertController and MDCDialogPresentationController


romoore:
When trying to deallocate properties in our tests, I noticed that there are some retained objects apparently unconnected to the test application.

Looks like MDCAlertController and MDCDialogPresentationController are retaining each other.

![screen shot 2018-10-11 at 9 38 35 am](https://user-images.githubusercontent.com/1753199/46808426-50ab4900-cd3a-11e8-807a-d1a15c368fb3.png)

andrewoverton:
![dialogs](https://user-images.githubusercontent.com/8020010/46812279-66247100-cd42-11e8-9619-67d8dcfe54c5.gif)

Things seem to be deallocated properly in the AlertComparison example at least.. Gif is a little blurry :/

romoore:
This is an artifact of never dismissing the AlertController, even though it's never presented.

romoore:
It may also be an artifact of the testing framework. I'm going to close this as obsolete unless we can produce it in an app.

PR #5397, opened 2018-10-11
{Tests} Deallocate test objects at the end of tests.

### LINKED ISSUE #5395: {Tests} Nil all properties in tearDown
All properties created in `-setUp` should be deallocated in `-tearDown`.

https://qualitycoding.org/xctestcase-teardown/

romoore:
Deallocating many of the properties created for each unit test. Since all
XCTestCase instances survive until the end of the test suite, any properties
created for a test will be preserved as well. This frees up around 3.7 MB
(~6%) of memory from the entire test suite, as measured on my desktop.

|Before|After|
|----|----|
|test-objcs-before|test-objcs-after|

See also: https://qualitycoding.org/xctestcase-teardown/

Closes #5395


romoore:
~Note that the Dialogs-related objects appear to be caused by a retain cycle #5396~

It looks like it may just be late deallocation in the tests.

components/FlexibleHeader/tests/unit/FlexibleHeaderNoTrackingScrollViewTests.swift
```diff
@@ -1,26 +0,0 @@
...
-
-// Tests confirming contract with no tracking scroll view
-class FlexibleHeaderNoTrackingScrollViewTests: FlexibleHeaderTrackingScrollViewTests {
-
-  override func setUp() {
-    super.setUp()
-
-    view.trackingScrollView = nil
-  }
-}
```
codeman7, line 26: Just want to confirm this file isn't being imported anywhere right?

codeman7 (approved):
_Tip:_ Consistency with 
```
property = nil;

[super teardown];
``` 
or 
```
property = nil;
[super teardown];
```
I believe the former is the preferred style but whatever you choose please just be consistent.

components/FlexibleHeader/tests/unit/FlexibleHeaderNoTrackingScrollViewTests.swift
romoore: It didn't cause any problems to delete it and I can't find any mentions of it in the code base.

romoore:
@codeman7 I agree with you, but in some cases I was maintaining within-file consistency. I believe keeping files internally consistent is more important (according to the style guide) than improving portions of them incrementally.  On more than one occasion, I've been told not to deviate from the file's style even if it didn't align with the project's style.  I believe we are only supposed to improve a file's style if we do it holistically across the file at once.

PR #4739, opened 2018-08-08
[Dialogs] Remove use of MDCFlatButton for MDCButton and MDCTextButtonThemer

codeman7:
Remove all instances of MDCFlatButton within [Dialogs] component
Replace with MDCTextButtonThemer and MDCButton

MDCFlatButton is the old API and MDCTextButtonThemer/MDCButton are the modern APIs.

Closes #3914

PR #5401, opened 2018-10-11
[Dialogs] Upgrade buttons class to MDCButton

galiak11:
MDCAlertController buttons currently use MDCFlatButton class for buttons. 
This soon-to-be-deprecated class will be switched with MDCButton.
This will allow alert theming to use different "types" for the alert buttons, including text & contained buttons.
Issue: b/117543195
TAP test: cl/217698229 (global pre-submits passed)
(Also see: #4739).

### DIFF
--- a/components/Dialogs/src/private/MDCAlertControllerView+Private.m
+++ b/components/Dialogs/src/private/MDCAlertControllerView+Private.m
@@ -49,8 +49,8 @@ static const CGFloat MDCDialogMessageOpacity = 0.54f;
 @end
 
 @implementation MDCAlertControllerView {
-    NSMutableArray<MDCFlatButton *> *_actionButtons;
-    BOOL _mdc_adjustsFontForContentSizeCategory;
+  NSMutableArray<MDCButton *> *_actionButtons;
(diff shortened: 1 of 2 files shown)


codeman7:
This was attempted in #4739 and was reverted in #5040 just be cautious.

galiak11:
Thanks for the warning. 
What was the client issue with this change? Who was the client?

components/Dialogs/src/private/MDCAlertControllerView+Private.m
```diff
@@ -120,9 +120,12 @@ - (void)setTitle:(NSString *)title {
 }
 
 - (MDCButton *)addActionButtonTitle:(NSString *)actionTitle target:(id)target selector:(SEL)selector {
-  MDCFlatButton *actionButton = [[MDCFlatButton alloc] initWithFrame:CGRectZero];
+  MDCButton *actionButton = [[MDCButton alloc] initWithFrame:CGRectZero];
+  // maintain the Dialogs's default buttons style (as text) after switching to the MDCButton class.
```
randallli: lets move the comment into the method.

randallli (approved):

PR #5366, opened 2018-10-09
{WIP} [Dialogs] Design Review: theming dialog buttons using custom presentation blocks

galiak11:
**Do Not Merge.**

Enable theming of dialog buttons by adding custom presentation blocks to MDCDialogAction, and invoking them before the dialog is presented, and after the buttons materialize and themed by the dialog's "model" properties like "alertController.buttonTitleColor" (so individual button theming overrides the dialog's model theming).

Pending Dialogs/Buttons design review.

Issue: b/116447500

Also see alternative theming approach using semantic meanings in PR #5387.

dialogs buttons model theming
dialogs buttons view theming

PR #5416, opened 2018-10-14
[Dialogs] Theming action buttons in DialogThemer 

galiak11:
Adding semantic emphasis to Dialog actions which allows conditional theming of buttons. The MDCDailogScheme now has a button scheme which can be used to theme buttons according to their action's assigned emphasis.
Issue: b/117608629
Design Review document (go/gmdc-ios-dialogs-design-doc).

Notes:
This is a prototype of the 4th approach to action theming in Dialogs. Option 4 is currently the leading favorite among the button theming approaches. See pros and cons discussion in [Dialogs Buttons API Alternatives - Design Review document (go/gmdc-ios-dialogs-design-doc)](go/gmdc-ios-dialogs-design-doc).

Issue: [b/117608629](https://b.corp.google.com/issues/117608629)
Followup issue: [#5447: [Dialogs] Followup work: changing buttons/actions memory structure](https://github.com/material-components/material-components-ios/pull/5447)

Note: also see [Option 3 PR: Custom presentation blocks](https://github.com/material-components/material-components-ios/pull/5366).

The approach taken by option 4 to theming buttons can be summarized with 3 API additions:
- Adding MDCButtonScheme property to MDCDailogScheme, which will be used by MDCAlertControllerThemer to theme action buttons.
- Adding semantic emphasis to MDCAlertActions, which are available to themers (through [MDCAlertController actions])
- Exposing action buttons to themers (through [MDCAlertController buttonForAction:])

This prototype was built in 2 steps, each in a separate commit, to make it easier to review the work (click the commit number of each step to view changes relevant only to the step):

1. [Dialogs] Creating detached buttons for alert actions before the alert is presented.
This step tracks actions and their buttons in a new MDCAlertActionManager class which is used by both MDCAlertController and MDCAlertControllerView to track actions and vend buttons.
The new [MDCAlertController buttonForAction:] button guarantees to return a button for a valid action. The button is created in detached mode (not attached to the view hierarchy) if requested before the alert is presented. Buttons will be created or attached (if already exist) during presentation time.

2.  [Dialogs] Adding emphasis to actions. Adapting Dialog themers to theme actions based on emphasis.
A MDCActionEmphasis enum was created and added as a property to MDCAlertAction.
Additionally, the MDCButtonScheme property was added to MDCAlertScheme.
In this step, MDCAlertControllerThemer is using the appropriate MDC_xxx_ButtonThemer along with the new MDCButtonScheme to style buttons based on their action's emphasis. Clients can now control button theming through manipulation of the buttonScheme property of the AlertScheme class.

Emphasis theming:
emphasis-theming
Alert-property theming (current):
traditional-theming
Un-themed alert:
unthemed-alert

### DIFF
--- a/components/Dialogs/src/ColorThemer/MDCAlertColorThemer.m
+++ b/components/Dialogs/src/ColorThemer/MDCAlertColorThemer.m
@@ -14,10 +14,9 @@
 
 #import "MDCAlertColorThemer.h"
 
-#import "MDCAlertController+ButtonForAction.h"
-#import "MaterialButtons+ColorThemer.h"
+#import "../../../Buttons/src/ColorThemer/MaterialButtons+ColorThemer.h"
+#import "../MDCAlertController+ButtonForAction.h"
 #import "MaterialButtons.h"
-#import "MaterialDialogs.h"
 
 @implementation MDCAlertColorThemer
 
--- a/components/Dialogs/src/ColorThemer/MDCAlertColorThemer.m
+++ b/components/Dialogs/src/ColorThemer/MDCAlertColorThemer.m
@@ -14,6 +14,9 @@
 
 #import "MDCAlertColorThemer.h"
 
+#import "MDCContainedButtonColorThemer.h"
+#import "MDCOutlinedButtonColorThemer.h"
+#import "MDCTextButtonColorThemer.h"
 #import "MaterialButtons.h"
 #import "MaterialDialogs.h"
 
@@ -23,9 +26,25 @@
                toAlertController:(nonnull MDCAlertController *)alertController {
   alertController.titleColor = [colorScheme.onSurfaceColor colorWithAlphaComponent:(CGFloat)0.87];
   alertController.messageColor = [colorScheme.onSurfaceColor colorWithAlphaComponent:(CGFloat)0.60];
-  alertController.buttonTitleColor = colorScheme.primaryColor;
   alertController.titleIconTintColor = colorScheme.primaryColor;
   alertController.scrimColor = [colorScheme.onSurfaceColor colorWithAlphaComponent:(CGFloat)0.32];
+
+  // Apply theming to buttons based on the action emphasis
+  for (MDCAlertAction *action in alertController.actions) {
+    MDCButton *button = [alertController buttonForAction:action];
+    switch (action.emphasis) {
+      case MDCActionEmphasisHigh:
+        [MDCContainedButtonColorThemer applySemanticColorScheme:colorScheme toButton:button];
+        break;
+      case MDCActionEmphasisMedium:
+        [MDCOutlinedButtonColorThemer applySemanticColorScheme:colorScheme toButton:button];
+        break;
+      case MDCActionEmphasisLow:  // fallthrough
+      default:
+        [MDCTextButtonColorThemer applySemanticColorScheme:colorScheme toButton:button];
+        break;
+    }
+  }
 }
 
 + (void)applyColorScheme:(id<MDCColorScheme>)colorScheme {
--- a/components/Dialogs/src/ColorThemer/MDCAlertColorThemer.m
+++ b/components/Dialogs/src/ColorThemer/MDCAlertColorThemer.m
@@ -14,9 +14,8 @@
 
 #import "MDCAlertColorThemer.h"
 
-#import "MDCContainedButtonColorThemer.h"
-#import "MDCOutlinedButtonColorThemer.h"
-#import "MDCTextButtonColorThemer.h"
+#import "MDCAlertController+ButtonForAction.h"
+#import "MaterialButtons+ColorThemer.h"
 #import "MaterialButtons.h"
 #import "MaterialDialogs.h"
 
--- a/components/Dialogs/src/DialogThemer/MDCAlertControllerThemer.m
+++ b/components/Dialogs/src/DialogThemer/MDCAlertControllerThemer.m
@@ -13,7 +13,9 @@
 // limitations under the License.
 
 #import "MDCAlertControllerThemer.h"
+
 #import "MDCAlertColorThemer.h"
+#import "../MDCAlertController+ButtonForAction.h"
 #import "MDCAlertTypographyThemer.h"
 #import "MaterialButtons+ButtonThemer.h"
 
--- a/components/Dialogs/src/DialogThemer/MDCAlertControllerThemer.m
+++ b/components/Dialogs/src/DialogThemer/MDCAlertControllerThemer.m
@@ -15,6 +15,7 @@
 #import "MDCAlertControllerThemer.h"
 #import "MDCAlertColorThemer.h"
 #import "MDCAlertTypographyThemer.h"
+#import "MaterialButtons+ButtonThemer.h" // todo: update podspec?
 
 @implementation MDCAlertControllerThemer
 
@@ -28,6 +29,24 @@
 
   alertController.cornerRadius = alertScheme.cornerRadius;
   alertController.elevation = alertScheme.elevation;
+
+  for (MDCAlertAction *action in alertController.actions) {
+    MDCButton *button = [alertController buttonForAction:action];
+    switch (action.emphasis) {
+      case MDCActionEmphasisHigh:
+        [MDCContainedButtonThemer applyScheme:alertScheme.buttonScheme toButton:button];
+        break;
+
+      case MDCActionEmphasisMedium:
+        [MDCOutlinedButtonThemer applyScheme:alertScheme.buttonScheme toButton:button];
+        break;
+
+      case MDCActionEmphasisLow:
+      default:
+        [MDCTextButtonThemer applyScheme:alertScheme.buttonScheme toButton:button];
+        break;
+    }
+  }
 }
 
 @end
--- a/components/Dialogs/src/DialogThemer/MDCAlertControllerThemer.m
+++ b/components/Dialogs/src/DialogThemer/MDCAlertControllerThemer.m
@@ -14,8 +14,8 @@
 
 #import "MDCAlertControllerThemer.h"
 
-#import "MDCAlertColorThemer.h"
 #import "../MDCAlertController+ButtonForAction.h"
+#import "MDCAlertColorThemer.h"
 #import "MDCAlertTypographyThemer.h"
 #import "MaterialButtons+ButtonThemer.h"
 
--- a/components/Dialogs/src/DialogThemer/MDCAlertControllerThemer.m
+++ b/components/Dialogs/src/DialogThemer/MDCAlertControllerThemer.m
@@ -15,7 +15,7 @@
 #import "MDCAlertControllerThemer.h"
 #import "MDCAlertColorThemer.h"
 #import "MDCAlertTypographyThemer.h"
-#import "MaterialButtons+ButtonThemer.h" // todo: update podspec?
+#import "MaterialButtons+ButtonThemer.h"
 
 @implementation MDCAlertControllerThemer
 
@@ -30,23 +30,22 @@
   alertController.cornerRadius = alertScheme.cornerRadius;
   alertController.elevation = alertScheme.elevation;
 
+  // Apply theming to buttons based on the action emphasis
   for (MDCAlertAction *action in alertController.actions) {
     MDCButton *button = [alertController buttonForAction:action];
+    // todo: b/117265609: Incorporate dynamic type support in semantic themers
     switch (action.emphasis) {
       case MDCActionEmphasisHigh:
         [MDCContainedButtonThemer applyScheme:alertScheme.buttonScheme toButton:button];
         break;
-
       case MDCActionEmphasisMedium:
         [MDCOutlinedButtonThemer applyScheme:alertScheme.buttonScheme toButton:button];
         break;
-
-      case MDCActionEmphasisLow:
+      case MDCActionEmphasisLow:  // fallthrough
       default:
         [MDCTextButtonThemer applyScheme:alertScheme.buttonScheme toButton:button];
         break;
     }
   }
 }
-
 @end
--- a/components/Dialogs/src/MDCAlertController.h
+++ b/components/Dialogs/src/MDCAlertController.h
@@ -71,12 +71,15 @@
 /** The color applied to the message of Alert Controller.*/
 @property(nonatomic, strong, nullable) UIColor *messageColor;
 
+// b/117717380: Will be deprecated
 /** The font applied to the button of Alert Controller.*/
 @property(nonatomic, strong, nullable) UIFont *buttonFont;
 
+// b/117717380: Will be deprecated
 /** The color applied to the button title text of Alert Controller.*/
 @property(nonatomic, strong, nullable) UIColor *buttonTitleColor;
 
+// b/117717380: Will be deprecated
 /** The color applied to the button ink effect of Alert Controller.*/
 @property(nonatomic, strong, nullable) UIColor *buttonInkColor;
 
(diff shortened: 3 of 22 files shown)


components/Dialogs/src/MDCAlertController.h
```diff
@@ -132,15 +116,66 @@
...
+/**
+ Set this flag to true to enable emphasis-based theming of dialog buttons. Setting the flag to
+ true ignores buttonFont, buttonTitleColor & buttonInkColor, and replaces these with button
+ theming applied by a dialog themer.
+
+ Default value: NO
+
+ b/117717380: will be deprecated
+ */
+@property(nonatomic, assign) BOOL mdc_enableEmphasisThemingOfActionButtons;
```
jverkoey: The `mdc_` prefix is surprising here - can we remove it? We typically only use this prefix in categories of classes we do not own.

components/Dialogs/src/MDCAlertController.h
```diff
@@ -167,10 +217,17 @@ typedef void (^MDCActionHandler)(MDCAlertAction *_Nonnull action);
 /**
  Title of the button shown on the alert dialog.
 
- Alert actions titles must be set in the actionWithTitle:handler: method.
+ Alert actions titles must be set using either of the actionWithTitle:[emphais:]handler: methods.
```
jverkoey: Consider deleting this line and the similar line in the `emphasis` docs - using an init method to initialize properties is assumed knowledge of working with Objective-C APIs.

components/Dialogs/src/MDCAlertController.h
```diff
@@ -152,13 +187,28 @@ typedef void (^MDCActionHandler)(MDCAlertAction *_Nonnull action);
...
  @param title The title of the button shown on the alert dialog.
+ @param emphasis The emphasis of the button that will be rendered in the alert dialog.
+        Unthemed actions will render all emphases as text. Apply themers to the alert
+        to achieve different appearance for different emphases.
  @param handler A block to execute when the user selects the action.
  @return An initialized MDCActionAlert object.
  */
 + (nonnull instancetype)actionWithTitle:(nonnull NSString *)title
+                               emphasis:(MDCActionEmphasis)emphasis
                                 handler:(__nullable MDCActionHandler)handler;
```
jverkoey, line 192: Consider annotating as the designated initializer.

components/Dialogs/src/MDCAlertController.h
```diff
@@ -132,15 +116,66 @@
...
+
+ Actions are the possible reactions of the user to the presented alert. Actions are added as a
+ button at the bottom of the alert. Affirmative actions should be added before dismissive actions.
+ Action buttons will be laid out from right to left if possible or top to bottom depending on space.
+
+ Material spec recommends alerts should not have more than two actions.
+
+ @param action Will be added to the end of MDCAlertController.actions.
+ */
+- (void)addAction:(nonnull MDCAlertAction *)action;
```
jverkoey, line 146: Consider moving the existing methods back up (`actions` and `addAction:`) to minimize amount of delta in this PR.

components/Dialogs/src/MDCAlertController.h
```diff
@@ -132,15 +116,66 @@
...
+ be attached to the view hierarchy at the time the method is called.
+
+ This method is commonly used by themers to style the button associated with the action.
+
+ @param action The action with which the button is associated. Must be an existing action that
+               had been previously added through addAction: to the alert.
+ @return The button associated with the action, or nil if the action doesn't exist (the action
+          must first be added to the alert).
+ */
+- (nullable MDCButton *)buttonForAction:(nonnull MDCAlertAction *)action;
```
jverkoey: Similar to UIGestureRecognizer's subclassing header `UIGestureRecognizerSubclass.h`, this API is a good example of something that we may want to move to a separate category header so that clients don't have access to it in their day-to-day use of the Dialogs component.

In order to access this API, a client will explicitly opt-in to importing the relevant header.

Practically speaking, this would mean adding an `MDCAlertController+ButtonForAction.h` header and *not* including that header in the component's umbrella header.

components/Dialogs/src/MDCAlertController.h
```diff
@@ -132,15 +116,66 @@
...
+ */
+@property(nonatomic, assign) BOOL mdc_enableEmphasisThemingOfActionButtons;
+
 @end
 
+typedef NS_ENUM(NSInteger, MDCActionEmphasis) {
+  MDCActionEmphasisLow = 0,
+  MDCActionEmphasisMedium = 1,
+  MDCActionEmphasisHigh = 2,
+};
```
jverkoey, line 157: Missing documentation for the enum and each of the values.

components/Dialogs/src/ColorThemer/MDCAlertColorThemer.m
```diff
@@ -38,4 +66,22 @@ + (void)applyColorScheme:(id<MDCColorScheme>)colorScheme {
   #endif
 }
 
+// b/117717380: Investigate: these may not be needed once traditional button theming is deprecaed
```
jverkoey: s/deprecaed/deprecated/

components/Dialogs/src/DialogThemer/MDCAlertControllerThemer.m
```diff
@@ -15,11 +15,19 @@
 #import "MDCAlertControllerThemer.h"
 #import "MDCAlertColorThemer.h"
 #import "MDCAlertTypographyThemer.h"
+#import "MaterialButtons+ButtonThemer.h"  // todo: update podspec?
```
jverkoey: Remove todo? Or is this something we need to do in a follow-up?

components/Dialogs/src/MDCAlertController.h
jverkoey: I'm also concerned about adding APIs that imply the existence of a theming system to our components. Can we move this logic to the themer APIs instead by introducing new themer APIs that handle the new, desired behavior?

components/Dialogs/src/private/MDCAlertActionManager.h
```diff
@@ -0,0 +1,61 @@
...
+// limitations under the License.
+
+#import <Foundation/Foundation.h>
+#import "MaterialButtons.h"
+#import "MaterialDialogs.h"
+
+@interface MDCAlertActionManager : NSObject
+
+@property(nonatomic, nonnull, strong, readonly) NSArray<MDCAlertAction *> *actions;
+/**
```
jverkoey, line 26: Tip: Style: empty lines above and below all methods.

components/Dialogs/src/private/MDCAlertActionManager.m
```diff
@@ -0,0 +1,95 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import "MDCAlertActionManager.h"
+
+@interface MDCAlertActionManager ()
+@property(nonatomic, nonnull, strong) NSMapTable *actionButtons;
```
jverkoey: Missing `<Key, Value>` types here.

components/Dialogs/src/private/MDCAlertActionManager.m
```diff
@@ -0,0 +1,95 @@
...
+  self = [super init];
+  if (self) {
+    _actions = [[NSMutableArray alloc] init];
+    _actionButtons = [NSMapTable mapTableWithKeyOptions:NSMapTableWeakMemory
+                                           valueOptions:NSMapTableStrongMemory];
+  }
+  return self;
+}
+
+- (NSArray<MDCButton *> *)sortedButtons {
```
jverkoey: The name implies that we're applying a sorting algorithm to the buttons, but it looks like we may be retrieving the buttons in action order. Consider `allButtonsInActionOrder`.

components/Dialogs/src/private/MDCAlertActionManager.m
```diff
@@ -0,0 +1,95 @@
...
+  NSMutableArray<MDCAlertAction *> *_actions;
+}
+
+@dynamic sortedButtons;
+
+- (instancetype)init {
+  self = [super init];
+  if (self) {
+    _actions = [[NSMutableArray alloc] init];
+    _actionButtons = [NSMapTable mapTableWithKeyOptions:NSMapTableWeakMemory
```
jverkoey, line 33: Do we need the weak keys for our action items? If not, I recommend replacing with NSMutableDictionary.

components/Dialogs/src/private/MDCAlertActionManager.m
```diff
@@ -0,0 +1,95 @@
...
+}
+
+- (nullable MDCButton *)buttonForAction:(nonnull MDCAlertAction *)action {
+  return [self.actionButtons objectForKey:action];
+}
+
+- (nullable MDCAlertAction *)actionForButton:(nonnull MDCButton *)button {
+  NSEnumerator *enumerator = [self.actionButtons keyEnumerator];
+  MDCAlertAction *action = nil;
+  while ((action = [enumerator nextObject])) {
```
jverkoey: Consider taking advantage of the fact that NSMapTable conforms to `NSFastEnumeration`:

`for (MDCAlertAction *action in self.actionButtons)`

jverkoey: Alternatively, consider using `allKeysForObject:` if you move to an NSDictionary.

components/Dialogs/src/private/MDCAlertActionManager.m
```diff
@@ -0,0 +1,95 @@
...
+  }
+  return action;
+}
+
+// creating a new buttons and associating it with the given action. the button is not added
+// to view hierarchy.
+- (nullable MDCButton *)addButtonForAction:(nonnull MDCAlertAction *)action
+                                    target:(nullable id)target
+                                  selector:(SEL _Nonnull)selector {
+  MDCButton *button = [self makeButtonForAction:action target:target selector:selector];
```
jverkoey: Consider returning nil early if the action is not already in the `_actions` array.

components/Dialogs/src/private/MDCAlertActionManager.m
```diff
@@ -0,0 +1,95 @@
...
+    if (currButton == button) {
+      return action;
+    }
+  }
+  return action;
+}
+
+// creating a new buttons and associating it with the given action. the button is not added
+// to view hierarchy.
+- (nullable MDCButton *)addButtonForAction:(nonnull MDCAlertAction *)action
```
jverkoey: Consider renaming to `createButtonForAction` to indicate that we're not necessarily adding a button each time (we may want to return early if there is already a button for this action).

jverkoey (dismissed):

components/Dialogs/src/MDCAlertController.h
romoore: +1 Please remove this property. Emphasis should always be used.

components/Dialogs/src/private/MDCAlertActionManager.m
galiak11: We don't necessarily need weak keys, however NSDictionary copies the keys (the actions in this case) which makes it harder to later access an action with a subscript. 
Alternatively, I could store the action objects in an NSDictionary as an NSValue. 
Not sure which solution is better. Any thoughts?

jverkoey: Ah - you may want to add `NSPointerFunctionsOpaquePersonality` to the key options to indicate that you're not implementing -hash and -isEqual.

components/Dialogs/src/MDCAlertController.h
jverkoey: Nevermind - this is a static method and can't be marked as a designated initializer.

components/Dialogs/src/private/MDCAlertActionManager.m
jverkoey: Speaking offline, we looked at this and felt it may be simpler to just have one array with an NSObject that stores an action and button property, something like so:

```objc
interface ActionUserInfo: NSObject {
  initWithAction:
  property(readonly) action;
  property(readwrite nullable) button;
}
```

galiak11: Followup discussion: https://github.com/material-components/material-components-ios/issues/5447

components/Dialogs/src/private/MDCAlertActionManager.m
romoore: Why can't we use an NSDictionary here?  Please add a comment to document why NSMapTable is required.

romoore (dismissed):
The design doc looks like it's not yet "In Review" and has some editing required to clean-up the options and something at the end was rejected.  

If this is just a prototype for one of the options, can you please adjust the base branch to be an experimental branch or something else instead of `develop`? As it stands, it looks like this is going to be changed in the current component rather than in the examples.

If the API review already took place, please add a link to it in the PR description for others to see.

components/Dialogs/src/MDCAlertController.h
randallli: Initially I believed that we can add an entry to the enum to use the legacy "emphasis" aka the old APIs.

However after reading through the PR I see this as a flag to turn on legacy/new way to style buttons. We could rename it to "useLegacyGlobalButtonStylingValues" or something to indicate this is a feature flag that will hold over while clients migrate over to the newer direct access to buttons. Implied here is the use of the emphasis meta data to style the buttons outside of the dialog since you now have full access to the buttons.

randallli (approved):

components/Dialogs/src/MDCAlertController.h
romoore: This should not be needed, since by default legacy will be "low emphasis" which equates to Text Buttons. Since Text Buttons are the "legacy" (or current) style for buttons in an Alert, we should expect no change to any clients that do not adopt the new styling.  

@randallli  Can you please explain why you believe a separate flag (in addition to the `emphasis` argument) is required?

components/Dialogs/src/MDCAlertController.h
galiak11: I'm thinking keeping all action-related logic together is slightly more important than minimizing delta in this case.

components/Dialogs/src/MDCAlertController.h
galiak11: Done

components/Dialogs/src/private/MDCAlertActionManager.h
galiak11: Done

galiak11 (commented):
Some work not done in this PR is moved to issue https://github.com/material-components/material-components-ios/issues/5447.

components/Dialogs/src/MDCAlertController.h
galiak11: This flag is now removed.

components/Dialogs/src/MDCAlertController.h
galiak11: Done.

components/Dialogs/src/MDCAlertController.h
galiak11: Agreed. will be resolved in a followup PR (issue: #5447)

components/Dialogs/src/private/MDCAlertActionManager.m
galiak11: Done

components/Dialogs/src/private/MDCAlertActionManager.m
galiak11: Done

components/Dialogs/src/private/MDCAlertActionManager.m
galiak11: Done

components/Dialogs/src/private/MDCAlertActionManager.m
galiak11: I've renamed it (to: buttonsInActionOrder), although this may require further consideration. Should this method create buttons that are missing, similar to "buttonForAction"?  I've renamed it, but didn't addressing the larger question in the PR. I moved the discussion to the followup issue #5447.

PR #5439, opened 2018-10-17
[List] SelfSizingStereoCell should invalidate its layout on re-use

### LINKED ISSUE #5436: [List] SelfSizingStereoCell should invalidate its layout on re-use
From the internal issue:

> Because MDCSelfSizingStereoCell isn't calling -setNeedsLayout in -prepareForReuse, the cells labels can get truncated due to faulty layout.

---

This is an internal issue. If you are a Googler, please visit [b/117883536](http://b/117883536) for more details.

---
#### Internal data
- Associated internal bug: [b/117883536](http://b/117883536)

jpoumailloux:
This change adds a call to -setNeedsLayout in -prepareForReuse. Without this call, the cell may not re-do a layout pass on re-use, causing truncation of the labels.

closes https://github.com/material-components/material-components-ios/issues/5436

### DIFF
--- a/components/List/src/MDCSelfSizingStereoCell.m
+++ b/components/List/src/MDCSelfSizingStereoCell.m
@@ -156,17 +156,15 @@ static const CGFloat kDetailColorOpacity = 0.6f;
 - (void)prepareForReuse {
   [super prepareForReuse];
 
-  [self invalidateCachedLayouts];
-
   self.titleLabel.text = nil;
   self.detailLabel.text = nil;
   self.leadingImageView.image = nil;
   self.trailingImageView.image = nil;
 
+  [self setNeedsLayout];
+
   [self mdc_setAdjustsFontForContentSizeCategory:NO];
   [self resetMDCSelfSizingStereoCellLabelProperties];
-
-  [self setNeedsLayout];
 }
 
 #pragma mark Layout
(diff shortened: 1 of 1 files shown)


components/List/src/MDCSelfSizingStereoCell.m
```diff
@@ -165,6 +165,8 @@ - (void)prepareForReuse {
 
   [self mdc_setAdjustsFontForContentSizeCategory:NO];
   [self resetMDCSelfSizingStereoCellLabelProperties];
+
+  [self setNeedsLayout];
```
andrewoverton: @jpoumailloux , would `[self invalidateCachedLayouts];` do the trick here? Do we need to call `setNeedsLayout`?

jpoumailloux: Calling -invalidateCachedLayouts would not be sufficient. This method is actually already called in -prepareForReuse a couple of lines above, and it still caused truncations of the labels. I've updated the change to remove the call to -invalidateCachedLayouts since calling -setNeedsLayout will also call it (the class overrides -setNeedsLayout to do that).

andrewoverton: Got it, thanks 👍

andrewoverton (approved):

PR #5453, opened 2018-10-19
[NavigationDrawer] Add Fakes files for tests

codeman7:
### Context
In #5423 my thinking was that we would need to test MDCBottomDrawerViewController but majority of the test are going to be related to MDCBottomDrawerContainerViewController. We may still want to test MDCBottomDrawerViewController so that is why I haven't completely removed that file for tests. But, we need fakes in a couple places so this adds _Fakes_ files so that code can be shared.

### The problem
We need fakes for test in MaterialNavigationDrawer because we can't present and that code should be shared.

### The fix
Add a `MDCNavigationDrawerFakes` files so we can have fakes and they can be shared across test files.

### Related bugs
#4911

### DIFF
--- a/components/NavigationDrawer/BUILD
+++ b/components/NavigationDrawer/BUILD
@@ -44,7 +44,10 @@ mdc_objc_library(
 mdc_objc_library(
     name = "unit_test_sources",
     testonly = 1,
-    srcs = native.glob(["tests/unit/*.m"]),
+    srcs = native.glob([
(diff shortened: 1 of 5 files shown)


romoore (dismissed):
Please keep this PR to just adding the flakes classes. Move the clean-up and tests changes to a follow-up PR.

yarneo (approved):

PR #5461, opened 2018-10-19
[NavigationDrawer] Add unit test for height related properties

codeman7:
### Context
We currently are missing a lot unit test for NavigationDrawer. In order to keep the PR's manageable I have decided to break them up into related groups. This group is the two properties that look at the headerViewControllers.preferredContentSize.height and tests those two.
### The problem
We are missing a lot of unit test for NavigationDrawer
### The fix
Add test for `topHeaderHeight` and `contentHeaderHeight`

### Bugs
Related to #4911

### DIFF
--- a/components/NavigationDrawer/tests/unit/MDCNavigationDrawerScrollViewTests.m
+++ b/components/NavigationDrawer/tests/unit/MDCNavigationDrawerScrollViewTests.m
@@ -111,14 +111,15 @@
   CGFloat mdcDeviceTopSafeArea = 20.f;
   CGSize fakePreferredContentSize = CGSizeMake(200, 300);
   MDCNavigationDrawerFakeHeaderViewController *fakeHeader =
-  [[MDCNavigationDrawerFakeHeaderViewController alloc] init];
+      [[MDCNavigationDrawerFakeHeaderViewController alloc] init];
   self.fakeBottomDrawer.headerViewController = fakeHeader;
 
   // When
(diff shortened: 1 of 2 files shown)


components/NavigationDrawer/tests/unit/MDCNavigationDrawerScrollViewTests.m
```diff
@@ -21,7 +21,8 @@ @interface MDCBottomDrawerContainerViewController (ScrollViewTests)
 
 @property(nonatomic) BOOL scrollViewObserved;
 @property(nonatomic, readonly) UIScrollView *scrollView;
-@property(nonatomic) BOOL currentlyFullScreen;
+@property(nonatomic, readonly) CGFloat topHeaderHeight;
```
rami-a, line 24: _Tip:_ These 2 properties appear to be unused. In that case, they should be removed.

rami-a (approved):

components/NavigationDrawer/tests/unit/MDCNavigationDrawerScrollViewTests.m
codeman7: In the test it's `fakeBottomDrawer.topHeaderHeight` that property has to be exposed via the category.

rami-a: Oops, I misread where these properties were declared. You're right.

PR #5465, opened 2018-10-22
[NavigationDrawer] Add test for contentHeaderTopInset and presentingViewBounds

codeman7:
### Content
NavigationDrawer is missing unit test. One of the _heavy lifting_ functions is `cacheLayoutCalculation`, which has many side effects. One of those side effects is setting the `contentHeaderTopInset`. In cacheLayoutCalculations there is two paths to set `contentHeaderTopInset` one is with scrollable content and one is without scrollable content. If the content isn't scrollable then the `contentHeaderTopInset` is the size of the `presentingViewBounds.size.height` minus the `headerViewController.preferredContentSize.height` and the `contentViewController.preferredContentSize.height`. If the content is scrollable then the `contentHeaderTopInset` is the size of the `presentingViewBounds` * `initialDrawerFactor`.  

### Work in the PR
Test `presentingViewBounds`

Test `contentHeaderTopInset` 
-  Scrollable
   - With large headerViewController
   - With large contentViewController
   - With large contentViewController and headerViewController
- Non-scrollable
  - With only a headerViewController
  - With only a contentViewController
  - With both a contentViewController and a headerViewController
  - With no contentViewController or headerViewController

### Remaining Work
For `cacheLayoutCalculations` we still need to test `contentHeightSurplus`
### Related issues
#4911

### DIFF
--- a/components/NavigationDrawer/tests/unit/MDCNavigationDrawerScrollViewTests.m
+++ b/components/NavigationDrawer/tests/unit/MDCNavigationDrawerScrollViewTests.m
@@ -23,6 +23,9 @@
 @property(nonatomic, readonly) UIScrollView *scrollView;
 @property(nonatomic, readonly) CGFloat topHeaderHeight;
 @property(nonatomic, readonly) CGFloat contentHeaderHeight;
+@property(nonatomic, readonly) CGFloat contentHeaderTopInset;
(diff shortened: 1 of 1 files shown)


yarneo (approved):
LGTM if CI passes

PR #5466, opened 2018-10-22
[NavigationDrawer] Test content height surplus and scrolls to reveal

codeman7:
### Context
NavigationDrawer is missing unit test. One of the _heavy lifting_ functions is `cacheLayoutCalculation`, which has many side effects. One of those side effects is setting the `contentHeightSurplus`. `contentHeightSurplus` effects `scrollsToReveal` so those test were also added. 

**_Note:_** I didn't add any comments in the test but those may be necessary to give future maintainers context on why the test are a given result.

### Related issues
#4911 
### Related PR
#5465


yarneo (approved):
LGTM if CI passes

PR #5475, opened 2018-10-22
[List] Do not enforce label text alignment based off effective user interface layout direction

### LINKED ISSUE #5472: [List] Stereo cell should not set textAlignment based off of effectiveInterfaceLayoutDirection
The Stereo cell shouldn't enforce textAlignment of its labels based off the effective user interface layout direction. It should be possible to make an English label LTR even if the system language is Arabic.

## Reproduction steps

1. Create a list using the MDCSelfSizingStereoCell.
2. Make the title text in Arabic and set the textAlignment to RTL.
3. Launch the app with an LTR scheme.

## Expected behavior

The title text should be RTL.

## Actual behavior

The title text is LTR.

## Additional Context

The way to solve this is to get rid of any calls to `-setTextAlignment:`.

andrewoverton:
The problem:

We were enforcing label text alignment based off effective user interface layout direction.
As a result, if the system language was English and you created an Arabic label (with its textAlignment set to Right) the label would ultimately end up with a textAlignment of Left.

The solution:

Don't set textAlignment from within the cell. Leave it to the client.

Closes #5472.

Before:
simulator screen shot - iphone x - 2018-10-22 at 16 23 26

After:
simulator screen shot - iphone x - 2018-10-22 at 16 23 58

### DIFF
--- a/components/List/src/MDCSelfSizingStereoCell.m
+++ b/components/List/src/MDCSelfSizingStereoCell.m
@@ -123,8 +123,6 @@ static const CGFloat kDetailColorOpacity = 0.6f;
   self.leadingImageView.frame = layout.leadingImageViewFrame;
   self.trailingImageView.frame = layout.trailingImageViewFrame;
   if (self.mdf_effectiveUserInterfaceLayoutDirection == UIUserInterfaceLayoutDirectionRightToLeft) {
-    self.titleLabel.textAlignment = NSTextAlignmentRight;
-    self.detailLabel.textAlignment = NSTextAlignmentRight;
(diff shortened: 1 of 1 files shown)


codeman7 (approved):

romoore:
@andrewoverton Do we have an Example that shows clients how to use MDFInternationalization to set the text alignment for these mixed-language use cases?

andrewoverton:
@romoore there is not currently an example showing such a thing

romoore:
Opened #5499 to track my request/suggestion.

PR #5482, opened 2018-10-23
[Dialogs] Refactor test to use setup and teardown

codeman7:
### Context
In working on #5296 I noticed a lot of the test had repeated code that can be pulled into a `setup` function. Additionally some of the test were missing _Given_, _When_, and _Then_ statements so those were added. This change hopefully makes this code more readable to future maintainers. I also noticed some of the test had extra work in the _Given_ that wasn't necessary so I removed that.
### The problem
Repeated steps in the _Given_ section of the test
### The fix
Implement the `setup` and `teardown` functions as well as add missing _Given_, _When_, and _Then_ comments.

### Related bugs
This isn't related to any bugs this is just something that I saw that could be improved.

### DIFF
--- a/components/Dialogs/tests/unit/MDCAlertControllerTests.m
+++ b/components/Dialogs/tests/unit/MDCAlertControllerTests.m
@@ -55,30 +55,35 @@ static NSString *const MDCAlertControllerSubclassValueKey = @"MDCAlertController
 #pragma mark - Tests
 
 @interface MDCAlertControllerTests : XCTestCase
-
+@property(nonatomic, nullable) MDCAlertController *alert;
 @end
 
 @implementation MDCAlertControllerTests
 
-- (void)testInit {
-  // Given
-  MDCAlertController *alert = [[MDCAlertController alloc] init];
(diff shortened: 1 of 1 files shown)


components/Dialogs/tests/unit/MDCAlertControllerTests.m
```diff
@@ -219,56 +225,54 @@ - (void)testAlertControllerColorSettingBetweenActions {
...
-  MDCAlertController *alert = [MDCAlertController alertControllerWithTitle:title
-                                                                   message:message];
-  alert.titleFont = [UIFont systemFontOfSize:25];
 
-  MDCAlertControllerView *view = (MDCAlertControllerView *)alert.view;
+  // When
+  self.alert.titleFont = [UIFont systemFontOfSize:25];
+
+  // Then
+  MDCAlertControllerView *view = (MDCAlertControllerView *)self.alert.view;
```
romoore, line 236: Tip: Move this into a property in a follow-up PR.

romoore (approved):

PR #5497, opened 2018-10-23
[AppBar] Initial frame wrong in MDCAppBarViewController for iPad split screen

### LINKED ISSUE #5479: Internal issue: b/118126496
This was filed as an internal issue. If you are a Googler, please visit [b/118126496](http://b/118126496) for more details.

---
#### Internal data
- Associated internal bug: [b/118126496](http://b/118126496)

rami-a:
### Context
When the migration from MDCAppBar to MDCAppBarViewController occurred, a small piece of code was left out, specifically in `addSubviewsToParent`.
### The problem
When running the catalog app in split screen on iPad and viewing the examples for MDCAppBarViewController, the App Bar is not sized correctly and overflows outside of the bounds.
### The fix
Re-add the frame width modifying logic from `addSubviewsToParent` into `didMoveToParentViewController:parent` for MDCAppBarViewController.
### Related bugs
closes #5479, b/118126496

### Screenshots
| Before | After |
| - | - |
|before|after|

### DIFF
--- a/components/AppBar/src/MDCAppBarViewController.m
+++ b/components/AppBar/src/MDCAppBarViewController.m
@@ -248,7 +248,7 @@ static NSString *const kMaterialAppBarBundle = @"MaterialAppBar.bundle";
   [self.navigationBar observeNavigationItem:parent.navigationItem];
 
   CGRect frame = self.view.frame;
-  frame.size.width = parent.view.bounds.size.width;
+  frame.size.width = CGRectGetWidth(parent.view.bounds);
   self.view.frame = frame;
 }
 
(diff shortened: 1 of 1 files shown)


components/AppBar/src/MDCAppBarViewController.m
```diff
@@ -246,6 +246,10 @@ - (void)didMoveToParentViewController:(UIViewController *)parent {
   [super didMoveToParentViewController:parent];
 
   [self.navigationBar observeNavigationItem:parent.navigationItem];
+
+  CGRect frame = self.view.frame;
+  frame.size.width = parent.view.bounds.size.width;
```
ianegordon: We need to use CGRectGetWidth() if we are using non-standardized CGRects.

ianegordon (changes requested):

components/AppBar/src/MDCAppBarViewController.m
rami-a: Done

ianegordon (approved):

PR #5500, opened 2018-10-23
[List] Add RTL Arabic text to list example

### LINKED ISSUE #5498: [List] Add label introspection with subsequent textAlignment setting to List example
## Feature Request/Suggestion

It would be nice to have some code in our List example showing how one might determine the correct textAlignment for a label based off the language of its text.

andrewoverton:
## The problem:

There wasn't any RTL text in our List example!

## The solution:

I added some RTL text to our List example.

Closes #5498

### Before:
simulator screen shot - iphone x - 2018-10-23 at 11 28 37

### After:
simulator screen shot - iphone x - 2018-10-23 at 11 28 02

### DIFF
--- a/components/List/examples/MDCSelfSizingStereoCellExample.m
+++ b/components/List/examples/MDCSelfSizingStereoCellExample.m
@@ -14,6 +14,8 @@
 
 #import "MDCSelfSizingStereoCellExample.h"
 
+#import <MDFInternationalization/MDFInternationalization.h>
+
 #import "MaterialList+ColorThemer.h"
 #import "MaterialList+TypographyThemer.h"
 #import "MaterialList.h"
@@ -110,6 +112,8 @@ static NSString *const kSelfSizingStereoCellExampleDescription =
                                                 forIndexPath:indexPath];
   cell.titleLabel.text = self.randomStrings[indexPath.item];
   cell.detailLabel.text = self.randomStrings[(indexPath.item + 1) % self.randomStrings.count];
+  cell.titleLabel.textAlignment = [self textAlignmentForText:cell.titleLabel.text];
+  cell.detailLabel.textAlignment = [self textAlignmentForText:cell.detailLabel.text];
   cell.leadingImageView.image =
       [[UIImage imageNamed:@"Cake"] imageWithRenderingMode:UIImageRenderingModeAlwaysTemplate];
   cell.trailingImageView.image =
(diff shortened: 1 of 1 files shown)


components/List/examples/MDCSelfSizingStereoCellExample.m
```diff
@@ -122,21 +126,50 @@ - (UICollectionViewCell *)collectionView:(UICollectionView *)collectionView
...
-  for (NSInteger i = 0; i < numberOfWords; i++) {
-    NSInteger lengthOfWord = 0 + arc4random() % (10 - 0);
-    NSMutableArray *letterArray = [[NSMutableArray alloc] initWithCapacity:lengthOfWord];
-    for (NSInteger j = 0; j < lengthOfWord; j++) {
-      int asciiCode = 97 + arc4random() % (122 - 97);
-      NSString *characterString = [NSString stringWithFormat:@"%c", asciiCode];
-      [letterArray addObject:characterString];
+- (NSTextAlignment)textAlignmentForText:(NSString *)text {
+  if (text.length > 0) {
+    if (text.mdf_calculatedLanguageDirection == NSLocaleLanguageDirectionLeftToRight) {
```
ianegordon, line 131: There may be some performance impact from using mdf_calculatedLanguageDirection in cellForIndexPath, but if everything looks okay, LGTM.

ianegordon (approved):

components/List/examples/MDCSelfSizingStereoCellExample.m
romoore: Was the performance impact measured in scrolling? Can we still hit 60 fps on an iPhone 5c/iPhone 4S?

romoore: I did a quick-and-dirty test on an iPhone 5c/iOS 10.3.3 and found that for 14 seconds worth of casual scrolling (over a 30 second period, depending on trace), it added about 116 ms worth of overhead (~0.8% increase).

mdc-list-example-mdf-overhead

andrewoverton: Interesting! Are we okay with this?

romoore: I dropped the performance to about half by caching the result in #5506

PR #5506, opened 2018-10-24
[ListItems] Improve BiDi example performance.

romoore:
The example showing how to use MDF Internationalization's language
direction code was somewhat inefficient, calling the method twice on the
same string. Instead it should be called once and the result checked.

In manual testing, dropped overhead from 0.8% of Main Thread time to
0.4% while scrolling the example.

Improves #5500


andrewoverton (approved):

andrewoverton:
Thanks Rob!

romoore:
Here's screenshots of the Instruments testing.

|Before|After|
|--|--|
|mdc-list-example-mdf-overhead-before|mdc-list-example-mdf-overhead-after|

ianegordon (approved):

PR #5517, opened 2018-10-24
[Snackbar] Update MDCSnackbarColorThemer to set correct background color to match the spec

### LINKED ISSUE #5512: Internal issue: b/118375446
This was filed as an internal issue. If you are a Googler, please visit [b/118375446](http://b/118375446) for more details.

---
#### Internal data
- Associated internal bug: [b/118375446](http://b/118375446)

rami-a:
### The problem
MDCSnackbarColorThemer is not aligned with the values for the background color from the spec and it should not appear transparent.

### The fix
Update the themer to blend the 2 colors from the spec to ensure Snackbar is aligned to the spec.

### Related bugs
Closes #5512 

### Screenshots
| Before | After |
| - | - |
|mdc snackbar old|mdc snackbar new|

**Note:** I rewrote the unit tests in Obj-c because of an issue with comparing colors that have alpha channels.


components/Snackbar/tests/unit/MDCSnackbarColorThemerTests.m
```diff
@@ -0,0 +1,93 @@
...
+
+  [super tearDown];
+}
+
+- (void)testSnackbarColorThemerChangesCorrectParameters {
+  // Given
+  MDCSemanticColorScheme *colorScheme = [[MDCSemanticColorScheme alloc] init];
+  colorScheme.surfaceColor = [UIColor redColor];
+  colorScheme.onSurfaceColor = [UIColor blueColor];
+  MDCSnackbarManager.snackbarMessageViewBackgroundColor = [UIColor whiteColor];
```
romoore, line 72: Please use an instance of the manager instead of the singleton.

romoore (changes requested):

components/Snackbar/tests/unit/MDCSnackbarColorThemerTests.m
romoore: See https://github.com/material-components/material-components-ios/blob/develop/components/Snackbar/tests/unit/MDCSnackbarManagerInstanceTests.m for examples, though we should use properties now instead of reallocating in every test.

components/Snackbar/tests/unit/MDCSnackbarColorThemerTests.m
```diff
@@ -0,0 +1,93 @@
...
+  MDCSnackbarManager.snackbarMessageViewBackgroundColor = self.snackbarMessageViewBackgroundColor;
+  for (NSNumber *state in self.titleColorForState.allKeys) {
+    if (self.titleColorForState[state] != nil) {
+      [MDCSnackbarManager setButtonTitleColor:self.titleColorForState[state]
+                                     forState:state.unsignedIntegerValue];
+    }
+  }
+
+  // Clean-up the test case
+  [self.titleColorForState removeAllObjects];
```
romoore, line 58: Please also assign `nil` to `titleColorForState` so that the array can be deallocated.

components/Snackbar/tests/unit/MDCSnackbarColorThemerTests.m
romoore: @rami-a Just reminded me that the themer does not currently take an instance of the SnackbarManager. This PR is fine as-is, but we should have a new API eventually to support theming an instance of SnackbarManager.

romoore (approved):

components/Snackbar/tests/unit/MDCSnackbarColorThemerTests.m
rami-a: done

PR #5570, opened 2018-11-01
[NavigationDrawer] Add scrim color to color themer

codeman7:
### Context
In #5458 we did not themer the scrim color because it wasn't a property on NavigationDrawer. The property was added in #5568. This PR themes the scrim color with the correct value. As more properties are added to navigation drawer more adjustments to the themer will need to be made.
### The problem
We aren't theming the scrim color
### The fix
Apply the correct values to the scrim color within the color themer
### Screenshots
| Before | After |
| - | - |
|simulator screen shot - iphone xs max - 2018-11-01 at 06 47 35|simulator screen shot - iphone xs max - 2018-11-01 at 06 46 35|

### DIFF
--- a/components/NavigationDrawer/src/ColorThemer/MDCBottomDrawerColorThemer.m
+++ b/components/NavigationDrawer/src/ColorThemer/MDCBottomDrawerColorThemer.m
@@ -14,12 +14,15 @@
 
 #import "MDCBottomDrawerColorThemer.h"
 
+static const CGFloat kScrimAlpha = (CGFloat)0.32;
+
 @implementation MDCBottomDrawerColorThemer
 
 + (void)applySemanticColorScheme:(id<MDCColorScheming>)colorScheme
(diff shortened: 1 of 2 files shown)


rami-a (approved):
Looks good.

One thing I noticed in the params doc, is that the scrim is marked as optional for navigation drawer component. Is this something we should implement?

codeman7:
@yarneo How do you feel about scrim being optional, I saw that it was optional in the document as well.

romoore:
I would be concerned if there were no scrim to hide the content behind the Navigation Drawer. Would that mean that users could reasonably expect to interact with the views not directly obscured by the Navigation Drawer? If yes, I think we'll have similar problems with UIAccessibility the way we would with Backdrop. If no, then I think users will be a bit confused that their Drawer gets dismissed while they're trying to touch a button elsewhere.

How would the navigation stack work without the scrim? Would the Navigation Drawer remain on-screen even as the VC that presented it is dismissed or popped?

PR #5339, opened 2018-10-06
[BottomNavigation] Change badge position

codeman7:
<h4>

```diff
- Warning - This may break screenshot testing -
```
</h4>

### Context
Currently the badge is centered on the top of the `iconImageView`. The `iconImageView`'s  `origin.y` is at 8.883 in MDC because we vertically center the _content_ in the MDCBottomNavigationItemView. We get the _content_ height by:

_contentHeight_ = icon.height + label.height + contentVerticalMargin

_contentVerticalMargin_ is the spacing between the icon and label.

The **_badge_** is 14dp tall. Centering the badge on the `iconImageView`'s `origin.y` puts the badge at 1.883.
### The problem
The badge is too high in relation to the top of the bottom navigation bar and doesn't match internal guidance.
### The fix
Make the badge and iconImageView have the same origin y, giving users more separation between the top of the navigation bar and the top of the badge.
### Related PR
Closes #4731 
### Screenshots
| Before | After |
| - | - |
|simulator screen shot - iphone x - 2018-10-06 at 11 10 04|simulator screen shot - iphone x - 2018-10-06 at 11 09 29|

PR #5571, opened 2018-11-01
[BottomNavigation] Fix selection animation for badge

codeman7:
### Context
In #5339 this bug was introduced. The motivation that lead to this bug was not having to recall this function in three different places. In doing it all in once place the animation broke.
### The problem
When _BottomNavigationItemView_ would animate the badge would jump to its new position while the title and icon would animate.
### The fix
Call `badgeCenterFromIconFrame` in the `UIView animation` block.

| Before | After | 
| - | - |
|before|after|

**_Code snippet for video_**
`static const NSTimeInterval kMDCBottomNavigationItemViewTransitionDuration = 2.180f;`


romoore:
Would you please add a screenshot or gif of the problem and solution?

romoore (approved):

PR #5717, opened 2018-11-08
WIP Global replace of single-precision floats with CGFloat casts.

romoore:
Global replace of single-precision floats with CGFloat casts.

Replacing all non-integral, single-precision floating point literals with double-precision values explicitly cast to CGFloat.  For example, `0.1f` should become `(CGFloat)0.1`.


.kokoro
```diff
@@ -49,6 +49,7 @@ fix_bazel_imports() {
   rewrite_tests "s/import MaterialComponents.Material(.+)/import components_\1_\1/"
   rewrite_source "s/import <Motion(.+)\/Motion.+\.h>/import \"Motion\1.h\"/"
   rewrite_source "s/import <MDF(.+)\/MDF.+\.h>/import \"MDF\1.h\"/"
+  rewrite_source "s/import <MaterialComponents\/Material(.+)\.h>/import\/\*framework import\*\/ \"Material\1.h\"/"
```
andrewoverton, line 52: Huh, this and [this](https://github.com/material-components/material-components-ios/pull/5717/files#diff-9b8844c11cf1465b9e2ab4e4a08e757cR67) were added in https://github.com/material-components/material-components-ios/pull/5541. Should they be here?

romoore: This was a WIP branch. I imported, modified, then exported develop. Apparently that flow isn't idempotent.

PR #5734, opened 2018-11-12
No longer using `-init` for Color Scheme.

romoore:
{MDC iOS} No longer using `-init` for Color Scheme.

Based on the discussions in go/mdc-ios-theming, we should not use the `-init` default initializer. Instead, we should use an explicit set of defaults.

Search regex: '\[\[MDCSemanticColorScheme alloc\] init\]'


rami-a (approved):

jverkoey (approved):

romoore:
Internal testing found no breakages caused by this change.

PR #5794, opened 2018-11-19
[Dialogs, Shape] Fix cross-component imports.

romoore:
Some of our examples and component code was depending on other components by
using relative path imports. Instead, they should be relying on the build
system to expose headers and importing those from the include paths.


romoore:
Re-application of https://github.com/material-components/material-components-ios/pull/5786

randallli (approved):

PR #5795, opened 2018-11-19
[Dialogs] These changes alter the way MDCAlertActions' titles are styled for b/119240836

andrewoverton:
This PR (which is imported from [cl/222102766](http://cl/222102766)) is part of a potential fix for [b/119240836](http://b/119240836). It changes the way MDCAlertActions' titles are styled to pave the way for a dependent CL, [cl/222102762](http://cl/222102762).  

~There are some unintended changes in this PR that are the result of a slightly broken export process 🙁The broken Swift import and the script deletions are not supposed to be there.~
_Edit_: I got rid of them


romoore (changes requested):
Please clean up the summary and description of this PR.

* If you need to reference internal links (b/, cl/), please also provide prose descriptions of their relevance 
* Clean up any strikeout text.

components/Dialogs/src/private/MDCAlertActionManager.m
```diff
@@ -89,6 +89,7 @@ - (MDCButton *)makeButtonForAction:(MDCAlertAction *)action
                             target:(id)target
                           selector:(SEL)selector {
   MDCButton *button = [[MDCButton alloc] initWithFrame:CGRectZero];
+  button.uppercaseTitle = NO;
```
romoore, line 92: I don't think we should make this change as the default. M1 styled dialogs still should use all-uppercase, and I believe it's still Material's "default" look for titles.  If we wanted to change Button default values, we should probably do it in Buttons (globally) rather than only in Dialogs.

components/Dialogs/src/private/MDCAlertController+Private.h
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <UIKit/UIKit.h>
+
+#import "MDCAlertController.h"
+
+@interface MDCAlertAction (Private)
```
romoore, line 19: Would you please add documentation for this category? What are the intended uses if it isn't required for the component and it's in `src/private`?

romoore (changes requested):

components/Dialogs/src/private/MDCAlertActionManager.m
andrewoverton: The M1 dialogs would still be all caps, because `-uppercaseString` would be called in the factory methods [here](https://github.com/material-components/material-components-ios/pull/5795/files#diff-7f12d761bb6d1e1c91da59ada3276844R51), but yeah, sounds like you all have a solution, so I'll close this

PR #5808, opened 2018-11-20
[Dialogs] Example: moved supplemental code into a single file:DialogsDismissingViewController 

galiak11:
Since the example was spread across many files, it made it difficult to keep track of where the code lived. Moving everything into the main VC header/implementation will simplify maintenance and future clean-up of the code.

Part of #5318


romoore:
Why are we moving everything to a single file? I'd prefer to keep one example = one file if possible. Ideally, even removing the "supplemental" files entirely.

galiak11:
Everything is not moved into a single file. The 2 implementation and 1 header files are consolidated into a single file. Whatever examples that were in these files remain as they are (no changes were made to the examples in these files).

randallli (approved):

PR #5821, opened 2018-11-22
[NavigationDrawer] Add the bottom drawer container view controller as a child view controller

guylivneh:
The bottom drawer container is missing from the view controller hierarchy. This causes issues with forwarding system calls such as UIViewController's dismissViewController:animated:completion:


yarneo (changes requested):
Hi Guy, thanks for contributing and opening a PR!

There are cases where the presentedViewController isn't MDCBottomDrawerViewController. If you notice the conditional above in line 148 `if ([self.presentedViewController isKindOfClass:[MDCBottomDrawerViewController class]]) {`, can you please put this line only in that conditional? Otherwise in cases where it isn't `MDCBottomDrawerViewController` the app can crash.

Also, please use the `self.presentedViewController` notation

Thanks!

guylivneh:
Fixed issue raised in comment by yarneo

yarneo (approved):

PR #5831, opened 2018-11-26
[Dialogs] Refactor low elevation example

galiak11:
Refactoring elevation example in Dragons.
Extending ShadowElevation instead of using an arbitrary constant.


components/Dialogs/examples/DialogsAlertCustomizationViewController.swift
```diff
@@ -270,6 +286,13 @@ class DialogsAlertCustomizationViewController: MDCCollectionViewController {
 
 }
 
+// app specific elevation constants
+extension ShadowElevation {
+  static var appDialog: ShadowElevation {
```
romoore: Please append a prefix unique to this example.  Maybe dialogsAlert_?

romoore (approved):

components/Dialogs/examples/DialogsAlertCustomizationViewController.swift
galiak11: Done

PR #5838, opened 2018-11-27
[Dialogs] Reduce & refactor Material Alert Controller Examples

galiak11:
Refactoring examples in DialogsAlertViewController.

* First commit removes no-longer-needed Material Alert examples from DialogsAlertViewController.m.
* Second commit consolidates the 2 implementation & 1 header files into a single implementation file (DialogsAlertViewController.m). 
* Third commit is for clang reformatting.

Issue: #5318.


romoore (approved):
Next time please separate the two changes into different PRs. Changing behavior (removing some Alerts) and refactoring should be independent.

PR #5861, opened 2018-11-28
[Snackbar] Disable potentially flaky test

rami-a:
### Context
The PR for [snapshot testing](https://github.com/material-components/material-components-ios/pull/5754) kept failing CI due to an unrelated test. 

### Solution
Disabling this test, fixes the problem

### Related bugs
[b/120145862](http://b/120145862)

### DIFF
--- a/components/Snackbar/tests/unit/MDCSnackbarMessageViewTests.m
+++ b/components/Snackbar/tests/unit/MDCSnackbarMessageViewTests.m
@@ -91,7 +91,8 @@
                         self.message.accessibilityLabel);
 }
 
-- (void)testAccessibilityHintDefaultIsNotNil {
+// TODO (b/120145862) Deflake and reenable this test
+- (void)_disabled_testAccessibilityHintDefaultIsNotNil {
   // When
   [self.manager showMessage:self.message];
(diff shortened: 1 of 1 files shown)


yarneo (approved):

components/Snackbar/tests/unit/MDCSnackbarMessageViewTests.m
```diff
@@ -91,7 +91,8 @@ - (void)testAccessibilityLabelSetFromSnackbarMessageProperty {
                         self.message.accessibilityLabel);
 }
 
-- (void)testAccessibilityHintDefaultIsNotNil {
+// TODO (b/120145862) Deflake and reenable this test
```
romoore, line 94: Tip: Next time please use a GitHub issue link instead of internal links.  I think it's easier for external contributors if we link to GitHub issues rather than internal links. Since they cannot see the `b/` content, it will be harder to find the actual work to be done later.

romoore (approved):

components/Snackbar/tests/unit/MDCSnackbarMessageViewTests.m
rami-a: Good point, will do that going forward

PR #5866, opened 2018-11-29
[ActionSheet] Rename example view controllers.

romoore:
Renaming each of the 3 example view controllers to include
"ExampleViewController" so they're easier to find. This is closer to our
typical naming pattern for example view controllers.


romoore:
Looks like clang-format failed but didn't post the results to GitHub.

rami-a (approved):

jverkoey (dismissed):
I think you're doing two related things in this set of changes:

1. Modifying the example view controllers to include `ExampleViewController` as a suffix.
2. Making the filename match the example view controller class's name.

Assuming the above is your intent, please match up the filenames with the class names in this PR and the others so that they are consistent. E.g. `ActionSheetComparisonViewController` should be `ActionSheetComparisonExampleViewController`.

romoore:
Hi Jeff. I think you might have misread the code. The examples previously had the Class and File names match - I'm not changing that behavior. I did have a bug with the VC you called-out, but I've corrected it in the latest commit.  If I misread your comments, would you please place a comment in the source where I'm changing the name-to-class mapping for one of the VCs?

romoore:
Ping @jverkoey

randallli:
I think romoore@ resolved the issue jverkoey@ found.

randallli (approved):

jverkoey (approved):

PR #5869, opened 2018-11-29
[Buttons] Floating Action Button Theming Extension

### LINKED ISSUE #5844: [Buttons] Internal issue: b/120086118
This was filed as an internal issue. If you are a Googler, please visit [b/120086118](http://b/120086118) for more details.

---
#### Internal data
- Associated internal bug: [b/120086118](http://b/120086118)

fluxminds:
Adds a new Theming Extension for MDCFloatingButton. Based on the design work in go/mdc-ios-theming-extensions, this themer will apply the Secondary Floating Action Button style to a FAB.  There should be no visible differences between this API and the current Themer APIs.

Closes #5844


romoore:
Please add to the Summary to describe this change.

components/BottomAppBar/examples/BottomAppBarTypicalUseExample.m
```diff
@@ -16,7 +16,7 @@
 
 #import "MaterialBottomAppBar+ColorThemer.h"
 #import "MaterialBottomAppBar.h"
-#import "MaterialButtons+ButtonThemer.h"
+#import "MaterialButtons+Theming.h"
```
romoore: Please update the examples in a follow-up PR.

components/Buttons/src/Theming/MDCFloatingButton+Theming.h
```diff
@@ -0,0 +1,29 @@
...
+
+#import "MaterialButtons.h"
+#import "MaterialContainerScheme.h"
+
+#import <Foundation/Foundation.h>
+
+@interface MDCFloatingButton (MaterialTheming)
+
+/**
+ Apply the secondary theme to this MDCFloatingButton instances.
```
romoore: Tip: Please add a link to the material.io article/section that describes the "secondary theme" for floating action buttons.

romoore: Nit: Please clean-up the comment a little bit.  Suggestion: "Applies the secondary Floating Action Button style to an MDCFloatingButton instance".

components/Buttons/src/Theming/MDCFloatingButton+Theming.m
```diff
@@ -0,0 +1,61 @@
...
+  [self applySecondaryThemeWithShapeScheme:shapeScheme];
+
+  id<MDCTypographyScheming> typographyScheme = scheme.typographyScheme;
+  if (!typographyScheme) {
+    typographyScheme =
+        [[MDCTypographyScheme alloc] initWithDefaults:MDCTypographySchemeDefaultsMaterial201804];
+  }
+  [self applySecondaryThemeWithTypographyScheme:typographyScheme];
+
+  [self setElevation:(CGFloat)6 forState:UIControlStateNormal];
```
romoore: Can MDCShadowElevations be used instead of literals?

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -84,4 +84,29 @@ class ButtonsThemingTest: XCTestCase {
...
+    // When
+    button.applySecondaryTheme(withScheme: scheme)
+    
+    // Then
+    XCTAssertEqual(button.backgroundColor(for: .normal), colorScheme.secondaryColor)
+    XCTAssertEqual(button.imageTintColor(for: .normal), colorScheme.onSecondaryColor)
+    
+    XCTAssertEqual(button.titleFont(for: .normal), typographyScheme.button)
+    
+    let buttonShape = button.shapeGenerator as! MDCRectangleShapeGenerator
```
romoore: Should this be a guard with an XCTAssertFail if it doesn't work?

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -84,4 +84,29 @@ class ButtonsThemingTest: XCTestCase {
     XCTAssertEqual(buttonShape.bottomRightCorner, shapeScheme.smallComponentShape.bottomRightCorner)
     XCTAssertEqual(buttonShape.bottomLeftCorner, shapeScheme.smallComponentShape.bottomLeftCorner)
   }
+  
+  func testFloatingButtonSecondaryTheme() {
```
romoore: This looks like a test with color + typography but without shape.  Please add the other 6 test cases if that's how we're handling things:

1. No custom themes
2. Custom color only
3. Custom typography only
4. Custom shape only
5. Custom color, typography (this test)
6. Custom color, shape
7. Custom typography, shape


However, I would strongly recommend testing only defaults (no customization), each scheme in isolation, and possibly all 3 schemes.

1. No custom schemes
2. Custom color only
3. Custom typography only
4. Custom shape only
5. [Optional] All 3 custom schemes

romoore (dismissed):
Please add the remaining tests and split the examples into separate PRs.

components/Buttons/src/Theming/MDCFloatingButton+Theming.h
fluxminds: Addressed.

components/Buttons/src/Theming/MDCFloatingButton+Theming.m
fluxminds: This is no corresponding MDCShadowElevation defined for this value, and from this PR's purpose, it is okay if we don't do the constant refactoring right now.

components/Buttons/src/Theming/MDCFloatingButton+Theming.h
fluxminds: Addressed. Thanks for the suggestion.

components/Buttons/src/Theming/MDCFloatingButton+Theming.m
romoore: Please use `MDCShadowElevationFABResting` as it is the correct elevation for the `resting` (`.normal`) state of the Floating Action Button (FAB).

components/Buttons/tests/unit/ButtonsThemingTest.swift
fluxminds: The shape testing was included, which followed the examples under shape testing. Will add the custom ones.

components/Buttons/src/Theming/MDCFloatingButton+Theming.m
fluxminds: Done.

components/Buttons/tests/unit/ButtonsThemingTest.swift
fluxminds: I addressed the custom color, typography cases. The custom shape case was not added, because the shape theme for FAB doesn't use things from shapeScheme, which mean there is no way to customize it. I can add a case to verify no values from shapeScheme is used if you would prefer. Originally there were no testing for that when the FAB themer was implemented.

components/Buttons/tests/unit/ButtonsThemingTest.swift
fluxminds: Addressed.

components/BottomAppBar/examples/BottomAppBarTypicalUseExample.m
fluxminds: Examples have been removed from this PR.

components/Buttons/tests/unit/ButtonsThemingTest.swift
romoore: I'm not sure how we could effectively test that the shape scheme "wasn't" used. Ideally, the shape scheme would apply the `cornerRadius` value for the buttons if that was the minimally-necessary condition for achieving the state. Although currently it might apply a masking layer to the component, testing for that would be very white-box testing of *another component*. I think it's best to leave it out for now.  Thanks for explaining!

components/Buttons/tests/unit/ButtonsThemingTest.swift
romoore: Let's please use an `if let` to ensure we have the correct type. I'd rather not have our tests crash.

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -84,4 +84,62 @@ class ButtonsThemingTest: XCTestCase {
...
+    let scheme = MDCContainerScheme()
+    let customColorScheme = MDCSemanticColorScheme()
+    customColorScheme.secondaryColor = .purple
+    customColorScheme.onSecondaryColor = .purple
+    scheme.colorScheme = customColorScheme
+
+    // When
+    button.applySecondaryTheme(withScheme: scheme)
+
+    // Then
```
romoore, line 277: Looks like this is missing the asserts for shape, typography, and the other miscellaneous properties.

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -84,4 +84,62 @@ class ButtonsThemingTest: XCTestCase {
...
+    let scheme = MDCContainerScheme()
+    let colorScheme = MDCSemanticColorScheme(defaults: .material201804)
+    let typographyScheme = MDCTypographyScheme(defaults: .material201804)
+
+    // When
+    button.applySecondaryTheme(withScheme: scheme)
+
+    // Then
+    XCTAssertEqual(button.backgroundColor(for: .normal), colorScheme.secondaryColor)
+    XCTAssertEqual(button.imageTintColor(for: .normal), colorScheme.onSecondaryColor)
```
romoore, line 232: Please add the remaining checks. Here's what I can think of quickly:
* inkColor
* titleColor
* shadowColor
* elevation
* borderWidth (if it's zero, borderColor has no effect)

You may want to write the assertion code once in a helper method and then call that method in each of these tests.   Otherwise, you could just copy-paste the identical asserts for all these tests, which seems to be the typography, color, and shape.

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -84,4 +84,62 @@ class ButtonsThemingTest: XCTestCase {
...
+    XCTAssertEqual(buttonShape.bottomLeftCorner, corner)
+  }
+
+  func testFloatingButtonSecondaryThemeWithCustomColorTheme() {
+    // Given
+    let button = MDCFloatingButton()
+    let scheme = MDCContainerScheme()
+    let customColorScheme = MDCSemanticColorScheme()
+    customColorScheme.secondaryColor = .purple
+    customColorScheme.onSecondaryColor = .purple
```
romoore, line 270: Please make these colors distinct so `secondaryColor` and `onSecondaryColor` can be distinguished.  Please also customize the rest of the scheme colors.  If it's easier, you can do something like:

```
customColorScheme.primaryColor = [UIColor colorWithWhite:0 alpha:(CGFloat)0.05];
customColorScheme.secondary = [UIColor colorWithWhite:0 alpha:(CGFloat)0.1];
customColorScheme.surfaceColor = [UIColor colorWithWhite:0 alpha:(CGFloat)0.15];
customColorScheme.backgroundColor = [UIColor colorWithWhite:0 alpha:(CGFloat)0.2];
...
```

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -84,4 +84,62 @@ class ButtonsThemingTest: XCTestCase {
...
+    let scheme = MDCContainerScheme()
+    let customTypographyScheme = MDCTypographyScheme()
+    customTypographyScheme.button = UIFont.systemFont(ofSize: 19)
+    scheme.typographyScheme = customTypographyScheme
+
+    // When
+    button.applySecondaryTheme(withScheme: scheme)
+
+    // Then
+    XCTAssertEqual(button.titleFont(for: .normal), customTypographyScheme.button)
```
romoore, line 332: This is also missing the other asserts.

romoore (dismissed):

components/Buttons/src/Theming/MDCFloatingButton+Theming.h
```diff
@@ -0,0 +1,32 @@
...
+#import <Foundation/Foundation.h>
+
+@interface MDCFloatingButton (MaterialTheming)
+
+/**
+ Applies the secondary Floating Action Button style to an MDCFloatingButton instance.
+ @param scheme A container scheme instance containing any desired customizations to the theming
+ system.
+
+ Learn more at the [Material
```
romoore, line 27: Please use "Material Guidelines" rather than "Material spec."

Suggestion: "The [Material Guidelines article for Floating Action Buttons](https://material.io/design/components/buttons-floating-action-button.html#theming) has more details about how Floating Action Buttons can be styled and used.

components/Buttons/tests/unit/ButtonsThemingTest.swift
fluxminds: Thanks for creating another PR to make it clear.

components/Buttons/src/Theming/MDCFloatingButton+Theming.h
fluxminds: Word `Guidelines` is used. The format has been kept to keep consistent with other places in the codebase.

components/Buttons/tests/unit/ButtonsThemingTest.swift
fluxminds: Those properties are not changed by the themer. Do we want to check all properties in this test?

components/Buttons/tests/unit/ButtonsThemingTest.swift
fluxminds: The rest of the scheme colors are not used by the themer. Do you think if it is necessary to customize and use NotEqual in this test case?

components/Buttons/tests/unit/ButtonsThemingTest.swift
fluxminds: Same as the above.

components/Buttons/tests/unit/ButtonsThemingTest.swift
fluxminds: Addressed. Assertions for the other subsystems were added:)

components/Buttons/tests/unit/ButtonsThemingTest.swift
fluxminds: Sure. Addressed in 41da40aaf53833e2d8b33b8c226c9fe30d462460.

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -162,34 +162,34 @@ class ButtonsThemingTest: XCTestCase {
...
     // When
     button.applyTextTheme(withScheme: scheme)
-
+    
     // Then
     XCTAssertEqual(button.titleColor(for: .normal), customColor)
     XCTAssertEqual(button.imageTintColor(for: .normal), customColor)
   }
-
+
```
romoore: Oh no! White space!

romoore:
@wenyuzhang666 Could you please add a bit more detail to this PR summary? That way when it's committed to git readers will have some context about why/how the change is being made.

components/Buttons/tests/unit/ButtonsThemingTest.swift
fluxminds: Swift file!
whitespaces removed in https://github.com/material-components/material-components-ios/pull/5869/commits/188c094c4a58a41e9364b6f2699b8c48fde8ad77

components/Buttons/tests/unit/ButtonsThemingTest.swift
fluxminds: Addressed in 3ea318570c68abea2a1fce745312dbf57da36679

components/Buttons/tests/unit/ButtonsThemingTest.swift
fluxminds: Addressed in 3ea318570c68abea2a1fce745312dbf57da36679

romoore:
@wenyuzhang666 I updated the PR summary with what I was thinking of including. Please take a look.

components/Buttons/src/Theming/MDCFloatingButton+Theming.h
romoore: Ok. I filed #5914 to fix this in a follow-up PR.

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -216,5 +216,134 @@ class ButtonsThemingTest: XCTestCase {
...
+    } else {
+      XCTFail("Button.shapeGenerator was not a MDCRectangularShapeGenerator")
+    }
+
+    // Test remaining properties
+    XCTAssertEqual(button.elevation(for: .normal), ShadowElevation.fabResting)
+    XCTAssertEqual(button.elevation(for: .highlighted), ShadowElevation.fabPressed)
+    XCTAssertEqual(button.elevation(for: .disabled), ShadowElevation.none)
+
+    XCTAssertEqual(button.borderWidth(for: .normal), defaultButton.borderWidth(for: .normal), accuracy: 0.001)
```
romoore: In this case, we want to be sure that a FAB has no border. Please assert that the border width is (nearly) 0.

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -216,5 +216,134 @@ class ButtonsThemingTest: XCTestCase {
...
+    let colorScheme = MDCSemanticColorScheme(defaults: .material201804)
+    let typographyScheme = MDCTypographyScheme(defaults: .material201804)
+
+    // When
+    button.applySecondaryTheme(withScheme: scheme)
+
+    // Then
+    XCTAssertEqual(button.backgroundColor(for: .normal), colorScheme.secondaryColor)
+    XCTAssertEqual(button.imageTintColor(for: .normal), colorScheme.onSecondaryColor)
+    // XCTAssertEqual(button.titleColor(for: .normal), colorScheme.onSecondaryColor)
```
romoore, line 235: Please add a TODO for this line so we can track the work to fix it.

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -216,5 +216,134 @@ class ButtonsThemingTest: XCTestCase {
...
+    scheme.colorScheme = customColorScheme
+    let typographyScheme = MDCTypographyScheme(defaults: .material201804)
+
+    // When
+    button.applySecondaryTheme(withScheme: scheme)
+
+    // Then
+    XCTAssertEqual(button.backgroundColor(for: .normal), customColorScheme.secondaryColor)
+    XCTAssertEqual(button.imageTintColor(for: .normal), customColorScheme.onSecondaryColor)
+    // XCTAssertEqual(button.titleColor(for: .normal), colorScheme.onSecondaryColor)
```
romoore, line 282: Please add the TODO here and below as well.

romoore (approved):

components/Buttons/tests/unit/ButtonsThemingTest.swift
fluxminds: Yes. I am going to mention your fix in the TODO to make it explicit.

fluxminds: Addressed in 1b3abf543a18d093735c7636cbe962a3831b11dd

components/Buttons/tests/unit/ButtonsThemingTest.swift
fluxminds: Addressed.

components/Buttons/tests/unit/ButtonsThemingTest.swift
fluxminds: Addressed in1b3abf543a18d093735c7636cbe962a3831b11dd

PR #5877, opened 2018-11-29
Add git-lfs instructions to tools documentation

rami-a:
Adding documentation describing installing git-lfs

### DIFF
--- a/contributing/tools.md
+++ b/contributing/tools.md
@@ -50,3 +50,11 @@ git clang-format -h
 https://github.com/material-foundation/github-squash-and-merge-pr-descriptions.
 
 This tool ensures that your pull request's merge commit matches your PR description.
+
+## Using git-lfs
+
+Snapshot tests utilize git-lfs to store the reference images. Please install and configure git-lfs in order to run and write snapshot tests. Perform the following steps:
+
+1. `brew install git-lfs`
+2. `git lfs install`
+3. `git lfs pull`


contributing/tools.md
```diff
@@ -50,3 +50,11 @@ git clang-format -h
...
 
 This tool ensures that your pull request's merge commit matches your PR description.
+
+## Using git-lfs
+
+Snapshot tests utilize git-lfs to store the reference images. Please install and configure git-lfs in order to run and write snapshot tests. Perform the following steps:
+
+1. `brew install git-lfs`
+2. `git lfs install`
+3. `git lfs pull`
```
codeman7, line 60: Do we need `git lfs pull`?

codeman7 (approved):

contributing/tools.md
rami-a: Yes, unless you install git-lfs before you clone the repo, the `pull` makes sure you have the actual image files rather than the pointer files

galiak11:
Is git-lfs a requirement for passing tests?  Should we also add a reference to this file in the releasing.md file?  or perhaps testing.md?  My concern is that "tools.md" is not the first place people will look for when tests start failing.

rami-a:
So the way I have set it up in #5887 is that the snapshot tests will only run for a specific iOS version (11.2 in this case), so people will not be running them locally unless they happen to be using that version. This was done so that we can have CI run the snapshot tests appropriately. 

So for most people git-lfs would not be a requirement for passing tests locally. Eventually this would change (#5888) or if the developer needs to make changes to the reference images. We could update these docs at a later time in that case.

galiak11 (approved):

PR #5885, opened 2018-11-30
[Buttons] Add text button themer

### LINKED ISSUE #5848: [Buttons] Internal issue: b/120083977
This was filed as an internal issue. If you are a Googler, please visit [b/120083977](http://b/120083977) for more details.

---
#### Internal data
- Associated internal bug: [b/120083977](http://b/120083977)

galiak11:
Adding a text themer to material themer category for Buttons. 
Closes #5848

### DIFF
--- a/components/Buttons/src/Theming/MDCButton+Theming.h
+++ b/components/Buttons/src/Theming/MDCButton+Theming.h
@@ -37,4 +37,11 @@
  */
 - (void)applyOutlinedThemeWithScheme:(nonnull id<MDCContainerScheming>)scheme;
 
+/**
+ Applies the Material Text Button style to an MDCButton instance.
+
+ @param scheme A container of material schemes used in customizing the appearance of the button.
+ */
+- (void)applyTextThemeWithScheme:(nonnull id<MDCContainerScheming>)scheme;
+
 @end
--- a/components/Buttons/src/Theming/MDCButton+Theming.m
+++ b/components/Buttons/src/Theming/MDCButton+Theming.m
@@ -21,6 +21,8 @@
 
 @implementation MDCButton (MaterialTheming)
 
+#pragma mark - Contained Button Themers
+
 - (void)applyContainedThemeWithScheme:(nonnull id<MDCContainerScheming>)scheme {
   id<MDCColorScheming> colorScheme = scheme.colorScheme;
   if (!colorScheme) {
@@ -61,6 +63,8 @@
   [MDCButtonShapeThemer applyShapeScheme:shapeScheme toButton:self];
 }
 
+#pragma mark - Outlined Button Themers
+
 - (void)applyOutlinedThemeWithScheme:(nonnull id<MDCContainerScheming>)scheme {
   id<MDCColorScheming> colorScheme = scheme.colorScheme;
   if (!colorScheme) {
@@ -100,6 +104,50 @@
 }
 
 - (void)applyOutlinedThemeWithShapeScheme:(id<MDCShapeScheming>)shapeScheme {
+  [MDCButtonShapeThemer applyShapeScheme:shapeScheme toButton:self];
+}
+
+#pragma mark - Text Button Themers
+
+- (void)applyTextThemeWithScheme:(nonnull id<MDCContainerScheming>)scheme {
+  id<MDCColorScheming> colorScheme = scheme.colorScheme;
+  if (!colorScheme) {
+    colorScheme =
+        [[MDCSemanticColorScheme alloc] initWithDefaults:MDCColorSchemeDefaultsMaterial201804];
+  }
+  [self applyTextThemeWithColorScheme:colorScheme];
+
+  id<MDCTypographyScheming> typographyScheme = scheme.typographyScheme;
+  if (!typographyScheme) {
+    typographyScheme =
+        [[MDCTypographyScheme alloc] initWithDefaults:MDCTypographySchemeDefaultsMaterial201804];
+  }
+  [self applyTextThemeWithTypographyScheme:typographyScheme];
+
+  id<MDCShapeScheming> shapeScheme = scheme.shapeScheme;
+  if (shapeScheme) {
+    [self applyTextThemeWithShapeScheme:shapeScheme];
+  } else {
+    self.layer.cornerRadius = (CGFloat)4.0;
+  }
+
+  NSUInteger maximumStateValue = UIControlStateNormal | UIControlStateSelected |
+                                 UIControlStateHighlighted | UIControlStateDisabled;
+  for (NSUInteger state = 0; state <= maximumStateValue; ++state) {
+    [self setElevation:MDCShadowElevationNone forState:state];
+  }
+  self.minimumSize = CGSizeMake(0, 36);
+}
+
+- (void)applyTextThemeWithColorScheme:(id<MDCColorScheming>)colorScheme {
+  [MDCTextButtonColorThemer applySemanticColorScheme:colorScheme toButton:self];
+}
+
+- (void)applyTextThemeWithTypographyScheme:(id<MDCTypographyScheming>)typographyScheme {
+  [MDCButtonTypographyThemer applyTypographyScheme:typographyScheme toButton:self];
+}
+
+- (void)applyTextThemeWithShapeScheme:(id<MDCShapeScheming>)shapeScheme {
   [MDCButtonShapeThemer applyShapeScheme:shapeScheme toButton:self];
 }
 
--- a/components/Buttons/tests/unit/ButtonsThemingTest.swift
+++ b/components/Buttons/tests/unit/ButtonsThemingTest.swift
@@ -135,5 +135,79 @@ class ButtonsThemingTest: XCTestCase {
     XCTAssertEqual(buttonShape.bottomLeftCorner, shapeScheme.smallComponentShape.bottomLeftCorner)
   }
 
+  func testTextThemeWithDefaultScheme() {
+    // Given
+    let button = MDCButton()
+    let scheme = MDCContainerScheme()
+
+    // When
+    button.applyTextTheme(withScheme: scheme)
+
+    // Then
+    helperAssertTextTheme(button: button)
+  }
+
+  func testTextThemeWithColorScheme() {
+    // Given
+    let button = MDCButton()
+    let scheme = MDCContainerScheme()
+    let colorScheme = MDCSemanticColorScheme(defaults: .material201804)
+    let customColor = UIColor.orange
+    colorScheme.primaryColor = customColor
+    scheme.colorScheme = colorScheme
+
+    // When
+    button.applyTextTheme(withScheme: scheme)
+
+    // Then
+    XCTAssertEqual(button.titleColor(for: .normal), customColor)
+    XCTAssertEqual(button.imageTintColor(for: .normal), customColor)
+  }
+
+  func testConvertContainedToTextTheme() {
+    // Given
+    let button = MDCButton()
+    let scheme = MDCContainerScheme()
+
+    // When
+    button.applyContainedTheme(withScheme: scheme)
+    button.applyTextTheme(withScheme: scheme)
+
+    // Then
+    helperAssertTextTheme(button: button)
+  }
+
+  // MARK: Helpers
+
+  func helperAssertTextTheme(button: MDCButton) {
+    let colorScheme = MDCSemanticColorScheme(defaults: .material201804)
+    let typographyScheme = MDCTypographyScheme(defaults: .material201804)
+
+    // Test colors
+    XCTAssertEqual(button.backgroundColor(for: .normal), .clear)
+    XCTAssertEqual(button.borderColor(for: .normal), nil)
+    XCTAssertEqual(button.inkColor, colorScheme.primaryColor.withAlphaComponent(0.16))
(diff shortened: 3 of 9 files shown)


components/Buttons/src/Theming/MDCButton+Theming.h
```diff
@@ -37,4 +37,11 @@
  */
 - (void)applyOutlinedThemeWithScheme:(nonnull id<MDCContainerScheming>)scheme;
 
+/**
+ Applies a Material text theme to the button.
```
romoore: Nit/suggestion: "Applies the Material Text Button style to an MDCButton instance."

components/Buttons/src/Theming/MDCButton+Theming.m
```diff
@@ -21,6 +21,8 @@
 
 @implementation MDCButton (MaterialTheming)
 
+// MARK: Contained Button Themers
```
romoore: Nit/consistency: Please use #pragma mark - Button Themers.  It's the general approach to code organization we've taken within MDC-iOS.

components/Buttons/src/Theming/MDCButton+Theming.m
```diff
@@ -103,4 +107,50 @@ - (void)applyOutlinedThemeWithShapeScheme:(id<MDCShapeScheming>)shapeScheme {
...
+  } else {
+    self.layer.cornerRadius = (CGFloat)4.0;
+  }
+
+  NSUInteger maximumStateValue = UIControlStateNormal | UIControlStateSelected |
+                                 UIControlStateHighlighted | UIControlStateDisabled;
+  for (NSUInteger state = 0; state <= maximumStateValue; ++state) {
+    [self setElevation:MDCShadowElevationNone forState:state];
+  }
+  self.minimumSize = CGSizeMake(0, 36.0);
```
romoore: Please use either `36` to be more consistent with the code base. If there is a reason not to, please cast to `CGFloat`.

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -135,5 +135,54 @@ class ButtonsThemingTest: XCTestCase {
...
+    // Test shape
+    XCTAssertEqual(button.layer.cornerRadius, 4.0, accuracy: 0.001)
+    // Test remaining properties
+    [.normal, .highlighted, .selected, .disabled].forEach {
+      XCTAssertEqual(button.elevation(for: $0), ShadowElevation.none)
+    }
+    XCTAssertEqual(button.minimumSize.height, 36.0, accuracy: 0.001)
+  }
+
+  func testTextThemeWithCustomScheme() {
```
romoore: This looks like it's only a custom color scheme. Please add tests for:

* Custom Typography scheme
* Custom Shape scheme

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -135,5 +135,54 @@ class ButtonsThemingTest: XCTestCase {
...
+    let colorScheme = MDCSemanticColorScheme(defaults: .material201804)
+    let typographyScheme = MDCTypographyScheme(defaults: .material201804)
+
+    // When
+    button.applyTextTheme(withScheme: scheme)
 
+    // Then
+    // Test colors
+    XCTAssertEqual(button.backgroundColor, .clear)
+    XCTAssertEqual(button.layer.borderColor, nil)
```
romoore: I think we should be checking `borderColorForState:` instead of the implementation of how it's applied. What do you think?

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -135,5 +135,54 @@ class ButtonsThemingTest: XCTestCase {
...
+    let scheme = MDCContainerScheme()
+    let colorScheme = MDCSemanticColorScheme(defaults: .material201804)
+    let typographyScheme = MDCTypographyScheme(defaults: .material201804)
+
+    // When
+    button.applyTextTheme(withScheme: scheme)
 
+    // Then
+    // Test colors
+    XCTAssertEqual(button.backgroundColor, .clear)
```
romoore: Please check the backgroundColorForState:

components/Buttons/src/Theming/MDCButton+Theming.m
```diff
@@ -103,4 +107,50 @@ - (void)applyOutlinedThemeWithShapeScheme:(id<MDCShapeScheming>)shapeScheme {
...
+}
+
+- (void)applyTextThemeWithTypographyScheme:(id<MDCTypographyScheming>)typographyScheme {
+  [MDCButtonTypographyThemer applyTypographyScheme:typographyScheme toButton:self];
+}
+
+- (void)applyTextThemeWithShapeScheme:(id<MDCShapeScheming>)shapeScheme {
+  [MDCButtonShapeThemer applyShapeScheme:shapeScheme toButton:self];
+}
+
```
romoore, line 153: Nit: Please remove the extra newlines.

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -135,5 +135,54 @@ class ButtonsThemingTest: XCTestCase {
...
+    XCTAssertEqual(button.inkColor, colorScheme.primaryColor.withAlphaComponent(0.16))
+    XCTAssertEqual(button.titleColor(for: [.normal, .highlighted]), colorScheme.primaryColor)
+    XCTAssertEqual(button.titleColor(for: .disabled),
+                   colorScheme.onSurfaceColor.withAlphaComponent(0.38))
+    [.normal, .highlighted].forEach {
+      XCTAssertEqual(button.imageTintColor(for: $0), colorScheme.primaryColor)
+    }
+    XCTAssertEqual(button.imageTintColor(for: .disabled),
+                   colorScheme.onSurfaceColor.withAlphaComponent(0.38))
+    // Test typography
```
romoore, line 199: Tip: Add blank lines between the "blocks" of assertions to make it easier to read.

romoore (changes requested):
Please break-out the examples into separate PRs. The unit tests seem to be missing some behavior.

components/Buttons/src/Theming/MDCButton+Theming.m
galiak11: Not sure which lines this refers to.

components/Buttons/tests/unit/ButtonsThemingTest.swift
galiak11: In general, leaving blank lines in between given/then/when blocks makes the tests more readable.

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -135,5 +135,54 @@ class ButtonsThemingTest: XCTestCase {
...
+    // Test shape
+    XCTAssertEqual(button.layer.cornerRadius, 4.0, accuracy: 0.001)
+    // Test remaining properties
+    [.normal, .highlighted, .selected, .disabled].forEach {
+      XCTAssertEqual(button.elevation(for: $0), ShadowElevation.none)
+    }
+    XCTAssertEqual(button.minimumSize.height, 36.0, accuracy: 0.001)
+  }
+
+  func testTextThemeWithColorScheme() {
```
galiak11: I renamed the test to reflect the content - similar to "testOutlinedThemeWithShapeScheme".

components/Buttons/src/Theming/MDCButton+Theming.m
romoore: There were several empty lines before the `@end`.

components/Buttons/tests/unit/ButtonsThemingTest.swift
romoore: Please add the remaining customized-scheme tests mentioned in my previous feedback.

* Custom Typography scheme
* Custom Shape scheme

components/Buttons/tests/unit/ButtonsThemingTest.swift
romoore: Agreed about Given/When/Then sections, but my comment was about grouping the tests within the "Given" section.

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -135,5 +135,54 @@ class ButtonsThemingTest: XCTestCase {
...
+    // Given
+    let button = MDCButton()
+    let scheme = MDCContainerScheme()
+    let colorScheme = MDCSemanticColorScheme(defaults: .material201804)
+    let customColor = UIColor.orange
```
romoore: Tip: Nit: `customColor` isn't required. You can just assign `UIColor.orange` to the scheme and then verify against `colorScheme.primaryColor` in the asserts.

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -135,5 +135,54 @@ class ButtonsThemingTest: XCTestCase {
...
+    XCTAssertEqual(button.titleFont(for: .normal), typographyScheme.button)
+    // Test shape
+    XCTAssertEqual(button.layer.cornerRadius, 4.0, accuracy: 0.001)
+    // Test remaining properties
+    [.normal, .highlighted, .selected, .disabled].forEach {
+      XCTAssertEqual(button.elevation(for: $0), ShadowElevation.none)
+    }
+    XCTAssertEqual(button.minimumSize.height, 36.0, accuracy: 0.001)
+  }
+
```
romoore, line 211: Tip: Consider adding a test that styles a Button as another type (like Contained) and then to Text and make sure the styling is correct.

components/Buttons/tests/unit/ButtonsThemingTest.swift
```diff
@@ -135,5 +135,54 @@ class ButtonsThemingTest: XCTestCase {
...
+    let colorScheme = MDCSemanticColorScheme(defaults: .material201804)
+    let customColor = UIColor.orange
+    colorScheme.primaryColor = customColor
+    scheme.colorScheme = colorScheme
+
+    // When
+    button.applyTextTheme(withScheme: scheme)
 
+    // Then
+    XCTAssertEqual(button.titleColor(for: .normal), customColor)
```
romoore: This seems to be missing the rest of the assertions. Consider extracting them into a helper method that can be reused across tests.

romoore (changes requested):
Please split the examples changes into a separate PR.

components/Buttons/tests/unit/ButtonsThemingTest.swift
galiak11: Unless I want to test that the title color is orange, not merely if the primary color has been applied to it.

components/Buttons/tests/unit/ButtonsThemingTest.swift
galiak11: I like this idea. I will add it as a separate PR.

galiak11: Great suggestion. Will add to a followup PR.

components/Buttons/tests/unit/ButtonsThemingTest.swift
galiak11: This test is be replaced with the re-styling test - see followup PR.

components/Buttons/tests/unit/ButtonsThemingTest.swift
galiak11: This test will be replaced - see followup PR.

romoore (dismissed):
Please split the examples into separate PRs.

components/Buttons/tests/unit/ButtonsThemingTest.swift
romoore: Isn't the test *that the primaryColor was assigned to it* from the customized scheme?  This was part of why I suggested refactoring the asserts into a helper method - they can be reused if the helper method is only comparing the values against the schemes (and any expected hard-coded values no captured in the schemes).

components/Buttons/tests/unit/ButtonsThemingTest.swift
romoore: Please include tests for new behavior in the same PR as the new behavior.

galiak11:
This PR is the same as the PRs for outline or contained buttons, so I fail to understand why examples here are to be broken into separate PRs but not for the outline/contained PRs. Can you please explain?

romoore:
@galiak11 I didn't review those, so I'm not sure why the reviewers might or might not have provided specific pieces of feedback. If this PR will be more thoroughly tested or narrower in scope than those, then that sounds like an improvement over what happened in the past.

andrewoverton:
Personally, I'd rather see this PR get merged in than see it split up as well, because it's blocking the dialog theming PR I want to issue

romoore:
@andrewoverton Agreed. I think it could have been merged already if the examples had been dropped after the first review instead of a lot of back-and-forth.  It feels like more effort has been spent trying to combine two different PRs (component + examples) into one than it would have taken to split them apart.

jverkoey (approved):
LGTM. Let’s get this merged as-is. We’ll discuss what happened in this PR’s feedback separately tomorrow.

PR #5886, opened 2018-11-30
[BottomNavigation] Initial implementation of the Bottom Navigation Controller

### LINKED ISSUE #4160: [BottomNavigation] MDCBottomNavigation should provide view controller containment

nokeeo:
Initial implementation of MDCBottomNavigationBarController. 

Closes #4160


components/BottomNavigation/examples/MDCBottomNavigationBarController.h
```diff
@@ -0,0 +1,99 @@
...
+ * in the navigation bar is selected.
+ * Available iOS 9.0+
+ */
+NS_CLASS_AVAILABLE_IOS(9_0)
+@interface MDCBottomNavigationBarController : UIViewController <MDCBottomNavigationBarDelegate>
+
+/**
+ * The bottom navigation bar that hosts the tab bar items.
+ * @warning This controller sets itself as the navigation bar's delegate.  If you would like to
+ * observe changes to the navigation bar, conform to \c MDCBottomNavigationBarControllerDelegate
```
romoore: General nit: We use `@c` instead of `\c` syntax in our header documentation.

components/BottomNavigation/examples/MDCBottomNavigationBarController.h
```diff
@@ -0,0 +1,99 @@
...
+ * If no tab item is selected it will be set to NSNotFound.
+ */
+@property(nonatomic) NSUInteger selectedIndex;
+
+- (void)viewDidLoad NS_REQUIRES_SUPER;
+
+#pragma mark - MDCBottomNavigationBarDelegate
+
+- (void)bottomNavigationBar:(nonnull MDCBottomNavigationBar *)bottomNavigationBar
+              didSelectItem:(nonnull UITabBarItem *)item NS_REQUIRES_SUPER;
```
romoore, line 70: How does `NS_REQUIRES_SUPER` work here? I don't understand how that would work if it's a delegate protocol method. Is there some superclass implementation to call out to?

components/BottomNavigation/examples/BottomNavigationControllerExampleViewController.swift
```diff
@@ -0,0 +1,43 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+
+@available(iOSApplicationExtension 9.0, *)
```
romoore: I think it's better to break out the example into a separate PR.

romoore:
Taking a look now - let's go over it in detail at our meeting today.

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
```diff
@@ -0,0 +1,265 @@
...
+
+- (void)setViewControllers:(NSArray<UIViewController *> *)viewControllers {
+  [self deselectCurrentItem];
+  NSMutableArray<UITabBarItem *> *tabBarItems = [NSMutableArray array];
+  for (UIViewController *viewController in viewControllers) {
+    [tabBarItems addObject:viewController.tabBarItem];
+  }
+
+  _navigationBar.items = tabBarItems;
+  _viewControllers = viewControllers;
```
romoore: Please make this an immutable copy so callers cannot modify the array underneath the VC.

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
```diff
@@ -0,0 +1,265 @@
...
+      selectedViewController == nil || [_viewControllers containsObject:selectedViewController],
+      @"Attempting to set BottomBarViewControllers to a view controller it does not contain");
+
+  // Early return if we are already set to the given VC
+  if (_selectedViewController == selectedViewController) {
+    return;
+  }
+
+  // Remove current VC and add new one.
+  [self removeContentViewController:_selectedViewController];
```
romoore: Nit: Please use property accessor: `[self removeContentViewController:self.selectedViewController];`

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
```diff
@@ -0,0 +1,265 @@
...
+  [self addContentViewController:selectedViewController];
+
+  // Set the iVar and update selected index
+  _selectedViewController = selectedViewController;
+  self.selectedIndex = [_viewControllers indexOfObject:selectedViewController];
```
romoore: Nit: Please use `self.viewControllers` where possible.

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
```diff
@@ -0,0 +1,265 @@
...
+  // Update the selected index value and views.
+  _selectedIndex = selectedIndex;
+  [self updateViewsForSelectedIndex:selectedIndex];
+}
+
+- (void)setViewControllers:(NSArray<UIViewController *> *)viewControllers {
+  [self deselectCurrentItem];
+  NSMutableArray<UITabBarItem *> *tabBarItems = [NSMutableArray array];
+  for (UIViewController *viewController in viewControllers) {
+    [tabBarItems addObject:viewController.tabBarItem];
```
romoore: Let's add a `nil` check here before adding it to the array?

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
```diff
@@ -0,0 +1,265 @@
...
+  return _selectedViewController;
+}
+
+#pragma mark - MDCBottomNavigationBarDelegate
+
+- (void)bottomNavigationBar:(MDCBottomNavigationBar *)bottomNavigationBar
+              didSelectItem:(UITabBarItem *)item {
+  // Early return if we cannot find the view controller.
+  NSUInteger index = [_navigationBar.items indexOfObject:item];
+  if (index >= [self.viewControllers count]) {
```
romoore: Let's also add an explicit check for `NSNotFound`?

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
```diff
@@ -0,0 +1,265 @@
...
+  if ([_delegate respondsToSelector:@selector(bottomNavigationBarController:
+                                                    didSelectViewController:)]) {
+    [_delegate bottomNavigationBarController:self didSelectViewController:selectedViewController];
+  }
+}
+
+- (BOOL)bottomNavigationBar:(MDCBottomNavigationBar *)bottomNavigationBar
+           shouldSelectItem:(UITabBarItem *)item {
+  NSUInteger index = [_navigationBar.items indexOfObject:item];
+  if (index >= [self.viewControllers count]) {
```
romoore: Please also check for `NSNotFound`.

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
```diff
@@ -0,0 +1,265 @@
...
+  [self.view.rightAnchor constraintEqualToAnchor:_navigationBar.rightAnchor].active = YES;
+
+  _navigationBarHeightConstraint =
+      [_navigationBar.heightAnchor constraintEqualToConstant:[self calculateNavigationBarHeight]];
+  _navigationBarHeightConstraint.active = YES;
+
+  [_navigationBar.topAnchor constraintEqualToAnchor:_content.bottomAnchor].active = YES;
+  [_navigationBar.bottomAnchor constraintEqualToAnchor:self.view.bottomAnchor].active = YES;
+
+  [_navigationBar setContentHuggingPriority:UILayoutPriorityDefaultHigh
```
romoore: Is this required given the left/right constraints?

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
```diff
@@ -0,0 +1,265 @@
...
+  view.translatesAutoresizingMaskIntoConstraints = NO;
+  [view.leadingAnchor constraintEqualToAnchor:_content.leadingAnchor].active = YES;
+  [view.trailingAnchor constraintEqualToAnchor:_content.trailingAnchor].active = YES;
+  [view.topAnchor constraintEqualToAnchor:_content.topAnchor].active = YES;
+  [view.bottomAnchor constraintEqualToAnchor:_content.bottomAnchor].active = YES;
+}
+
+/** Returns the desired height of the navigation bar. **/
+- (CGFloat)calculateNavigationBarHeight {
+  return [_navigationBar sizeThatFits:self.view.bounds.size].height;
```
romoore: Please calculate the `CGSize` with `CGRectGetHeight` and `CGRectGetWidth` instead of accessing the bounds directly.  It's possible for the bounds to be non-standardized with a negative size.

romoore (approved):
LGTM with some minor comments to be addressed.

romoore:
+ @material-components/mdc-ios in case anyone else is interested in this work.

components/BottomNavigation/examples/MDCBottomNavigationBarController.h
nokeeo: Done.

components/BottomNavigation/examples/BottomNavigationControllerExampleViewController.swift
nokeeo: Done.

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
nokeeo: Done.

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
nokeeo: Done.

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
nokeeo: Done.

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
nokeeo: Done.

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
nokeeo: Done.

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
nokeeo: Done.

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
nokeeo: Correct.  Removed.

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
nokeeo: Done.

components/BottomNavigation/examples/MDCBottomNavigationBarController.h
nokeeo: The implementation of this method is the core functionality of this controller.  If a subclass overrides this method and fails to call super, this controller would not be functional.  Adding ```NS_REQUIRES_SUPER``` to this method's signature ensures that the subclass calls the superclass's implementation if they override this method.

nokeeo:
Tests will be added this week.

components/BottomNavigation/examples/MDCBottomNavigationBarController.h
romoore: Ah, got it. I was reading it wrong. I understood it to mean that the **implementation** of **THIS CLASS** had to call super. Not subclasses. Makes total sense now!

components/BottomNavigation/examples/MDCBottomNavigationBarController.h
```diff
@@ -0,0 +1,99 @@
...
+ * The bottom navigation bar that hosts the tab bar items.
+ * @warning This controller sets itself as the navigation bar's delegate.  If you would like to
+ * observe changes to the navigation bar, conform to @c MDCBottomNavigationBarControllerDelegate
+ * and set the delegate property of this controller.
+ */
+@property(nonatomic, strong, readonly, nonnull) MDCBottomNavigationBar *navigationBar;
+
+/**
+ * An array of view controllers to display when their corresponding tab bar item is selected in the
+ * navigation bar.  When this property is set, the navigation bar's \c items property will be set to
```
romoore: Please do a find/replace of `\c` to `@c`

romoore (approved):

components/BottomNavigation/examples/BottomNavigationController/MDCBottomNavigationBarController.h
```diff
@@ -0,0 +1,97 @@
+// Copyright 2016-present the Material Components for iOS authors. All Rights Reserved.
```
romoore: Looks like this file was added in error.

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
```diff
@@ -0,0 +1,278 @@
...
+    _selectedIndex = NSNotFound;
+  }
+
+  return self;
+}
+
+- (void)viewDidLoad {
+  [super viewDidLoad];
+
+  _navigationBar.delegate = self;
```
romoore: Please use `self.navigationBar` outside of the initializer.  Same for the other properties where it makes sense.

components/BottomNavigation/examples/MDCBottomNavigationBarController.m
```diff
@@ -0,0 +1,278 @@
...
+}
+
+- (void)setSelectedIndex:(NSUInteger)selectedIndex {
+  // If we are setting to NSNotFound deselect the items
+  if (selectedIndex == NSNotFound) {
+    [self deselectCurrentItem];
+    return;
+  }
+
+  BOOL outOfBounds = selectedIndex >= [self.viewControllers count] ||
```
romoore: Nit: `count` is a property, please use dot notation for access (e.g., `self.viewControllers.count`)

romoore (approved):

romoore (approved):
LGTM

mradzinski:
@romoore Is this something to expect to be released soon? I'm in the process of starting a project which would truly benefit from this PR. Not rushing, just wondering :)

romoore:
Hi @mradzinski . Not in the next few weeks. December is a quieter period for us due to several corporate holidays in our US offices.  The next steps for this work would be a design review with the core team and then another API review on GitHub—likely in January 2019. Once those pass we'll be working on migrating this work to the  ~MaterialComponentsAlpha~ MaterialComponentsBeta pod spec.  If everything looks good after some internal testing and integration, it will be moved out to the main Material Components pod spec. I wouldn't anticipate it being in the main pod spec before February 2019.

mradzinski:
Hi @romoore I see, thanks for the detailed info. I actually went ahead and copy-pasted the classes manually, bridged them (I'm on Swift) and it crashes [at this line](https://github.com/nokeeo/material-components-ios/blob/8973b07eb8bc0b6cfcfe3f16ff6bdeb6cdfec362/components/BottomNavigation/examples/MDCBottomNavigationBarController.m#L221) (Seems `navigationBar` is nil by then). Just thought you guys might wanted to know :)

Cheers and thanks!

romoore:
Thanks. I suspect that `init` is not being called and the class failed to implement the superclass designated VCs.  How did you initialize the view controller, did you use Interface Builder?

mradzinski:
I'm actually extending from `MDCBottomNavigationBarController`, but indeed the controller is present in a Storyboard. I don't know if you're referring to that when mentioning Interface Builder?

mradzinski:
@romoore d5a4e10 fixes it. Thanks.

PR #5897, opened 2018-12-03
[NavigationDrawer] Remove clipping view

codeman7:
### Context
In working on navigation drawer I noticed the addition of the clippingScrollView. This additional view just makes the component more complex with no known benefit. This change removes that view in order to simply the component and remove the extra subview in the hierarchy.
### The problem
Adding the clippingScrollView adds extra complexity to the navigation drawer.
### The fix
Remove the clipping scroll view.

### DIFF
--- a/components/NavigationDrawer/src/private/MDCBottomDrawerContainerViewController.m
+++ b/components/NavigationDrawer/src/private/MDCBottomDrawerContainerViewController.m
@@ -127,9 +127,6 @@ static UIColor *DrawerShadowColor(void) {
 // The main scroll view.
 @property(nonatomic, readonly) UIScrollView *scrollView;
 
-// View that functions as the superview of |scrollView|. Used to clip the top of the scroll view.
-@property(nonatomic, readonly) UIView *scrollViewClippingView;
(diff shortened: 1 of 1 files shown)


guylivneh:
scrollViewClippingView is there for the cases when the drawer header's background is semi-transparent to ensure that the content is not seen behind it while scrolling in scrollview. If this use case is tested and the behavior is still satisfying without it - scrollViewClippingView could be potentially removed.

guylivneh:
If eventually scrollViewClippingView is not removed - I suggest maybe update the comment to explain what it's for (that could definitely help future readers)

codeman7:
@guylivneh Thanks! @yarneo and I couldn't figure out what it is for, we will talk to design and see what they prefer, only because a transparent AppBar should show the content behind it.

codeman7:
After talking to internal designers, we have concluded we do want to show the contentViewController behind the headerViewController if the headerViewController is semi-transparent. @guylivneh

yarneo (approved):
thanks for simplifying the code!!

PR #5905, opened 2018-12-04
[Dialogs] Add dialog category for theming

### LINKED ISSUE #5836: [Dialogs] Internal issue: b/118160386
This was filed as an internal issue. If you are a Googler, please visit [b/118160386](http://b/118160386) for more details.

---
#### Internal data
- Associated internal bug: [b/118160386](http://b/118160386)
- Blocked by: https://github.com/material-components/material-components-ios/issues/5882

andrewoverton:
These changes add the Dialogs theming category.
Closes #5836.


MaterialComponentsAlpha.podspec
```diff
@@ -79,6 +79,17 @@ Pod::Spec.new do |mdc|
...
   end
 
+  mdc.subspec "Dialogs+Theming" do |extension|
+    extension.ios.deployment_target = '8.0'
+    extension.public_header_files = "components/#{extension.base_name.split('+')[0]}/src/#{extension.base_name.split('+')[1]}/*.h"
+    extension.source_files = "components/#{extension.base_name.split('+')[0]}/src/#{extension.base_name.split('+')[1]}/*.{h,m}", "components/#{extension.base_name.split('+')[0]}/src/#{extension.base_name.split('+')[1]}/private/*.{h,m}"
+    extension.dependency "MaterialComponents/#{extension.base_name.split('+')[0]}"
+    extension.dependency "MaterialComponents/#{extension.base_name.split('+')[0]}+ColorThemer"
+    extension.dependency "MaterialComponents/#{extension.base_name.split('+')[0]}+TypographyThemer"
+    extension.dependency "MaterialComponents/ShadowElevations"
```
codeman7, line 89: Do we need to depend on Button+Theming?

codeman7 (approved):

components/Dialogs/src/Theming/MDCAlertController+Theming.h
```diff
@@ -0,0 +1,32 @@
...
+// limitations under the License.
+
+#import "MaterialContainerScheme.h"
+#import "MaterialDialogs.h"
+
+/**
+ This category is used to style MDCAlertController instances to a specific Material style which can
+ be found within the [Material Guidelines](https://material.io/design/components/buttons.html).
+ */
+@interface MDCAlertController (MaterialTheming)
```
galiak11: Tip: Should the file name match the category name? 
i.e.: MDCAlertController+MaterialTheming.h

components/Dialogs/tests/unit/DialogsThemingTests.swift
```diff
@@ -0,0 +1,49 @@
...
+
+import XCTest
+
+import MaterialComponents.MaterialDialogs
+import MaterialComponents.MaterialColorScheme
+import MaterialComponents.MaterialTypographyScheme
+import MaterialComponents.MaterialShadowElevations
+import MaterialComponentsAlpha.MaterialDialogs_Theming
+
+class DialogsThemerTests: XCTestCase {
```
galiak11: Can we also add a test for dialog-buttons? Verify that buttons are styled based on their emphases?

components/Dialogs/examples/DialogsAlertCustomizationViewController.swift
```diff
@@ -57,14 +57,7 @@ class CustomAlertViewController: UIViewController {
...
 
-  var colorScheme = MDCSemanticColorScheme()
-  var typographyScheme = MDCTypographyScheme()
-  var alertScheme: MDCAlertScheme {
-    let scheme = MDCAlertScheme()
-    scheme.colorScheme = self.colorScheme
-    scheme.typographyScheme = self.typographyScheme
-    return scheme
-  }
+  var scheme: MDCContainerScheme = MDCContainerScheme()
```
galiak11: I believe all examples need to use the same name for the instance scheme - so it can be injected by our example app. Something like "containerScheme" or "materialScheme", etc would work better in that case.

components/Dialogs/examples/DialogsAlertViewController.m
```diff
@@ -26,8 +26,6 @@
 
 @interface DialogsAlertViewController : MDCCollectionViewController
 @property(nonatomic, strong, nullable) NSArray *modes;
-@property(nonatomic, strong, nullable) MDCSemanticColorScheme *colorScheme;
```
galiak11, line 29: A materialScheme / containerScheme instance should replace the color/typography instances, so all examples can have consistent theming.
Make sure to set a default scheme in either the init method or in viewDidLoad (which I think is becoming the common place for that in our examples).

components/Dialogs/examples/DialogsTypicalUseViewController.m
```diff
@@ -44,22 +41,6 @@ - (void)viewDidLoad {
     self.colorScheme =
```
galiak11, line 41: Is there a reason to keep the color scheme?

MaterialComponentsAlpha.podspec
andrewoverton: Yes! Thanks

components/Dialogs/src/Theming/MDCAlertController+Theming.h
andrewoverton: Yes, it should. We should fix this with buttons too...

components/Dialogs/examples/DialogsTypicalUseViewController.m
andrewoverton: It's used [elsewhere](https://github.com/material-components/material-components-ios/pull/5905/files#diff-31c3277d8e88d585f221a2c199ec8140R44) in the view controller so I decided to keep it

components/Dialogs/examples/DialogsAlertCustomizationViewController.swift
andrewoverton: I'll rename it to `containerScheme` 👍

components/Dialogs/src/Theming/MDCAlertController+Theming.h
```diff
@@ -0,0 +1,32 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import "MaterialContainerScheme.h"
+#import "MaterialDialogs.h"
+
+/**
+ This category is used to style MDCAlertController instances to a specific Material style which can
+ be found within the [Material Guidelines](https://material.io/design/components/buttons.html).
```
romoore: Drive-by nit: Looks like this is the wrong link to the guidelines article?

Suggestion: https://material.io/design/components/dialogs.html

andrewoverton: Thank you 🙏

components/Dialogs/tests/unit/DialogsThemingTests.swift
andrewoverton: Done.

components/Dialogs/examples/DialogsAlertViewController.m
andrewoverton: done

PR #5913, opened 2018-12-04
[Buttons] Improve color themer tests.

romoore:
The MDCFloatingButtonColorThemer test was not checking all states for
the image tint color. Also adding a TODO for the broken
`backgroundColorForState:` API.

Noticed when working on #5912


fluxminds (approved):
LGTM. Thanks for catching this issue.

PR #5922, opened 2018-12-05
[Buttons] Update secondary floating action button themer examples

fluxminds:
Update examples to use secondary floating button theming extension method instead of calling themer's class methods.

Related https://github.com/material-components/material-components-ios/issues/5844

### DIFF
--- a/components/FeatureHighlight/examples/FeatureHighlightShownViewExample.m
+++ b/components/FeatureHighlight/examples/FeatureHighlightShownViewExample.m
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 #import "MaterialButtons.h"
-#import "MaterialButtons+ButtonThemer.h"
+#import "MaterialButtons+Theming.h"
 #import "MaterialFeatureHighlight.h"
 #import "MaterialFeatureHighlight+ColorThemer.h"
 #import "MaterialFeatureHighlight+FeatureHighlightAccessibilityMutator.h"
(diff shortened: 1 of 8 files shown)


components/FeatureHighlight/examples/FeatureHighlightShownViewExample.m
```diff
@@ -32,17 +32,20 @@ - (id)init {
...
+}
 
+- (void)didTapButton:(id)sender {
   MDCFloatingButton *fab = [[MDCFloatingButton alloc] init];
   [fab setImage:[UIImage imageNamed:@"Plus"] forState:UIControlStateNormal];
   [fab sizeToFit];
   fab.center = _button.center;
 
-  [MDCFloatingActionButtonThemer applyScheme:buttonScheme toButton:fab];
+  [fab applySecondaryThemeWithScheme:self.containerScheme];
```
rami-a, line 48: _Nit:_ Please call as a method: `[self containerScheme]`

rami-a (approved):

components/FeatureHighlight/examples/FeatureHighlightShownViewExample.m
fluxminds: Good catch and sorry about the confusion here. The reason `self.` is used here is becauuse `@property(nonatomic, readonly, strong) MDCContainerScheme *containerScheme;` is defined in `FeatureHighlightExampleSupplemental.h`. And the reason I created this property was because a category needs to call `containerScheme` and I don't feel it is good to expose the containerScheme as a method.

rami-a: oh ok, that makes sense

PR #5932, opened 2018-12-06
 [Dialogs] Renaming example view controllers

galiak11:
Adding "Example" to names of all view controllers - in files/class names.

### DIFF
--- a/components/Dialogs/examples/DialogsAlertCustomizationExampleViewController.swift
+++ b/components/Dialogs/examples/DialogsAlertCustomizationExampleViewController.swift
@@ -20,41 +20,6 @@ import MaterialComponents.MaterialCollections
 import MaterialComponents.MaterialDialogs_DialogThemer
 import MaterialComponents.MaterialTypographyScheme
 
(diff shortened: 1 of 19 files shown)


components/Dialogs/examples/DialogsAlertCustomizationExampleViewController.swift
```diff
@@ -20,42 +20,7 @@ import MaterialComponents.MaterialCollections
 import MaterialComponents.MaterialDialogs_DialogThemer
 import MaterialComponents.MaterialTypographyScheme
 
-class CustomAlertViewController: UIViewController {
```
rami-a, line 23: Was this removed because it wasn't be used?

rami-a (approved):

components/Dialogs/examples/DialogsAlertCustomizationExampleViewController.swift
galiak11: yes. it wasn't used anywhere, and there was no point in just renaming it...

PR #5933, opened 2018-12-06
[Snapshot] Tweak snapshot test base class to handle the verify in tearDown

rami-a:
This change simplifies the way a developer would have to write a snapshot test. The logic to verify the view is in the tearDown and will enforce the developer to write the test correctly.

### DIFF
--- a/components/Cards/tests/snapshot/MDCCardSnapshotTests.m
+++ b/components/Cards/tests/snapshot/MDCCardSnapshotTests.m
@@ -27,10 +27,9 @@
 
   // Given
   MDCCard *card = [[MDCCard alloc] initWithFrame:CGRectMake(0, 0, 50, 50)];
-  UIView *backgroundView = [self addBackgroundViewToView:card];
 
-  // Then
-  [self snapshotVerifyView:backgroundView];
+  // When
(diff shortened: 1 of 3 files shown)


components/Cards/tests/snapshot/MDCCardSnapshotTests.m
```diff
@@ -27,10 +27,9 @@ - (void)testDefaultCard {
 
   // Given
   MDCCard *card = [[MDCCard alloc] initWithFrame:CGRectMake(0, 0, 50, 50)];
-  UIView *backgroundView = [self addBackgroundViewToView:card];
 
-  // Then
-  [self snapshotVerifyView:backgroundView];
+  // When
```
romoore: This seems like the "Then". You're verifying here.  Maybe it's When + Then?

romoore (approved):

components/Cards/tests/snapshot/MDCCardSnapshotTests.m
rami-a: Actually yeah, the `then` is better, `when` can be reserved for applying themes or similar in future tests

PR #5938, opened 2018-12-07
[Shape] add hash function to MDCCornerTreatment

fluxminds:
The MDCCornerTreatment already supports equality. hash function is added in this PR to supplement it.

### DIFF
--- a/components/private/ShapeLibrary/tests/unit/ShapeLibraryTests.m
+++ b/components/private/ShapeLibrary/tests/unit/ShapeLibraryTests.m
@@ -39,7 +39,6 @@ void GetCGPathAddLineToPointValues(void *info, const CGPathElement *element);
   MDCCornerTreatment *cornerTreatment = [MDCCornerTreatment cornerWithCurve:CGSizeMake(2, 5)];
 
   // When
-  XCTAssertNotEqual(curvedCorner.hash, curvedCorner2.hash);
   XCTAssertNotEqualObjects(curvedCorner, curvedCorner2);
   curvedCorner2.size = CGSizeMake(2, 5);
 
--- a/components/private/ShapeLibrary/tests/unit/ShapeLibraryTests.m
+++ b/components/private/ShapeLibrary/tests/unit/ShapeLibraryTests.m
@@ -39,10 +39,12 @@ void GetCGPathAddLineToPointValues(void *info, const CGPathElement *element);
   MDCCornerTreatment *cornerTreatment = [MDCCornerTreatment cornerWithCurve:CGSizeMake(2, 5)];
 
   // When
+  XCTAssertNotEqual(curvedCorner.hash, curvedCorner2.hash);
   XCTAssertNotEqualObjects(curvedCorner, curvedCorner2);
   curvedCorner2.size = CGSizeMake(2, 5);
 
   // Then
+  XCTAssertEqual(curvedCorner.hash, curvedCorner2.hash);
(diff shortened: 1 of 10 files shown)


components/private/ShapeLibrary/tests/unit/ShapeLibraryTests.m
```diff
@@ -39,10 +39,12 @@ - (void)testCurvedCornerEquality {
   MDCCornerTreatment *cornerTreatment = [MDCCornerTreatment cornerWithCurve:CGSizeMake(2, 5)];
 
   // When
+  XCTAssertNotEqual(curvedCorner.hash, curvedCorner2.hash);
```
romoore: Please remove these `NotEqual` asserts. Hashes can be not equal when the objects are equal. It's the reverse (Equality requires identical hashes) that has to be enforced.

components/private/Shapes/tests/unit/ShapesNoopTest.m
```diff
@@ -27,4 +27,23 @@ - (void)testInit {
...
 
+- (void)testCornerTreatmentEquality {
+  // Given
+  MDCCornerTreatment *cornerTreatment1 = [[MDCCornerTreatment alloc] init];
+  cornerTreatment1.valueType = MDCCornerTreatmentValueTypeAbsolute;
+  MDCCornerTreatment *cornerTreatment2 = [[MDCCornerTreatment alloc] init];
+  cornerTreatment2.valueType = MDCCornerTreatmentValueTypePercentage;
+
+  // Then
+  XCTAssertNotEqual(cornerTreatment1.hash, cornerTreatment2.hash);
```
romoore: Please remove this `NotEqual` assert as well.

components/private/Shapes/tests/unit/ShapesNoopTest.m
```diff
@@ -27,4 +27,23 @@ - (void)testInit {
   XCTAssertNotNil(view);
 }
 
+- (void)testCornerTreatmentEquality {
```
romoore: Please move this to a new test class. The "noop" test can be removed later since we now actually have tests for shapes.

romoore (approved):

components/private/Shapes/tests/unit/ShapesNoopTest.m
wenyuzhang666: This is the only test file under Shapes directory, the other one I assume you mentioned is for ShapeLibrary/, which is not under Shapes/. I think it makes sense to either rename this class, or create a new test file.

components/private/ShapeLibrary/tests/unit/ShapeLibraryTests.m
fluxminds: Addressed in https://github.com/material-components/material-components-ios/pull/5954

components/private/Shapes/tests/unit/ShapesNoopTest.m
fluxminds: Addressed in https://github.com/material-components/material-components-ios/pull/5954

components/private/Shapes/tests/unit/ShapesNoopTest.m
fluxminds: Addressed in https://github.com/material-components/material-components-ios/pull/5954

fluxminds:
Closing this one, since a dependency PR has been merged in. A new PR has been sent in https://github.com/material-components/material-components-ios/pull/5954.

PR #5940, opened 2018-12-07
[Snapshot] Deallocate `testView` in teardown

romoore:
To make sure we aren't retaining too much in memory after tests
complete, the `-tearDown` method of Snapshot tests should remove any
remaining references created during their set up.

Found while working on #5939


rami-a (approved):

romoore:
Made obsolete by #5947

PR #5945, opened 2018-12-07
[Theming] Update our theming extensions

codeman7:
### Context
All of our theming targets should be `MDC<#Component>+Theming` but all files should be `MDC<#Component>+MaterialTheming.m/h`. We got out of sync and caught this early.

### The problem
Our targets and file names weren't correct in some cases

### The fix
Update our targets and files to be what we have decided as a team.


romoore (approved):
LGTM

PR #5946, opened 2018-12-07
[Shape] add NSCopying support to ShapeCategory

fluxminds:
Adding NSCopying support to ShapeCategory would facilitate deep copying for any class that uses it.

### DIFF
--- a/components/schemes/Shape/tests/unit/MDCShapeSchemeTests.m
+++ b/components/schemes/Shape/tests/unit/MDCShapeSchemeTests.m
@@ -25,19 +25,19 @@
 
 - (void)testShapeCategoryEquality {
   // Given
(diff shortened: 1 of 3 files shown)


components/schemes/Shape/tests/unit/MDCShapeSchemeTests.m
```diff
@@ -68,4 +69,20 @@ - (void)testInitWithMaterialDefaults {
                                                                 andSize:4]);
 }
 
+- (void)testShapeCategoryCopy {
+  // Given
+  MDCShapeCategory *cat = [[MDCShapeCategory alloc] initCornersWithFamily:MDCShapeCornerFamilyCut
```
romoore, line 74: Nit: Please name this something like "shapeCategory" or "testCategory" or "originalCategory" or even "category". A cat is also an animal and it confused me for a moment when I read the code.

romoore (approved):

Issue #5762, opened 2018-11-15
 [TextField] Error Text Overflows Text Field Width 


material-mirror:
When setting MDCTextField error text to a long string, the error text runs past the width of the text field itself.

---
#### Internal data
- Associated internal bug: [b/119567486](http://b/119567486)
- Blocked by: https://github.com/material-components/material-components-ios/issues/2153

enthuan:
Hi, is this feature still on the roadmap? I will soon need to make error label on multiple lines and I don't want to use another lib ;)
Thanks

romoore:
I'll ask @randallli to reëvaluate the priority of this work in our backlog.

kaal-dam:
Hello, did you have any idea of when this will be fixed ?

zishanj:
Any update on this issue?

sv02:
Any updates?

Fernandez-Aldo:
How do I implement multiline error text?

silvaric:
You have that feature at **v93.2.0**
```
MDCTextField *textField = [[MDCTextField alloc] init];
textField.leadingUnderlineLabel.numberOfLines = 0;
textField.leadingUnderlineLabel.lineBreakMode = NSLineBreakByWordWrapping;
```

Fernandez-Aldo:
> You have that feature at **v93.2.0**
> 
> ```
> MDCTextField *textField = [[MDCTextField alloc] init];
> textField.leadingUnderlineLabel.numberOfLines = 0;
> textField.leadingUnderlineLabel.lineBreakMode = NSLineBreakByWordWrapping;
> ```
> 
> Release notes : https://github.com/material-components/material-components-ios/releases/tag/v93.2.0

Thanks Silvaric

Fernandez-Aldo:
Is there a way to customize the border color of the texfield when setting errors?

andrewoverton:
@Fernandez-Aldo for MDCTextFields you would do this by setting the `errorColor` on the controller

PR #5967, opened 2018-12-10
[TextFields] Fix `active` snapshots for Outlined Baseline

romoore:
The `active` snapshots failed to be in the correct state because an important
notification wasn't posted correctly. Adding the notification in the test fake
results in more accurate snapshot images for text fields.

Part of #5762


rami-a (approved):

romoore:
https://github.com/material-components/material-components-ios/pull/5972 is a better fix for the notifications.

PR #5969, opened 2018-12-10
[Snapshot] Add support for image tolerance.

romoore:
Some snapshot tests are flaky (which is a bug). Allow those tests to
allow some small percentage of pixels to change to eliminate the flake
while still detecting larger changes.

Caused by #5970

### DIFF
--- a/components/private/Snapshot/MDCSnapshotTestCase.h
+++ b/components/private/Snapshot/MDCSnapshotTestCase.h
@@ -25,7 +25,20 @@
 /**
  * This will call FBSnapshotVerifyView but first check for supported iOS versions. Additionally,
  * this will use UIGraphicsImageRenderer to render the view correctly (including shadows).
+ *
+ * Permits no changed pixels.
+ *
+ * @param view The view to use for snapshot comparison.
  */
 - (void)snapshotVerifyView:(UIView *)view;
+
+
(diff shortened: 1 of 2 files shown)


jverkoey (approved):
Cool!

components/private/Snapshot/MDCSnapshotTestCase.h
```diff
@@ -25,7 +25,20 @@
...
+ * @param view The view to use for snapshot comparison.
  */
 - (void)snapshotVerifyView:(UIView *)view;
 
+
+/**
+ * This will call FBSnapshotVerifyView but first check for supported iOS versions. Additionally,
+ * this will use UIGraphicsImageRenderer to render the view correctly (including shadows).
+ * @param view the view to use for snapshot comparison.
+ * @param tolerancePercent the percentage of pixels that can differ while still passing.
```
rami-a: _Tip:_ I think we should specify that it should be 0-1 here so we don't have to dig through the library itself.

rami-a (approved):

components/private/Snapshot/MDCSnapshotTestCase.h
romoore: Done.

PR #5972, opened 2018-12-11
[TextFields] Post notifications when `isEditing` changes.

romoore:
To ensure that clients can remain the UITextFieldDelegate of MDCTextField
objects, the MDCTextInputControllers respond to NSNotifications coming from
UITextField when its state changes. To improve correct layout and appearance
of text fields in snapshot tests, any test fakes should post the relevant
notifications when changing the value of `isEditing`.

Part of #5762


jverkoey (approved):
git lfs image diffs are working again - hurray!

Also: so many visual bugs :(

PR #5981, opened 2018-12-12
[TextFields] Add `isEditing` snapshots for Filled style

romoore:
Filled controller text fields should have `isEditing` (`active`) snapshots.

Part of #5762


romoore:
The layout is buggy because we're not allowing animations to occur.

PR #5987, opened 2018-12-12
[TextFields] Allow animations to execute in tests.

### LINKED ISSUE #5970: [TextFields] Outlined controllers with placeholders longer than the text field are flaky
When an Outlined controller's placeholder string would be longer than the textfield, the position of the floating placeholder is flaky. This can be seen in the snapshot test `testOutlinedTextFieldWithLongPlaceholderTextIsEditing` which requires a 5% tolerance to pass without any functional changes.

romoore:
Some of TextFields' layout changes only take place inside animation blocks. To
move us closer toward a more "life-like" rendering of the text fields in our
snapshot tests, the test should allow the main run loop to execute any pending
animations before attempting to take a snapshot of the view.

This seems to fix the flakiness seen in #5970, so the `withTolerance` APIs added to support those flaky tests are no longer needed.

Part of #5762
Fixes #5970

### DIFF
--- a/components/TextFields/tests/snapshot/MDCTextFieldOutlinedControllerBaselineSnapshotTests.m
+++ b/components/TextFields/tests/snapshot/MDCTextFieldOutlinedControllerBaselineSnapshotTests.m
@@ -116,10 +116,7 @@
   [self.textField MDCtest_setIsEditing:YES];
 
   // Then
-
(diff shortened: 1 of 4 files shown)


fluxminds:
Followed up with Rob offline, some comments explaining why/how we are using `drainMainRunLoop` as a practical way to test the layout with animation right now.

fluxminds (approved):

rami-a (approved):

PR #6010, opened 2018-12-13
[NavigationDrawer] Attempt to fix header shadow not being removed

### LINKED ISSUE #6011: [NavigationDrawer] Header Shadow Layer is not removed upon dismissal
This was filed as an internal issue. If you are a Googler, please visit [b/120989705](http://b/120989705) for more details.

---
#### Internal data
- Associated internal bug: [b/120989705](http://b/120989705)

m-cheung:
This pull request attempts to fix an issue where in some cases, the Navigation Drawer can be dismissed, but the header shadow will not be removed from the super layer. The cause of this is because the bottomDrawerContainerViewController in MDCBottomDrawerPresentationController is removed before the completeTransition in MDCBottomDrawerTransitionController (which triggers viewDidDisappear in MDCBottomDrawerContainerViewController).

Internal issue: b/120989705
Closes: #6011 

#### Don't forget:
- [ ] Sign the CLA bot. You can do this once the pull request is opened.

pull request.

### DIFF
--- a/components/NavigationDrawer/src/MDCBottomDrawerPresentationController.m
+++ b/components/NavigationDrawer/src/MDCBottomDrawerPresentationController.m
@@ -191,6 +191,9 @@ static CGFloat kTopHandleTopMargin = (CGFloat)5.0;
 
 - (void)dismissalTransitionDidEnd:(BOOL)completed {
   if (completed) {
+    if ([self.presentedViewController isKindOfClass:[MDCBottomDrawerViewController class]]) {
+      [self.bottomDrawerContainerViewController removeFromParentViewController];
+    }
     [self.scrimView removeFromSuperview];
     [self.topHandle removeFromSuperview];
   }
--- a/components/NavigationDrawer/src/MDCBottomDrawerPresentationController.m
+++ b/components/NavigationDrawer/src/MDCBottomDrawerPresentationController.m
@@ -191,9 +191,6 @@ static CGFloat kTopHandleTopMargin = (CGFloat)5.0;
 
 - (void)dismissalTransitionDidEnd:(BOOL)completed {
   if (completed) {
-    if ([self.presentedViewController isKindOfClass:[MDCBottomDrawerViewController class]]) {
-      [self.bottomDrawerContainerViewController removeFromParentViewController];
-    }
     [self.scrimView removeFromSuperview];
     [self.topHandle removeFromSuperview];
   }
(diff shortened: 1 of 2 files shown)


m-cheung:
@yarneo, I'm not too sure where the best place to make this fix would be. This was one of the fixes that I was able to come up with, but I'm not certain it is the best way of doing it. Please let me know what your thoughts are.

Cheers!

yarneo:
Thank you for finding this issue and opening a fix @m-cheung !

I was wondering, have you found cases where the invocation of the `completeTransition` method in the transition controller isn't necessarily calling `dismissalTransitionDidEnd` of the presentation controller after? There are cases where `dismissalTransitionDidEnd` is invoked before the transition controller finished?

Here is a picture for clarification of what I'd assume would happen every time:
screen shot 2018-12-14 at 10 19 30 am

m-cheung:
> Thank you for finding this issue and opening a fix @m-cheung !
> 
> I was wondering, have you found cases where the invocation of the `completeTransition` method in the transition controller isn't necessarily calling `dismissalTransitionDidEnd` of the presentation controller after? There are cases where `dismissalTransitionDidEnd` is invoked before the transition controller finished?
> 
> Here is a picture for clarification of what I'd assume would happen every time:
> screen shot 2018-12-14 at 10 19 30 am

I have not observed any cases where the transition controller isn't calling `dismissalTransitionDidEnd` nor have I seen cases where `dismissalTransitionDidEnd` is invoked before the transition controller is finished.

The issue is with regards to when `viewDidDisappear` is called. `viewDidDisappear` is also invoked when `completeTransition` occurs. However, `viewDidDisappear` seems to occurs after `dismissalTransitionDidEnd`. Since the presentation controller removes itself from the parent before `viewDidDisappear` occurs, it never ends up getting executed.

This is the flow that I observe when the view controller is not removed during `dismissalTransitionDidEnd` (I apologize for the slow, low quality GIF ahead of time): 
flow

yarneo:
Thank you!
The reason I am investigation this a bit more is that I would prefer to avoid the view controller removing itself from the parent view controller. How about moving the block of code from viewDidDisappear to the view controller's dealloc? I just tested it and it seems to work well. Curious to know your thoughts and if that solves the current issue.

m-cheung:
I can attest that putting this code block in dealloc also solves that issue (it was actually one of the solutions that I explored as well). Out of the two solutions, I couldn't decide on which one seemed the most logical and just ended up placing the code block in `viewDidDisappear`.

I have made that change. PTAL and let me know what you think.

yarneo (approved):
Thank you!

m-cheung:
Noticed the tests were failing due to `removeScrollViewObserver` being missing from `viewDidDisappear`. The latest changes should resolve that issue.

PR #6012, opened 2018-12-14
[BottomNavigation]Adds Initial tests for MDCBottomNavigationBarController

nokeeo:


MaterialComponents.podspec
```diff
@@ -181,7 +181,7 @@ Pod::Spec.new do |mdc|
 
     component.test_spec 'tests' do |tests|
       tests.test_spec 'unit' do |unit_tests|
-        unit_tests.source_files = "components/#{component.base_name}/tests/unit/*.{h,m,swift}", "components/#{component.base_name}/tests/unit/supplemental/*.{h,m,swift}"
+        unit_tests.source_files = "components/#{component.base_name}/tests/unit/*.{h,m,swift}", "components/#{component.base_name}/tests/unit/supplemental/*.{h,m,swift}", "components/#{component.base_name}/examples/*.{h,m,swift}", "components/#{component.base_name}/examples/supplemental/*.{h,m,swift}", "components/#{component.base_name}/examples/tests/unit/*.{h,m,swift}"
```
nokeeo, line 184: Should we create an issue to revert this change after this lands as including examples in the source files will no longer be needed?

romoore: Hmm, I wonder if we should just add these to an independent spec or something.

@jverkoey who is a Cocoapods wiz and @andrewoverton who has been working on our test spec recently.

We have a Prototype/Alpha (proper Alpha) component in `BottomNavigation/examples` and want to add unit tests. What's the best way to do so?

components/BottomNavigation/examples/tests/unit/MDCBottomNavigationBarControllerTests.m
```diff
@@ -0,0 +1,351 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <XCTest/XCTest.h>
+
+#import "MDCBottomNavigationBarController.h"
+
+CGFloat const kDefaultExpectationTimeout = 15.f;
```
romoore: Nit: For consistency with the code base, please use `15` for integral values and `(CGFloat)1.5` for non-integrals.

romoore: Nit: Please make this a `static` variable to keep its scope to the test file.

components/BottomNavigation/examples/tests/unit/MDCBottomNavigationBarControllerTests.m
```diff
@@ -0,0 +1,351 @@
...
+ */
+@property(nonatomic, strong, nullable) id returnValue;
+
+@end
+
+@implementation MDCBottomNavigationControllerTests
+
+- (void)setUp {
+  [super setUp];
+  _bottomNavigationBarController = [[MDCBottomNavigationBarController alloc] init];
```
romoore: Nit: Please use property accessors here and throughout.

```objc
self.bottomNavigationBarController = [[MDCBottomNavigationBarController alloc] init];
```

components/BottomNavigation/examples/tests/unit/MDCBottomNavigationBarControllerTests.m
```diff
@@ -0,0 +1,351 @@
...
+}
+
+- (void)testSetViewControllers {
+  // Setup items.
+  UITabBarItem *tabBarItem1 = [[UITabBarItem alloc] initWithTitle:@"Title1" image:nil tag:0];
```
romoore, line 62: Please add comments with the names "Given", "When", and "Then" to mark the different portions of the test.  It helpers readers understand the preconditions, test conditions, and verifications in a standard way.

Blog post talking about it: https://martinfowler.com/bliki/GivenWhenThen.html
Example use: https://github.com/material-components/material-components-ios/blob/develop/components/BottomAppBar/tests/unit/BottomAppBarColorThemerTests.m#L49

components/BottomNavigation/examples/tests/unit/MDCBottomNavigationBarControllerTests.m
```diff
@@ -0,0 +1,351 @@
...
+  UIViewController *viewController1 = [[UIViewController alloc] init];
+  UIViewController *viewController2 = [[UIViewController alloc] init];
+
+  // Assign the tab bar items to the view controllers
+  viewController1.tabBarItem = tabBarItem1;
+  viewController2.tabBarItem = tabBarItem2;
+
+  _bottomNavigationBarController.viewControllers = @[ viewController1, viewController2 ];
+
+  XCTAssert([tabBarItems count] == [_bottomNavigationBarController.navigationBar.items count],
```
romoore: Nit: Please use `XCTAssertEqual` instead of the generic `XCTAssert` function. You can add a contextual message to the end of the default message if you want, although in this case it's probably not necessary.

e.g.,

```objc
XCTAssertEqual(self.bottomNavigationBarController.navigationBar.items.count, tabBarItems.count,
@" - The number of items in the bottom navigation bar's `items` should be set by the number of view controllers.");
```

romoore: Nit: Please put the "tested variable" on the left side of the `XCTAssert...` method.  That way the default error message will be something like,  "(test value) is not equal to (expected value)".

components/BottomNavigation/examples/tests/unit/MDCBottomNavigationBarControllerTests.m
```diff
@@ -0,0 +1,351 @@
...
+            @"The number of items in the bottom navigation bar's items property does not match the "
+            @"number of items set in the test");
+  int count = (int)[tabBarItems count];
+
+  // Check each item in the navigation bar.
+  for (int i = 0; i < count; i++) {
+    UITabBarItem *expectedItem = [tabBarItems objectAtIndex:i];
+    UITabBarItem *item = [_bottomNavigationBarController.navigationBar.items objectAtIndex:i];
+    XCTAssert([expectedItem.title isEqualToString:item.title] && expectedItem.tag == item.tag,
```
romoore: Same suggested use of `XCTAssertEqual` here and throughout.  In this case, two distinct asserts would be preferable since it increases the visibility of which assertion failed (equality vs. tag value).

components/BottomNavigation/examples/tests/unit/MDCBottomNavigationBarControllerTests.m
```diff
@@ -0,0 +1,351 @@
...
+
+@end
+
+@implementation MDCBottomNavigationControllerTests
+
+- (void)setUp {
+  [super setUp];
+  _bottomNavigationBarController = [[MDCBottomNavigationBarController alloc] init];
+}
+
```
romoore, line 59: Please add the `tearDown` method to nil-out the bottom navigation bar property. Since they are not deallocated automatically after every test method, there is some risk by this test class being the delegate of so many controllers.

Good article talking about how `tearDown` behaves: https://qualitycoding.org/xctestcase-teardown/

components/BottomNavigation/examples/tests/unit/MDCBottomNavigationBarControllerTests.m
```diff
@@ -0,0 +1,351 @@
...
+/**
+ * An array of the values of the expected arguments the delegate method should be called with.
+ * The order of the array is the order the arguments appear in the method signature.
+ */
+@property(nonatomic, strong, nullable) NSArray<id> *expectedArguments;
+
+/**
+ * The return value a delegate method should return with.
+ */
+@property(nonatomic, strong, nullable) id returnValue;
```
romoore: Nit/naming: Please name this something like `delegateReturnValue` to make the variable's use a little clearer when reading tests.

romoore (approved):
Some nits and code improvement to be made but overall LGTM.

MaterialComponents.podspec
jverkoey: This looks good for now; we can explore making this more formal in a follow-up PR.

jverkoey (approved):

MaterialComponents.podspec
romoore: This is removing the examples from the MaterialComponentsExamples development pod in Xcode:

screen shot 2018-12-18 at 10 01 52 am

PR #6028, opened 2018-12-15
[TextFields] Add `leadingImage` snapshots.

romoore:
Adding snapshots for single-line text fields because they support
leading views.

Part of #5762


components/TextFields/tests/snapshot/resources/TextFieldsSnapshotTests.xcassets/ic_done.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
+{
```
rami-a, line 1: This asset appears unused, is there a reason for adding it?

rami-a (approved):

components/TextFields/tests/snapshot/resources/TextFieldsSnapshotTests.xcassets/ic_done.imageset/Contents.json
romoore: It was simpler to just copy-paste everything from the examples to the tests. We can make use of it later.

Issue #5511, opened 2018-10-24
[EPIC] Add support for Chips


material-mirror:
Text Fields need to support Input Chips as seen at https://material.io/design/components/chips.html#input-chips

## Definitions of Done ##
1.  Users can append a new Chip to a Text Field.
2. Users can delete the most recently-added Chip from a Text Field.
3. The Chips are visually coherent with the Placeholder or User-entered text of the Text Field.
4.  Support for mixing Chips containing either LTR or RTL text.  Each Chip only has LTR or RTL text.  The position of the Chips within the Text Field will match that of the current device Locale (not the Keyboard Input language).
5.  All Chips and the input area of the Text Field support minimum touch target sizes (48 points square).

### Initial research spike ###
https://github.com/material-components/material-components-ios/issues/4178

### Initial prototype ###
https://github.com/romoore/material-components-ios/tree/issue-4178

---
#### Internal data
- Associated internal bug: [b/118378923](http://b/118378923)
- Blocked by: https://github.com/material-components/material-components-ios/issues/6211
- Blocked by: https://github.com/material-components/material-components-ios/issues/4283
- Blocked by: https://github.com/material-components/material-components-ios/issues/4285
- Blocked by: https://github.com/material-components/material-components-ios/issues/4286

material-mirror:
The internal issue [b/118378923](http://b/118378923) is now closed. This issue is being closed as a result.

PR #6032, opened 2018-12-17
[Chips] Rename internal MDCTextField subclass in MDCChipField

galiak11:
### Background
Adding chip support to MDCTextfield (b/118378923) introduces a new MDCChipTextField API.
This PR renames internal classes that start with "MDCChipTextField..." to allow adding the new API. 

There is no change to the MDCChipField public API.

Prework for #5511

### DIFF
--- a/components/Chips/src/MDCChipField.m
+++ b/components/Chips/src/MDCChipField.m
@@ -107,8 +107,9 @@ const UIEdgeInsets MDCChipFieldDefaultContentEdgeInsets = {
 
 @end
 
-@interface MDCChipField ()
-    <MDCChipFieldTextFieldDelegate, MDCTextInputPositioningDelegate, UITextFieldDelegate>
+@interface MDCChipField () <MDCChipFieldTextFieldDelegate,
+                            MDCTextInputPositioningDelegate,
+                            UITextFieldDelegate>
 @end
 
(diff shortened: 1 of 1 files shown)


components/Chips/src/MDCChipField.m
```diff
@@ -122,7 +122,7 @@ - (instancetype)initWithFrame:(CGRect)frame {
 
     _chips = [NSMutableArray array];
 
-    MDCChipTextField *chipTextField = [[MDCChipTextField alloc] initWithFrame:self.bounds];
+    MDCChipFieldTextField *chipTextField = [[MDCChipFieldTextField alloc] initWithFrame:self.bounds];
```
rami-a: Should the property by renamed as well?

rami-a (approved):

romoore (dismissed):
Have we considered naming the new one "MDCChipsTextField"? After all, the component is called "Chips" and it supports multiple chips.

If the alternate (non-conflicting) name is not possible for some reason, then please create an internal CL with all relevant fixes so that the release engineer will be able to apply the fixes when the release is cut.

components/Chips/src/MDCChipField.m
galiak11: Done

PR #6036, opened 2018-12-17
[BottomAppBar] Enable tapping on area above Navigation Bar

KirkSpaziani:
The view has a touch deadzone due to the optional button. This change causes the bottom app bar view to ignore touches that don't touch any of its subviews.

Closes #5606


randallli (approved):

KirkSpaziani:
Looks like I'm not authorized to merge, @randallli can you fire it off?

PR #6041, opened 2018-12-18
[Cards] Add snapshot tests for Card component with variations

### LINKED ISSUE #5493: [Cards] Add snapshot test to Cards
Definition of done:

- There is a snapshot test for Cards demonstrating a "default" card with zero configuration.

---
#### Internal data
- Associated internal bug: [b/118193097](http://b/118193097)
- Blocked by: https://github.com/material-components/material-components-ios/issues/5496

rami-a:
### Context
Since adding the ability to do snapshot tests, we want to increase our snapshot test coverage.

### The problem
We currently do not have snapshot tests that for cards

### The fix
* Add snapshot tests for MDCCard in unthemed, baseline themed, and outlined themed
* Add snapshot tests for MDCCardCollectionCell in unthemed, baseline themed, and outlined themed

### Bug
Closes #5493


components/Cards/tests/snapshot/MDCCardSnapshotTests.m
```diff
@@ -13,24 +13,92 @@
 // limitations under the License.
 
 #import "MDCSnapshotTestCase.h"
+#import "MaterialCards+Theming.h"
 #import "MaterialCards.h"
 
 @interface MDCCardSnapshotTests : MDCSnapshotTestCase
 
+@property(nonatomic, strong) MDCSemanticColorScheme *colorScheme;
```
romoore: Shouldn't the container scheme hold the color scheme property?

rami-a: Actually yeah that is better. Will make that change

romoore (approved):
Please file a bug for the cards with outlines AND elevation. I believe a Card should either be outlined OR elevated.

rami-a:
@romoore noticed that the outlined card should not have an elevation. I've opened #6045 to track and fix this.

PR #6044, opened 2018-12-18
[Scripts] Exit on error in release script

romoore:
The release script was continuing after errors failed to generate the
correct local clone of the repository for the API Diff step. This is
because the main release script suppresses non-zero return codes by
default. Instead, we should explicitly exit of any command returns an
unhandled non-zero value.

### DIFF
--- a/scripts/release
+++ b/scripts/release
@@ -1,4 +1,4 @@
-#!/bin/bash +e
+#!/bin/bash
 #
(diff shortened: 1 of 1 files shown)


jverkoey (approved):
It's likely this will result in things now failing when they shouldn't be, so we'll need to keep an eye out.

romoore:
I'll merge after the current release is finished.

PR #6063, opened 2018-12-19
[ActivityIndicator] Stop activity indicator animating when hidden

### LINKED ISSUE #5955: [ActivityIndicator] Stop animating the Activity Indicator when it is hidden
This was filed as an internal issue. If you are a Googler, please visit [b/120683366](http://b/120683366) for more details.

---
#### Internal data
- Associated internal bug: [b/120683366](http://b/120683366)

rami-a:
### The problem
When a client hides an activity indicator, the animation would continue despite not being necessary

### The solution
Override `setHidden` and stop the animation when hidden is true.

### Bugs
Closes #5955

### DIFF
--- a/components/ActivityIndicator/src/MDCActivityIndicator.m
+++ b/components/ActivityIndicator/src/MDCActivityIndicator.m
@@ -221,6 +221,11 @@ static const CGFloat kSingleCycleRotation =
   return CGSizeMake(edge, edge);
 }
 
+- (void)setHidden:(BOOL)hidden {
+  [super setHidden:hidden];
+  [self stopAnimating];
+}
+
 #pragma mark - Public methods
 
 - (void)startAnimating {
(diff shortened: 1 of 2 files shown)


components/ActivityIndicator/src/MDCActivityIndicator.m
```diff
@@ -221,6 +221,14 @@ - (CGSize)sizeThatFits:(__unused CGSize)size {
   return CGSizeMake(edge, edge);
 }
 
+- (void)setHidden:(BOOL)hidden {
+  [super setHidden:hidden];
+
+  if (hidden) {
```
romoore, line 227: Please test this PR against internal clients before submitting to make sure we don't break any screenshot tests.

romoore (approved):
LGTM pending internal tests.

components/ActivityIndicator/src/MDCActivityIndicator.m
rami-a: Tested internally and nothing is broken.

PR #6071, opened 2018-12-20
[ActionSheet] Remove duplicate build target.

romoore:
The ActionSheet BUILD file had a duplicate `private` target that isn't
necessary. `privateHeaders` already exists and doesn't use `includes`.


codeman7 (approved):

romoore:
The target should probably include the `includes` unless we want to retain the relative path usage in our headers. Although exporting "includes" can be expensive, limiting it to test targets (which should not be dependencies of other targets) should not have a significant effect.

romoore:
I'd like to use #6075 instead.

PR #6072, opened 2018-12-20
[ActionSheet] Add examples bazel targets.

romoore:
Adding BUILD targets for Objective-C and Swift examples. Also updating
.kokoro rewrite rules to transform the necessary imports. Combined with
the new build-only presubmit jobs on kokoro, we should be able to detect
non-building code before it is released internally.

### DIFF
--- a/components/ActionSheet/BUILD
+++ b/components/ActionSheet/BUILD
@@ -102,6 +102,35 @@ mdc_objc_library(
     visibility = ["//visibility:private"],
 )
 
+mdc_objc_library(
+    name = "ObjcExamples",
+    srcs = glob([
+        "examples/*.m",
+        "examples/*.h",
+        "examples/supplemental/*.m",
+        "examples/supplemental/*.h",
+    ]),
+    deps = [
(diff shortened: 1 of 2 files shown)


components/ActionSheet/BUILD
```diff
@@ -102,6 +102,35 @@ mdc_objc_library(
     visibility = ["//visibility:private"],
 )
 
+mdc_objc_library(
+    name = "ObjcExamples",
+    srcs = glob([
+        "examples/*.m",
+        "examples/*.h",
+        "examples/supplemental/*.m",
```
codeman7, line 110: _Question:_ Why not use `hdrs`?

codeman7 (approved):

components/ActionSheet/BUILD
romoore: [The `hdrs` argument is used to export headers to other targets.](https://docs.bazel.build/versions/master/be/objective-c.html#objc_library.hdrs) Since nothing depends on this example code, we don't need to export anything.

PR #6086, opened 2018-12-20
[bazel] Add component "extensions" rule.

romoore:
Since our extensions should generally follow the file system convention
for naming, we can create a custom bazel rule that reduces some
boilerplate code.

----

Example
-------

**Before**

```
mdc_objc_library(
    name = "Theming",
    srcs = native.glob(["src/Theming/*.m"]),
    hdrs = native.glob(["src/Theming/*.h"]),
    includes = ["src/Theming"],
    visibility = ["//visibility:public"],
    deps = [
        ":ActionSheet",
        ":ColorThemer",
        ":TypographyThemer",
        "//components/schemes/Container",
    ],
)
```

**After**

```
mdc_extension_objc_library(
    name = "Theming",
    deps = [
        ":ActionSheet",
        ":ColorThemer",
        ":TypographyThemer",
        "//components/schemes/Container",
    ],
)
```

### DIFF
--- a/material_components_ios.bzl
+++ b/material_components_ios.bzl
@@ -56,6 +56,43 @@ def mdc_public_objc_library(
       enable_modules = 1,
       **kwargs)
 
+def mdc_extension_objc_library(
+    name,
+    deps = [],
+    sdk_frameworks = [],
+    visibility = ["//visibility:public"],
+    **kwargs):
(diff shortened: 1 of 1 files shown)


jverkoey (approved):
Love it!

PR #6105, opened 2018-12-21
{TEST} [AppBar] Bad formatting! Bad formatting!

romoore:
Do not submit. WIP. Testing only.


romoore:
Looks like ~git clone failed on~ my forced rebase broke kokoro test runners.

romoore:
Looks like the history is too munged for kokoro to handle. Going to push up a separate PR to verify.

PR #6114, opened 2018-12-26
[TextFields] Arabic RTL snapshot tests

### LINKED ISSUE #6022: [TextFields] Snapshot tests need a way to force RTL layout
The snapshot tests for Text Fields should be able to render RTL script in RTL layout. However, setting neither the `effectiveUserInterfaceLayoutDirection` nor the `semanticContentAttribute`  seems to convince the text input that the text is in RTL.  The best that happens is some constraints updates, but neither the helper text nor the actual input text is positioned correctly.

## (Broken) Example:
failed_testtextfieldwithshortinputplaceholderhelpertextsisediting_11_2 2x

---
#### Internal data
- Associated internal bug: [b/121040596](http://b/121040596)

romoore:
Adding a single set of Arabic-language RTL snapshot tests for Text Fields. 

Closes #6022


components/TextFields/tests/snapshot/supplemental/MDCAbstractTextFieldSnapshotTests+I18N.m
```diff
@@ -20,6 +20,34 @@
...
 
+- (void)MDCForceTextInputRightToLeft:(UIView *)view {
+  if ([view conformsToProtocol:@protocol(UITextInput)]) {
+    id<UITextInput> textInput = (id<UITextInput>)view;
+    UITextRange *textRange = [textInput textRangeFromPosition:textInput.beginningOfDocument
+                                                   toPosition:textInput.endOfDocument];
+    if (textRange) {
+      [textInput setBaseWritingDirection:UITextWritingDirectionRightToLeft forRange:textRange];
+    } else {
+      XCTAssertNotNil(textRange);
```
fluxminds, line 32: I might misunderstand things here. What are `textRange` expected to look like here?

fluxminds (approved):
One small question left on the pull request.

components/TextFields/tests/snapshot/supplemental/MDCAbstractTextFieldSnapshotTests+I18N.m
romoore: textRange is the UITextRange of the currently-input text of the text field.

fluxminds: Followed up offline, we want an assertion failure to be thrown if the `else` block is executed. And because XCTAssert doesn't stop execution. Moving this line above the `setBase..` would still require the `if (textRange)` guard. Putting a comment there would make it clear:)

romoore: Done

PR #6130, opened 2018-12-27
[TextFields] Add Arabic snapshot tests for Filled style.

romoore:
Adding snapshot tests to capture Arabic strings with right-to-left layout with
Filled controller and without floating placeholders.

Part of #5762


components/TextFields/tests/snapshot/MDCTextFieldFilledControllerLeadingImageArabicSnapshotTests.m
```diff
@@ -0,0 +1,55 @@
...
+@implementation MDCTextFieldFilledControllerLeadingImageArabicSnapshotTests
+
+- (void)setUp {
+  [super setUp];
+
+  // Uncomment below to recreate the golden images for all test methods. Add it to a test method to
+  // update only that golden image.
+  //  self.recordMode = YES;
+
+  self.textField.clearButtonMode = UITextFieldViewModeAlways;
```
galiak11, line 33: Tip: Does it make sense to include some screenshot tests with the clear button off?

galiak11 (approved):

components/TextFields/tests/snapshot/MDCTextFieldFilledControllerLeadingImageArabicSnapshotTests.m
romoore: We could certainly add some, but I figured the clear button alignment/position is a more "complex" layout than without.  As a result, the tests I wrote so far tend to turn it on.

PR #6136, opened 2018-12-28
[NavigationBar] Refactor tests 

codeman7:
### Context
In working on #6118  I noticed some of the test had the same setup in the _Given_ section.  This duplication of code reduces the readability of the code and the test added in #6118 will also use this function to further reduce the duplication of _Given_ in the tests.
### The problem
Some of the test had the same _Given_.
### The fix
Move the duplicate code to it's own function to improve readability and make it so test later can be simpler.
### Related bug
This was noticed when working on #6133


randallli (approved):

randallli:
I reran the kokoro CI.

randallli:
Merging this as this change has no effect on the website and all other CI has passed.

PR #6137, opened 2018-12-28
[Snapshot] Extract TextFields strings for reuse in all Snapshot tests

codeman7:
### Context
Currently `MDCTextField` snapshot tests use a helper file for different strings. This would be helpful if we could use this for other components snapshot test as well as MDCTextField, as other components will need to test different values in their labels.  This would be helpful if other components can use this and not have to depend on MDCTextFieldSnapshot test but if this file was in a more general folder.
### The problem
Currently we have to redefine string values similar to what `MDCTextField` has or depend on MDCTextField snapshot test to get the values.
### The fix
Move this file to a more general folder and rename the values to not include _TextField_.

### DIFF
--- a/components/TextFields/tests/snapshot/MDCTextFieldFilledControllerLeadingImageSnapshotTests.m
+++ b/components/TextFields/tests/snapshot/MDCTextFieldFilledControllerLeadingImageSnapshotTests.m
@@ -16,7 +16,6 @@
 
 #import "MDCAbstractTextFieldSnapshotTests+LeadingImage.h"
 #import "MDCAbstractTextFieldSnapshotTests.h"
-#import "MDCSnapshotTestsStrings.h"
 #import "MaterialTextFields.h"
 
 @interface MDCTextFieldFilledControllerLeadingImageSnapshotTests : MDCAbstractTextFieldSnapshotTests
--- a/components/TextFields/tests/snapshot/MDCTextFieldFilledControllerLeadingImageSnapshotTests.m
+++ b/components/TextFields/tests/snapshot/MDCTextFieldFilledControllerLeadingImageSnapshotTests.m
@@ -16,7 +16,7 @@
 
 #import "MDCAbstractTextFieldSnapshotTests+LeadingImage.h"
 #import "MDCAbstractTextFieldSnapshotTests.h"
-#import "MDCTextFieldSnapshotTestsStrings.h"
+#import "MDCSnapshotTestsStrings.h"
 #import "MaterialTextFields.h"
 
 @interface MDCTextFieldFilledControllerLeadingImageSnapshotTests : MDCAbstractTextFieldSnapshotTests
--- a/components/TextFields/tests/snapshot/MDCTextFieldFilledControllerSnapshotTests.m
+++ b/components/TextFields/tests/snapshot/MDCTextFieldFilledControllerSnapshotTests.m
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 #import "MDCAbstractTextFieldSnapshotTests.h"
-#import "MDCSnapshotTestsStrings.h"
 #import "MaterialTextFields.h"
 
 @interface MDCTextFieldFilledControllerSnapshotTests : MDCAbstractTextFieldSnapshotTests
--- a/components/TextFields/tests/snapshot/MDCTextFieldFilledControllerSnapshotTests.m
+++ b/components/TextFields/tests/snapshot/MDCTextFieldFilledControllerSnapshotTests.m
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 #import "MDCAbstractTextFieldSnapshotTests.h"
-#import "MDCTextFieldSnapshotTestsStrings.h"
+#import "MDCSnapshotTestsStrings.h"
 #import "MaterialTextFields.h"
 
 @interface MDCTextFieldFilledControllerSnapshotTests : MDCAbstractTextFieldSnapshotTests
--- a/components/TextFields/tests/snapshot/MDCTextFieldFilledFloatingControllerSnapshotTests.m
+++ b/components/TextFields/tests/snapshot/MDCTextFieldFilledFloatingControllerSnapshotTests.m
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 #import "MDCAbstractTextFieldSnapshotTests.h"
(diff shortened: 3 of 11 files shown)


romoore:
Just curious where "we have to redefine string values similar to what `MDCTextField` has"? Which other tests are redeclaring these strings?   Could you please mention the specific other tests as well?

components/private/Snapshot/MDCSnapshotTestsStrings.h
```diff
@@ -14,30 +14,30 @@
 
 #pragma mark - Latin text
 
-static NSString *const MDCTextFieldSnapshotTestsPlaceholderShortTextLatin = @"P text";
-static NSString *const MDCTextFieldSnapshotTestsPlaceholderLongTextLatin =
+static NSString *const MDCSnapshotTestsPlaceholderShortTextLatin = @"P text";
```
romoore, line 17: These English strings don't really make sense outside of Text Fields usage. These strings (at least the English ones) are unique to their placement within the text field. For example, "P text" represents Placeholder text and "I text" is Input text.  The Arabic strings are just short/long Lorem Ipsum generated at http://generator.lorem-ipsum.info/_arabic . 

I would suggest the following:
*  Copy all the strings into `private/Snapshot` but regenerate the Latin text from Lorem Ipsum.
* In TextFields snapshot strings, keep the current Latin values (they would be unique to Text Fields) and use the `private/Snapshot` strings for Arabic.
* Other components can use the more general Lorem Ipsum values defined in `private/Snapshot` when appropriate and can generate custom strings as needed.

What do you think?

romoore (approved):
Approving, but I'd you to either keep the Text Fields Latin strings unique to Text Fields or if you have a better solution, we can talk through that.

components/TextFields/tests/snapshot/MDCTextFieldFilledControllerLeadingImageSnapshotTests.m
```diff
@@ -16,7 +16,7 @@
 
 #import "MDCAbstractTextFieldSnapshotTests+LeadingImage.h"
 #import "MDCAbstractTextFieldSnapshotTests.h"
-#import "MDCTextFieldSnapshotTestsStrings.h"
```
romoore, line 19: Let's remove this import. It's left over from earlier code.

components/TextFields/tests/snapshot/MDCTextFieldFilledControllerSnapshotTests.m
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 #import "MDCAbstractTextFieldSnapshotTests.h"
-#import "MDCTextFieldSnapshotTestsStrings.h"
```
romoore, line 16: Let's remove this import. It's unused.

components/TextFields/tests/snapshot/MDCTextFieldFilledFloatingControllerSnapshotTests.m
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 #import "MDCAbstractTextFieldSnapshotTests.h"
-#import "MDCTextFieldSnapshotTestsStrings.h"
```
romoore, line 16: Let's remove this import. It's unused.

components/TextFields/tests/snapshot/MDCTextFieldFullWidthCharacterCountSnapshotTests.m
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 #import "MDCAbstractTextFieldSnapshotTests.h"
-#import "MDCTextFieldSnapshotTestsStrings.h"
```
romoore, line 16: Let's remove this import. It's unused.

components/TextFields/tests/snapshot/MDCTextFieldFullWidthControllerLeadingImageSnapshotTests.m
```diff
@@ -14,7 +14,7 @@
 
 #import "MDCAbstractTextFieldSnapshotTests+LeadingImage.h"
 #import "MDCAbstractTextFieldSnapshotTests.h"
-#import "MDCTextFieldSnapshotTestsStrings.h"
```
romoore, line 17: Let's remove this import. It's unused.

components/TextFields/tests/snapshot/MDCTextFieldFullWidthControllerSnapshotTests.m
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 #import "MDCAbstractTextFieldSnapshotTests.h"
-#import "MDCTextFieldSnapshotTestsStrings.h"
+#import "MDCSnapshotTestsStrings.h"
```
romoore: Let's remove this import. It's unused.

components/TextFields/tests/snapshot/MDCTextFieldOutlinedControllerSnapshotTests.m
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 #import "MDCAbstractTextFieldSnapshotTests.h"
-#import "MDCTextFieldSnapshotTestsStrings.h"
+#import "MDCSnapshotTestsStrings.h"
```
romoore: Let's remove this import. It's unused.

romoore (approved):

codeman7:
No other specific test use these but we only have 4 test outside of MDCTextFields, so I am guessing when all other components do more rigorous testing this will be needed.

components/private/Snapshot/MDCSnapshotTestsStrings.h
codeman7: I can agree with that, I was going to write some snapshot test and needed strings so instead of making a new file or declaring new variables I tried to just move these. I think they should all be lorem-ipsum.

romoore:
If we're not working on a specific change that would warrant this refactor, it might be worth holding off on it for now. I've added the last languages text fields should probably test (Cyrillic, Hindi, and Korean), and I suspect for many components, it may be better to use a more general string structure instead of things called "shortInputText" and "longHelperText".

I suspect other components will find things like localized images, text, or even ready-made schemes convenient. All of those could be added gradually as we identify specific needs for them.

PR #6138, opened 2018-12-28
[Snapshot] Move files to src

codeman7:
### Context
In working on #6137 I noticed snapshot test helpers were in the private directory but weren't in a _src_ folder. This doesn't align with any of our other components in the _private_ folder.
### The problem
Snapshot test helpers weren't in a _src_ folder.
### The fix
Move the files into _src_.
### Additional notes
I have removed the BUILD file as we will need to add a BUILD file to [ios-snapshot-test-case](https://github.com/uber/ios-snapshot-test-case) before we can add a BUILD file.


romoore (approved):
LGTM. I cleaned-up the summary description since it included some strikeout text that wouldn't have been useful in the git logs later.

PR #6141, opened 2018-12-28
[TextFields] Add Hindi snapshot tests.

romoore:
Adding snapshot tests for Hindi script across multiple controller styles.
Tests are based on existing Arabic tests and use the same controller styles.

Part of #5762


codeman7:
Similar question to #6147, I am confused why we need to test multiple LTR languages?

codeman7:
Why was the [Snapshot] label removed?

romoore:
The `[Snapshot]` label is for the Snapshot component in `components/private/Snapshot`. We don't currently have a label unique to snapshot tests within other components.

romoore:
LTR and RTL are not the only important things to test. Alignment with baseline of scripts, ascender/descender heights and overlaps with other UI elements, and any potential manual calculation of widths/heights can be impacted by the choice of script.  In the case of Hindi, there are several rendered glyphs that are made up of multiple "characters". Depending on how we perform layout within the components, these different scripts can interact differently with the layout.

rami-a (approved):

PR #6143, opened 2018-12-28
[Slider] Deflake unit tests.

romoore:
Several unit tests were comparing the outcome of arithemtic by using
`XCTAssertEqual`, which requires absolute precision for floating point numbers.
Due to rounding errors, this is neither important nor desirable. Instead,
`XCTAssertEqualWithAccuracy` should be used.

Also corrected the order of the assertion parameters. The test value should
come before the expected value because the failure message has the template,
@"(%@) is not equal to (%@)" and when read it implies that the second value
(sentence object) is the correct value.

Removed some debug messages that managed to creep in there a while back.

### DIFF
--- a/components/Slider/tests/unit/SliderTests.m
+++ b/components/Slider/tests/unit/SliderTests.m
@@ -906,12 +906,12 @@ static const CGFloat kEpsilonAccuracy = (CGFloat)0.001;
 
 - (void)testThumbRadiusDefault {
   // Then
-  XCTAssertEqual(self.slider.thumbRadius, 6);
+  XCTAssertEqualWithAccuracy(self.slider.thumbRadius, 6, kEpsilonAccuracy);
 }
 
 - (void)testThumbElevationDefault {
   // Then
-  XCTAssertEqual(self.slider.thumbElevation, MDCShadowElevationNone);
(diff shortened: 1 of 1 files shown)


components/Slider/tests/unit/SliderTests.m
```diff
@@ -1103,7 +1098,7 @@ - (void)testAccessibilityIncrementDiscreteSlider {
     CGFloat stepValue = (self.slider.maximumValue - self.slider.minimumValue) /
         (self.slider.numberOfDiscreteValues - 1);
     CGFloat expectedValue = self.slider.minimumValue + stepValue;
-    XCTAssertEqualWithAccuracy(self.slider.value, expectedValue, 0.0001,
+    XCTAssertEqualWithAccuracy(self.slider.value, expectedValue, kEpsilonAccuracy,
```
codeman7, line 1101: Why the change from `0.001` to `kEpsilonAccuracy`, I thought `0.001` was okay in tests?

codeman7 (approved):

components/Slider/tests/unit/SliderTests.m
romoore: I was just reusing the accuracy value that was already in-use within the file. It didn't seem like the difference (0.0001 to 0.001) would have any effect on the calculations being performed.

PR #6147, opened 2018-12-28
[TextFields] Adding Korean snapshot tests.

romoore:
Adding Korean text for snapshot tests for all current controller types.

Part of #5762


codeman7:
What is the difference between these tests and the Latin tests?

codeman7 (approved):

romoore:
These use Korean strings in the snapshots instead of English text. By using Arabic text, we've found some layout errors in Outlined controllers (not related to RTL). The idea of using different scripts is to try and identify those errors and/or prevent regressions of the same kind in the future.

codeman7:
I understand the need to test RTL and LTR but I am confused on why we would need to test LTR twice?

PR #6152, opened 2018-12-30
[BottomNavigation] Lays out the bar when titleVisibility is changed

### LINKED ISSUE #6151: [BottomNavigation] Title visibility changes don't happen immediately
When bottomNavigation.titleVisibility is set for an already-displayed MDCBottomNavigationBar, there is no visible change until a tab is selected.

## Reproduction steps

1. Make an app with an MDCBottomNavigationBar.
2. Put a button in your app that toggles bottomNavigation.titleVisibility between .always and .never.

## Expected behavior

When the toggling button is tapped, the title immediately becomes visible or invisible.

## Actual behavior

When the toggling button is tapped, nothing visible happens. The next time a tab is tapped, the title visibility change happens.

This is particularly egregious when switching visibility on tab selection. You can end up with the title and the image both centered, and thus on top of each other.

## Platform

 - Device: Simulated iPhone 5s
 - OS: iOS 12.1

---
#### Internal data
- Associated internal bug: [b/122264214](http://b/122264214)

Ruberik:
Closes #6151.

### DIFF
--- a/components/BottomNavigation/examples/BottomNavigationTitleVisibilityChangeExample.swift
+++ b/components/BottomNavigation/examples/BottomNavigationTitleVisibilityChangeExample.swift
@@ -34,7 +34,7 @@ class BottomNavigationTitleVisibilityChangeExample: UIViewController, MDCBottomN
   }
   
   func commonBottomNavigationTypicalUseSwiftExampleInit() {
-    view.backgroundColor = .lightGray
+    view.backgroundColor = colorScheme.backgroundColor
     view.addSubview(bottomNavBar)
     
     // Always show bottom navigation bar item titles.
@@ -87,6 +87,9 @@ class BottomNavigationTitleVisibilityChangeExample: UIViewController, MDCBottomN
   override func viewWillLayoutSubviews() {
     super.viewWillLayoutSubviews()
     layoutBottomNavBar()
+  }
+  
+  override func viewDidLoad() {
     addInstructionLabel()
   }
   
--- a/components/BottomNavigation/examples/BottomNavigationTitleVisibilityChangeExample.swift
+++ b/components/BottomNavigation/examples/BottomNavigationTitleVisibilityChangeExample.swift
@@ -109,7 +109,7 @@ extension BottomNavigationTitleVisibilityChangeExample {
   
   class func catalogMetadata() -> [String: Any] {
     return [
-      "breadcrumbs": ["Bottom Navigation", "Change title visibility"],
+      "breadcrumbs": ["Bottom Navigation", "Bottom Navigation Title Visibility (Swift)"],
       "primaryDemo": false,
       "presentable": false,
     ]
--- a/components/BottomNavigation/examples/BottomNavigationTitleVisibilityChangeExample.swift
+++ b/components/BottomNavigation/examples/BottomNavigationTitleVisibilityChangeExample.swift
@@ -90,6 +90,7 @@ class BottomNavigationTitleVisibilityChangeExample: UIViewController, MDCBottomN
   }
   
   override func viewDidLoad() {
+    super.viewDidLoad()
     addInstructionLabel()
   }
   
--- a/components/BottomNavigation/examples/BottomNavigationTitleVisibilityChangeExample.swift
+++ b/components/BottomNavigation/examples/BottomNavigationTitleVisibilityChangeExample.swift
@@ -1,0 +1,117 @@
+// Copyright 2018-present the Material Components for iOS authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import MaterialComponents.MaterialBottomNavigation_ColorThemer
+import MaterialComponents.MaterialColorScheme
+
+class BottomNavigationTitleVisibilityChangeExample: UIViewController, MDCBottomNavigationBarDelegate {
+  
+  var colorScheme = MDCSemanticColorScheme()
+  
+  // Create a bottom navigation bar to add to a view.
+  let bottomNavBar = MDCBottomNavigationBar()
+  
+  init() {
+    super.init(nibName: nil, bundle: nil)
+    commonBottomNavigationTypicalUseSwiftExampleInit()
+  }
+  
+  @available(*, unavailable)
+  required init?(coder aDecoder: NSCoder) {
+    super.init(coder: aDecoder)
+  }
+  
+  func commonBottomNavigationTypicalUseSwiftExampleInit() {
+    view.backgroundColor = .lightGray
+    view.addSubview(bottomNavBar)
+    
+    // Always show bottom navigation bar item titles.
+    bottomNavBar.titleVisibility = .always
+    
+    // Cluster and center the bottom navigation bar items.
+    bottomNavBar.alignment = .centered
+    
+    // Add items to the bottom navigation bar.
+    let tabBarItem1 = UITabBarItem(title: "Home", image: UIImage(named: "Home"), tag: 0)
+    let tabBarItem2 =
+      UITabBarItem(title: "Messages", image: UIImage(named: "Email"), tag: 1)
+    let tabBarItem3 =
+      UITabBarItem(title: "Favorites", image: UIImage(named: "Favorite"), tag: 2)
+    bottomNavBar.items = [ tabBarItem1, tabBarItem2, tabBarItem3 ]
+    
+    // Select a bottom navigation bar item.
+    bottomNavBar.selectedItem = tabBarItem2;
+    
+    bottomNavBar.delegate = self
+  }
+  
+  func layoutBottomNavBar() {
+    let size = bottomNavBar.sizeThatFits(view.bounds.size)
+    var bottomNavBarFrame = CGRect(x: 0,
+                                   y: view.bounds.height - size.height,
+                                   width: size.width,
+                                   height: size.height)
+    if #available(iOS 11.0, *) {
+      bottomNavBarFrame.size.height += view.safeAreaInsets.bottom
+      bottomNavBarFrame.origin.y -= view.safeAreaInsets.bottom
+    }
+    bottomNavBar.frame = bottomNavBarFrame
+  }
+  
+  func addInstructionLabel() {
+    let instructionLabel = UILabel()
+    instructionLabel.numberOfLines = 0
+    instructionLabel.lineBreakMode = .byWordWrapping
+    instructionLabel.text = "Choose the Home tab to make all titles disappear, and any other tab to make them reappear."
+    view.addSubview(instructionLabel)
+    let size = instructionLabel.sizeThatFits(view.bounds.size)
+    let instructionFrame = CGRect(x: 0,
+                                  y: view.bounds.height / 2 - size.height / 2,
+                                  width: size.width,
+                                  height: size.height)
+    instructionLabel.frame = instructionFrame
+  }
+  
+  override func viewWillLayoutSubviews() {
+    super.viewWillLayoutSubviews()
+    layoutBottomNavBar()
+    addInstructionLabel()
+  }
+  
+  override func viewWillAppear(_ animated: Bool) {
+    super.viewWillAppear(animated)
+    self.navigationController?.setNavigationBarHidden(true, animated: animated)
+  }
+  
+  func bottomNavigationBar(_ bottomNavigationBar: MDCBottomNavigationBar, didSelect item: UITabBarItem) {
+    if item == bottomNavigationBar.items[0] {
+      bottomNavigationBar.titleVisibility = .never
+    } else {
+      bottomNavigationBar.titleVisibility = .always
+    }
+  }
+}
+
+// MARK: Catalog by convention
+extension BottomNavigationTitleVisibilityChangeExample {
+  
+  class func catalogMetadata() -> [String: Any] {
+    return [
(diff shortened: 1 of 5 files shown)


Ruberik:
FYI, I just updated this change so that the demo is in its own file, instead of in BottomNavigationTypicalUseExample.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
```diff
@@ -383,6 +383,7 @@ - (void)setTitle:(NSString *)title {
 -(void)setTitleVisibility:(MDCBottomNavigationBarTitleVisibility)titleVisibility {
   _titleVisibility = titleVisibility;
   [self updateLabelVisibility];
+  [self centerLayoutAnimated:YES];
```
romoore: Does it not work to call `setNeedsLayout` here instead so any layout related to visibility can be performed in a standard UIKit layout pass?

romoore (changes requested):
Thanks for the PR! Is there a unit test that could be added to test this condition? If not, is there a small snippet that could be turned into a snapshot test?  Perhaps changing the bounds/frame while the titles are not visible?

It would be great to find the code path that causes this bug and make sure we've gotten it squashed.

components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
Ruberik: Yes, that's better. I'm actually going to move that into updateLabelVisibility, since that seems like a better place for that to live.

romoore: Sounds good! 👍

Ruberik:
@romoore Test added! I'll consider this a helpful introduction to Objective-C; all the hair will grow back.

Ruberik:
BTW, while I have your attention on this test file, does testSetSelectedItemTintColorUpdatesInkColor look right to you? It has the same check for item1 and item2, even though one is selected, the other isn't, and the point of the test is to make sure selection changes something.

romoore:
Hi @Ruberik . That test changes the `selectedItemTintColor` on both a selected and unselected item. It then confirms that both have changed.  Since how the Ripple (here called "ink") color's actual value is documented publicly, the test doesn't make any assumptions about what the new color should be - only that it isn't the original color.  The reason for that test (and the related fix in https://github.com/material-components/material-components-ios/pull/4950 was to fix a bug in the implementation and update how Ripple color is applied to Bottom Navigation items.

I could certainly see an improvement being made where we assert that both items have the same (changed) color. I've opened https://github.com/material-components/material-components-ios/pull/6190 to add that check.

Was there something else in the test that you think could be improved?

components/BottomNavigation/examples/BottomNavigationTitleVisibilityChangeExample.swift
```diff
@@ -0,0 +1,117 @@
...
+                                  y: view.bounds.height / 2 - size.height / 2,
+                                  width: size.width,
+                                  height: size.height)
+    instructionLabel.frame = instructionFrame
+  }
+  
+  override func viewWillLayoutSubviews() {
+    super.viewWillLayoutSubviews()
+    layoutBottomNavBar()
+    addInstructionLabel()
```
romoore, line 94: I may be mistaken, but this looks like it creates a new UILabel every time the view is resized or rotated?   Did you mean to put the call to `addInstructionLabel()` in `viewDidLoad`?

components/BottomNavigation/examples/BottomNavigationTitleVisibilityChangeExample.swift
```diff
@@ -0,0 +1,117 @@
...
+    commonBottomNavigationTypicalUseSwiftExampleInit()
+  }
+  
+  @available(*, unavailable)
+  required init?(coder aDecoder: NSCoder) {
+    super.init(coder: aDecoder)
+  }
+  
+  func commonBottomNavigationTypicalUseSwiftExampleInit() {
+    view.backgroundColor = .lightGray
```
romoore: Nit: Please use `colorScheme.backgroundColor` for the background color.

romoore (approved):
LGTM. Just a pair of minor comments about the example.

randallli:
I reran the kokoro CI.

randallli:
error from kokoro (Internal CI):

Test Suite 'BottomNavigationItemViewTests' started at 2019-01-07 07:58:32.626
Test Case '-[BottomNavigationItemViewTests testBadgeCenterIsCorrectWithoutRTL]' started.
Test Case '-[BottomNavigationItemViewTests testBadgeCenterIsCorrectWithoutRTL]' passed (0.002 seconds).
Test Case '-[BottomNavigationItemViewTests testBadgeCenterIsCorrectWithRTL]' started.
Test Case '-[BottomNavigationItemViewTests testBadgeCenterIsCorrectWithRTL]' passed (0.002 seconds).
Test Case '-[BottomNavigationItemViewTests testContentInsetLayout]' started.
Test Case '-[BottomNavigationItemViewTests testContentInsetLayout]' passed (0.016 seconds).
Test Case '-[BottomNavigationItemViewTests testHorizontalMarginLayout]' started.
Test Case '-[BottomNavigationItemViewTests testHorizontalMarginLayout]' passed (0.003 seconds).
Test Case '-[BottomNavigationItemViewTests testSetSelectedItemTintColorUpdatesInkColor]' started.
Test Case '-[BottomNavigationItemViewTests testSetSelectedItemTintColorUpdatesInkColor]' passed (0.002 seconds).
Test Case '-[BottomNavigationItemViewTests testSetTitleVisibilityUpdatesLayout]' started.
components/BottomNavigation/tests/unit/BottomNavigationItemViewTests.m:141: error: -[BottomNavigationItemViewTests testSetTitleVisibilityUpdatesLayout] : ((imageMidY) equal to (32.75) +/- (0.001)) failed: ("32.5") is not equal to ("32.75") +/- ("0.001")

components/BottomNavigation/tests/unit/BottomNavigationItemViewTests.m
```diff
@@ -127,6 +127,28 @@ - (void)testSetSelectedItemTintColorUpdatesInkColor {
...
+  MDCBottomNavigationItemView *view = [[MDCBottomNavigationItemView alloc] init];
+  view.title = @"Test Content";
+  view.image = fakeImage();
+  view.bounds = CGRectMake(0, 0, 100, 100);
+  view.contentVerticalMargin = 20;
+  view.contentHorizontalMargin = 20;
+  view.titleVisibility = MDCBottomNavigationBarTitleVisibilityAlways;
+  [view layoutIfNeeded];
+  CGFloat imageMidY = CGRectGetMidY(view.iconImageView.frame);
+  XCTAssertEqualWithAccuracy(imageMidY, 32.75, 0.001);
```
romoore: Tip: Consider making this a less-than-equal and greater-than-equal so you can verify the "approximate" location.  It seems like we don't care too much about the exact placement, but rather only that it is "to the left of center" by some reasonable amount.

components/BottomNavigation/examples/BottomNavigationTitleVisibilityChangeExample.swift
Ruberik: Thank you for spotting that. Moved to viewDidLoad().

components/BottomNavigation/examples/BottomNavigationTitleVisibilityChangeExample.swift
Ruberik: Fixed, here and in other BottomNav Swift examples.

components/BottomNavigation/tests/unit/BottomNavigationItemViewTests.m
Ruberik: Good point. I appreciate you giving a broader view; I admit I was happy just to get a passing test that wouldn't have passed before.

Ruberik:
I'll preface this by saying that I'm a Git noob who, so far, has managed to make things... sort of work.

I've made the changes, but I can't seem to get them on github.com/Ruberik/material-components-ios. Here's what happened:

1. When I created this PR, I (unwisely?) made it in my own `develop` branch.
2. When I rebase my local `develop` branch onto material-components' `develop` branch, it works fine. When I try to force push (is that really what I should be doing?) onto github.com/Ruberik/material-components-ios's `develop` branch, here's what I get. The commandline is coming from Sourcetree, for what it's worth:

```
git --no-optional-locks -c color.branch=false -c color.diff=false -c color.status=false -c diff.mnemonicprefix=false -c core.quotepath=false -c credential.helper=sourcetree -c lfs.customtransfer.bitbucket-media-api.path=/Applications/Sourcetree.app/Contents/Resources/git_local/git-lfs/git-lfs-bitbucket-media-api push -v --force origin refs/heads/develop:refs/heads/develop 
Pushing to https://github.com/Ruberik/material-components-ios.git
open /Users/Bartholomew/dev/material-components-ios/snapshot_test_goldens/goldens_64/MDCTextAreaOutlinedControllerBaselineCharacterCountArabicSnapshotTests/testTextFieldEmptyDisabled_11_2@2x.png: no such file or directory
error: failed to push some refs to 'https://github.com/Ruberik/material-components-ios.git'
Completed with errors, see above
```

Sorry to drag you into my own personal Git troubles.

romoore:
Can you try  `git lfs fetch --all` and then push?  Let me know if that helps.

romoore:
In general, if you found you've created a commit in the wrong branch, you can do the following:

1. Use `git log` to find the commits since you branched from upstream's `develop`. Grab the commit hash.
2. Use `git reset --soft <COMMIT HASH>` to undo ALL of your work and have it sitting as if it's just been modified on the file system.
3. Use `git stash` to put all modified files (but not newly-created files) into the stash
4. Use `git checkout -b my-new-branch` to create the new branch, `my-new-branch`
5. Use `git stash apply` to unstash your changes and put them in the working directory
6. Commit your changes to the new branch
7. Push (forced?) to the existing remote branch or push to a new remote branch (and open a new PR).

As clean-up, you'll also want to do `git stash drop` to drop the latest stash entry. **Be careful!**  If you aren't sure, you can always `git stash list` to take a look at what's in your stash.

romoore:
Oops, I had the wrong `git lfs` command above. It's been correct to spell "all" correctly.

components/BottomNavigation/tests/unit/BottomNavigationItemViewTests.m
```diff
@@ -127,6 +127,27 @@ - (void)testSetSelectedItemTintColorUpdatesInkColor {
...
+  [view layoutIfNeeded];
+  CGFloat imageMidYWithTitle = CGRectGetMidY(view.iconImageView.frame);
+
+  // When
+  view.titleVisibility = MDCBottomNavigationBarTitleVisibilityNever;
+  [view layoutIfNeeded];
+
+  // Then
+  CGFloat imageMidYWithoutTitle = CGRectGetMidY(view.iconImageView.frame);
+  XCTAssertGreaterThan(imageMidYWithoutTitle, imageMidYWithTitle + 1);  // We add 1 to make sure the change in position is greater than a rounding error. Practically, it's likely to be greater than 10.
```
romoore: Nit: This line looks a bit long. Consider breaking it up or running clang-format on your changes.

romoore (dismissed):

components/BottomNavigation/tests/unit/BottomNavigationItemViewTests.m
Ruberik: Broken up and clang-formatted.

Ruberik:
Thanks! That `git lfs` command fixed me up nicely. I believe the currently-pushed thing in my fork is properly rebased and has addressed all comments.

Thanks for all the help writing this. I've definitely become a more competent git user after this.

components/BottomNavigation/examples/BottomNavigationTitleVisibilityChangeExample.swift
```diff
@@ -0,0 +1,120 @@
...
+    instructionLabel.frame = instructionFrame
+  }
+  
+  override func viewWillLayoutSubviews() {
+    super.viewWillLayoutSubviews()
+    layoutBottomNavBar()
+  }
+  
+  override func viewDidLoad() {
+    addInstructionLabel()
```
romoore, line 94: Did you forget to call `super.viewDidLoad()`?

Ruberik: I did! Did a lint pick that up, or did you spot it?

romoore:
;_; Looks like the force-push lost the commit history.  I'll take another review at this today.

components/BottomNavigation/examples/BottomNavigationTitleVisibilityChangeExample.swift
romoore: That was just me. I've often made the same mistake in the past, so I try to catch it where I can. ;-)

Ruberik:
Aww! Sorry. Dumb question: is there a better way to handle catching up with your `develop` branch than rebasing my changes in my local repository, then force pushing them? I tried to find a way to directly rebase my GitHub repository, but couldn't find one.

romoore:
If you do your work on a local branch, you can then `git merge develop` to pull changes into your development branch.  Here's my typical workflow:

1. `git checkout -b issue-1234`
2. Do work....
3. `git commit`
4. `git push origin issue-1234`
5. `git checkout develop`
6. `git pull upstream develop` (this keeps my local `develop` fast-forwarded with changes upstream)
7. `git push origin develop`
7. `git checkout issue-1234`
8. `git merge develop`  (resolve any conflicts)
9. `git push origin issue-1234`

I will repeat 2-9 for any additional work, feedback, etc.  If a PR is still sitting in GitHub in the morning, I try to do a merge from `develop` before doing any more work.  I just opened https://github.com/material-components/material-components-ios/issues/6280 to reduce the number of errors not caught by outdated PRs.

Ruberik:
Thanks for explaining that! All the guides I could find seemed to assume nothing was happening to the repository other than the reader's changes, and skipped 4-10. Is there a Google guide to GitHub workflow somewhere external? Or a good guide somewhere that you used to learn this stuff?

romoore (approved):

romoore:
I'm not sure, but Atlassian has a good description of the differences.

https://www.atlassian.com/git/tutorials/merging-vs-rebasing

Issue #5825, opened 2018-11-26
[Cocoapods beta] pod try unable to install


ykws:
Unable to install the `MaterialComponents` pod, when selections are 1 and 2 .
Now no longer support them?

The selection's 3, 4 and 5 had succeed.

## Reproduction steps

```
$ pod try MaterialComponents
Updating spec repositories
warning: inexact rename detection was skipped due to too many files.
warning: you may want to set your diff.renameLimit variable to at least 4282 and retry the command.

Trying MaterialComponents
1: catalog/MDCDragons.xcodeproj
2: catalog/MDCCatalog.xcworkspace
3: demos/Bare/Bare.xcworkspace
4: demos/Pesto/Pesto.xcworkspace
5: demos/Shrine/Shrine.xcworkspace
Which project would you like to open
1
Performing CocoaPods Installation
Fetching podspec for `MaterialCatalog` from `MaterialCatalog/`
Fetching podspec for `MaterialComponents` from `../`
Fetching podspec for `MaterialComponentsAlpha` from `../`
Fetching podspec for `MaterialComponentsExamples` from `../`
Installing CatalogByConvention (2.5.1)
Installing MDFInternationalization (1.1.0)
Installing MDFTextAccessibility (1.2.0)
Installing MaterialCatalog (70.0.0)
Installing MaterialComponents (70.0.0)
Installing MaterialComponentsAlpha (70.0.0)
Installing MaterialComponentsExamples (70.0.0)
Installing MotionAnimator (2.8.1)
Installing MotionInterchange (1.6.0)
Installing MotionTransitioning (5.0.0)
[!] Unable to install the `MaterialComponents` pod, because the `MaterialComponents-Unit-private-Application-tests-unit` target in Xcode would have no sources to compile.
```

## Platform

 - OS: Mac OS Mojave 10.14.1
 - CocoaPods: 1.6.0.beta.2

---
#### Internal data
- Associated internal bug: [b/121041042](http://b/121041042)

ykws:
@material-automation I had set title, would you check again?

PR #6155, opened 2018-12-30
[Documentation] Adds a "getting started" section to contributing/README.md

Ruberik:
Thanks to @codeman7 for his help getting me started on #6151. I found this hard enough to do on my own that I thought others might have trouble too. I realize there are many questions of taste that go into this sort of documentation:
- Does it belong here, or somewhere else?
- Are any of the steps too detailed, or not detailed enough?
- Does it start too early? In other words, would anybody who's reading this actually need the first few steps?
- Should I mention how to work around #5825 as a Known Issue, or should I assume that will get fixed soon enough that it isn't worth mentioning?

I'm assuming I don't have this written *exactly* the way you'd like it, so this PR is intended to start a discussion.

### DIFF
--- a/contributing/README.md
+++ b/contributing/README.md
@@ -67,16 +67,16 @@ To find an issue to work on, filter the issues list by the ["is:fixit" label](ht
 If you're new to iOS development and have a computer running OS X, these steps should help you get all the prerequisites. More experienced users may just need to see the last steps.
 
 1. We assume you already have installed on you mac
-  1. XCode and the Xcode Command Line Tools which comes with [git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git).
-  1. [cocoapods](https://guides.cocoapods.org/using/getting-started.html#installation).
+    1. XCode and the Xcode Command Line Tools which comes with [git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git).
+    1. [cocoapods](https://guides.cocoapods.org/using/getting-started.html#installation).
 1. Install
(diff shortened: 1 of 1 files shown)


randallli:
Thanks for the PR that tries to fix our confusing docs. Clearly our docs are missing a couple things because we had to give you more explicit instructions than our existing docs. 

> Does it belong here, or somewhere else?

Some of this is elsewhere already. I would like to use links to our docs where we talk about it as well as GitHub references that explain things too. I will inline my thoughts.
> Are any of the steps too detailed, or not detailed enough?

We may be able to convert some of the details to links to GitHub help docs. (I will inline)
> Does it start too early? In other words, would anybody who's reading this actually need the first few steps?

Some things we can assume the user has already installed. Xcode and git to name a few. I will inline exactly which ones i feel can be omitted.
> Should I mention how to work around #5825 as a Known Issue, or should I assume that will get fixed soon enough that it isn't worth mentioning?

No need to call this out just yet. I believe @andrewoverton put in a fix for. We can confirm its status in the issue.

randallli:
Instead of in lining the comments I tried to do a PR on top of yours but accidentally merged the changes directly into your branch. If your ok with my additions we can merge this in.

randallli (approved):

Ruberik:
LGTM. That looks way better. Thank you!

randallli:
I reran the kokoro CI.

Ruberik:
Thanks for taking this over and making it happen, @randallli!

PR #6165, opened 2019-01-02
[TextFields] Truncate helper/error text.

romoore:
When helper or error text is too long to fit within the bounds of the text
input, it should be truncated instead of running outside the frame.

Part of #5762


rami-a (approved):

romoore:
Running internal tests to identify potential breakages. Googlers can check cl/227528742 to follow the progress.

PR #6167, opened 2019-01-02
[ChipTextField] Use MDCChipTextFieldScrollView to replace leftView functionality for flexibility and extensibility (proof of concept)

fluxminds:
This PR is dependent on https://github.com/material-components/material-components-ios/pull/6035. And there is no new API change on MDCChipTextField class level.

This PR is meant to solve the problem mentioned in https://github.com/material-components/material-components-ios/pull/6035 -> "Enabling scrolling of the Chips and the remaining alignment issues will be resolved in followup PRs." 
The issues it solves:
(1) Alignment (a way to adjust it also)
(2) clear button
(3) the coupling issue with apple's leftView in the original PR.

Some remaining problem with the original issues (Will be updated in the TODO in the code too):
(1) The chip removal indentation doesn't work well (Solvable by adding an additional constraint).
(2) The scrolling behavior of Chips was not added (Shouldn't be hard to add on this base - 1 day workload, but needs confirmation from design, it's probably not a feature we want to support right now)
(3) Fix a hack number added on https://github.com/material-components/material-components-ios/pull/6167/files#diff-6c02a09694ee9585406c9dc9da7621efR117.
(4) A known bug where the space between text and clear button is uncertain when the chip is very long, it should be automatically resolved after solving (3).
(5) RTL is not the purpose of this PR. It will be included in https://github.com/material-components/material-components-ios/pull/6035
(6) Screenshot Testing is not the purpose of this PR, but it will be very helpful. Something to work on after solving (1) and (3).

Screenshots:
simulator screen shot - iphone xs - 2019-01-02 at 11 30 50
simulator screen shot - iphone xs - 2019-01-02 at 11 30 54
simulator screen shot - iphone xs - 2019-01-02 at 11 30 57


components/ChipTextField/examples/MDCChipTextField.m
```diff
@@ -29,144 +37,254 @@ @implementation MDCChipTextField
...
+                                                                                          toItem:self
+                                                                                       attribute:NSLayoutAttributeBottom
+                                                                                      multiplier:1
+                                                                                        constant:0];
+  [NSLayoutConstraint activateConstraints:@[chipContainerViewConstraintTop, chipContainerViewConstraintBottom]];
+  [self updateChipViewLeadingConstraints];
+  [self updateChipViewTrailingConstraints];
+}
+
+- (void)handleTextFieldReturnWithText:(NSString *)text {
```
rami-a: I think we should keep the public API specific to appending a chip and not use the wording of return since the user can choose when to append a chip.

components/ChipTextField/examples/MDCChipTextField.m
```diff
@@ -13,14 +13,22 @@
 // limitations under the License.
 
 #import "MDCChipTextField.h"
+
 #import "MaterialChips.h"
+#import "MDCChipTextFieldScrollView.h"
+
+//static CGFloat const kChipsSpacing = 10.0f;
```
rami-a: Should we remove this?

components/ChipTextField/examples/MDCChipTextField.m
```diff
@@ -29,144 +37,254 @@ @implementation MDCChipTextField
...
+             action:@selector(chipTextFieldTextDidChange)
+   forControlEvents:UIControlEventEditingChanged];
   }
   return self;
 }
 
+- (void)setupChipsContainerView {
+  MDCChipTextFieldScrollView *chipsContainerView = [[MDCChipTextFieldScrollView alloc] initWithFrame:CGRectZero];
+  chipsContainerView.translatesAutoresizingMaskIntoConstraints = NO;
+//  chipsContainerView.chipSpacing = kChipsSpacing;
```
rami-a: Should we remove this?

components/ChipTextField/examples/MDCChipTextField.m
```diff
@@ -29,144 +37,254 @@ @implementation MDCChipTextField
...
-  CGRect textRect = [self textRectForBounds:self.bounds];
+  CGRect textRect = [self editingRectForBounds:self.bounds];
   UITextRange *textRange = [self textRangeFromPosition:self.beginningOfDocument
                                             toPosition:self.endOfDocument];
   CGRect inputRect = [self firstRectForRange:textRange];
+  // HACK: use a constant here
+  CGFloat textHorizontalDelta = textRect.size.width - inputRect.size.width - 17;
+  CGFloat offsetAdjustment = textHorizontalDelta;
+//  // HACK: this assumption is also bad
+//  UIScrollView *fieldEditor = nil;
```
rami-a: Should we remove all of this?

components/ChipTextField/examples/MDCChipTextFieldScrollView.m
```diff
@@ -0,0 +1,199 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import "MDCChipTextFieldScrollView.h"
+
+@interface MDCChipTextFieldScrollView ()
+
+@property (nonatomic, readwrite, weak) UIView *contentView;
```
rami-a: I think these three properties should all be strong not weak.

components/ChipTextField/examples/MDCChipTextField.m
```diff
@@ -13,14 +13,22 @@
...
+//static CGFloat const kChipsSpacing = 10.0f;
 
-@interface MDCChipTextField ()
+@interface MDCChipTextField () <MDCChipTextFieldScrollViewDataSource>
 
-@property(nonatomic, strong) UIView *chipsView;
-@property(nonatomic) CGFloat insetX;
-@property(nonatomic, strong) NSLayoutConstraint *leadingConstraint;
-@property(nonatomic, strong) NSMutableArray<MDCChipView *> *chips;
+@property (nonatomic, weak) MDCChipTextFieldScrollView *chipsContainerView;
```
rami-a: I think this should be strong not weak.

components/ChipTextField/examples/MDCChipTextField.m
```diff
@@ -13,14 +13,22 @@
...
+@property (nonatomic) CGFloat chipsContainerHorizontalOffsetForTextEditing;
+@property (nonatomic) CGFloat chipsContainerHorizontalWidthForTextEditing;
+
+@property (nonatomic, readwrite, weak) NSLayoutConstraint *chipContainerViewConstraintLeading;
```
rami-a: I think this should be strong not weak.

components/ChipTextField/examples/MDCChipTextField.m
fluxminds: Addressed in efcb81c4eed97ffe93b9378271fa5c91d227a63a. 

I added the support for the spacing between chips back, but set the value to be 0 for now, since removing this influence factor would be easier for us to debug before all other alignment issues are fixed.

components/ChipTextField/examples/MDCChipTextField.m
fluxminds: Addressed in efcb81c4eed97ffe93b9378271fa5c91d227a63a.

components/ChipTextField/examples/MDCChipTextField.m
fluxminds: Addressed in 233e4691a7c8f02864dbca983d6d6203b87caf50 to make the comment clear. We probably need to follow up and solve the problem here. Leaving the code here would be good for reference for now since it is in Alpha.

components/ChipTextField/examples/MDCChipTextFieldScrollView.m
fluxminds: Addressed in 6d8a1c1dca231edfe812462981d84d9f003d47fa.

components/ChipTextField/examples/MDCChipTextField.m
fluxminds: Addressed in 6d8a1c1dca231edfe812462981d84d9f003d47fa

components/ChipTextField/examples/MDCChipTextField.m
fluxminds: Discussed offline. Will keep the properties for constraints to be `weak`.

components/ChipTextField/examples/MDCChipTextField.m
fluxminds: Addressed in 4f0eaed1b2831b8095d6e612a53c35a810deb668

components/ChipTextField/examples/MDCChipTextField.m
romoore: For posterity (and those of us in the audience), could you briefly explain why these should be `weak`?

components/ChipTextField/examples/MDCChipTextFieldScrollView.h
```diff
@@ -0,0 +1,48 @@
...
+
+#import <UIKit/UIKit.h>
+
+#import "MaterialChips.h"
+
+@class MDCChipTextFieldScrollView;
+
+@protocol MDCChipTextFieldScrollViewDelegate <NSObject, UIScrollViewDelegate>
+
+//TODO: delegate methods to handle touch event
```
galiak11: Is this todo in our sprint/future todo list in buganzier?

components/ChipTextField/examples/MDCChipTextFieldScrollView.h
```diff
@@ -0,0 +1,48 @@
...
+@interface MDCChipTextFieldScrollView : UIScrollView
+
+@property (nonatomic, weak, nullable) id <MDCChipTextFieldScrollViewDelegate> delegate;
+@property (nonatomic, weak, nullable) id <MDCChipTextFieldScrollViewDataSource> dataSource;
+
+@property(nonatomic) CGFloat chipSpacing;
+
+- (void)scrollToRight;
+
+// TODO: remove these methods
```
galiak11: Is this in our "todo/future" list on buganizer?

components/ChipTextField/examples/MDCChipTextFieldScrollView.m
```diff
@@ -0,0 +1,200 @@
...
+}
+
+- (void)commonInit {
+  // set up content view
+  UIView *contentView = [[UIView alloc] initWithFrame:self.frame];
+  contentView.translatesAutoresizingMaskIntoConstraints = NO;
+  [self addSubview:contentView];
+  self.contentView = contentView;
+
+  NSLayoutConstraint *contentConstraintTop = [NSLayoutConstraint constraintWithItem:contentView
```
galiak11: Can you run clang formatter? It's hard to review these lines with a laptop screen size...

components/ChipTextField/examples/MDCChipTextFieldScrollView.m
```diff
@@ -0,0 +1,200 @@
...
+                                                                              attribute:NSLayoutAttributeCenterY
+                                                                              relatedBy:NSLayoutRelationEqual
+                                                                                 toItem:chipView
+                                                                              attribute:NSLayoutAttributeCenterY
+                                                                             multiplier:1
+                                                                               constant:0];
+  
+  NSLayoutConstraint *chipViewConstraintLeading = nil;
+  NSInteger numberOfChips = [self.dataSource numberOfChipsOnScrollView:self];
+  if (numberOfChips == 0) {
```
galiak11, line 154: This assumes that adding the left most chip is always done before all other chips are added. Is that a valid assumption in all use cases?

components/ChipTextField/examples/MDCChipTextFieldScrollView.m
```diff
@@ -0,0 +1,200 @@
...
+
+  // Set fixed values for this scroll view.
+  [self setShowsVerticalScrollIndicator:NO];
+  [self setShowsHorizontalScrollIndicator:NO];
+  self.bounces = NO;
+  self.scrollEnabled = NO;
+}
+
+- (void)appendChipView:(MDCChipView *)chipView {
+  self.trailingConstraint.active = NO;
```
galiak11, line 123: Why not just remove the constraint, instead of deactivating it?

components/ChipTextField/examples/MDCChipTextFieldScrollView.h
fluxminds: Yes, it is meant for the scrolling feature and touch handling events.

components/ChipTextField/examples/MDCChipTextFieldScrollView.m
fluxminds: Addressed in cac37a9518c71227ae5869aa8e5955ec23255c04

components/ChipTextField/examples/MDCChipTextFieldScrollView.m
fluxminds: It seems setting the `active` is encouraged post iOS8 referencing from https://developer.apple.com/documentation/uikit/uiview/1622659-removeconstraint?language=objc

components/ChipTextField/examples/MDCChipTextFieldScrollView.m
fluxminds: I think it would not be a correct assumption once we support inserting chip. But for now I think it is okay? I can add a comment here if that would make it clear.

components/ChipTextField/examples/MDCChipTextFieldScrollView.h
fluxminds: It is not. It will be a refactor issue. I can add this one.

components/ChipTextField/examples/MDCChipTextFieldScrollView.m
galiak11: Yes. I think adding a comment would make it clear.

components/TextFields/src/MDCTextField.m
```diff
@@ -537,7 +545,7 @@ - (CGRect)textRectForBounds:(CGRect)bounds {
     clearButtonWidth += 2 * MDCTextInputClearButtonImageBuiltInPadding;
 
     // Clear buttons are only shown if there is entered text or programatically set text to clear.
-    if (self.text.length > 0) {
+    if (self.hasTextContent > 0) {
```
galiak11: hasTextContent is a BOOL - no need to compare with 0 (appears in multiple places).
Tip: "isEmpty" might be a more conventional name for this.

components/TextFields/src/private/MDCTextInputCommonFundament.m
```diff
@@ -593,7 +598,7 @@ - (void)clearButtonDidTouch {
     }
   }
 
-  self.text = nil;
+  [self.textInput clearText];
```
galiak11, line 629: Why clear the text in a "clearButtonAlpha" method ? Looks like a code smell

components/ChipTextField/examples/MDCChipTextFieldScrollView.h
```diff
@@ -0,0 +1,48 @@
...
+- (MDCChipView *)scrollView:(MDCChipTextFieldScrollView *)scrollView chipForIndex:(NSInteger)index;
+
+@end
+
+@interface MDCChipTextFieldScrollView : UIScrollView
+
+@property(nonatomic, weak, nullable) id<MDCChipTextFieldScrollViewDelegate> delegate;
+@property(nonatomic, weak, nullable) id<MDCChipTextFieldScrollViewDataSource> dataSource;
+
+@property(nonatomic) CGFloat chipSpacing;
```
galiak11, line 43: Is this space between chips? Perhaps adding a comment?
Tip: Should this be added to the API, so it's configurable by themers?

components/ChipTextField/examples/MDCChipTextFieldScrollView.m
```diff
@@ -0,0 +1,223 @@
...
+                                    constant:0];
+  NSLayoutConstraint *contentConstraintWidth =
+      [NSLayoutConstraint constraintWithItem:contentView
+                                   attribute:NSLayoutAttributeWidth
+                                   relatedBy:NSLayoutRelationEqual
+                                      toItem:self
+                                   attribute:NSLayoutAttributeWidth
+                                  multiplier:1
+                                    constant:0];
+  contentConstraintWidth.priority = UILayoutPriorityDefaultLow;
```
galiak11, line 98: What's the reason for the lower priority?

components/ChipTextField/examples/MDCChipTextFieldScrollView.m
```diff
@@ -0,0 +1,223 @@
...
+  [NSLayoutConstraint activateConstraints:@[
+    contentConstraintTop, contentConstraintBottom, contentConstraintLeading,
+    contentConstraintTrailing, contentConstraintHeight, contentConstraintWidth
+  ]];
```
galiak11, line 102: Why activating the constraints? I also don't see where these are being added to the view hierarchy.
Perhaps adding a comment explaining the constraint logic?

components/TextFields/src/MDCTextField.m
fluxminds: Oh. I missed it when changing it I think.

components/TextFields/src/private/MDCTextInputCommonFundament.m
fluxminds: This is for overwriting in the subclass.

components/ChipTextField/examples/MDCChipTextFieldScrollView.m
fluxminds: Activating the constraints automatically add the constraints to the correct views.

components/ChipTextField/examples/MDCChipTextFieldScrollView.m
fluxminds: Discussed offline. Right now, it is not a required constraint so the containerView's trailing can satisfy the one restrict its width by textfield width.

components/ChipTextField/examples/MDCChipTextFieldScrollView.h
fluxminds: Good point. Comments added in 8668014bafcee88f34296598f39b87de73aa958b

components/ChipTextField/examples/MDCChipTextFieldScrollView.m
fluxminds: Addressed in 8668014bafcee88f34296598f39b87de73aa958b

components/TextFields/src/MDCTextField.m
fluxminds: Addressed in 8668014bafcee88f34296598f39b87de73aa958b

components/ChipTextField/examples/ChipTextFieldExampleViewController.swift
```diff
@@ -0,0 +1,126 @@
...
+
+  override init(nibName nibNameOrNil: String?, bundle nibBundleOrNil: Bundle?) {
+    chipInputController = MDCTextInputControllerOutlined(textInput: chipTextField)
+    chipInputController.placeholderText = "With Chips"
+    inputController = MDCTextInputControllerOutlined(textInput: textField)
+    inputController.placeholderText = "Regular MDCTextField"
+    super.init(nibName: nibNameOrNil, bundle: nibBundleOrNil)
+  }
+
+  required init?(coder aDecoder: NSCoder) {
```
galiak11: Tip: A cleaner Approach in swift to initialization (in general and of view controllers in particular) is to use self-executing closures to initialize properties instead of doing that in an init. (ie: lazy initializers for "chipInputController" and "inputController"). Then you will not need any initializers for the view controller.

components/ChipTextField/examples/ChipTextFieldExampleViewController.swift
```diff
@@ -0,0 +1,126 @@
...
+      chipTextField.leadingAnchor.constraint(equalTo: guide.leadingAnchor, constant: 20.0).isActive = true
+      chipTextField.trailingAnchor.constraint(equalTo: guide.trailingAnchor, constant: -20.0).isActive = true
+      chipTextField.topAnchor.constraint(equalTo: guide.topAnchor, constant: 40.0).isActive = true
+    } else if #available(iOSApplicationExtension 9.0, *) {
+      chipTextField.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: 20.0).isActive = true
+      chipTextField.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -20.0).isActive = true
+      chipTextField.topAnchor.constraint(equalTo: view.topAnchor, constant: 40.0).isActive = true
+    } else {
+      // Fallback on earlier versions
+      print("This example is supported on iOS version 9 or later.")
```
galiak11: This example should support all supported iOS versions.

components/ChipTextField/examples/MDCChipTextField.m
```diff
@@ -0,0 +1,360 @@
...
+  [self.chipsContainerView appendChipView:chipView];
+
+  // recalculate the layout to get a correct chip frame values
+  [self.chipsContainerView layoutIfNeeded];
+  [self.chipViews addObject:chipView];
+
+  [self clearChipsContainerOffsetWithConstant:kTextToEnterPlaceholderLength];
+}
+
+// TODO: the constant here reflects the margin, places calling this method need to be refactored.
```
galiak11, line 111: Is this in our "todo"?

components/ChipTextField/examples/MDCChipTextField.m
```diff
@@ -0,0 +1,360 @@
...
+  [self layoutIfNeeded];
+  [self.chipsContainerView setNeedsLayout];
+  [self.chipsContainerView layoutIfNeeded];
+  [self.chipsContainerView scrollToRight];
+}
+
+#pragma mark - Constraints
+
+- (void)updateConstraints {
+  // TODO: This is not optimized for performance, but due to how MDCTextInputController works, we
```
galiak11, line 153: Not clear what's the "todo" (the action that needs to be taken) here. Can you explain? Also, please add to our "todo" list if there's one.

components/ChipTextField/examples/MDCChipTextField.m
```diff
@@ -0,0 +1,360 @@
...
+static CGFloat const kChipsSpacing = 0.0f;
+static CGFloat const kTextToEnterPlaceholderLength = 16.0f;
+
+@interface MDCChipTextField () <MDCChipTextFieldScrollViewDataSource>
+
+@property(nonatomic, strong) MDCChipTextFieldScrollView *chipsContainerView;
+
+@property(nonatomic, readwrite, weak) NSLayoutConstraint *chipContainerViewConstraintLeading;
+@property(nonatomic, readwrite, weak) NSLayoutConstraint *chipContainerViewConstraintTrailing;
+@property(nonatomic) CGFloat chipContainerViewConstraintTrailingConstant;
```
galiak11, line 29: Why is there both chipContainerViewConstraintTrailingConstant and kChipsSpacing? Aren't they the same thing?

components/ChipTextField/examples/MDCChipTextField.m
```diff
@@ -0,0 +1,360 @@
...
+}
+
+- (void)textFieldDidEndEditingWithNotification:(NSNotification *)notification {
+  [self clearChipsContainerOffsetWithConstant:0.0f];
+  [self.chipsContainerView scrollToLeft];
+}
+
+#pragma mark - MDCChipTextFieldScrollViewDataSource
+
+- (NSInteger)numberOfChipsOnScrollView:(MDCChipTextFieldScrollView *)scrollView {
```
galiak11: Tip: May "in" be better than "on" in this case?

components/ChipTextField/examples/MDCChipTextFieldScrollView.h
```diff
@@ -0,0 +1,48 @@
...
+
+#import <UIKit/UIKit.h>
+
+#import "MaterialChips.h"
+
+@class MDCChipTextFieldScrollView;
+
+@protocol MDCChipTextFieldScrollViewDelegate <NSObject, UIScrollViewDelegate>
+
+// TODO: delegate methods to handle touch event
```
galiak11, line 23: Please add to our "todo" list.

components/ChipTextField/examples/MDCChipTextFieldScrollView.m
```diff
@@ -0,0 +1,223 @@
...
+  self.bounces = NO;
+  self.scrollEnabled = NO;
+}
+
+- (void)appendChipView:(MDCChipView *)chipView {
+  self.trailingConstraint.active = NO;
+  [self.contentView addSubview:chipView];
+  [chipView setContentHuggingPriority:UILayoutPriorityRequired
+                              forAxis:UILayoutConstraintAxisHorizontal];
+  [chipView setContentCompressionResistancePriority:UILayoutPriorityRequired
```
galiak11: Are these changing priority?

components/TextFields/src/private/MDCTextInputCommonFundament.m
galiak11: I'm not sure I understand the response. My concern was that it is an unexpected side effect (the field being cleared when clearing the button's alpha). Is that not the case?

components/ChipTextField/examples/MDCChipTextField.m
fluxminds: chipContainerViewConstraintTrailingConstant is used for the trailing space of containerView, which is different from the spacing between chipViews.

components/ChipTextField/examples/MDCChipTextFieldScrollView.h
fluxminds: This one belongs to ticket: https://github.com/material-components/material-components-ios/issues/6208

components/TextFields/src/private/MDCTextInputCommonFundament.m
fluxminds: Communicated offline. The `clearText` exists in `clearButtonDidTouch`

components/ChipTextField/examples/ChipTextFieldExampleViewController.swift
fluxminds: This one has been ticketed in the epic..

components/ChipTextField/examples/MDCChipTextField.m
fluxminds: It is a code refactor but the comment here is not clear enough. Going to add more explanation here:)

components/ChipTextField/examples/MDCChipTextField.m
fluxminds: Addressed in 0f433959bc04c76f0573eef55b8d2c880d5caea7.

components/ChipTextField/examples/MDCChipTextField.m
fluxminds: Addressed in 0f433959bc04c76f0573eef55b8d2c880d5caea7.

components/ChipTextField/examples/ChipTextFieldExampleViewController.swift
fluxminds: Seems there is a problem with the github diff. Going to rebase this branch and check the file diff there.

galiak11 (approved):

components/ChipTextField/examples/MDCChipTextField.m
fluxminds: Good point. Addressed in327c22b99d72535ea92b932b3286dbf257e542bb. No new constraints are created in this method now.

romoore:
This was reverted in https://github.com/material-components/material-components-ios/pull/6346 because it does not compile with bazel.

components/TextFields/src/MDCTextInput.h
```diff
@@ -201,6 +201,12 @@ typedef NS_ENUM(NSUInteger, MDCTextInputTextInsetsMode) {
 /** The underline view */
 @property(nonatomic, nullable, strong, readonly) MDCTextInputUnderlineView *underline;
 
+/** A boolean value that indicates whether the text input contains content **/
```
romoore, line 204: If this goes back into TextFields, please break these changes out into separate PRs. This broke internally and doesn't compile at all with bazel.

PR #6169, opened 2019-01-02
[Collections] Add delegate method to check whether swiping should be limited to one direction.

randallli:
[Collections] Add delegate method to check whether swiping should be limited to one direction.


components/NavigationDrawer/examples/BottomDrawerExpandFullscreenExample.swift
```diff
@@ -17,7 +17,7 @@ import MaterialComponents.MaterialBottomAppBar
 import MaterialComponents.MaterialBottomAppBar_ColorThemer
 import MaterialComponents.MaterialColorScheme
 import MaterialComponents.MaterialNavigationDrawer
-import MaterialComponents.MaterialNavigationDrawer_ColorThemer
```
rami-a: These changes appear unrelated, perhaps they should be split out into a separate PR

contributing/bazel_kokoro.md
```diff
@@ -117,22 +117,19 @@ If your component depends on non-source resources, you can add a bundle target t
 with the following additions:
 
 ```
+load("//tools/build_defs/apple:resources.bzl", "apple_bundle_import")
+
 mdc_public_objc_library(
     name = "ComponentName",
-    bundles = [":Bundle"],
```
rami-a, line 122: These changes appear unrelated, perhaps they should be split out into a separate PR

.gitattributes
```diff
@@ -1 +1,2 @@
-snapshot_test_goldens/**/*.png filter=lfs diff=lfs merge=lfs -text
+# Do not add git-lfs attributes to this files. It causes problems with Cocoapods.
```
rami-a, line 1: We do not want this change, looks like a carry-over from stable branch

randallli:
moved this into https://github.com/material-components/material-components-ios/pull/6170

components/NavigationDrawer/examples/BottomDrawerExpandFullscreenExample.swift
randallli: Removed. this was an artifact of the export that I have now fixed.

contributing/bazel_kokoro.md
randallli: https://github.com/material-components/material-components-ios/pull/6171

.gitattributes
randallli: Done. in other pr by rebasing.

PR #6175, opened 2019-01-02
[TextFields] Add underline multiline snapshots tests.

romoore:
Adding Underline style multi-line text field snapshot tests. This style is in
the Kitchen Sink example and should be verified for breaking changes going
forward.

Part of #5762


codeman7:
Is there a way we can have the empty text field test their own subclass so we don't repeat them for each language?

codeman7 (approved):

romoore:
@codeman7 - You'd prefer for an extra class for each style being tested? Something like:

* MDCTextAreaOutlinedControllerBaselineCharacterCountEmptySnapshotTests
* MDCTextAreaOutlinedControllerEmptySnapshotTests
* MDCTextFieldFilledControllerLeadingImageEmptySnapshotTests
* MDCTextFieldFilledControllerEmptySnapshotTests
* MDCTextFieldFilledFloatingControllerLeadingImageEmptySnapshotTests
* MDCTextFieldFilledFloatingControllerEmptySnapshotTests
* ...and the other 12 or so?

codeman7:
Everything is subclassed from MDCAbstractTextFieldSnapshotTests and can't we have a MDCAbstractTextFieldEmptySnapshotTest so that we don't have empty text fields of every style and all the languages? 

Just an idea, you know more about snapshots than I do. I just worry about having multiple languages for multiple styles on multiple components and all those test taking more and more time.

romoore:
I think having an empty one in each style is reasonable, but I agree that per-language is overkill.

romoore:
I tried breaking them into a category that's only included in some subclasses, but the methods were executed regardless. I suspect Xcode does a very thorough job of runtime checking for method names.

https://github.com/material-components/material-components-ios/issues/6177

Issue #4937, opened 2018-08-28
[BottomNavigation] Ink color only matches tint for first tab item.


bambaragoogle:
After setting a MDCBottomNavigationBar's selectedItemTintColor property, only the first tab item's ink changes to that tint color. The documentation doesn't specify, but I would expect this property to affect either all tab ink or none of it. If it's not intended to change the ink color, please expose an additional property that exposes ink color.

romoore:
Thanks, @bambaragoogle . Could you please provide steps to reproduce? We recently merged #4420 that should have demonstrated the fix.

romoore:
Oh, whoops, you want the itemTintColor to affect the ink color. #4420 is unrelated.

romoore:
I tried to reproduce this in a default-themed MDCCatalog (not using our black/grey theme customizations) and could not reproduce the issue you are describing. It seems as though the ink color is changing and is correctly based on the `selectedItemTintColor` when I test several situations:

1. The initial (default) state
2. When I change the `selectedItemTintColor` and `unselectedItemTintColor`
3. When I reapply the original values (idempotency)
4. With a white background to show which tint is being used for the Ripple

|State|Default | Changed | Default Reapplied (idempotency) | Tint highlighted|
|--|--|--|--|--|
|Unselected|![unselected-1-default](https://user-images.githubusercontent.com/1753199/44790580-af8f7700-ab6d-11e8-83fb-ede9955ae3e1.png)|![unselected-2-changed](https://user-images.githubusercontent.com/1753199/44790585-b322fe00-ab6d-11e8-9281-566c963aac22.png)|![unselected-3-reapply](https://user-images.githubusercontent.com/1753199/44790592-b6b68500-ab6d-11e8-91bb-3574576b97af.png)|![unselected-4-tintcolor](https://user-images.githubusercontent.com/1753199/44790691-ea91aa80-ab6d-11e8-89e7-9114109eec19.png)|
[...]

PR #6190, opened 2019-01-07
[BottomNavigation] Confirm (un)selected ink is same.

romoore:
When changing the value of `selectedItemTintColor`, we should confirm that
both selected and unselected items receive the same ink color.

Follow-up for #4937

### DIFF
--- a/components/BottomNavigation/tests/unit/BottomNavigationItemViewTests.m
+++ b/components/BottomNavigation/tests/unit/BottomNavigationItemViewTests.m
@@ -125,6 +125,7 @@ static UIImage *fakeImage(void) {
   // Then
   XCTAssertNotEqualObjects(item1.inkView.inkColor, item1DefaultInkColor);
   XCTAssertNotEqualObjects(item2.inkView.inkColor, item2DefaultInkColor);
+  XCTAssertEqualObjects(item1.inkView.inkColor, item2.inkView.inkColor);
 }
 
 - (void)testBadgeCenterIsCorrectWithoutRTL {


codeman7 (approved):

Ruberik:
@romoore I guess I'm confused about this test. Here's how it reads to me:

GIVEN that you have two MDCBottomNavigationItemViews, and the first one is selected and the second one isn't...

WHEN you change the colour that's displayed iff each view is selected (which the second one isn't)...

THEN the colour that's currently displayed should be different from the initial colour in both views.

My guess is that I'm wrong about the THEN clause -- why would the colour have changed for view2? -- but I don't know how. :-)

romoore:
@Ruberik I think there might be a misunderstanding about how the property `selectedItemTintColor` applies to the (Ink) Ripple.  Based on the changes made in https://github.com/material-components/material-components-ios/pull/4950 in order to fix https://github.com/material-components/material-components-ios/issues/4937, the same ripple color should be used for both views. The ripple color for any/all views is based on the selected item's tint color, it's not based on the tint color of the specific view.

To address your reading of the test directly, your *WHEN* and *THEN* clauses are incorrect. The test should be:

1. **GIVEN** a selected and one or more unselected items (view).
2. **WHEN** the selected item's tint color changes.
3. **THEN** the (Ink) Ripple color changes for all items.

Does this make more sense?

Ruberik:
Ahh, got it. I made a bad assumption about what a Ripple colour is -- that it would just be the displayed colour, rather than something that might depend on other colours that might show. Thanks for taking the time to explain it.

PR #6258, opened 2019-01-09
[BottomNavigation] Add theming extension

codeman7:
### Context
As the team pivots to using theming within extensions we need to move all components to extensions. This will address the needs for BottomNavigation. This adds a new method to `MDCBottomNavigationBar - applyThemeWithScheme:(id<MDCContainerScheming>)scheme`.
_Note_ [Please see](go/mdc-ios-theming-v2)

### The problem
We currently do not theme [MDCBottomNavigationBar](https://github.com/material-components/material-components-ios/tree/develop/components/BottomNavigation) the way the team has decided to theme components.

### The fix
This themes [MDCBottomNavigationBar](https://github.com/material-components/material-components-ios/tree/develop/components/BottomNavigation) with the new style we have all agreed on.


romoore:
I thought we paused adding new theming extension code until the current ones had time to be used by clients and we could get feedback.  Is there a specific need for this theming extension?

components/BottomNavigation/tests/unit/BottomNavigationThemingTests.m
```diff
@@ -0,0 +1,118 @@
...
+@property(nonatomic, strong) NSMutableArray<MDCBottomNavigationItemView *> *itemViews;
+- (void)applyThemeWithColorScheme:(id<MDCColorScheming>)colorScheme;
+- (void)applyThemeWithTypographyScheme:(id<MDCTypographyScheming>)typographyScheme;
+@end
+
+@interface MDCBottomNavigationItemView (Testing)
+@property(nonatomic, strong) UILabel *label;
+@end
+
+@implementation BottomNavigationThemingTests
```
galiak11, line 36: Does the new theming pass existing tests (to make sure there are no visual changes)?

galiak11 (commented):
I don't see any fixed examples.  Do we have any examples that currently theme bottom navigation bars? If so, those should be fixed.

Should we also update the readme file?

components/BottomNavigation/tests/unit/BottomNavigationThemingTests.m
codeman7: I am confused by what you mean "existing tests", can you please explain what tests you are referring to?

codeman7:
Examples have been added, I don't know if we want to do the README yet. I still need to talk to @randallli to see if we want more theming to land.

galiak11 (approved):
Looks good to me.
One note - if we're pacing adding theming extensions, can we focus first on components that have multiple states, like bottom nav (item), list (item), tabs, textfields, etc.  ?

PR #6282, opened 2019-01-11
[Math] Add Math tests to bazel

### LINKED ISSUE #6231: Internal issue: b/122532042
This was filed as an internal issue. If you are a Googler, please visit [b/122532042](http://b/122532042) for more details.

---
#### Internal data
- Associated internal bug: [b/122532042](http://b/122532042)

fluxminds:
closes #6231

### DIFF
--- a/components/private/Math/BUILD
+++ b/components/private/Math/BUILD
@@ -12,10 +12,38 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("//:material_components_ios.bzl", "mdc_public_objc_library")
+load(
+    "//:material_components_ios.bzl",
+    "mdc_objc_library",
+    "mdc_public_objc_library",
+    "mdc_unit_test_suite",
+)
 
 licenses(["notice"])  # Apache 2.0
 
 mdc_public_objc_library(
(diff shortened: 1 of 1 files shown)


components/private/Math/BUILD
```diff
@@ -12,10 +12,38 @@
...
+    name = "unit_test_sources",
+    srcs = native.glob([
+        "tests/unit/*.m",
+        "tests/unit/*.h",
+    ]),
+    sdk_frameworks = [
+        "CoreGraphics",
+        "XCTest",
+    ],
+    visibility = ["//visibility:private"],
```
romoore: Tip: Nit: Private is the default visibility and so it's not necessary to provide it explicitly.

https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes

romoore (approved):

components/private/Math/BUILD
fluxminds: Good to know! Addressed in https://github.com/material-components/material-components-ios/pull/6282/commits/8c495c753a2f2316889306d0476c84abb6db630f.

Issue #6287, opened 2019-01-11
{kokoro} Get https://github.com/material-foundation/ios-snapshot-test-case building with bazel


material-mirror:
Now that we've forked ios-snapshot-test-case, we need it to compile using bazel.

---
#### Internal data
- Associated internal bug: [b/122714776](http://b/122714776)

PR #6297, opened 2019-01-12
WIP - {kokoro} Getting Snapshot tests working with bazel

romoore:
WIP - This needs to be broken-up into several PRs.

PR #6299, opened 2019-01-13
[TextFields] Fix snapshot test imports.

romoore:
In order to prepare snapshot tests to compile using bazel, all of the
import statements need to be fixed to use the correct paths.

Part of #6287


romoore:
This was split-out from https://github.com/material-components/material-components-ios/pull/6297

andrewoverton (approved):

Issue #6235, opened 2019-01-08
 [TextFields] Add snapshot tests to bazel


material-mirror:
This was filed as an internal issue. If you are a Googler, please visit [b/122530393](http://b/122530393) for more details.

---
#### Internal data
- Associated internal bug: [b/122530393](http://b/122530393)
- Blocked by: https://github.com/material-components/material-components-ios/issues/6230

PR #6307, opened 2019-01-14
{bazel} Add rule for snapshot Objective-C library.

romoore:
Creating a BUILD file and public "Snapshot" target for clients to depend on.
The primray use of this target will be in material_components_ios.bzl to
generate snapshot build rules.

Part of #6235

### DIFF
--- a/material_components_ios.bzl
+++ b/material_components_ios.bzl
@@ -167,6 +167,29 @@ def mdc_examples_swift_library(
       ]),
       **kwargs)
 
+def mdc_snapshot_objc_library(
+    name,
+    extra_srcs = [],
+    deps = [],
+    sdk_frameworks = [],
+    visibility = ["//visibility:private"],
+    testonly = 1,
+    **kwargs):
+  """Declare an mdc_objc_library for snapshot test source."""
+  mdc_objc_library(
+      name = name,
+      srcs = native.glob([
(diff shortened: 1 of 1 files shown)


codeman7:
I believe this is closely related to #6308, it seems like we will only need this for #6308. Will this be the same rule used for `<#component>` snapshot test? So something like:
```
mdc_snapshot_objc_library(
    name = "SnapshotTests",
    extra_srcs = ["tests/snapshots/*.m"],
    deps = [":<#component>"],
    testonly = 1,
 )
```
throughout the component `BUILD` files? If it is only going to be used to add `private/snapshot` and it's a one off I don't think we should add a rule.

codeman7 (approved):

romoore:
#6308 is related but not the same change - it introduces the BUILD targets for `components/private/Snapshot`, which is used by this rule.  An example of the new rule in use is in the "WIP" PR I've been breaking-out into these PRs.

https://github.com/material-components/material-components-ios/blob/0d4f9d7b5f00837d94e76a5a9f82fc3bbd2b464a/components/TextFields/BUILD#L100

codeman7:
I miss read `mdc_objc_library` as `mdc_snapshot_objc_library` in #6308. Please ignore my comment.

romoore:
Makes sense. 👍

PR #6310, opened 2019-01-14
{kokoro} Install git-lfs for bazel tests.

romoore:
In order to support snapshot testing we should install git-lfs whenever we
perform a "test" action. Ideally we would identify all bazel targets that
depend on `//components/private/Snapshot`, but this is good enough for a
start.

Prework for #6235


fluxminds (approved):
Tip: might be more clear to comment inline that it is only required for test targets depending on snapshot tests.
Putting the PR description is in the commit message is also fine too.

romoore:
Opened https://github.com/material-components/material-components-ios/issues/6312 so we don't lose track of the improvement.

PR #6311, opened 2019-01-14
[BottomNavigation] Elevates the BottomNavigationBarController to beta status and re-adds its unit tests.

nokeeo:
Contributes to #4160


MaterialComponentsBeta.podspec
```diff
@@ -66,6 +66,19 @@ Pod::Spec.new do |mdc|
     extension.dependency "MaterialComponents/schemes/Typography"
   end
 
+  mdc.subspec "BottomNavigation" do |component|
+    component.ios.deployment_target = '8.0'
```
romoore, line 70: Should this be 9.0? I thought some of the APIs were only 9+?

components/BottomNavigation/tests/unit/beta/MDCBottomNavigationBarControllerTests.m
```diff
@@ -0,0 +1,348 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <XCTest/XCTest.h>
+
+#import "MDCBottomNavigationBarController.h"
+
+static CGFloat const kDefaultExpectationTimeout = 15;
```
romoore, line 20: Please import CoreGraphics to use CGFloat.

components/BottomNavigation/tests/unit/beta/MDCBottomNavigationBarControllerTests.m
```diff
@@ -0,0 +1,348 @@
...
+
+@implementation MDCBottomNavigationControllerTests
+
+- (void)setUp {
+  [super setUp];
+  self.bottomNavigationBarController = [[MDCBottomNavigationBarController alloc] init];
+}
+
+- (void)tearDown {
+  [super tearDown];
```
romoore: Nit: Please call `[super tearDown]` at the end of your `tearDown` method. It's a safer pattern in case the superclass destroys something you depend on.

romoore (approved):
LGTM with a few little feedback.

MaterialComponentsBeta.podspec
nokeeo: The controller is 9+, however when I set the deployment target to 9.0, when installing the pods for the catalog, the installer could not find a compatible version for the unit tests, because the catalog's target deployment version is 8.0.

We are currently using ```NS_CLASS_AVAILABLE_IOS``` to restrict the controller to 9+.

I am a bit unfamiliar with CocoaPods.  Anything I missed or any tips?

romoore: Hmm, summoning @ianegordon  and @andrewoverton (tomorrow) in case either knows better. Andrew's been working with CocoaPods a lot recently and likely knows better than I do.  Ian knows more about iOS versioning and CocoaPods, I believe.

I'm not sure if we need to specify iOS 9+ for the target or if we think it's sufficient to put it in the headers.

components/BottomNavigation/tests/unit/beta/MDCBottomNavigationBarControllerTests.m
nokeeo: Can do.  Just curious, why the redundant import if CoreGraphics is a dependency of UIKit?

romoore: Importing direct dependencies is a habit I've picked up over the years and think it's a good practice in general.  

For example, if I import "MaterialButtons+Theming.h", it could import "MaterialButtons.h" for me. However in the future that might change to a forward declaration (it won't), and then my code would break.  If I explicitly import all of my direct dependencies, my code remains buildable.

Also, as a library developer I've experienced the pain of having to fix many clients' code because they depended on transitive imports for their direct dependencies. :-\

romoore: As an aside, I don't see any imports for CoreGraphics in UIKit.h. Here's what my SDK headers have:

* `Foundation`
* `UIKitCore`
* A conditional import of `UIKit/DocumentManager`

nokeeo: Makes sense!

UIGraphics, UIBezierPath, and others import CG.  Those are both imported by UIKitCore.

MaterialComponentsBeta.podspec
ianegordon: If you have gated any use of iOS 9 APIs, I expect this would work as intended.

Please verify on an iOS 8.0 simulator or device.

romoore: It crashes on an iOS 8.3 simulator as-is. We'll have to find a way to put a guard in the catalog code.

romoore: Got a version not crashing (but not great).

simulator screen shot - iphone 4s - 2019-01-17 at 22 24 05


```
diff --git a/components/BottomNavigation/examples/BottomNavigationBarControllerExampleViewController.swift b/components/BottomNavigation/examples/BottomNavigationBarControllerExampleViewController.swift
index 66f1ab8e45..3281d6af9f 100644
--- a/components/BottomNavigation/examples/BottomNavigationBarControllerExampleViewController.swift
+++ b/components/BottomNavigation/examples/BottomNavigationBarControllerExampleViewController.swift
@@ -30,6 +30,11 @@ class BottomNavigationControllerExampleViewController: MDCBottomNavigationBarCon
   }
 
   override func viewDidLoad() {
+    if #available(iOS 9, *) {
+
+    } else {
+      return
+    }
     super.viewDidLoad()
 
     let viewController1 = BaseCellExample()
@@ -46,7 +51,7 @@ class BottomNavigationControllerExampleViewController: MDCBottomNavigationBarCon
 
   class func catalogMetadata() -> [String: Any] {
     return [
-      "breadcrumbs": ["Bottom Navigation", "Bottom Navigation Controller"],
+      "breadcrumbs": ["Bottom Navigation", "Bottom Navigation Controller (iOS 9+)"],
       "presentable": false
     ]
   }
```

PR #6317, opened 2019-01-14
[ChipTextField] add touch delegate methods into MDCChipTextField

fluxminds:
Currently MDCChipTextField doesn't have ability to handle touch interaction, this PR adds a MDCChipTextFieldDelegate for supporting touch delegate methods.


components/ChipTextField/examples/MDCChipTextField.h
```diff
@@ -22,6 +51,16 @@
 */
 @interface MDCChipTextField : MDCTextField
 
+/**
+ The chip views in the textfield.
+ */
+@property(nonatomic, readonly, copy) NSArray<MDCChipView *> *chipViews;
```
galiak11: Can we guarantee that the order of the chips in the array is the same as the visual order of the chips? If so, than that needs to be documented in the comment. Alternatively there can be a "chipForIndex" API (plus numberOfChips).

components/ChipTextField/examples/MDCChipTextField.h
```diff
@@ -45,4 +84,12 @@
  */
 - (void)appendChipWithText:(nonnull NSString *)text NS_SWIFT_NAME(appendChip(text:));
 
+/**
+ Set a chip view at particular index to be selected or not.
+
+ @param selected selection status to set
+ @param index The index of the chip view to set the selection status
+ */
+- (void)setChipViewSelected:(BOOL)selected atIndex:(NSInteger)index;
```
galiak11: What is the reason for this API?  Can we add an explanation to the design doc?

components/ChipTextField/examples/MDCChipTextField.m
```diff
@@ -20,15 +20,16 @@
...
+                                MDCChipTextFieldScrollViewDelegate>
 
 @property(nonatomic, strong) MDCChipTextFieldScrollView *chipsContainerView;
 
 @property(nonatomic, readwrite, weak) NSLayoutConstraint *chipContainerViewConstraintLeading;
 @property(nonatomic, readwrite, weak) NSLayoutConstraint *chipContainerViewConstraintTrailing;
 @property(nonatomic) CGFloat chipContainerViewConstraintTrailingConstant;
 // Chip view models
-@property(nonatomic, strong) NSMutableArray<MDCChipView *> *chipViews;
+@property(nonatomic, readwrite, copy) NSMutableArray<MDCChipView *> *mutableChipViews;
```
galiak11, line 32: Why are we using a different name? Is it possible to use a mutable chipViews array in the implementation, which uses the same name as the readonly interface?

components/ChipTextField/examples/MDCChipTextField.h
```diff
@@ -45,4 +84,12 @@
  */
 - (void)appendChipWithText:(nonnull NSString *)text NS_SWIFT_NAME(appendChip(text:));
```
galiak11: Should this return the chipView that was appended?

components/ChipTextField/examples/MDCChipTextField.m
```diff
@@ -50,6 +51,54 @@ - (instancetype)initWithFrame:(CGRect)frame {
...
+  MDCChipView *chipView = [[MDCChipView alloc] init];
+  chipView.titleLabel.text = text;
+  chipView.translatesAutoresizingMaskIntoConstraints = NO;
+
+  [self appendChipView:chipView];
+}
+
+- (void)appendChipView:(MDCChipView *)chipView {
+  if ([self.chipTextFieldDelegate respondsToSelector:@selector(chipField:shouldAddChip:)]) {
+    NSInteger indexToAppend = self.mutableChipViews.count;
```
galiak11, line 74: Why is there an index, if we're always adding to the end?  Also, if we're always adding to the end, we might as well call it "shouldAppendChip" and remove the index argument.

components/ChipTextField/examples/MDCChipTextField.m
```diff
@@ -50,6 +51,54 @@ - (instancetype)initWithFrame:(CGRect)frame {
...
+    }
+  }
+
+  [self.chipsContainerView appendChipView:chipView];
+
+  // recalculate the layout to get a correct chip frame values
+  [self.chipsContainerView layoutIfNeeded];
+  [self.mutableChipViews addObject:chipView];
+
+  [self clearChipsContainerOffsetWithConstant:kTextToEnterPlaceholderLength];
```
galiak11, line 89: What is this constant (kTextToEnterPlaceholderLength) for?

components/ChipTextField/examples/MDCChipTextField.h
```diff
@@ -12,8 +12,37 @@
...
 // limitations under the License.
 
+#import "MDCChipView.h"
 #import "MDCTextField.h"
 
+@class MDCChipTextField;
+
+@protocol MDCChipTextFieldDelegate <NSObject>
+
+@optional
```
galiak11, line 22: I'm not sure how much I like a protocol that all its methods are optional. One way to avoid it is to use multiple protocols, or perhaps an aggregator of sub protocols. Let's rethink this.

components/ChipTextField/examples/MDCChipTextFieldScrollView.h
```diff
@@ -20,33 +20,74 @@
 
 @protocol MDCChipTextFieldScrollViewDelegate <NSObject, UIScrollViewDelegate>
 
-// TODO: delegate methods to handle touch event
+@optional
```
galiak11: Same comment as before - one method doesn't need to be optional.

galiak11 (commented):
Can we add examples that use this API to perform tasks from the spec?
Can you please also elaborate more in the main comment? And also an issue number is missing.

components/ChipTextField/examples/MDCChipTextField.h
fluxminds: Sure. I will add some comment, the index represents the order

components/ChipTextField/examples/MDCChipTextField.m
fluxminds: I don't think it is possible to change the type for a property in implementation file.

components/ChipTextField/examples/MDCChipTextFieldScrollView.h
fluxminds: Addressed in 0003b1507a139f7d11899e583f895bac70dd94e6

components/ChipTextField/examples/MDCChipTextField.h
fluxminds: Followed up offline. Since these methods are optional, we need to keep make sure it is explicit.

components/ChipTextField/examples/MDCChipTextField.m
fluxminds: It is a constant number currently used to fix the weird typing offset issue.

components/ChipTextField/examples/MDCChipTextField.h
fluxminds: Addressed in 0003b1507a139f7d11899e583f895bac70dd94e6

components/ChipTextField/examples/MDCChipTextField.h
fluxminds: Removed this API in 0003b1507a139f7d11899e583f895bac70dd94e6

components/ChipTextField/examples/MDCChipTextField.h
fluxminds: Addressed in 0003b1507a139f7d11899e583f895bac70dd94e6

fluxminds:
@galiak11 Comments addressed in 0003b1507a139f7d11899e583f895bac70dd94e6.

components/ChipTextField/examples/MDCChipTextField.m
galiak11: One way to avoid using 2 different names to define only a getter for chipViews - by making it an accessor method:
- NSArray<MDCChipView *> *chipViews;

romoore: If you declare a public `readonly` NSArray, you can just return a copy of the internally-declared NSMutableArray in the custom getter.  Any reason not to do that?

galiak11 (approved):
Can you please add an issue number and some better description when merging this PR?

components/ChipTextField/examples/MDCChipTextField.h
```diff
@@ -43,6 +79,6 @@
 
  @param text The string to display in the chip.
  */
-- (void)appendChipWithText:(nonnull NSString *)text NS_SWIFT_NAME(appendChip(text:));
+- (MDCChipView *)appendChipWithText:(nonnull NSString *)text NS_SWIFT_NAME(appendChip(text:));
```
romoore: So if a client wants an image on the chip (for example because they found an Address Book entry), they are expected to modify the MDCChipView directly after it's appended?

components/ChipTextField/examples/MDCChipTextField.m
fluxminds: Eh. I am confused. I guess there were two issues you were talking about. 
@romoore Yes, and that is how it is currently implemented.
@galiak11 Directly exposing the accessor method of a private property is dangerous, and I think there will also be compiler issues if [self.chipsViews addObject:] is called in this case. Avoiding using two names is doable if we swap the synthesized instance with a mutable one, but I think using property is always a better practice in this case.

components/ChipTextField/examples/MDCChipTextField.h
fluxminds: The original thought was it should be handled in a delegate method (e.g. shouldAddChipView:chipView or willAddChipView:chipView) if user wants to handle it before the chip is added as a subview.

This code review change was still debatable.

romoore: The delegate method sounds good. If we can prevent adding a Chip (for example with misspelled contact name) then the client can support a broader range of behavior.

For example:
Types: "rsmoore"
Result: Chip("rsmoore")

Types: "rsmooore"
Result: Text remains set to "rsmooore" and Error is shown

components/ChipTextField/examples/MDCChipTextField.m
romoore: @wenyuzhang666 So if there's no reason not to go with the current approach, why are you proposing changing it?

fluxminds: Sorry. The changes in the PR was what I meant by `currently implemented`.

components/ChipTextField/examples/MDCChipTextField.h
fluxminds: Right. That would be supported in this case.

components/ChipTextField/examples/MDCChipTextField.m
romoore: Thanks for clearing that up. Makes sense. 👍

romoore:
+1 to feedback from @galiak11.  Please improve the summary of this PR before you merge.

fluxminds:
Sure. I updated the description.

PR #6325, opened 2019-01-14
{kokoro} Only install git-lfs on kokoro

romoore:
We shouldn't be performing git-lfs operations locally, only on kokoro.


randallli (approved):

romoore:
Merging this anyway. The failure is the ongoing timeouts for snapshot tests. but adding this check will fix running `.kokoro` locally for some.

PR #6333, opened 2019-01-15
{kokoro} Deflake snapshot tests on kokoro

romoore:
Snapshot tests were timing-out because of two causes:
1. Simulators were not starting up correctly
2. Text Fields snapshot tests were taking too long.
3. Snapshot tests were sometimes taking longer than 1 second.

The solutions to these problems seem to be:
1. Reset and pre-start simulators before calling bazel
2. Increasing the time-out for text fields snapshot tests
3. Turning off parallel snapshot tests to prevent CPU/resource starvation for tests.

Work-around for #6335

### DIFF
--- a/components/TextFields/BUILD
+++ b/components/TextFields/BUILD
@@ -115,6 +115,9 @@ mdc_snapshot_objc_library(
 
 mdc_snapshot_test(
     name = "snapshot_tests",
+    # Tests have sometimes taken more than 5 minutes to launch the simulator
+    # and execute
+    timeout = "long",
     deps = ["snapshot_test_lib"],
 )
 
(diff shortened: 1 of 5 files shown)


randallli (dismissed):

components/TextFields/BUILD
```diff
@@ -115,6 +115,9 @@ mdc_snapshot_objc_library(
 
 mdc_snapshot_test(
     name = "snapshot_tests",
+    # Tests have sometimes taken more than 5 minutes to launch the simulator
+    # and execute
+    timeout = "long",
```
romoore: Included here to help deflake, but will be merged separately in https://github.com/material-components/material-components-ios/pull/6337

codeman7 (approved):

randallli (commented):
lgtm

PR #6415, opened 2019-01-18
[Buttons] Add fallback for `borderWidthForState:`

codeman7:
## Related links
* Component: [MDCButton](https://github.com/material-components/material-components-ios/tree/develop/components/Buttons/src)
* Bug: [Buttons] Unset `borderWidthForState:` reverts to 0, not .normal width value #3408
* API: [`borderWidthForState`](https://github.com/material-components/material-components-ios/blob/dfd657d694ad5a8612406815376f253310d9bfae/components/Buttons/src/MDCButton.h#L240)
* Related Apple docs: [`UIControlState`](https://developer.apple.com/documentation/uikit/uicontrol/state)
* Guidelines: [Material Outlined Button](https://material.io/design/components/buttons.html#outlined-button)
## Introduction

Currently we have an API `borderWidthForState:` on `MDCButton` that returns the borderWidth for a given `UIControlState` will return the _borderWidth_. This is used to render a border around the button as shown in the link above.

## The problem

If a client sets the _borderWidth_ for `UIControlStateNormal` then that value will be rendered in all states but the API for any state other than `UIControlStateNormal` will return 0. We want our API to match the rendering behavior.

## The fix

If a client sets a value for `UIControlStateNormal` then return that value if there is no value for the given `UIControlState` they have _asked_ for when calling `borderWidthForState:`.

## Testing

- [x] Default behavior if a client hasn't set any values
- [x] Behavior if a client only sets `UIControlStateNormal`
- [x] Behavior if a client sets every control state to be a unique value

## Screenshots
|    | No border for both states | 2pt border same for both states | 2pt border for `.normal` 4pt border for `.highlighted` |
| --- | --- | --- | --- |
| `.normal` |simulator screen shot - iphone xs max - 2019-01-18 at 09 49 14|simulator screen shot - iphone xs max - 2019-01-18 at 09 48 38|simulator screen shot - iphone xs max - 2019-01-18 at 09 49 47|
| `.highlighted` | simulator screen shot - iphone xs max - 2019-01-18 at 09 49 16|simulator screen shot - iphone xs max - 2019-01-18 at 09 48 39|simulator screen shot - iphone xs max - 2019-01-18 at 09 49 50|

_Note:_ This is an API change not a rendering change

### DIFF
--- a/components/Buttons/src/MDCButton.m
+++ b/components/Buttons/src/MDCButton.m
@@ -628,6 +628,10 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
 #pragma mark - Border Width
 
 - (CGFloat)borderWidthForState:(UIControlState)state {
+  // If the `.highlighted` flag is set, turn off the `.disabled` flag
+  if ((state & UIControlStateHighlighted) == UIControlStateHighlighted) {
+    state = state & ~UIControlStateDisabled;
+  }
   NSNumber *borderWidth = _borderWidths[@(state)];
   if (borderWidth != nil) {
     return (CGFloat)borderWidth.doubleValue;
@@ -636,9 +640,17 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
 }
 
 - (void)setBorderWidth:(CGFloat)borderWidth forState:(UIControlState)state {
-  _borderWidths[@(state)] = @(borderWidth);
-
-  [self updateBorderWidth];
+  UIControlState storageState = state;
+  if ((state & UIControlStateHighlighted) == UIControlStateHighlighted) {
+    storageState = state & ~UIControlStateDisabled;
+  }
+  // Only update the backing dictionary if:
+  // 1. The `state` argument is the same as the "storage" state, OR
+  // 2. There is already a value in the "storage" state.
+  if (storageState == state || _backgroundColors[@(storageState)] != nil) {
+    _borderWidths[@(state)] = @(borderWidth);
+    [self updateBorderWidth];
+  }
 }
 
 - (void)updateBorderWidth {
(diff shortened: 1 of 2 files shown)


romoore:
Relevant issue: https://github.com/material-components/material-components-ios/issues/3408

components/Buttons/src/MDCButton.m
```diff
@@ -632,7 +632,7 @@ - (CGFloat)borderWidthForState:(UIControlState)state {
   if (borderWidth != nil) {
     return (CGFloat)borderWidth.doubleValue;
   }
-  return 0;
+  return (CGFloat)[_borderWidths[@(UIControlStateNormal)] doubleValue];
```
romoore, line 639: Please account for UIButton's behavior around `UIControlStateDisabled | UIControlStateHighlighted`

See https://github.com/material-components/material-components-ios/pull/5919 for more details.

romoore (dismissed):
Please make sure we match UIButton's behavior for all states. See inline comment for details.

components/Buttons/src/MDCButton.m
codeman7: Done

components/Buttons/tests/unit/ButtonsTests.m
```diff
@@ -265,6 +265,50 @@ - (void)testElevationNormalZeroElevation {
   XCTAssertEqual([button elevationForState:UIControlStateNormal], 0);
 }
 
+- (void)testDefaultBorderWidth {
+  // Given
+  MDCButton *button = [[MDCButton alloc] init];
+
+  // Then
+  for (NSUInteger controlState = 0; controlState < kNumUIControlStates; ++controlState) {
```
romoore: Looks like this isn't testing the `.selected | .highlighted | .disabled` state. Please make the comparison `<=` so it will test the combination of all UIControl states.  Same for the other 2 tests being added.

romoore (approved):
LGTM with a nit about the test loops.

components/Buttons/tests/unit/ButtonsTests.m
codeman7: Done

codeman7:
Internal testing cl/230144000

PR #6432, opened 2019-01-21
[Buttons] Add setup and teardown to tests

codeman7:
## Related links
* Component: [MDCButton](https://github.com/material-components/material-components-ios/tree/develop/components/Buttons/src)
* Bug: This was discovered when working on #3408
* Related Apple docs: [`setUp & tearDown`](https://developer.apple.com/documentation/xctest/xctestcase/understanding_setup_and_teardown_for_test_methods)
## Introduction
While working on #6415 I noticed a lot of the tests within `ButtonTests.m` created an MDCButton in the given. This change implements a _setUp_ and _tearDown_ function that creates an MDCButton that all of the tests can then use.
## The problem
In almost every test we create a button and then test it. We can simply this by just doing a _setup_ and _teardown_.
## The fix
Add _setup_ and _teardown_ methods so that all the test that created a button in the _Given_ section didn't have to do that.
## Additional notes
Some of the test created a button and some created a button with a frame. If the frame had no effect on the test then setting the frame was removed.

### DIFF
--- a/components/Buttons/tests/unit/ButtonsTests.m
+++ b/components/Buttons/tests/unit/ButtonsTests.m
@@ -113,99 +113,104 @@ static NSString *controlStateDescription(UIControlState controlState) {
 @end
 
 @interface ButtonsTests : XCTestCase
+@property(nonatomic, strong, nullable) MDCButton *button;
 @end
 
 @implementation ButtonsTests
 
+- (void)setUp {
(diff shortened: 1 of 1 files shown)


codeman7:
## **Do not merge** 
Waiting to rebase once #6415 is merged.

romoore (approved):

Issue #5522, opened 2018-10-25
Internal issue: b/118419714


material-mirror:
This was filed as an internal issue. If you are a Googler, please visit [b/118419714](http://b/118419714) for more details.

---
#### Internal data
- Associated internal bug: [b/118419714](http://b/118419714)

material-mirror:
The internal issue [b/118419714](http://b/118419714) is now closed. This issue is being closed as a result.

PR #6445, opened 2019-01-22
[Catalog] Add logic for custom post launch actions

codeman7:
## Related links
* Related bug: #5522
* Additional link: cl/229879820
## Introduction
In the future we may want to add custom launch actions to the application that will be performed at startup. This PR adds the ability to add an extension to `AppDelegate` that adds a function `performPostLaunchAction` to perform custom launch actions in the future if we need to.
## The problem
There is currently no hook to perform post launch actions within `MDCCatalog`.
## The fix
Add logic to check if a `Selector` exist and if so respond to it within `AppDelegate`. For additional context please see links above.


catalog/MDCCatalog/AppDelegate.swift
```diff
@@ -57,6 +57,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate, MDCAppBarNavigationContro
       name: AppTheme.didChangeGlobalThemeNotificationName,
       object: nil)
 
+    if self.responds(to: Selector(("performPostLaunchAction"))) {
```
romoore, line 60: Since this is reused, please put it into a constant in a follow-up PR.

romoore (approved):

romoore:
Spoke offline, @codeman7 wants to merge now to get this into the week's release.

Issue #6437, opened 2019-01-22
[FlexibleHeader] Incorrect insets during animated transitions


brianjmoore:
During animated transitions between tracking scroll views, MDCFlexibleHeader provides incorrect content insets.

## Reproduction steps

1. Open MDCDragons app, go to "App Bar" > "Manual Tabs Jump (Animated)" example. (Provided in https://github.com/material-components/material-components-ios/pull/6447)
2. Tap on "Second" tab.
3. Tap on "Third" tab.
4. Tap on "Second" tab.
5. Tap on "First" tab.

(Generally: This happens when the view is added to the view hierarchy and visible between calls to `trackingScrollWillChange` and `setTrackingScrollView`)

## Expected behavior
Consistent content insets and offsets during and after the animation in. (The animation is a fade-in in this case, but might also be a slide-in managed by UIPageViewController for example).

## Actual behavior
When displaying a tab for the first time, it has insets that are ~40pt too tall. This creates a jumping effect when switching between tabs.

## Platform

 - Device: iPhone XR simulator
 - OS: iOS 12.1

## Additional context

![insets](https://user-images.githubusercontent.com/395995/51554264-ddd66180-1e42-11e9-9b36-65da73dd069c.gif)

---
#### Internal data
- Associated internal bug: [b/123297383](http://b/123297383)

material-mirror:
The internal issue [b/123297383](http://b/123297383) is now closed. This issue is being closed as a result.

PR #6447, opened 2019-01-22
[AppBar] Add animated tab switching example

brianjmoore:
Test case demonstrating issues with MDCFlexibleHeader and animated tab transitions in MDCAppBar.

Original issue:
https://github.com/material-components/material-components-ios/issues/6437


components/AppBar/examples/AppBarAnimatedGlitchExample.swift
```diff
@@ -0,0 +1,280 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import CoreGraphics
+import UIKit
+import third_party_objective_c_material_components_ios_components_AppBar_AppBar
```
romoore: These imports look like they were rewritten by the `.kokoro` script. It's best to have them start with something like:

```swift
import MaterialComponents.AppBar
import MaterialComponents.AppBar_ColorThemer
```

brianjmoore: Good catch, done.

components/AppBar/examples/AppBarAnimatedGlitchExample.swift
```diff
@@ -0,0 +1,281 @@
...
+  lazy var tabBar: MDCTabBar = {
+    let tabBar = MDCTabBar()
+
+    tabBar.items = self.tabs.map { $0.tabBarItem }
+
+    tabBar.delegate = self
+    return tabBar
+  }()
+
+  override init(nibName nibNameOrNil: String?, bundle nibBundleOrNil: Bundle?) {
```
romoore, line 49: Does this get called from `-init`? CatalogByConvention doesn't call this initializer directly, it only calls `-init`.

components/AppBar/examples/AppBarAnimatedGlitchExample.swift
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+
+    for (offset, tab) in zip(0..., tabs) {
+      tab.tabBarItem.tag = offset
+    }
+
+    view.backgroundColor = .white
```
romoore: Tip: This can be removed since you reassign it below.

romoore (approved):

components/AppBar/examples/AppBarAnimatedGlitchExample.swift
brianjmoore: Yes, UIViewController.init calls through to `init(nibName:bundle:)`. As the designated initializer it's the preferred override point.

components/AppBar/examples/AppBarAnimatedGlitchExample.swift
brianjmoore: Good point, thx.

PR #6449, opened 2019-01-23
[Tabs] Move the clipsToBounds behavior from MDCTabBar to MDCItemBar

mrhappyasthma:
The primary benefit here is that it allows the user to add a shadow to the CALayer of MDCTabBar without it getting clipped to bounds. (Where as setting `.clipsToBounds=NO` on MDCTabBar then the underlying itemBar/collectionView will extend outside of the bounds of the parent.)

By moving this clipsToBounds up a layer, it allows both of these behaviors to work.

For example:

- Adding a shadow to tabs: https://i.imgur.com/2CY15ip.png

- Tabs extending out of bounds if attempting to set MDCTabBar clipsToBounds=NO as it exists today: https://i.imgur.com/YYfkIMq.png
- Tabs fixed when setting MDCTabBar clipsToBounds=NO with this change: https://i.imgur.com/7QJhFBx.png

### DIFF
--- a/components/Tabs/src/MDCTabBar.m
+++ b/components/Tabs/src/MDCTabBar.m
@@ -125,7 +125,6 @@ static MDCItemBarAlignment MDCItemBarAlignmentForTabBarAlignment(MDCTabBarAlignm
   _unselectedTitleColor = _unselectedItemTintColor;
   _inkColor = [UIColor colorWithWhite:1 alpha:(CGFloat)0.7];
 
-  self.clipsToBounds = YES;
   _barPosition = UIBarPositionAny;
   _hasDefaultItemAppearance = YES;
   _hasDefaultAlignment = YES;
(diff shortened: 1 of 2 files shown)


romoore:
I'm a little concerned about this change. It looks like it's motivated by a desire to "nest" Tabs within one another and provide separation. This is already called-out in the [Material Guidelines article for Tabs](https://material.io/design/components/tabs.html#placement) as a "caution", so I'm wary of making support explicit.

My other concern is that you're adding a shadow below the Tabs themselves. From all of the design mocks I'm seeing in the article, it seems that providing an optional hairline/divider at the bottom edge of the Tabs might be acceptable, but the only shadows/elevation I see are from the AppBars containing the Tabs. Double-checking with a designer, only the "Primary variant" (where the Tabs surface is colored with the Theme's `primaryColor`) has elevation (4 dp). The example you're providing using `surfaceColor` should have an elevation of 0.

Although I think shadow support should be added to Tabs, it should be added via an elevation API similar to our other components. Adding an `elevation` property of type `MDCShadowElevation` is probably the best way forward.

jverkoey (changes requested):
Marking as changes requested due to @romoore's concerns.

romoore:
I'm closing this PR for now. Based on feedback from Design, I think we have a good reason to support a hairline separator between "layered" Tabs. My understanding is that Tabs shouldn't have shadows.

Issue #6466, opened 2019-01-24
{kokoro} bazel jobs are not building catalog/ targets


romoore:
For example, https://github.com/material-components/material-components-ios/pull/6458 passed builds but was missing dependencies when run locally via the `.kokoro` script.

romoore:
Oops, fixed by https://github.com/material-components/material-components-ios/pull/6468

PR #6468, opened 2019-01-24
{kokoro} Build everything when using run_bazel

romoore:
If no target is provided, the `-d bazel` (`run_bazel` within `.kokoro`) dependency system only
built (or tested) the components. Instead, it should build the entire
workspace.

Closes #6458

### DIFF
--- a/.kokoro
+++ b/.kokoro
@@ -346,7 +346,7 @@ run_bazel() {
       perform_pre_test_cleanup "iPhone"
     fi
   fi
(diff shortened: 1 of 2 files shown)


codeman7 (approved):

jverkoey (approved):

romoore:
Oops, this actually fixed https://github.com/material-components/material-components-ios/issues/6466

Issue #6305, opened 2019-01-14
[BottomDrawer] Drawer animation is broken at certain heights 


material-mirror:
This was filed as an internal issue. If you are a Googler, please visit [b/122819771](http://b/122819771) for more details.

---
#### Internal data
- Associated internal bug: [b/122819771](http://b/122819771)

material-mirror:
The internal issue [b/122819771](http://b/122819771) is now closed. This issue is being closed as a result.

PR #6491, opened 2019-01-27
[NavigationDrawer] Fix issue where drawer opens passed `initialDrawerFactor`

codeman7:
## Related links
* Bug: #6305 
* Component: [NavigationDrawer](https://github.com/material-components/material-components-ios/tree/develop/components/NavigationDrawer)
* Related PRs: #5928 #5587 #5520 #5443
## Introduction
The navigation drawer should support resizable content. This content should be able to change and the drawer stay at the same height or not _jump_ to another height. There was a problem where the drawer would jump between heights for certain `preferredContentSize`'s of the `contentViewController`. This was thought to be addressed in #5928. Upon investigation the root cause was not that the drawer jumped when `preferredContentSize` changed but that some how the drawer was able to open passed it's `initialDrawerFactor`. Therefore when the `preferredContentSize` did change the drawer go to the correct `initialDrawerFactor` and when it changed back it would go to the incorrect value. Additionally, when the drawer got into this state it would be in an `.expanded` drawer state because the drawer thought it should open up passed the `initialDrawerFactor`.
## The problem
The drawer thought it had less content that it actually had and therefore would go into the incorrect state and override `initialDrawerFactor`.
## The fix
Upon investigation fixing the drawer's internal calculation as to how much _content_ it actually has vs how tall the `presentingViewBounds.height` multiplied by `initialDrawerFactor`.
## Tests
Tests were updated as well as comments around them to understand why they were updated.
## Code snippet
In order to test the before and after behavior edit `BottomDrawerInfiniteScrollingExample.swift` as outlined below.
```diff
let bottomDrawerViewController = MDCBottomDrawerViewController()
+ contentViewController.supportsScrollToTop = false
bottomDrawerViewController.contentViewController = contentViewController
```
Doing the above will showcase how the drawer got into it's broken state.

### DIFF
--- a/components/NavigationDrawer/examples/BottomDrawerExpandFullscreenExample.swift
+++ b/components/NavigationDrawer/examples/BottomDrawerExpandFullscreenExample.swift
@@ -66,9 +66,7 @@ class BottomDrawerExpandFullscreenExample: UIViewController {
     let bottomDrawerViewController = MDCBottomDrawerViewController()
     bottomDrawerViewController.contentViewController = contentViewController
     contentViewController.drawerVC = bottomDrawerViewController
-    bottomDrawerViewController.setTopCornersRadius(12, for: .collapsed)
     bottomDrawerViewController.headerViewController = headerViewController
-    bottomDrawerViewController.isTopHandleHidden = false
     bottomDrawerViewController.trackingScrollView = contentViewController.tableView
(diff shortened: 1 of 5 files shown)


yarneo (approved):
Just tested it locally on our examples and everything looks good!

Is it possible to add a dragons with an example showcasing the existing problem, and then we can show how your fix solves the issue?

PR #6531, opened 2019-01-31
[BottomNavigation] Add more snapshot tests.

romoore:
Adding snapshot tests for longer titles. Adding Ink/Ripple touches to
every test.


components/BottomNavigation/tests/snapshot/MDCBottomNavigationBarSnapshotTests.m
```diff
@@ -108,6 +114,71 @@ - (void)generateAndVerifySnapshot {
...
+                                                       CGRectGetMidY(itemView.bounds))
+                                  animated:NO
+                            withCompletion:nil];
+}
+
+#pragma mark - Title length
+
+- (void)testJustifiedUnspecifiedAlwaysWithFiveLongTitleItemsiPadWidthTypicalHeight {
+  // When
+  self.navigationBar.alignment = MDCBottomNavigationBarAlignmentJustified;
```
codeman7, line 131: _Question:_Do you think this would be better as a helper method? 
```
- (void)helperSetupNavigationBarWithAlignment:(MDCBottomNavigationBarAlignment)alighnment {
  self.navigationBar.alighnment = alignment;
  self.navigationBar.titleVisibility = MDCBottomNavigationBarTitleVisibilityAlways;
  self.tabItem1.title = kLongTitle;
  self.tabItem2.title = kLongTitle;
  self.tabItem3.title = kLongTitle;
  self.tabItem4.title = kLongTitle;
  self.tabItem5.title = kLongTitle;
  self.navigationBar.frame = CGRectMake(0, 0, kWidthiPad, kHeightTypical);
  [self performInkTouchOnBar:self.navigationBar item:self.tabItem1];
}
```

Then all three of these methods could call it.

codeman7 (approved):

components/BottomNavigation/tests/snapshot/MDCBottomNavigationBarSnapshotTests.m
romoore: I like keeping the conditions of each test explicit at this point.  I considered adding stuff to set-up, helpers, etc. as I wrote these (it's tedious). But I do like the general approach of tests being as explicit as possible about what they're testing. 

I could do a little follow-up with a slightly different helper that takes more parameter.

Issue #6520, opened 2019-01-29
[BottomNavigation] Center alignment does not match material.io guidelines


romoore:
The `.center` alignment doesn't cluster items sufficiently close to match [the designs on material.io](https://material.io/design/components/bottom-navigation.html#placement).

## Expected
![bn-expected-center](https://user-images.githubusercontent.com/1753199/51946676-4e0b5700-23f0-11e9-87ac-94e42f3c7fb0.png)

## Current
![bn-clustered-actual](https://user-images.githubusercontent.com/1753199/51946721-71ce9d00-23f0-11e9-8afc-d280eb03156d.png)

---
#### Internal data
- Associated internal bug: [b/123643147](http://b/123643147)

PR #6543, opened 2019-02-02
[BottomNavigation] Improve snapshot tests.

romoore:
Making snapshot tests more realistic by adding badges.

Preparation for #6520


romoore:
The odd placement of the badges (see image below) in the ItemView snapshots is a result of the Badge view manipulating its own frame during layout. When placed within the BottomNavigationBar a layout pass is performed BEFORE positioning the badge and so it's rendered in the correct place.  #6544 fixes the bug and will update the snapshots after this lands.

testnarrowwidthtypicalheightlongtitlestackedrtl_11

components/BottomNavigation/tests/snapshot/MDCBottomNavigationBarSnapshotTests.m
```diff
@@ -139,6 +142,11 @@ - (void)changeToRTLAndArabicWithTitle:(NSString *)title {
       view.semanticContentAttribute = UISemanticContentAttributeForceRightToLeft;
     }
   }
+  if (self.navigationBar.items.count >= 2U) {
```
galiak11, line 145: What's special about this magic number?

components/BottomNavigation/tests/snapshot/MDCBottomNavigationBarSnapshotTests.m
```diff
@@ -35,6 +35,8 @@
 static NSString *const kLongTitleArabic =
     @"دول السيطرة استطاعوا ٣٠. مليون وفرنسا أوراقهم انه تم, نفس قد والديون العالمية. دون ما تنفّس.";
 static NSString *const kShortTitleArabic = @"ما تنفّس.";
+static NSString *const kBadgeTitleLatin = @"888+";
+static NSString *const kBadgeTitleArabic = @"أورا";
```
galiak11, line 39: What's the different between a shot title and a badge title?

galiak11 (approved):
It would be helpful to add this comment to the description of the PR. Explaining what badges are how they are used would also help.

components/BottomNavigation/tests/snapshot/MDCBottomNavigationBarSnapshotTests.m
romoore: The badge is limited to 4 characters  (that's the max I've seen in the Bottom Navigation article) but I was too lazy to grab Arabic numerals and just reused *lorem ipsum* from Text Fields.  The shorter title text is closer to a "long" title that's typical for Bottom Navigation. For example, "Favorites", "Settings" or "Streaming".

components/BottomNavigation/tests/snapshot/MDCBottomNavigationBarSnapshotTests.m
romoore: I wouldn't consider this a particularly magic number, it's just a bounds check.  Item 1 is where most examples have the ink touch and I wanted to prefer a different item when possible.  Since the next number after 1 is 2, I went with that.

romoore:
> It would be helpful to add this comment to the description of the PR. Explaining what badges are how they are used would also help.

I didn't want to put this in the git logs since it's only a relevant comment until the follow-up changes land.  The poor alignment is limited to snapshot tests (live code has sufficient layout passes to avoid these bugs) so I felt it would not be worth putting in the permanent history.

PR #6547, opened 2019-02-03
[Dialogs] buttons theming and emphasis documentation

galiak11:
### Goal
Documenting emphasis of actions in MDCAlertController. 

### The new content
**High**, **medium** and **low** emphasis actions generate **high**, **medium** and **low** emphasis buttons, respectively.

### Note
Issue: b/118394935
Replacing PR: #6419

### DIFF
--- a/components/Dialogs/README.md
+++ b/components/Dialogs/README.md
@@ -30,6 +30,7 @@ involve multiple tasks.
   <li class="icon-list-item icon-list-item--link">Class: <a href="https://material.io/components/ios/catalog/dialogs/api-docs/Classes/MDCAlertController.html">MDCAlertController</a></li>
   <li class="icon-list-item icon-list-item--link">Class: <a href="https://material.io/components/ios/catalog/dialogs/api-docs/Classes/MDCAlertControllerView.html">MDCAlertControllerView</a></li>
   <li class="icon-list-item icon-list-item--link">Class: <a href="https://material.io/components/ios/catalog/dialogs/api-docs/Classes/MDCDialogPresentationController.html">MDCDialogPresentationController</a></li>
+  <li class="icon-list-item icon-list-item--link">Protocol: <a href="https://material.io/components/ios/catalog/dialogs/api-docs/Protocols/MDCDialogPresentationControllerDelegate.html">MDCDialogPresentationControllerDelegate</a></li>
   <li class="icon-list-item icon-list-item--link">Enumeration: <a href="https://material.io/components/ios/catalog/dialogs/api-docs/Enums.html">Enumerations</a></li>
   <li class="icon-list-item icon-list-item--link">Enumeration: <a href="https://material.io/components/ios/catalog/dialogs/api-docs/Enums/MDCActionEmphasis.html">MDCActionEmphasis</a></li>
 </ul>
@@ -46,7 +47,8 @@ involve multiple tasks.
   - [Typical use: modal dialog](#typical-use-modal-dialog)
   - [Typical use: alert](#typical-use-alert)
 - [Extensions](#extensions)
-  - [Theming extensions](#theming-extensions)
+  - [Theming Extensions](#theming-extensions)
+  - [Theming Actions](#theming-actions)
   - [Using a Themer](#using-a-themer)
 - [Accessibility](#accessibility)
   - [MDCPresentationController Accessibility](#mdcpresentationcontroller-accessibility)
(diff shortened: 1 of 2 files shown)


components/Dialogs/README.md
```diff
@@ -231,6 +233,73 @@ MDCContainerScheme *containerScheme = [[MDCContainerScheme alloc] init];
 ```
 <!--</div>-->
 
+### Theming Actions
+
+Actions in MDCAlertController have emphasis which affects their theming.
+High emphasis actions generate contained buttons, medium emphasis actions generate outlined buttons and low emphasis actions generate text buttons.
```
romoore: Tip: I think it's best to omit these specifics since it binds us to the implementation of emphasis. Until the design guidelines articles have the styling shown publicly, I think it's best to keep the actual style as an implementation detail. That will give us the freedom to change the styling without announcements, deprecations, etc.  What do you think?

components/Dialogs/README.md
```diff
@@ -231,6 +233,73 @@ MDCContainerScheme *containerScheme = [[MDCContainerScheme alloc] init];
...
+
+<div class="article__asset article__asset--screenshot">
+  <img src="docs/assets/dialogButtons.png" alt="An alert presented with a title, body, high-emphasis 'OK' button and low-emphasis 'Cancel' button." width="320">
```
romoore, line 242: Awesome alt text, thanks!

components/Dialogs/README.md
```diff
@@ -231,6 +233,73 @@ MDCContainerScheme *containerScheme = [[MDCContainerScheme alloc] init];
...
+  let scheme = MDCContainerScheme()
+
+  // Create an Alert dialog
+  let alert = MDCAlertController(title: "Button Theming",
+                                 message: "Lorem ipsum dolor sit amet, sit consectetur adipiscing")
+
+  // Add actions with emphases that will generate buttons with the desired appearance. 
+
+  // An example of a high emphasis action (and button appearance):
+  alert.addAction(MDCAlertAction(title:"All Right", emphasis: .high, handler: handler))
```
romoore: Do we have an example from a design article we can use instead?  The "Not Now" and "Later" options seem to have the same meaning.

components/Dialogs/README.md
```diff
@@ -231,6 +233,73 @@ MDCContainerScheme *containerScheme = [[MDCContainerScheme alloc] init];
...
+
+  // An example for a medium exmphasis:
+  alert.addAction(MDCAlertAction(title:"Not Now", emphasis: .medium, handler: handler))
+
+  // Low exmphasis:
+  alert.addAction(MDCAlertAction(title:"Later", emphasis: .low, handler: handler))
+
+  // Make sure to spply theming after all actions are added, so they are themed too!
```
romoore: Nit: Spelling: s/spply/apply/

components/Dialogs/README.md
```diff
@@ -231,6 +233,73 @@ MDCContainerScheme *containerScheme = [[MDCContainerScheme alloc] init];
...
+      [MDCAlertController alertControllerWithTitle:@"Button Theming" 
+                    message:@"Lorem ipsum dolor sit amet, sit consectetur adipiscing"];
+
+  // Add actions with different emphasis, creating buttons with different themes.
+  MDCAlertAction *primaryAction = [MDCAlertAction actionWithTitle:@"All Right"
+                                                          emphasis:MDCActionEmphasisHigh
+                                                           handler:handler];
+  [alert addAction:primaryAction];
+
+  MDCAlertAction *cancelAction = [MDCAlertAction actionWithTitle:@"Not Now"
```
romoore: The Swift snippet has 3 actions and Objective-C has 2. I think using only the two from the Objective-C example is better and then they would be consistent.

components/Dialogs/README.md
```diff
@@ -231,6 +233,73 @@ MDCContainerScheme *containerScheme = [[MDCContainerScheme alloc] init];
...
+                                                         emphasis:MDCActionEmphasisMedium
+                                                          handler:handler];
+  [alert addAction:cancelAction];
+
+  // Make sure to call the themer after all actions are added, so they are themed too!
```
romoore: Looks like the comment might have been left over from Themers?

romoore (approved):
LGTM except the wording of actions might be improved if we draw from an existing Guidelines mock.

components/Dialogs/README.md
galiak11: No, this is specific to actions, to make sure action buttons are uppercase and all other theming is applied.

romoore: Then why are we referring to "Themer" instead of theming extension?

PR #6548, opened 2019-02-04
[States] Material State Scheme (Experimental Prototype)

galiak11:
### Material State Scheme

This is an experimental implementation of a State Scheme. Main implementation ideas:

* Storing state related information outside of themers, in a dedicated data source. This will allow sharing state information across components/themers.
* The State Scheme stores its data in a dictionary, which can be serialized and/or shared with external systems (in the future).
* Using a ColorResource class which represents colors and reduces dependency between subsystems (more info below).
* Colors that don't pass contrast ratio can be overridden and customized by clients, who can pick  specific colors that work with their color schemes.  This concept is demonstrated in the StateScheme initializer. 

### Color Resources

The Color Resource class is a concept borrowed from Android's Resources, which are being used heavily in their theming implementation. 

On iOS, we use Color Resources to serialized/de-serialized colors from a dictionary.  They are used independently by the different schemes, reducing cross dependency between them.  

Internally, the State scheme manages all state-related values as strings in a single dictionary and uses Color Resources to return structured and validated color data which can then be converted to an actual UIColor (and to guarantee the success of this conversion).
This allows the state scheme to return abstract information, like a semantic color name, without knowing the specific value, which will be assigned to it later by a specific color scheme.  

### State Map Implementation

The State Scheme's map stores state-related values which determine appearance of states in components.  State appearance is different for "on-primary" and "on white" themes, therefore a separate map is stored for each theme.

Each theme map is made of a pre-defined list of attributes, which determine its appearance in the different states. Currently supported state-related attributes include: `overlay color, container color, content color, border color and elevation.`  Each of the 5 attributes includes one or more values, one for each state. 

The State Scheme map is a private dictionary, with values stored as strings. This is to allow easy archiving or loading from external sources. The scheme exposes public access methods (getters and setters) for each attribute and their different states.  

The State scheme's getters returns color values as ColorResources, which can then be passed to specific color schemes to generate UIColor values from that scheme. In contrast, ShadowElevation getters return ShadowElevation type. 

The State Scheme also defines 2 setters for each color, one accepting Palette colors, and another which accepts a semantic color name. 

### MDCControlState

A MDCControlState enumeration is used to represent all supported Material States. It is also used to convert between Material States and UIControlStates. Components that define their own State enumerations will have to provide a conversion method between this enumeration and their own's (Alternatively, all components should share this single enumeration).

### StateExampleColorScheme

StateExampleColorScheme is a temporary place to add logic without making changes directly in MDCSemanticColorScheme in. We expect the color scheme to have an additional "overlayColor" and possibly an "outlineColor" (which may already exist on some systems).  Additionally, it defines a method that maps ColorResources to "real" UIColors.

### Public API Considerations

A public API that can be used by themers has not yet been determined, and is not included in this PR.  The current example file demonstrates getting a ColorResource from the StateScheme, and passes it to a ColorScheme to get an final UIColor.

We are considering adding a public API to the color scheme which will return a UIColor (hiding the specific implementation of colors as resources from themers and client apps). For instance:

```
In MDCSemanticColorScheme.h:

- (UIColor *)contentColorForState:(MDCControlState)state  forTheme:(StateTheme)theme;
```

This can be achieved by adding a StateScheme property to the color scheme, and initializing it with the same defaults enumeration (to allow for color-scheme specific overrides of states). For example:

```
In MDCSemanticColorScheme.m:

 init(withDefaults defaults: MDCColorSchemeDefaults) {

    MDCColorScheme = MDCSemanticColorScheme(defaults: .material201804)
    MDCStateScheme = MDCStateScheme(defaults: .material201804)
   ...
}
```

### Issue
b/122521260

### DIFF
--- a/MaterialComponentsExamples.podspec
+++ b/MaterialComponentsExamples.podspec
@@ -1,6 +1,7 @@
 experimental_sources = [
   'components/*/examples/experimental/supplemental/*.{h,m,swift}',
   'components/*/examples/experimental/*.{h,m,swift}',
+  'components/private/State/src/*.{h,m,swift}'
 ]
 
 experimental_headers = [
--- a/components/private/State/src/MDCControlState.swift
+++ b/components/private/State/src/MDCControlState.swift
@@ -1,0 +1,123 @@
+// Copyright 2019-present the Material Components for iOS authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+/**
+ MDCControlState represents all Material states, including the special interactive
+ state, which is any state that is not normal, disabled or error.
+ * Eight distinctive states are supported.
+ * Material's enabled and pressed states are renamed as their UIControlState state names (which
+   is normal and highlighted).
+ * We are still experimenting with state combinations. Whther adding names for state combinations
+   (like: selectedHighlighted) is needed, is stil in debate.
+ */
+struct MDCControlState: OptionSet, CustomStringConvertible {
+
+  /// Material names of states
+  var name: String {
+    switch self {
+    case .normal:      return "enabled"
+    case .highlighted: return "pressed"
+    case .selected:    return "selected"
+    case .active:      return "active"
+    case .focused:     return "focused"
+    case .dragged:     return "dragged"
+    case .disabled:    return "disabled"
+    case .error:       return "error"
+
+    case .interactive: return "interactive"
+
+    case .selectedHighlighted: return "selected+pressed"
+    case .selectedFocused:     return "selected+focused"
+    case .selectedDragged:     return "selected+dragged"
+
+    case .activeHighlighted: return "active+pressed"
+    case .activeFocused:     return "active+focused"
+    case .activeDragged:     return "active+dragged"
+
+    default: return "<undefined>"
+    }
+  }
+
+  var description: String {
+    var state = ""
+    if self.contains(.normal) {state += " enabled"}
+    if self.contains(.highlighted) {state += " pressed"}
+    if self.contains(.selected) {state += " selected"}
+    if self.contains(.active) {state += " active"}
+    if self.contains(.focused) {state += " focused"}
+    if self.contains(.dragged) {state += " dragged"}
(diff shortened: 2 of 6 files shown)


components/private/State/src/MDCControlState.swift
```diff
@@ -0,0 +1,123 @@
...
+/**
+ MDCControlState represents all Material states, including the special interactive
+ state, which is any state that is not normal, disabled or error.
+ * Eight distinctive states are supported.
+ * Material's enabled and pressed states are renamed as their UIControlState state names (which
+   is normal and highlighted).
+ * We are still experimenting with state combinations. Whther adding names for state combinations
+   (like: selectedHighlighted) is needed, is stil in debate.
+ */
+struct MDCControlState: OptionSet, CustomStringConvertible {
```
andrewoverton, line 26: Whatever we ultimately merge in won't be in Swift though right?

galiak11: The public ColorScheme API will continue to be in objective C. Examples will remain in Swift. I'm not sure yet about non-public API.

MaterialComponentsExamples.podspec
```diff
@@ -1,6 +1,7 @@
 experimental_sources = [
   'components/*/examples/experimental/supplemental/*.{h,m,swift}',
   'components/*/examples/experimental/*.{h,m,swift}',
+  'components/private/State/src/*.{h,m,swift}'
```
andrewoverton, line 4: Should all the state stuff go here or should it go in `components/schemes/` with all the other subsystem stuff?

components/private/State/src/MDCControlState.swift
andrewoverton: Would an Objective-C MDCControlState be an NS_Options? Or an NSObject?

components/private/State/src/StateScheme.swift
```diff
@@ -0,0 +1,407 @@
...
+    switch self {
+    case .outline: return .onWhite
+    case .elevated: return .onWhite
+    default: return nil
+    }
+  }
+}
+
+// State dictionary is a list of attributes. An attribute is a list of resources:
+typealias ResourceDictionary = [String:String]
```
andrewoverton, line 34: I worry about the robustness of a `typealiased` dictionary, could it be something else?

MaterialComponentsExamples.podspec
galiak11: Yes, I moved it to components/schemes in the alpha version (see #6585) which is going to land in develop.

components/private/State/src/MDCControlState.swift
galiak11: All implementation will be in objective C, at least until swift 5 is out, to maintain backward compatibility.  MDCControlState will be written in objective C, and is going to be an option-set/bit-mask. See #6585.

components/private/State/src/StateScheme.swift
galiak11: It was easier to keep it a String in this implementation, but for the actual implementation, I agree this might be too limiting. I will add it as a todo item for the Alpha implementation.

PR #6550, opened 2019-02-04
[BottomNavigation] Vary badges in snapshot tests.

romoore:
The BottomNavigation item view snapshot tests were always using the same
badge value. Instead, we should have some variation so we can capture
different alignment renderings.

Prework for #6520


components/BottomNavigation/tests/snapshot/MDCBottomNavigationItemViewSnapshotTests.m
```diff
@@ -65,154 +68,159 @@ - (void)setUp {
...
+  self.itemView.badgeValue = badgeValue;
 }
 
 #pragma mark - Varied widths
 
-- (void)testNarrowWidthTypicalHeightLongTitleStackedLTR {
+- (void)testNarrowWidthTypicalHeightLongTitleEmptyBadgeStackedLTR {
   // When
   self.itemView.titleBelowIcon = YES;
   self.itemView.frame = CGRectMake(0, 0, kWidthNarrrow, kHeightTypical);
```
galiak11, line 92: How do you choose typical/narrow/short for size?

galiak11 (approved):

components/BottomNavigation/tests/snapshot/MDCBottomNavigationItemViewSnapshotTests.m
romoore: Are you asking how I got the values?
They're from [the Material Guidelines article for Bottom Navigation](https://material.io/design/components/bottom-navigation.html#specs).  The "narrow" width is below minimum because it is 320 / 5, which is an older iPhone in landscape with 5 items.

Or are you asking how I choose the combination for each test? I believe I varied them so cover each combination of width + height (or at least most).

Issue #6546, opened 2019-02-02
{kokoro} website-generation job is failing


romoore:
Looks like `--without-node` is not a valid option for brew.

```
Homebrew 2.0.0
Homebrew/homebrew-core (git revision 98017e; last commit 2019-02-03)
Homebrew/homebrew-cask (git revision e126e; last commit 2019-02-02)
Usage: brew install [options] formula

Install formula.

formula is usually the name of the formula to install, but it can be specified
in several different ways.

    -d, --debug                      If brewing fails, open an interactive
                                     debugging session with access to IRB or a
                                     shell inside the temporary build directory
        --env                        If std is passed, use the standard build
                                     environment instead of superenv.If super
                                     is passed, use superenv even if the formula
                                     specifies the standard build environment.
        --ignore-dependencies        Skip installing any dependencies of any
                                     kind. If they are not already present, the
                                     formula will probably fail to install.
        --only-dependencies          Install the dependencies with specified
                                     options but do not install the specified
                                     formula.
        --cc                         Attempt to compile using provided
                                     compiler. compiler should be the name
                                     of the compiler's executable, for instance
                                     gcc-7 for GCC 7. In order to use LLVM's
                                     clang, use llvm_clang. To specify the
                                     Apple-provided clang, use clang. This
                                     parameter will only accept compilers that
                                     are provided by Homebrew or bundled with
                                     macOS. Please do not file issues if you
                                     encounter errors while using this flag.
    -s, --build-from-source          Compile the specified formula from source
                                     even if a bottle is provided. Dependencies
                                     will still be installed from bottles if
                                     they are available.
        --force-bottle               Install from a bottle if it exists for the
                                     current or newest version of macOS, even if
                                     it would not normally be used for
                                     installation.
        --include-test               Install testing dependencies required to
                                     run brew test.
        --devel                      If formula defines it, install the
                                     development version.
        --HEAD                       If formula defines it, install the HEAD
                                     version, aka. master, trunk, unstable.
        --fetch-HEAD                 Fetch the upstream repository to detect if
                                     the HEAD installation of the formula is
                                     outdated. Otherwise, the repository's HEAD
                                     will be checked for updates when a new
                                     stable or development version has been
                                     released.
        --keep-tmp                   Don't delete the temporary files created
                                     during installation.
        --build-bottle               Prepare the formula for eventual bottling
                                     during installation.
    -f, --force                      Install without checking for previously
                                     installed keg-only or non-migrated
                                     versions.
[...]

PR #6551, opened 2019-02-04
[BottomNavigation] Use chrome_reader in example.

romoore:
The typical use example for Bottom Navigation should use the icon
`chrome_reader_mode` to make it easier to compare the implementation to
[the Material.io guidelines article for Bottom Navigation](https://material.io/design/components/bottom-navigation.html#behavior).

|Before|After|
|---|---|
|bn-typical-search-before|bn-typical-reader-after|

Preparation for #6520


romoore:
website job failing due to https://github.com/material-components/material-components-ios/issues/6546

fluxminds (approved):

PR #6559, opened 2019-02-05
[BottomNavigation] Add sizeThatFits: to Item View

romoore:
The MDCBottomNavigationItemView did not implement sizeThatFits: and so it was
not possible for the MDCBottomNavigationBar to interrogate the view for its
appropriate size during layout.

Part of #6520


yarneo (approved):
If there could be an example that checks the flow for titleBelowIcon = NO that would be great, thanks

PR #6568, opened 2019-02-06
v76.1.0

ianegordon:
Release PR

DO NOT SUBMIT


ianegordon:
Verified all committers / authors have signed the CLA.

CHANGELOG.md
```diff
@@ -1,3 +1,49 @@
+# 76.1.0
+
+Add additional snapshot tests and mark outdated theming API as deprecated.
```
romoore, line 3: Tip: I don't know that we should call these APIs "outdated", we are just moving toward a newer approach.  Maybe the more neutral, "...and marking Dialogs Themer APIs as deprecated." would work?

CHANGELOG.md
```diff
@@ -1,3 +1,49 @@
...
+
+## New deprecations
+
+The previous theming API is being deprecated in favor of MDCContainerScheme.
```
romoore, line 7: In favor of Theming Extensions? Maybe link to the docs on GitHub?

romoore (approved):

CHANGELOG.md
romoore: Docs here: 
https://github.com/material-components/material-components-ios/blob/develop/components/schemes/README.md

romoore: Why wasn't this feedback addressed before merging?

CHANGELOG.md
ianegordon: I think the newer approach validates calling the old API outdated, but I can definitely see the negative connotation.
s/outdated/previous/

CHANGELOG.md
ianegordon: Follow-up PR incoming.

PR #6571, opened 2019-02-07
v76.1.1

ianegordon:
Pull in some missing feedback.
Ensure kokoro is working as expected.


CHANGELOG.md
```diff
@@ -1,10 +1,14 @@
+# 76.1.1
+
+Fix build failures.
+
 # 76.1.0
 
-Add additional snapshot tests and mark outdated theming API as deprecated.
+Add additional snapshot tests and mark previous Dialog Themer APIs as deprecated.
```
romoore, line 7: Changing these won't have any effect on the tagged 76.1.0 release and would make the CHANGELOG no longer be an "append" operation for each release. I suggest we leave the old version as-is and focus this release on any fixes to Cocoapods.

components/BottomSheet/tests/unit/BottomSheetShapeThemerTests.swift
```diff
@@ -15,7 +15,6 @@
 import XCTest
 import MaterialComponents.MaterialBottomSheet
 import MaterialComponents.MaterialBottomSheet_ShapeThemer
-import MaterialComponents.MaterialShapeLibrary
```
romoore, line 18: Isn't this import still being used for `MDCCornerTreatment.corner(withRadius: 3)`? Did you mean to switch to the "New" version?

romoore (changes requested):
I don't think code should rely on transitive imports for symbols it's using directly.

CHANGELOG.md
ianegordon: SGTM. Done.

components/BottomSheet/tests/unit/BottomSheetShapeThemerTests.swift
ianegordon: Relying on a non-required import was causing pod lib lint to fail.
It works as expected now.
We may need to switch to the New version in the future, but it don't appear to be required yet.

Xcode tests pass as expected as do cocoapods.

```
$ pod lib lint --subspec=BottomSheet MaterialComponents.podspec 
MaterialComponents passed validation.
```

romoore: I believe we should be using `MaterialComponent.MaterialShapeLibraryNew`.  That's where Yarden put the source until we can remove the original (private) ones.

romoore:
How was this landed without a review and approval?

romoore:
This change cannot build - the Presubmits failures indicated this:

```
Use --sandbox_debug to see verbose messages from the sandbox
components/Cards/tests/unit/MDCCardShapeThemerTests.swift:26:55: error: type 'MDCCornerTreatment' has no member 'corner'
    shapeScheme.mediumComponentShape.topRightCorner = MDCCornerTreatment.corner(withRadius: 3)
```

romoore:
This release will need to be overwritten and the changes reversed and corrected. The Cocoapods failures were determined offline to be related to the `.kokoro` script updating CocoaPods automatically. Although MDC-iOS does not currently support Cocoapods 1.6, the current release is broken due to missing imports.

PR #6577, opened 2019-02-08
[Releasing] Require presubmits to pass

romoore:
The instructions didn't previously require presubmit checks to pass. Making it clear that this is a required step before merging.

### DIFF
--- a/contributing/releasing.md
+++ b/contributing/releasing.md
@@ -271,7 +271,7 @@ go-ahead from the following clients:
 
 ## Merge the release candidate branch
 
-Once the release-candidate has passed all tests by clients, you may merge the release into the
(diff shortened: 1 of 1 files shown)


yarneo (approved):

ianegordon:
While I very much support the spirit of this change, I want to point out that **this PR**, with a simple text change to a document, did **NOT** pass all the presubmits.

randallli (approved):

romoore:
@ianegordon It passed all required presubmits. The website generator is unsupported, broken, and it's not clear if we should continue using it. It was marked as not Required and https://github.com/material-components/material-components-ios/issues/6546 is tracking the problem.

I've updated the wording to address your concerns.

PR #6580, opened 2019-02-08
[Cocoapods] Upgrade to Cocoapods 1.6.0

### LINKED ISSUE #6356: Move to Cocoapods 1.6
When the time comes (probably not now) we should update to Cocoapods 1.6. Leading up to that, I want to clarify some of the impact of https://github.com/material-components/material-components-ios/pull/5966 and give some context on remaining work to do.

#5966 was titled "Make pod install work on Cocoapods 1.6.0.beta.2", because that was originally what I thought it could do. By the time it was merged a more accurate title for it might've been "Greatly reduce the amount of work needed to get our workspace working optimally with Cocoapods 1.6.0.beta.2, while maintaining adequate functionality with Cocoapods 1.5.0, and also restoring previously broken Swift IDE functionality in Xcode".

Here are some things I think we need to do when we move to 1.6.0 for real:
1. Re-enable the new build system. When the new build system was introduced we quickly learned that it didn't work with our project and we opted out of it. I found in my explorations for #5966 that the reverse was true with 1.6.0--the legacy BUILD system no longer worked, but the new one did, provided I did the next two things in this list.
2. Uncomment [this line](https://github.com/material-components/material-components-ios/blob/develop/catalog/Podfile#L5).
3. Make sure all the validations work on 1.6.0. At one point, while on 1.6, I had all the validations passing for MaterialComponents and MaterialComponentsBeta. I didn't look at the MaterialComponentsExamples podspec I don't think. Anyway, when these things were passing for 1.6.0 you could command click on Swift modules and it would actually take you to them instead of showing the question mark. It was really exciting. Unfortunately, I remember having to table some of the changes that went into that to get things working on 1.5 again. When we move past 1.5, we can go back to optimizing for 1.6.

---
#### Internal data
- Associated internal bug: [b/122969056](http://b/122969056)

andrewoverton:
These changes prepare our repo for Cocoapods 1.6.0. This work is a continuation of https://github.com/material-components/material-components-ios/pull/5966.

Our `post_install` step in the Podfile modifies xcconfig files in the Target Support Files directory. When our CI updated to Cocoapods 1.6.0 `pod install` started failing because the xcconfig files we expected to be there were no longer there. This wasn't initially an issue for me locally because I didn't delete the Pods directory after I installed Cocoapods 1.6.0, so all the files left by `pod install` commands on Cocoapods 1.5 that our post_install step relied on were still there. When I deleted the Pods directory I began to see the issue. Updating the `post_install` steps to reflect the target changes made by Cocoapods 1.6.0 seems to have resolved the issue.

General note: In Cocoapods 1.5.0 all the various test_specs in a given podspec file would be grouped into one test target. In Cocoapods 1.6.0, each test_spec in a podspec gets its own test target. To make sure we are able to run all the tests with a ⌘U, I updated the Dragons and Catalog schemes to include these many new test targets.

Note for release engineer: Maybe we should add a note to the changelog about this? Something like this:

>This version updates our project to Cocoapods 1.6.0. While a degree of backwards compatibility exists, for best results consider updating to Cocoapods 1.6.0.

Closes #6356.

### DIFF
--- a/.kokoro
+++ b/.kokoro
@@ -16,7 +16,7 @@
 
 # Fail on any error
 set -e
-COCOAPODS_VERSION="1.5.3"
+COCOAPODS_VERSION="1.6.0"
 
 if [ -n "$KOKORO_BUILD_NUMBER" ]; then
   repo_dir='github/repo'
--- a/catalog/MDCCatalog.xcodeproj/project.pbxproj
+++ b/catalog/MDCCatalog.xcodeproj/project.pbxproj
@@ -409,7 +409,7 @@
 			files = (
 			);
 			inputPaths = (
-				"${SRCROOT}/Pods/Target Support Files/Pods-MDCCatalog/Pods-MDCCatalog-frameworks.sh",
+				"${PODS_ROOT}/Target Support Files/Pods-MDCCatalog/Pods-MDCCatalog-frameworks.sh",
 				"${BUILT_PRODUCTS_DIR}/CatalogByConvention/CatalogByConvention.framework",
 				"${BUILT_PRODUCTS_DIR}/MDFInternationalization/MDFInternationalization.framework",
 				"${BUILT_PRODUCTS_DIR}/MDFTextAccessibility/MDFTextAccessibility.framework",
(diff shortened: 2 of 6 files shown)


romoore (approved):
Please just add a comment for this week's release engineer to paste into the release CHANGELOG.  Specifically, I believe this is backwards compatible for users but for *contributors* they should be using Pods 1.6+.

ianegordon (approved):
Please shout this change loudly as it has a high likelihood of breaking people who are unaware.

romoore:
@andrewoverton Told us offline that it should be compatible with Pods 1.5 for consumers but that contributors would need to upgrade to 1.6 to receive all extra warnings in XCode.

romoore:
I'm no longer seeing our Snapshot tests in the Unit Tests explorer. Is there another way to see and run the Snapshot tests in Xcode?

PR #6586, opened 2019-02-10
[BottomNavigation] Extend ripple to bar bounds.

romoore:
The ripple effect was being constrained to the layout container of the
BottomNavigationBar. Instead, it should be allowed to fill the visible
bar. It also correct a bug where the ripple effect would extend beyond
the bounds of the bar because the item views were taller than the bar
view.

The effect was most prominent on newer iPhones where `safeAreaInsets` pushes the item views away from the edge of the screen.

## iPhone XS Max (Simulator)

|Before|After|
|---|---|
|bn-ripple-ipx-develop|bn-ripple-ipx-fixed|


components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -101,17 +102,22 @@ - (void)commonMDCBottomNavigationBarInit {
...
+  _itemsLayoutView.clipsToBounds = NO;
+  [self addSubview:_inkClippingView];
+  [_inkClippingView addSubview:_itemsLayoutView];
 #if defined(__IPHONE_10_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_10_0)
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wunguarded-availability"
 #pragma clang diagnostic ignored "-Wtautological-pointer-compare"
   if (&UIAccessibilityTraitTabBar != NULL) {
-    _containerView.accessibilityTraits = UIAccessibilityTraitTabBar;
+    _itemsLayoutView.accessibilityTraits = UIAccessibilityTraitTabBar;
```
codeman7, line 120: _Tip:_ Should we mark the `inkClippingView` as `accessibilityElement = NO`?

components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -164,27 +172,27 @@ - (void)layoutLandscapeModeWithBottomNavSize:(CGSize)bottomNavSize
...
+      [self sizeitemsLayoutViewItemsDistributed:NO
                             withBottomNavSize:bottomNavSize
                                containerWidth:containerWidth];
       self.titleBelowItem = YES;
       break;
   }
 }
 
-- (void)sizeContainerViewItemsDistributed:(BOOL)itemsDistributed
+- (void)sizeitemsLayoutViewItemsDistributed:(BOOL)itemsDistributed
```
codeman7: _Tip:_ Shouldn't it be `sizeItemsLayoutViewItemsDistributed..`? `sizeItems` ~`sizeitems`~?

codeman7 (approved):

components/BottomNavigation/src/MDCBottomNavigationBar.m
romoore: For UIView, the default value is `NO`. 

From UIAccessibility.h:

```objc
/*
 Return YES if the receiver should be exposed as an accessibility element. 
 default == NO
 default on UIKit controls == YES 
 Setting the property to YES will cause the receiver to be visible to assistive applications. 
 */
@property (nonatomic) BOOL isAccessibilityElement;
```

components/BottomNavigation/src/MDCBottomNavigationBar.m
romoore: Good catch - that was a find/replace mistake.

PR #6589, opened 2019-02-10
[BottomNavigation] Add blur example

romoore:
Adds a new Dragons example to demonstrate the blur effect available for
Bottom Navigation Bar.

dragons-bn-blur-example

Follow-up to #6587
Part of #5217


romoore:
This depends on #6587 and won't build until then.

andrewoverton (approved):

PR #6594, opened 2019-02-11
[BottomNavigation] Badge size grows with value

romoore:
When the badge value grows to a larger string than the initial value,
the badge view will now correctly have its bounds updated.

|Before|After|
|---|---|
|bn-badge-grow-before|bn-badge-grow-after|


yarneo (approved):
Much Wow, better visual result by removing code. Good job.

jverkoey (approved):

codeman7 (approved):

PR #6602, opened 2019-02-12
[BottomNavigation] Fix badge position and font.

romoore:
The item badge font was too large based on the mocks provided on the
Material.io guidelines article.

It was also incorrectly positioned when the number of characters increased.
The position of the badge is determined by its top-leading corner and it
grows with the label toward its trailing edge. The previous alignment
accidentally positioned the label's center at an offset from the icon
frame.

|State|Screenshot|
|---|---|
|Before|bn-catalog-develop|
|After|bn-catalog-after|

## Animation comparing this change with Material Guidelines
bn-badge-compare-guidelines

Part of #6520


components/BottomNavigation/src/private/MDCBottomNavigationItemView.m
```diff
@@ -27,7 +27,14 @@
 static const CGFloat kMaxSizeDimension = 1000000;
 static const CGFloat MDCBottomNavigationItemViewInkOpacity = (CGFloat)0.150;
 static const CGFloat MDCBottomNavigationItemViewTitleFontSize = 12;
-static const CGFloat kMDCBottomNavigationItemViewBadgeYOffset = 4;
+
+// The fonts available on iOS differ from that used on Material.io.  When trying to approximate
+// the position on iOS, it seems like a horizontal inset of 10 points looks pretty close.
+static const CGFloat kBadgeXOffsetFromIconEdgeWithTextLTR = -8;
+
+// However, hen the badge has no visible text, its horizontal center should be 1 point inset from
```
romoore: Nit/spelling: s/hen/when/

codeman7 (approved):

PR #6604, opened 2019-02-12
[BottomNavigation] Fix "centered" alignment.

### LINKED ISSUE #6520: [BottomNavigation] Center alignment does not match material.io guidelines

romoore:
The `.centered` alignment now correctly groups the items together in the
center of the screen.  It makes each item as wide as the widest, up to
the Material Guidelines spec's maximum width of 168 points.  The
`.centered` alignment is only used in horizontally-Regular environments
but provides a visually-appealing alternative to
`justifiedAdjacentTitles`.

Previously, using `.centered` in some circumstances would spread the
items over too wide a horizontal area, leading to a nearly "justified"
look. When grouped correctly, it would sometimes fail to attempt to make
the items the default 120 points wide, instead making them less than 90
points wide.

Fixes #6520


romoore:
Merge after #6603 and #6602

randallli (approved):

PR #6611, opened 2019-02-12
[Cocoapods] Fix snapshot testing after Cocoapods 1.6.0 update

### LINKED ISSUE #6609: Snapshot tests aren't running after the Cocoapods 1.6.0 update
Snapshot tests are no longer run with the other tests following https://github.com/material-components/material-components-ios/pull/6580

---
#### Internal data
- Associated internal bug: [b/124382701](http://b/124382701)

andrewoverton:
Snapshot tests stopped running after https://github.com/material-components/material-components-ios/pull/6580. I think this PR fixes it.

Cocoapods 1.6.0 creates a selected "white lego" test target for each unit test `test_spec` in a podspec. However, it did not do so for our snapshot tests. Instead, it created a single unselectable  `MaterialComponentSnapshotTests` build target with the red and white target icon.

I moved the snapshot testing related files like `MaterialSnapshot.h` to a regular `subspec`, as opposed to a `test_spec`. After this change, Cocoapods 1.6.0 created selected "white lego" snapshot testing targets, just like with the unit tests. I guess it's bad to have test_specs within subspecs that don't have any source files?

EDIT: What I described above was only the beginning. I had to do a lot more to get everything working. Because Cocoapods 1.6.0 makes each `test_spec` into its own test target I had to create additional Swift dummy files to evade the [-lswiftSwiftOnoneSupport](https://github.com/material-components/material-components-ios/blob/develop/components/Cards/tests/snapshot/Dummy.swift#L17) linker error that Rami discovered when first adding Snapshot testing. I also had to add these Swift files to BUILD files. We might be able to get rid of these if we only depend on `iOSSnapshotTestCase/Core` (which is the same thing as `iOSSnapshotTestCase` but without Swift support) and commit to writing only Obj-C snapshot tests.

I also had to add two Objective-C dummy source files at the top level of the `MaterialComponentsSnapshotTests` podspec, and to the `private/Snapshot` BUILD file. I don't know why these source files were required. I arrived at this after much trial and error, which is not ideal. I want to try to get rid of them if possible, but I think they should go in for now, just so we can get this stuff working again.

Lastly, I had to disable bitcode everywhere. For some reason, the fact that `iOSSnapshotTestCase` explicitly disables bitcode is now a problem for us, where it wasn't before. From what I read on various blog posts, if an app target's dependency explicitly disables bitcode, an app target cannot enable it. Again, I don't know why this is only now a problem. This change on my part is the result of reading linker error messages and taking their advice. Right now I'm just disabling it everywhere in the `post_install`. We might be able to get away with only disabling it in MDCDragons and MDCDragons and just committing that to version control. I plan to look into that after this gets merged in.

I'm hoping that some of the changes in this PR can be undone at a later date, preferably soon. There is a chance that only depending on `iOSSnapshotTestCase/Core` and removing our snapshot tests from the Dragons and Catalog schemes and putting them in their own app target _could_ go a long way towards doing this.

Closes #6609


MaterialComponentsSnapshotTests.podspec
```diff
@@ -10,73 +10,80 @@ Pod::Spec.new do |s|
   s.requires_arc = true
   s.dependency 'MaterialComponents'
   s.dependency 'MaterialComponentsBeta'
+  s.dependency 'iOSSnapshotTestCase', '2.2.0'
+  s.pod_target_xcconfig = { 'ENABLE_BITCODE' => 'NO' }
+
+  # These source dummies are required at the top level to build successfully with Cocoapods 1.6.0.
+  # Swift dummy source files are required at the subspec level.
+  s.source_files = "components/private/Snapshot/src/SourceDummies/*.{h,m}"
```
romoore, line 18: Tip: Please file a bug to try and use the actual `src` in the future.

components/Cards/tests/snapshot/CardsSnapshotTestsSwiftSourceDummy.swift
```diff
@@ -0,0 +1,27 @@
+// Copyright 2018-present the Material Components for iOS authors. All Rights Reserved.
```
romoore, line 1: 2019?

components/Chips/tests/snapshot/ChipsSnapshotTestsSwiftSourceDummy.swift
```diff
@@ -0,0 +1,27 @@
+// Copyright 2018-present the Material Components for iOS authors. All Rights Reserved.
```
romoore, line 1: 2019?

components/Ripple/tests/snapshot/RippleSnapshotTestsSwiftSourceDummy.swift
```diff
@@ -0,0 +1,27 @@
+// Copyright 2018-present the Material Components for iOS authors. All Rights Reserved.
```
romoore, line 1: 2019?

components/TextFields/tests/snapshot/TextFieldsSnapshotTestsSwiftSourceDummy.swift
```diff
@@ -0,0 +1,27 @@
+// Copyright 2018-present the Material Components for iOS authors. All Rights Reserved.
```
romoore, line 1: 2019?

components/private/Snapshot/src/SourceDummies/MDCGlobalSnapshotSourceDummy.h
```diff
@@ -0,0 +1,18 @@
+// Copyright 2018-present the Material Components for iOS authors. All Rights Reserved.
```
romoore, line 1: 2019?

components/private/Snapshot/src/SourceDummies/MDCGlobalSnapshotSourceDummy.m
```diff
@@ -0,0 +1,19 @@
+// Copyright 2018-present the Material Components for iOS authors. All Rights Reserved.
```
romoore, line 1: 2019?

romoore (approved):

andrewoverton:
Thanks for the PR review @romoore, I'm going to correct the copyright and try to improve things generally in a subsequent PR

PR #6618, opened 2019-02-14
[Cocoapods] Fixing Snapshot Tests

### LINKED ISSUE #6609: Snapshot tests aren't running after the Cocoapods 1.6.0 update

### LINKED ISSUE #6700: Snapshot Testing Not Working After Cocoapods 1.6.0 Update
After the Cocoapods 1.6.0 update snapshot tests are not being run in Xcode. There are no snapshot test targets included in the MDCDragons/MDCCatalog scheme.

andrewoverton:
Initially this PR was supposed to consist of a revert of #6619 (which reverted #6611) with some small tweaks to fix whatever was wrong in #6611. It turned into something else! I had a hard time figuring out a way to both run MDCDragons/MDCCatalog with ⌘R _and_ run snapshot tests within either of the two apps' schemes. The only way I was able to manage it was by abandoning our convention for podspecs (only for snapshot tests), which relies heavily on nested subspecs and test_specs. In its place, I wrote a probably not very idiomatic ruby module that aims to allow for all snapshot testing source files to funnel into a single unit test target but still give us some flexibility in how we configure snapshot tests on a per component basis.

Closes #6700.


yarneo (approved):

MaterialComponentsSnapshotTests.podspec
```diff
@@ -1,139 +1,88 @@
...
-    component.ios.deployment_target = '8.0'
-    component.test_spec 'tests' do |tests|
-      tests.test_spec 'snapshot' do |snapshot_tests|
-        snapshot_tests.requires_app_host = true
-        snapshot_tests.source_files = "components/#{component.base_name}/tests/snapshot/*.{h,m,swift}"
-        snapshot_tests.dependency "MaterialComponentsSnapshotTests/private/Snapshot"
-      end
-    end
-  end
+module SnapshotPodspecHelper
```
codeman7, line 1: This file is absolutely amazing!

MaterialComponentsSnapshotTests.podspec
```diff
@@ -1,139 +1,88 @@
...
+  # launch because XCTest isn't found.
+  s.source_files = ["components/private/Snapshot/src/SourceDummies/*.{h,m}"]
 
-  s.subspec "private" do |private_spec|
-    private_spec.test_spec "Snapshot" do |snapshot|
-      snapshot.ios.deployment_target = '8.0'
-      snapshot.source_files = "components/private/#{snapshot.base_name}/src/*.{h,m,swift}"
-      snapshot.dependency 'iOSSnapshotTestCase', '2.2.0'
-    end
+  s.test_spec "SnapshotTests" do |snapshot_tests|
```
romoore, line 80: So they're basically all bundled into one big target? If it works, sounds good!

MaterialComponentsSnapshotTests.podspec
```diff
@@ -1,139 +1,88 @@
...
-        snapshot_tests.source_files = "components/#{component.base_name}/tests/snapshot/*.{h,m,swift}", "components/#{component.base_name}/tests/snapshot/supplemental/*.{h,m,swift}"
-        snapshot_tests.resources = "components/#{component.base_name}/tests/snapshot/resources/*"
-        snapshot_tests.dependency "MaterialComponentsSnapshotTests/private/Snapshot"
-        snapshot_tests.dependency "MDFInternationalization"
-      end
-    end
-  end
+Pod::Spec.new do |s|
+  s.name         = "MaterialComponentsSnapshotTests"
+  s.version      = "77.0.0"
```
romoore: Should be 78.0.0 now. :-\

components/Cards/BUILD
```diff
@@ -182,21 +182,9 @@ mdc_snapshot_objc_library(
     ],
 )
 
-swift_library(
```
romoore, line 185: This looks unrelated - is this just a clean up for empty sources?

components/Ripple/BUILD
```diff
@@ -19,7 +19,10 @@ load(
     "mdc_objc_library",
     "mdc_public_objc_library",
     "mdc_unit_test_suite",
+    "mdc_snapshot_objc_library",
```
romoore: Can these BUILD file changes be moved to a separate PR or are they somehow required for the Podspec changes?

components/Cards/BUILD
andrewoverton: I got rid of the Cards Swift source dummy file so I had to make this change otherwise CI wouldn't pass

components/Ripple/BUILD
andrewoverton: In a previous iteration of this PR I had added a Swift source dummy to `components/Ripple/src/` so I had to add the snapshot tests to the BUILD otherwise CI wouldn't pass. I can get rid of this now I guess...

PR #6694, opened 2019-02-22
[Dialogs] Use test Typography scheme

romoore:
Using a test-specific Typography scheme to simplify custom scheme
creation. Removes a unit test that was testing the unit test setUp
method.


yarneo (commented):
What is MaterialComponentsTestingSupport?

romoore:
@yarneo Do you mean the podspec?
https://github.com/material-components/material-components-ios/blob/develop/MaterialComponentsTestingSupport.podspec

Could you point out which line you're asking about?

yarneo (approved):

Issue #6622, opened 2019-02-15
[Dialogs] Internal issue: b/124469770


material-mirror:
This was filed as an internal issue. If you are a Googler, please visit [b/124469770](http://b/124469770) for more details.

---
#### Internal data
- Associated internal bug: [b/124469770](http://b/124469770)
- Blocked by: https://github.com/material-components/material-components-ios/issues/6751

material-mirror:
The internal issue [b/124469770](http://b/124469770) is now closed. This issue is being closed as a result.

PR #6696, opened 2019-02-22
[Dialogs] Fix presented corner radius in custom view controllers

galiak11:
### Background

MDCDailogPresentationController renders the shadow layer of a presented view controller.  Its theming extension themes all presentation aspects, but it cannot ensure that the corner radius of the presentation controller matches the corner radius of the view it presents.

I've been experimenting with 3 different solutions to this problem, eventually deciding the last one is the optimal.  Feedback welcome!

### Solution A [REJECTED]

Apply a presentation theme directly to the custom view controller. Unlike MDCDailogPresentationController extension, a UIViewController extension has access to both the presented view and the presentation controller, and can ensure the corner radius is set on both.

**Rejected because**: Adding a theming extension to a UIKit class is less optimal.

### Solution B [REJECTED]

Going back to a MDCDailogPresentationController theming extension.
This solution sets the corner radius of the presentedViewController in MDCDialogPresentationController theming extension (thanks @wenyuzhang666 and @ianegordon  for your helpful input).

My concern with this approach is that by accessing the view property of the presentedViewController we may trigger "viewDidLoad" inside the themer.

**Rejected because**: Its a little brittle. It is not clear what effect view.layer.cornerRadius has over mdc_presentationController.dialogCornerRadius.

**Implementation in commit**: "Fix presented corner radius in custom view controllers"

### Solution C [ACCEPTED]

Changing the way MDCDailogPresentationController's dialogCornerRadius is used to determine corner radius.  This solution uses it to conditionally set the corner radius for both its shadow layer and the presented view layer.

We ensure internally that these 2 match by setting the corner radius on both - dynamically when the presentation animation begins (in MDCDailogPresentationController). 

Which corner radius "wins" is determined by the following logic:

dialogCornerRadius is configured by default to be disabled (it has a default value of -1), which allows the presented view's cornerRadius to take precedence and be used for the shadow layer.

When dialogCornerRadius has a non-negative value, typically when set by a themer, it becomes the preferred value that's applied to both the shadow layer and the presented view layer.

**Accepted because**:
* The corner radius of the shadow layer is guaranteed to always match the view's without any additional steps (as long as neither of them is changed again after the end of the run loop after the view is presented).
* Default UIKit behavior is preserved when a themer is not called: setting view.layer.cornerRadius alone will result in correct radius for both shadow and presented view.
* The themer can override the corner radius by setting dialogCornerRadius alone, and it will result in correct radius for both shadow and presented view.

Note: Accessing the presentedViewController.view property did not seem to be a problem. The view seems to be always be loaded by the time [MDCDailogPresentationController presentationTransitionWillBegin] is called.

**Implementation in commit**: "Changing purpose of dialogCornerRadius to fix the presented corner-radius issue" (or final diff of this PR).

### Before and After

Note: this Before & After view works for all 3 solutions.

Before (Incorrect shadow shape with a custom corner radius value of 24):

dialogbefore

After (Correct shadow shape with a custom corner radius value of 24):

dialogafter

### Issues

#6622
b/124469770
Followup issue: b/126934531

### DIFF
--- a/components/Dialogs/examples/DialogsAlertExampleViewController.m
+++ b/components/Dialogs/examples/DialogsAlertExampleViewController.m
@@ -127,36 +127,6 @@ static NSString *const kReusableIdentifierItem = @"cell";
   [self presentViewController:materialAlertController animated:YES completion:NULL];
 }
 
-- (IBAction)didTapNondismissingAlert {
-  NSString *titleString = @"This alert requires an action.";
-  NSString *messageString = @"You can't dismiss it by tapping the background. You must choose "
-                             "one of the actions available.";
-
-  MDCAlertController *materialAlertController =
-      [MDCAlertController alertControllerWithTitle:titleString message:messageString];
-  [self themeAlertController:materialAlertController];
-
-  MDCAlertAction *agreeAction = [MDCAlertAction actionWithTitle:@"AGREE"
-                                                        handler:^(MDCAlertAction *action) {
-                                                          NSLog(@"%@", @"AGREE pressed");
-                                                        }];
-  [materialAlertController addAction:agreeAction];
-
-  MDCAlertAction *disagreeAaction = [MDCAlertAction actionWithTitle:@"DISAGREE"
-                                                            handler:^(MDCAlertAction *action) {
-                                                              NSLog(@"%@", @"DISAGREE pressed");
-                                                            }];
-  [materialAlertController addAction:disagreeAaction];
-
-  // This code accesses the presentation controller and turns off dismiss on background tap.
-  MDCDialogPresentationController *presentationController =
-      materialAlertController.mdc_dialogPresentationController;
-  [presentationController applyThemeWithScheme:self.containerScheme];
-  presentationController.dismissOnBackgroundTap = NO;
-
-  [self presentViewController:materialAlertController animated:YES completion:NULL];
-}
-
 - (IBAction)didTapDynamicAlert {
   // The following strings are extra verbose to better demonstrate the dynamic handling of an alert.
   NSString *titleString = @"This alert supports Dynamic Type font sizing.";
--- a/components/Dialogs/src/MDCDialogPresentationController.h
+++ b/components/Dialogs/src/MDCDialogPresentationController.h
@@ -16,8 +16,6 @@
 
 #import "MaterialShadowElevations.h"
 
-extern CGFloat const MDCUsePresentedViewCornerRadius;
-
 @class MDCDialogPresentationController;
 
 /**
@@ -71,22 +69,22 @@ extern CGFloat const MDCUsePresentedViewCornerRadius;
 /**
  Customize the corner radius of the shadow to match the presented view's corner radius.
 
- By default (when the value is MDCUsePresentedViewCornerRadius), the dialogCornerRadius is ignored,
- and the presented view will be presented with it original corner radius value that is assigned
- by the client. The presentation controller will ensure in that case to match its shadow
- shape to the corner radius of the presented view.
+ By default, the corner radius of the presented shadow is adjusted to match the corner radius of the
+ view being presented.  This behavior is achieved without making any changes to dialogCornerRadius.
+ Once a value is set on dialogCornerRadius, that value will be used to determine the radius of both
+ the presetned view and the shadow. That means that any further changes to the presented view's
+ corner radius (yourViewController.view.layer.cornerRadius) will be ignored once dialogCornerRadius
+ is set.
 
- Modifying dialogCornerRadius to a positive value, will present the presented view with this
- positive value, overriding previous corner radius values that may have been previously
- set by the client. The presentation controller will ensure in that case to match both the
- presetned view's and the shadow's corner radius to the value assigned in dialogCornerRadius.
+ In either cases, the presentation controller ensures that the shadow layer's corner radius matches
+ the presented view's.
 
  Material themers use dialogCornerRadius for setting the corner radius, therefore, when applying
  a themer to your custom UIViewController, any value you assign to your view's corner radius
-(through view.layer.cornerRadius) will be ignored. In that case, if you wish to further
- customize the corner radius of the presented view, set dialogCornerRadius instead.
+ will be ignored. If you wish to override the corner radius after a themer is called, make sure
+ to set it to dialogCornerRadius, and not to the presented view's corner radius.
 
- Defaults to: MDCUsePresentedViewCornerRadius (-1.0).
+ Defaults to: The presented view's corner radius.
  */
 @property(nonatomic, assign) CGFloat dialogCornerRadius;
 
--- a/components/Dialogs/src/MDCDialogPresentationController.h
+++ b/components/Dialogs/src/MDCDialogPresentationController.h
@@ -68,10 +68,19 @@
 
 /**
  Customize the corner radius of the shadow to match the presented view's corner radius.
- If the presented view corner radius and dialogCornerRadius are different, the rendered shadow will
- not match.
 
- Defaults to 0.0.
+ The default value (-1) will use the presented view's corner radius as the corner radius of the
+ shadow. Setting any value to myPresentedViewConteroller.view.layer.cornerRadius will be used
+ as the shadow's corner radius.
+
+ Changing the dialogCornerRadius will work in the opposite way. When a view controller is pesented,
+ it's view's corner radius will be adjusted to match this value.
+
+ Applying a themer on a custom dialog or on an alert (MDCAlertController) will adjust the
+ dialogCornerRadius value, and will override any value manually set for the
+ presented view controller.
+
+ Defaults to -1.0.
  */
 @property(nonatomic, assign) CGFloat dialogCornerRadius;
 
--- a/components/Dialogs/src/MDCDialogPresentationController.h
+++ b/components/Dialogs/src/MDCDialogPresentationController.h
@@ -16,7 +16,7 @@
 
 #import "MaterialShadowElevations.h"
 
-extern CGFloat const kPresentedViewCorenerRadius;
+extern CGFloat const MDCUsePresentedViewCornerRadius;
 
 @class MDCDialogPresentationController;
 
@@ -71,7 +71,7 @@ extern CGFloat const kPresentedViewCorenerRadius;
 /**
  Customize the corner radius of the shadow to match the presented view's corner radius.
 
- By default (when the value is kPresentedViewCorenerRadius), the dialogCornerRadius is ignored,
+ By default (when the value is MDCUsePresentedViewCornerRadius), the dialogCornerRadius is ignored,
  and the presented view will be presented with it original corner radius value that is assigned
  by the client. The presentation controller will ensure in that case to match its shadow
  shape to the corner radius of the presented view.
@@ -86,7 +86,7 @@ extern CGFloat const kPresentedViewCorenerRadius;
 (through view.layer.cornerRadius) will be ignored. In that case, if you wish to further
  customize the corner radius of the presented view, set dialogCornerRadius instead.
 
- Defaults to: kPresentedViewCorenerRadius (-1.0).
+ Defaults to: MDCUsePresentedViewCornerRadius (-1.0).
  */
 @property(nonatomic, assign) CGFloat dialogCornerRadius;
 
--- a/components/Dialogs/src/MDCDialogPresentationController.h
+++ b/components/Dialogs/src/MDCDialogPresentationController.h
@@ -16,6 +16,8 @@
 
 #import "MaterialShadowElevations.h"
 
+extern CGFloat const kPresentedViewCorenerRadius;
+
 @class MDCDialogPresentationController;
 
 /**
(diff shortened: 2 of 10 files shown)


components/Dialogs/examples/DialogsAlertExampleViewController.m
```diff
@@ -127,36 +127,6 @@ - (IBAction)didTapShowLongAlert {
   [self presentViewController:materialAlertController animated:YES completion:NULL];
 }
 
-- (IBAction)didTapNondismissingAlert {
```
galiak11, line 130: No point in fixing this example, since it wasn't used anywhere.

ianegordon (dismissed):

fluxminds:
@galiak11 Does the PresentationController extension have access to presentedViewController? Seems it is a property of UIPresentationController?

components/Dialogs/src/Theming/UIViewController+MaterialTheming.h
```diff
@@ -16,18 +16,17 @@
...
 /**
- Applies a container scheme to this instance.
+ Applies a container scheme to a custom view controller, to theme its presentation style
 
- @param scheme A container scheme instance containing any desired customizations to the theming
- system.
+ @param scheme A container scheme used for customizing the theme
  */
-- (void)applyThemeWithScheme:(nonnull id<MDCContainerScheming>)scheme;
+- (void)applyPresentationThemeWithScheme:(nonnull id<MDCContainerScheming>)scheme;
```
ianegordon: methods added to UIKit classes via categories should have a prefix.
mdc_applyPresentationThemeWithScheme:

components/Dialogs/src/Theming/UIViewController+MaterialTheming.m
```diff
@@ -12,24 +12,28 @@
...
-- (void)applyThemeWithScheme:(nonnull id<MDCContainerScheming>)scheme {
+- (void)applyPresentationThemeWithScheme:(nonnull id<MDCContainerScheming>)scheme {
   // Color
   id<MDCColorScheming> colorScheme = scheme.colorScheme;
   if (!colorScheme) {
     colorScheme =
         [[MDCSemanticColorScheme alloc] initWithDefaults:MDCColorSchemeDefaultsMaterial201804];
   }
-  self.scrimColor = [colorScheme.onSurfaceColor colorWithAlphaComponent:(CGFloat)0.32];
+  MDCDialogPresentationController *presentationController = self.mdc_dialogPresentationController;
```
ianegordon: I don't think this handles the case where a view controller is themed before it is presented.  Won't self.mdc_dialogPresentationController be nil?

ianegordon:
I concerned that we are moving the theming of the presentationController to UIViewController.  Is there no way to theme the view controller through the presentation controller?
Concerns in line.

ianegordon (dismissed):

components/Dialogs/src/Theming/UIViewController+MaterialTheming.h
galiak11: Will do.

components/Dialogs/src/Theming/MDCDialogPresentationController+MaterialTheming.m
```diff
@@ -27,8 +27,15 @@ - (void)applyThemeWithScheme:(nonnull id<MDCContainerScheming>)scheme {
   }
   self.scrimColor = [colorScheme.onSurfaceColor colorWithAlphaComponent:(CGFloat)0.32];
 
-  // Other properties
-  self.dialogCornerRadius = kCornerRadius;
+  // Corner Radius
+  if ([self.presentedViewController isKindOfClass:[MDCAlertController class]]) {
+    ((MDCAlertController *)self.presentedViewController).cornerRadius = kCornerRadius;
+  } else {
```
galiak11: My concern with this approach is that by accessing the view property of the presentedViewController we may trigger "viewDidLoad" inside the themer.  
Some solutions are:
*  I could call isViewLoaded first, but then the corner radius will have to be set elsewhere after the view is loaded.
* set the corner radius in the presentation controller's presentationTransitionWillBegin which is called when the view is being presented (should we match the view.layer.cornerRadius to be the same as dialogCornerRadius, or the other way around?)
* Just document this fact in the theming extension's header file.  

Any suggestions/thoughts?

components/Dialogs/src/MDCDialogPresentationController.m
```diff
@@ -82,6 +72,7 @@ - (instancetype)initWithPresentedViewController:(UIViewController *)presentedVie
   self = [super initWithPresentedViewController:presentedViewController
                        presentingViewController:presentingViewController];
   if (self) {
+    _dialogCornerRadius = -1.0;
```
codeman7: _Question:_ Why is this defaulting to `-1.0` instead of `0.0`?

components/Dialogs/src/MDCDialogPresentationController.m
```diff
@@ -146,6 +137,20 @@ - (void)presentationTransitionWillBegin {
   // presentation and dismissal of the dialog.
   // https://material.io/guidelines/motion/choreography.html#choreography-creation
 
+  // Ensure corner radius matches between the tracking view and the view being presented
```
codeman7: _Question:_ Can we remove the `_trackingView` all together?

components/Dialogs/src/MDCDialogPresentationController.m
```diff
@@ -146,6 +137,20 @@ - (void)presentationTransitionWillBegin {
...
+  if (self.dialogCornerRadius < 0) {
+    // If dialogCornerRadius is not set, use the presented view's cornerRadius for the shadow layer.
+    _trackingView.layer.cornerRadius = self.presentedViewController.view.layer.cornerRadius;
```
fluxminds: This changes the behavior to be: if `dialogCornerRadius` is not set, then force cornerRadius to be the view's layerCornerRadius. If this behavior is okay, I am fine with the change.

components/Dialogs/src/MDCDialogPresentationController.m
```diff
@@ -146,6 +137,20 @@ - (void)presentationTransitionWillBegin {
   // presentation and dismissal of the dialog.
   // https://material.io/guidelines/motion/choreography.html#choreography-creation
 
+  // Ensure corner radius matches between the tracking view and the view being presented
+  if (self.dialogCornerRadius < 0) {
```
fluxminds: Maybe using a private BOOL ivar dialogCornerRadiusIsSet would be more clear? Using negative number might be fragile?

components/Dialogs/src/MDCDialogPresentationController.m
```diff
@@ -146,6 +137,20 @@ - (void)presentationTransitionWillBegin {
...
+    // If dialogCornerRadius is not set, use the presented view's cornerRadius for the shadow layer.
+    _trackingView.layer.cornerRadius = self.presentedViewController.view.layer.cornerRadius;
+  } else {
+    // If dialogCornerRadius is set, use its value for the shadow layer as well as the presented
+    // view layer, overriding any direct assignments to the presented view's cornerRadius.
+    _trackingView.layer.cornerRadius = self.dialogCornerRadius;
+    // Note: For MDCAlertController, this assumes that the "view" property points to the same
+    // instance as the "alertView" property. Therefore, we are safe to not set its cornerRadius
+    // property (its ".cornerRadius", rather then its ".view.layer.cornerRadius")
+    self.presentedViewController.view.layer.cornerRadius = self.dialogCornerRadius;
```
fluxminds: Maybe self.presentedView can be used here directly?

fluxminds (changes requested):

components/Dialogs/src/MDCDialogPresentationController.m
galiak11: This value is ignored if negative, so the presented view's corner radius takes precedence. See full explanation in the PR description (solution C). I also updated the comment explaining this in MDCDialogPresentationController.h - hopefully it will make this more clear.

components/Dialogs/src/MDCDialogPresentationController.m
galiak11: That would require one more variable for setting a corner radius - in addition to the 2 that are already in use (the view's cornerRardius and the dialogCornerRadius).

components/Dialogs/src/MDCDialogPresentationController.m
galiak11: Yes, that is intentional. 
The goal is to make sure that we respond correctly to default UIKit behavior - meaning, if you set the corner radius on your view - it will work "out of the box" without additional actions. That's a UIKit behavior that users will expect. See more in the PR's description (solution C).

components/Dialogs/src/MDCDialogPresentationController.m
galiak11: I think that's the view that renders the shadow, so it's needed for the presentation.

components/Dialogs/src/MDCDialogPresentationController.m
fluxminds: It would be a private variable, so it shouldn't affect API. And adding it would help the clearance of the code, so the number of variable is not a concern in this case. `-1` implicitly delivered the same idea.

components/Dialogs/src/MDCDialogPresentationController.m
fluxminds: ping this one.

components/Dialogs/src/MDCDialogPresentationController.m
fluxminds: Cool. It sounds good if that's what UIKit suggests.

fluxminds (changes requested):
Replied to some comments.

components/Dialogs/src/MDCDialogPresentationController.m
galiak11: dialogCornerRadius is a public API, so when clients modify it, these 2 will get out of sync.  The necessary syncing logic to resolve this doesn't payoff - because you now have to test for this -1 in a different place.
Instead I added a public constant to make the use of this case more clear. Please see updated comment in MDCDialogPresentationController.h - hopefully this issue is explained better now.

components/Dialogs/src/Theming/UIViewController+MaterialTheming.m
galiak11: self.mdc_dialogPresentationController gets initialized first time it's referenced, if the transitionController is set. so it is not nil, even before the it is presented.
Also - I abandoned this implementation in favor of solution C (see in description) - going back to old theming extension on MDCDialogPresentationController.

components/Dialogs/src/MDCDialogPresentationController.h
```diff
@@ -16,6 +16,8 @@
 
 #import "MaterialShadowElevations.h"
 
+extern CGFloat const kPresentedViewCorenerRadius;
```
ianegordon: nit: Corner

Consider removing the k as this is not a constant from CoreFoundation.  (I know we are not consistent on this)

components/Dialogs/src/MDCDialogPresentationController.m
```diff
@@ -146,6 +138,20 @@ - (void)presentationTransitionWillBegin {
...
+    // If dialogCornerRadius is not set, use the presented view's cornerRadius for the shadow layer.
+    _trackingView.layer.cornerRadius = self.presentedView.layer.cornerRadius;
+  } else {
+    // If dialogCornerRadius is set, use its value for the shadow layer as well as the presented
+    // view layer, overriding any direct assignments to the presented view's cornerRadius.
+    _trackingView.layer.cornerRadius = self.dialogCornerRadius;
+    // Note: For MDCAlertController, this assumes that the "view" property points to the same
+    // instance as the "alertView" property. Therefore, we are safe to not set its cornerRadius
+    // property (its ".cornerRadius", rather then its "presentedView.layer.cornerRadius")
+    self.presentedView.layer.cornerRadius = self.dialogCornerRadius;
```
ianegordon, line 163: If we want to be extremely cautious here, we could store the previousPresentedViewCornerRadius before overwriting it.  When we stop presenting the view, we could set self.presentedView.layer.cornerRadius back to its original value.

ianegordon (dismissed):

components/Dialogs/src/MDCDialogPresentationController.m
fluxminds: I understand the syncing logic, what I meant was we could add a private instance variable like `isDialogCornerRadiusManuallySet` to avoid the constant issue, because using a default value as a way to indicate a instance hasn't been changed is not stable.

fluxminds (dismissed):
Left a follow-up comment.

components/Dialogs/src/MDCDialogPresentationController.h
romoore: Also naming: s/Corener/Corner/

romoore: Since this is a public global constant, please start with the component prefix, so something like `MDCDialogsPresentedViewCornerRadius`

http://google.github.io/styleguide/objcguide.html#constants

components/Dialogs/src/MDCDialogPresentationController.m
galiak11: Sounds good.

galiak11: Sounds good.

components/Dialogs/src/MDCDialogPresentationController.m
galiak11: I agree with your concern, but im not sure your suggestions will solve it.  This condition now compares with "MDCUsePresentedViewCornerRadius" instead of -1, which makes it a little clearer for the next person (using a constant to imitate swift's nil value, which unfortunately doesn't exist in obj c floats.).  
MDCUsePresentedViewCornerRadius is used in more than one place, and creating 2 variables to test this condition seems like an overkill.
How strongly are you feeling about this?

fluxminds: Followed up offline. I feel not exposing the `-1` value for clients would reduce the complexity and remove the unnecessary functionality we provide.

fluxminds (changes requested):
Replied to a comment after the offline discussion.

fluxminds (approved):

PR #6699, opened 2019-02-22
[Container] Fix colorScheme and typographyScheme to be nonnull.

codeman7:
## Introduction
Before we move container scheme to ready we need to have both the `colorScheme` and `typographyScheme` nonnull. By making colorScheme and typographyScheme of MDCContainerScheming nonnull, clients will reduce the amount of conditional checks required in their apps and reduce a category of potential errors (accidentally passing nil to a nonnull parameter).

## The problem
Both `colorScheme` and `typographyScheme` were nullable in the container scheme making it harder for clients to use the container scheme throughout their app on non Material UIElements, i.e. UIView.backgroundColor

## The fix
Make both of these properties nonnull.

## Additional work
- [x]  I had to update the swift examples to not be optional and I additionally removed the nil checks from the ObjC examples.


yarneo (approved):

components/ActionSheet/examples/ActionSheetComparisonExampleViewController.m
```diff
@@ -46,15 +46,7 @@ - (instancetype)init {
 - (void)viewDidLoad {
   [super viewDidLoad];
 
-  id<MDCColorScheming> colorScheme;
```
romoore: If this code was generating a compiler error, then I think it's good to change it. If not, I think it should be moved to a separate PR and this should focus only on the critical path to make changes to MDCContainerScheming.

components/Cards/examples/EditReorderCollectionViewController.swift
```diff
@@ -34,11 +34,11 @@ class EditReorderCollectionViewController: UIViewController,
   var containerScheme: MDCContainerScheming
 
   var colorScheme: MDCColorScheming {
```
romoore: Seems like these two computed vars can be removed entirely and their use replaced with `containerScheme.colorScheme`.

components/schemes/Container/src/MDCContainerScheme.h
```diff
@@ -49,8 +49,8 @@
 __attribute__((objc_subclassing_restricted)) @interface MDCContainerScheme
```
romoore, line 47: Please be sure to update the header documentation now that properties are no longer `nil`.  The defaults should be explicitly documented.

romoore (dismissed):

components/ActionSheet/examples/ActionSheetTypicalUseExampleViewController.m
```diff
@@ -44,15 +44,7 @@ - (instancetype)init {
...
-  id<MDCColorScheming> colorScheme;
-  if (self.containerScheme.colorScheme != nil) {
-    colorScheme = self.containerScheme.colorScheme;
-  } else {
-    colorScheme =
-        [[MDCSemanticColorScheme alloc] initWithDefaults:MDCColorSchemeDefaultsMaterial201804];
-  }
-
-  self.view.backgroundColor = colorScheme.backgroundColor;
+  self.view.backgroundColor = self.containerScheme.colorScheme.backgroundColor;
```
galiak11: I think checking if self.containerScheme is nil is still required (and assigning it a default container-scheme if it's nil)), because it may not be injected.
This is true for all examples.

components/schemes/Container/src/MDCContainerScheme.h
```diff
@@ -26,12 +26,12 @@
 /**
  The color scheme to use for the container scheme.
  */
-@property(nonatomic, nullable, readonly) id<MDCColorScheming> colorScheme;
+@property(nonatomic, nonnull, readonly) id<MDCColorScheming> colorScheme;
```
galiak11, line 29: Should we provide an initializer now that these are not null? 
How are we expecting client apps to assign these values on a newly initialized container scheme?

romoore: @galiak11  There is an `init` method for `MDCContainerScheme` in the `.m` file. I don't think there's much value dictating initialization/construction in the protocol - implementations may use builders, factories, `init` methods, etc.   The protocol is focused on consumption - requiring the properties, their types and attributes.  

Clients can assign the properties directly to the implementation properties.

components/schemes/Container/src/MDCContainerScheme.h
codeman7: Done

components/ActionSheet/examples/ActionSheetComparisonExampleViewController.m
codeman7: Removed

randallli (approved):

components/schemes/Container/src/MDCContainerScheme.h
codeman7: I think we just intend to have the `init` in the `.m` file as @romoore has outlined. Do you agree with that?

galiak11: Yes.

components/Cards/examples/EditReorderCollectionViewController.swift
codeman7: These have been updated.

galiak11 (approved):

components/schemes/Container/src/MDCContainerScheme.h
romoore: This is still not done - please fix the header documentation above.

romoore (approved):
Please fix the header comment before merging.

components/schemes/Container/src/MDCContainerScheme.h
codeman7: Good catch, I didn't see that. It has been fixed.

codeman7:
This has been tested in cl/235759497

components/ActionSheet/examples/ActionSheetTypicalUseExampleViewController.m
codeman7: This has been updated.

PR #6706, opened 2019-02-23
[BottomNavigation] Adds the ability to detect long presses on the bottom navigation controller item

nokeeo:
This will allow the bottom navigation controller to display a dialog of the tab bar item when dynamic type is enabled.

### DIFF
--- a/components/BottomNavigation/src/MDCBottomNavigationBar.h
+++ b/components/BottomNavigation/src/MDCBottomNavigationBar.h
@@ -184,7 +184,7 @@ typedef NS_ENUM(NSInteger, MDCBottomNavigationBarAlignment) {
  Called when the user long holds a tab bar item. A user may drag their finger along the bottom
  navigation bar to a new tab bar item and this method will be called again with the new item.
  */
-- (void)bottomNavigationBar:(MDCBottomNavigationBar *)bottomNavigationBar
+- (void)bottomNavigationBar:(nonnull MDCBottomNavigationBar *)bottomNavigationBar
                didPressItem:(nonnull UITabBarItem *)item;
 
 /**
@@ -193,7 +193,7 @@ typedef NS_ENUM(NSInteger, MDCBottomNavigationBarAlignment) {
  tab bar item.
  This call is not necessarily balanced with every didPressItem call.
  */
-- (void)bottomNavigationBar:(MDCBottomNavigationBar *)bottomNavigationBar
+- (void)bottomNavigationBar:(nonnull MDCBottomNavigationBar *)bottomNavigationBar
              didReleaseItem:(nonnull UITabBarItem *)item;
 
 @end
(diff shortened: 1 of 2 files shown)


romoore:
The website-generation failure seems to be from an outdated `develop` branch. @nokeeo  Can you please merge the latest changes from `develop` into your branch?

components/BottomNavigation/src/MDCBottomNavigationBar.h
```diff
@@ -180,4 +180,20 @@ typedef NS_ENUM(NSInteger, MDCBottomNavigationBarAlignment) {
 - (void)bottomNavigationBar:(nonnull MDCBottomNavigationBar *)bottomNavigationBar
               didSelectItem:(nonnull UITabBarItem *)item;
 
+/**
+ Called when the user long holds a tab bar item. A user may drag their finger along the bottom
+ navigation bar to a new tab bar item and this method will be called again with the new item.
+ */
+- (void)bottomNavigationBar:(nonnull MDCBottomNavigationBar *)bottomNavigationBar
```
romoore, line 187: How does this API work with UIAccessibility (VoiceOver, SwitchControl)?

romoore (commented):
This is an interesting change, but can the goal be achieved with the Controller directly by using the `viewForItem:` API and attaching gesture recognizers that way?

components/BottomNavigation/src/MDCBottomNavigationBar.h
nokeeo: > Can the goal be achieved with the Controller directly by using the `viewForItem:` API and attaching gesture recognizers that way?

I believe it can be done, but I think it may result in a more complicated implementation.  If we attach a gesture recognizer to each view corresponding with an item, the controller will receive all touch/state event updates through a method.  From that method we will have to map the view back to the item.  Currently there is no API on the navigation bar to provide an item given a view.  Doing this muddies the responsibility boundary between the controller and the navigation bar.  In my mind, the navigation bar manages the relationship between the items and the views, and the controller manages the relationship between children view controllers and items.  Having the container controller manage the navigation bar's subviews is a leak in this abstraction.  

[Apple's docs](https://developer.apple.com/library/archive/featuredarticles/ViewControllerPGforiPhoneOS/) (User Interaction Section) also recommend that views handle touch events and propagate an event to the view controller via a delegate.

Another possible gotchas I see with this approach is how to handle a user sliding from one item to another.  I think a user must touch down in a long press gesture recognizer's bounds for the recognizer to start.

> How does this API work with UIAccessibility (VoiceOver, SwitchControl)?

In terms of the larger feature(#4159) I don't believe support for VoiceOver is necessary, because the view is transient . The user is primarily interacting with the application using audio and the screen reader will read the item title.  

However someone else may use this API for a different purpose.  I am not super familiar with SwitchControl.  What are your thoughts on how this API should work under those conditions?

One possible solution is to gate this API behind a private header?

romoore: Thanks for the response.  Let's talk offline about this tomorrow. I'd like us to follow UITabBar's behavior as much as possible, excepting for any Material Theming that might be required on the pop-up view.

In either case, I think we should sit down and propose a design document for this behavior and also make sure the Designers are happy with any proposed look and feel. I'm sure they'll want to consider platform adaptation to other platforms (Android, Flutter, Web), so it might not be exactly the way UITabBar handles it in the end.

nokeeo: Ok!  Sounds good to me!

PR #6718, opened 2019-02-27
[BottomNavigation] Examples clean-up.

romoore:
This is a collection of internal clean-ups that were overdue on Bottom Navigation examples.  Nearly all examples (8 of 9) were updated. The majority of the improvements focused on visual layout and appearance, but some code clean-up took place as well.

* Making sure buttons/labels remain on-screen after rotations.
* Correctly positioning the BottomNavigationBar with respect to the safe
  area.
* Not adding subviews or configuring views in `-init`
* Allow Catalog/Dragons to inject an App Bar.
* Stop overriding titles.
* Use property access.

## Before/After Screenshots

Not all examples visibly changed as a result of the code clean-up (e.g., "Bottom Navigation" and "Blur Effect"), but the before and after images are present to make it easier to review.

|Example|Before|After|
|---|---|---|
|Badge Value Test|bn-badgevalue-develop|bn-badgevalue-after|
|Blur Effect|bn-blur-develop|bn-blur-after|
|Bottom Navigation|bn-typical-develop|bn-typical-after|
|Bottom Navigation (Swift)|bn-swift-develop|bn-swift-after|
|Bottom Navigation Reorder (Swift)|bn-reorder-develop|bn-reorder-after|
|Bottom Navigation Selected|bn-selected-develop|bn-selected-after|
|Bottom Navigation Set Color (Swift)|bn-setcolor-develop|bn-setcolor-after|
|Bottom Navigation Title Visibility (Swift)|bn-titlevis-develop|bn-titlevis-after|


components/BottomNavigation/examples/BottomNavigationTitleVisibilityChangeExample.swift
```diff
@@ -71,21 +72,26 @@ class BottomNavigationTitleVisibilityChangeExample: UIViewController, MDCBottomN
...
   }
   
   override func viewWillLayoutSubviews() {
     super.viewWillLayoutSubviews()
+
+    var viewBounds = view.bounds;
+    if #available(iOS 11.0, *) {
+      viewBounds = UIEdgeInsetsInsetRect(viewBounds, view.safeAreaInsets)
+    }
+    let labelWidth = min(viewBounds.size.width - 32, 480);
```
galiak11, line 92: Magic numbers... -32, 480 - What do these mean?

galiak11 (commented):
It looks like the height of the bottom nav grew bigger after rotation (looking at the "After" screenshots). Shouldn't it remain the same?

components/BottomNavigation/examples/BottomNavigationTitleVisibilityChangeExample.swift
romoore: 32 is 16 point margins on either side if the label spans the full width of the view. I believe this matches [Material.io guidelines for margins](https://material.io/design/layout/responsive-layout-grid.html#columns-gutters-margins).

480 looked pretty nice when the layout was very wide.  If you have a reason to pick a different max width, I'd be happy to use it. :)

romoore:
This PR changes the VCs layout of the BottomNavigationBar. The VC *should* be including any bottom safe area in the height of the BottomNavigation so it can extend down into the safe area and meet the screen edge.  However, there's currently a bug in MDCBottomNavigationBar's `sizeThatFits:` implementation where it includes the safe area insets, and that's being fixed in https://github.com/material-components/material-components-ios/pull/6717.

After this change (and before the fix in #6717), these examples will be "too tall". Some examples already have this behavior. After the fix, all examples should be correctly positioning/sizing the BottomNavigationBar.

jverkoey (commented):
The "Selected Item" after image looks incorrect; is the bottom navigation bar supposed to have that much padding at the bottom?

romoore:
@jverkoey I had the same discussion with @galiak11 .  The BottomNavigationBar has the wrong `sizeThatFits:` depending on its position in the view hierarchy (it accounts for safeAreaInsets). Several of our examples are already performing the correct layout, resulting in the disproportionate height. Once #6717 lands, these can be updated and they will look correct.  Unfortunately, feedback on that PR means that the bug fix has to be behind an opt-in toggle rather than having opt-out behavior for clients, so this PR will require follow-up work.

codeman7 (approved):

Issue #6608, opened 2019-02-12
MDCBottomNavigationControllerTests lacks proper availability guards


andrewoverton:
`MDCBottomNavigationBarController` has a `NS_CLASS_AVAILABLE_IOS(9_0)`, but its accompanying protocol and tests do not. I believe this is causing pod lib lint validation failures.

---
#### Internal data
- Associated internal bug: [b/124304557](http://b/124304557)

romoore:
@andrewoverton  Can you determine if you're still seeing the same errors that caused this ticket?

andrewoverton:
I just did `pod lib lint MaterialComponents.podspec --subspec=NavigationDrawer` and everything was fine so I'm going to close this ticket I guess.

PR #6719, opened 2019-02-27
[BottomNavigation] Make Controller available on iOS 8.

romoore:
Adds constraints safe for use on iOS 8. This removes the availability
checks previously set on the MDCBottomNavigationBarController class that
may have been causing CocoaPods lint errors.

Related to #6608


jverkoey (approved):
LGTM, though I feel we should consider the adding of iOS 8 support to code to be very low priority (P4).

romoore:
@jverkoey  This PR was opened because @andrewoverton suspected the missing availability macros on the tests and protocol were causing CocoaPods lib lint problems. It was easier to add iOS 8 support than it was to go around making everything iOS 9+ (which has already caused challenges).

PR #6733, opened 2019-02-28
[Typography] Dynamic Type 2.0

ianegordon:
Next iteration of our Dynamic Type.

Mimics Apple's new APIs.  (UIFontMetrics)
Adds MDCFontScalar to attach curves.
Adds UIFont category to expose fontScaledForCategory methods.
Adds new MDCTypographyScheme with scalable fonts.


components/Typography/src/MDCFontScaler.h
```diff
@@ -0,0 +1,60 @@
...
+extern NSString * _Nonnull MaterialTextStyleHeadline4;
+extern NSString * _Nonnull MaterialTextStyleHeadline5;
+extern NSString * _Nonnull MaterialTextStyleHeadline6;
+extern NSString * _Nonnull MaterialTextStyleSubtitle1;
+extern NSString * _Nonnull MaterialTextStyleSubtitle2;
+extern NSString * _Nonnull MaterialTextStyleBody1;
+extern NSString * _Nonnull MaterialTextStyleBody2;
+extern NSString * _Nonnull MaterialTextStyleButton;
+extern NSString * _Nonnull MaterialTextStyleCaption;
+extern NSString * _Nonnull MaterialTextStyleOverline;
```
jverkoey: Can we annotate this as an enum?

components/Typography/src/MDCFontScaler.h
```diff
@@ -0,0 +1,60 @@
+#import <UIKit/UIKit.h>
+
+// Material's text styles, which are similar, but not quite equivalent to Apple's UIFontTextStyle.
+extern NSString * _Nonnull MaterialTextStyleHeadline1;
```
jverkoey: We don't have a lot of precedent for use of Material in our prefixes; `MDCTextStyle` feels like the expected prefix here.

components/Typography/src/MDCFontScaler.h
```diff
@@ -0,0 +1,60 @@
...
+extern NSString * _Nonnull MaterialTextStyleSubtitle2;
+extern NSString * _Nonnull MaterialTextStyleBody1;
+extern NSString * _Nonnull MaterialTextStyleBody2;
+extern NSString * _Nonnull MaterialTextStyleButton;
+extern NSString * _Nonnull MaterialTextStyleCaption;
+extern NSString * _Nonnull MaterialTextStyleOverline;
+
+
+/**
+ MDCFontScaler is used to attach a scaling curve to a UIFont via an associated object on that
```
jverkoey: `MDCFontScaler attaches`

components/Typography/src/MDCFontScaler.h
```diff
@@ -0,0 +1,60 @@
...
+ font instance.
+
+ This interface is similar to UIFontMetrics, but the font returned from MDCFontScaler do *not*
+ automatically adjust when the device's text size / content size category is changed.
+ */
+@interface MDCFontScaler : NSObject
```
jverkoey, line 45: Can we annotate this as subclassing-restricted?

components/Typography/src/MDCFontScaler.h
```diff
@@ -0,0 +1,60 @@
...
+
+
+/**
+ Initializes a font scaler object for the specified text style.
```
jverkoey: `Initializes a font scaler object with the specified text style.`

components/Typography/src/UIFont+MaterialScalable.h
```diff
@@ -0,0 +1,73 @@
+// Copyright 2017-present the Material Components for iOS authors. All Rights Reserved.
```
jverkoey: Copyright headers here and throughout.

components/Typography/src/UIFont+MaterialScalable.h
```diff
@@ -0,0 +1,73 @@
...
+ This can be used to return a font for a text element that should *not* be scaled with Dynamic
+ Type.
+
+ @return Font sized for UIContentSizeCategoryLarge OR self if there is no associated curve
+ */
+- (nonnull UIFont *)mdc_scaledFontAtDefaultSize;
+
+
+/**
+ Scales an arbitraty value based on the current Dynamic Type settings and the scaling curve.
```
jverkoey: `arbitrary`

components/schemes/Typography/src/MDCTypographyScheme.h
```diff
@@ -94,7 +100,15 @@ typedef NS_ENUM(NSInteger, MDCTypographySchemeDefaults) {
   /**
    The Material defaults, circa April 2018.
    */
-  MDCTypographySchemeDefaultsMaterial201804
+  MDCTypographySchemeDefaultsMaterial201804,
+  /**
```
jverkoey, line 105: Add a newline above this comment.

components/schemes/Typography/src/MDCTypographyScheme.h
```diff
@@ -94,7 +100,15 @@ typedef NS_ENUM(NSInteger, MDCTypographySchemeDefaults) {
...
+   The Material defaults, circa February 2019.
+
+   This scheme implements fonts with the same matrics as MDCTypographySchemeDefaultsMaterial201804
```
jverkoey: `metrics`

components/Typography/src/MDCFontScaler.h
```diff
@@ -0,0 +1,60 @@
...
+- (nonnull instancetype)initForMaterialTextStyle:(nonnull NSString *)textStyle;
+
+/**
+ Creates and returns a font scaler object for the specified text style.
+
+ @param textStyle The style that will be used to determine the scaling curver associated with the
+ returned font.  For example, MaterialTextStyleBody1.
+ @return An initialized font scaler object.
+ */
++ (nonnull instancetype)scalerForMaterialTextStyle:(nonnull NSString *)textStyle;
```
jverkoey: Do we expect clients to create and retain font scalers for each text style they care about?

I ask because looking at the implementation I wager this API could be just as effectively accomplished as a set of static methods:


```objc
// This:
   fontScaler = [[MDCFontScaler alloc] initForMaterialTextStyle:MaterialTextStyleHeadline2];
  _headline2 = [fontScaler scalableFontWithFont:_headline2];

// Becomes this:
  _headline2 = [MDCFontScaler scalableFontWithFont:_headline2 textStyle:MaterialTextStyleHeadline2];
```

jverkoey (dismissed):

codeman7:
Can we remove the `.m` files from this PR? Is it ready for review?

components/Typography/src/MDCFontScaler.h
ianegordon: We will need to allow subclassing to add the Internal-specific curves.

components/Typography/src/MDCFontScaler.h
ianegordon: Done.

components/Typography/src/MDCFontScaler.h
ianegordon: Done.

components/Typography/src/MDCFontScaler.h
ianegordon: Done here and below.

components/Typography/src/UIFont+MaterialScalable.h
ianegordon: Done.

components/Typography/src/UIFont+MaterialScalable.h
ianegordon: Done.

components/schemes/Typography/src/MDCTypographyScheme.h
ianegordon: Done.

components/schemes/Typography/src/MDCTypographyScheme.h
ianegordon: Done.

components/Typography/src/MDCFontScaler.h
ianegordon: This interface is intentionally similar to UIKit's UIFontMetrics.

ianegordon:
This PR is not yet ready for general review.

components/Typography/src/MDCFontScaler.h
jverkoey: Got it. Thanks!

https://developer.apple.com/documentation/uikit/uifontmetrics

components/Typography/src/UIFont+MaterialScalable.h
```diff
@@ -0,0 +1,73 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <UIKit/UIKit.h>
+
+#import "MDCFontTextStyle.h"
+
+@interface UIFont (MaterialScalable)
+
+
```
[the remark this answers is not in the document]
jverkoey: Looks like there's still an extra newline here?

components/Typography/src/UIFont+MaterialScalable.h
```diff
@@ -0,0 +1,69 @@
...
+/**
+ An associated scaling curve that is a dictionary that maps UISizeCategory to fontsize
+ */
+@property(nonatomic, nullable, setter=mdc_setScalingCurve:)
+    NSDictionary<UIContentSizeCategory, NSNumber *> *mdc_scalingCurve;
+
+/**
+ Is there a scaling curve associated with the font
+ */
+@property(nonatomic, readonly, getter=mdc_hasScalingCurve) BOOL mdc_scalingCurveAssociated;
```
jverkoey: Do we need this API? In other words: can a client rely on a nil return value from mdc_scalingCurve to indicate the same thing?

components/Typography/src/MDCFontScaler.h
ianegordon: Done.

components/Typography/src/UIFont+MaterialScalable.h
ianegordon: This is a convenience method.  

To be clear, the underlying implementation is exactly what you mention above, there is no associated storage:
```
- (BOOL)mdc_hasScalingCurve {
  if (self.mdc_scalingCurve != nil) {
    return YES;
  } else {
    return NO;
  }
}
```

During discussions with clients, some wanted to use a typography scale without scalable fonts to indicate that a dynamic type should not be enabled.  They will end up querying a UIFont for it's scalingCurve a decent amount in their code.  It's the future readability of:

if (font.mdc_hasScalingCurve)
if (font.mdc_scalingCurve != nil)

I'm happy to remove if you feel strongly.

components/Typography/src/MDCFontScaler.h
```diff
@@ -0,0 +1,71 @@
...
+extern MDCTextStyle MDCTextStyleSubtitle1;
+extern MDCTextStyle MDCTextStyleSubtitle2;
+extern MDCTextStyle MDCTextStyleBody1;
+extern MDCTextStyle MDCTextStyleBody2;
+extern MDCTextStyle MDCTextStyleButton;
+extern MDCTextStyle MDCTextStyleCaption;
+extern MDCTextStyle MDCTextStyleOverline;
+
+/**
+ MDCFontScaler attaches a scaling curve to a UIFont via an associated object on that font instance.
```
randallli, line 35: make this more clear what we are attaching (not a MDCFontScaler)

ianegordon: Added clarity.

components/Typography/src/MDCFontScaler.h
ianegordon: I'm going to resolve this, but in case I wasn't clear, let me add a bit more context.
Clients may have their own implementation of type scaling or hierarchies they want to use.  So PanAmMaterial may have a FlightInfo text style (MDCTextStyle).  We need to allow PanAm to subclass to implement their own curves.  The common case would be to just call the existing MDC implementation (return super.typeScaleForTextStyle:) unless text style is unique to PanAm.

components/Typography/src/UIFont+MaterialScalable.h
andrewoverton: I'm not strongly opposed to this convenience accessor, but I feel like this:
`if (font. mdc_hasScalingCurve)`
isn't actually any more convenient than:
`if (font.mdc_scalingCurve)`

components/Typography/src/MDCFontScaler.h
```diff
@@ -0,0 +1,76 @@
...
+
+ Instances of fonts processed through MDCFontScaler will have an associated dictionary that maps
+ UIFontTextStyle to Font Size.  Category methods on UIFont allow clients to get instances of
+ resized fonts baed on this associated dictionary.  Note that an instance of MDCFontScaler is
```
andrewoverton: typo -- "baed"

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
```diff
@@ -0,0 +1,203 @@
...
+    @"Caption", @"Caption (Scalable)",
+    @"Button", @"Button (Scalable)",
+    @"Overline", @"Overline (Scalable)",
+    @"Subtitle1", @"Subtitle1 (Scalable)",
+    @"Subtitle2", @"Subtitle2 (Scalable)",
+    @"Headline1", @"Headline1 (Scalable)",
+    @"Headline2", @"Headline2 (Scalable)",
+    @"Headline3", @"Headline3 (Scalable)",
+    @"Headline4", @"Headline4 (Scalable)",
+    @"Headline5", @"Headline5 (Scalable)",
```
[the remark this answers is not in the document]
ianegordon: Well that's terrible advice.
When I get a moment I'll see if there is a flag in clang-format that maintains the number of elements per line.  I believe it's possible in gofmt and swiftlint, so I am hopeful.

ianegordon:
:/ In this particular case, Kokoro is slightly less than informative.

ci-error

I propose we merge, barring other comments / concerns / questions.
I may be biased :)

yarneo:
The reason why it is failing:
```
Test Case '-[MDCAlertControllerTypographyThemerTests testApplyingTypographyScheme]' started.
components/Dialogs/tests/unit/MDCAlertControllerTypographyThemerTests.m:35: error: -[MDCAlertControllerTypographyThemerTests testApplyingTypographyScheme] : ((view.titleLabel.font) equal to (typographyScheme.headline6)) failed: ("<900644f5 bf7f0000>") is not equal to ("<20f177f5 bf7f0000>")
Test Case '-[MDCAlertControllerTypographyThemerTests testApplyingTypographyScheme]' failed (0.002 seconds).
Test Suite 'MDCAlertControllerTypographyThemerTests' failed at 2019-03-06 19:00:51.566.
	 Executed 1 test, with 1 failure (0 unexpected) in 0.002 (0.002) seconds
Test Suite 'unit_tests_IPHONE_X_IN_11_0.xctest' failed at 2019-03-06 19:00:51.566.
	 Executed 41 tests, with 1 failure (0 unexpected) in 0.217 (0.221) seconds
Test Suite 'All tests' failed at 2019-03-06 19:00:51.566.
	 Executed 41 tests, with 1 failure (0 unexpected) in 0.217 (0.222) seconds
```

Let me know if I can help with anything

ianegordon:
Strange,... let's take a look tomorrow.

romoore:
Rerunning kokoro jobs since they seemed to be marked passing this morning but @ianegordon and @yarneo saw failures.

romoore:
Looks like the test failure was a result of a bad test assertion. It was using `XCTAssertEqual` instead of `XCTAssertEqualObjects` to compare two UIFont objects.

ianegordon:
Travis returned one pass and two timeouts.
I'm willing to submit with those numbers if you +1 the PR.

components/Typography/src/UIFont+MaterialScalable.m
```diff
@@ -0,0 +1,121 @@
...
+
+  NSNumber *fontSizeNumber;
+  if (sizeCategory) {
+    fontSizeNumber = self.mdc_scalingCurve[sizeCategory];
+  }
+
+  // If you have queried the table for a sizeCategory that doesn't exist, we will return the
+  // traits for XXXL.  This handles the case where the values are requested for one of the
+  // accessibility size categories beyond XXXL such as
+  // UIContentSizeCategoryAccessibilityExtraLarge.  Accessbility size categories are only
```
jverkoey: This comment + comment hints at some brittleness here; e.g. what if a custom scaling curve value is missing?

If there is no font size defined for a given scaling curve I'd expect the font not to be adjusted at all; is it possible for us to go with a behavior like that and delete this comment + subsequent code?

```
  if (fontSizeNumber == nil) {
    fontSizeNumber = self.mdc_scalingCurve[UIContentSizeCategoryExtraExtraExtraLarge];
  }
```

components/Typography/src/UIFont+MaterialScalable.m
```diff
@@ -0,0 +1,121 @@
...
+
+- (nonnull UIFont *)mdc_scaledFontAtDefaultSize {
+  return [self mdc_scaledFontForSizeCategory:UIContentSizeCategoryLarge];
+}
+
+- (CGFloat)scaledValueForValue:(CGFloat)value {
+  // If it is available, query the preferredContentSizeCategory.
+  UIContentSizeCategory defaultSizeCategory = UIContentSizeCategoryLarge;
+  UIContentSizeCategory currentSizeCategory = UIContentSizeCategoryLarge;
+  if ([UIApplication mdc_safeSharedApplication]) {
```
jverkoey: Lines 77-81 here are the same as lines 60-61; please extract to a shared method.

components/Typography/src/UIFont+MaterialScalable.m
```diff
@@ -0,0 +1,121 @@
...
+  NSNumber *defaultFontSizeNumber = self.mdc_scalingCurve[defaultSizeCategory];
+
+  NSNumber *currentFontSizeNumber = self.mdc_scalingCurve[currentSizeCategory];
+  // If you have queried the table for a sizeCategory that doesn't exist, we will return the
+  // traits for XXXL.  This handles the case where the values are requested for one of the
+  // accessibility size categories beyond XXXL such as
+  // UIContentSizeCategoryAccessibilityExtraLarge.
+  if (currentFontSizeNumber == nil) {
+    currentFontSizeNumber = self.mdc_scalingCurve[UIContentSizeCategoryExtraExtraExtraLarge];
+  }
```
jverkoey: Similarly to the above comment; can we delete this fallback behavior?

jverkoey (approved):

components/Typography/src/UIFont+MaterialScalable.m
ianegordon: A scaling curve should be complete.  By complete I mean containing values for all sizes from XXS to XXXL.  The accessibility values are optional, which is why there is the fallback behavior of returning XXXL.  This matches Apple's previous behavior of only scaling body past XXXL.  I'd be happy to change the definition of complete to mean containing values for all sizes from XXS - AXXXL.  (Standard Size categories & A11y extensions)  This would remove the need for fallback behavior.
I think it is acceptable to throw an exception if a curve is missing a size category.
Thoughts?

jverkoey: Can we enforce this by creating an object with nonnull properties for each of the expected values, rather than a dictionary? I think this would simultaneously make the code less brittle while also allowing us to encourage good practice when using these APIs.

In other words, instead of using  a `NSDictionary<NSString *, NSNumber *>` as the internal storage, turn that into an object that can be used to store well-typed nonnull values.

ianegordon: To be (overly?) specific the key in the dictionary is a UIContentSizeCategory, which admittedly is an NSString * :)  But  NSDictionary<UIContentSizeCategory, NSNumber *> has a lot more context than a vanilla NSString * key.

At some point we are going to have to convert from a UIContentSizeCategory to a CGFloat.  If history indicates we are the *only* people who will be making these curves.  I'm confident that an NSDictionary is as simple as possible and will be easily understood by future devs.  We do give up some safety, but I think the balance is right.

components/Typography/src/UIFont+MaterialScalable.m
ianegordon: Don.e

ianegordon:
Merging w/o Travis +1 since it has been timing out recently.
All tests pass locally on Xcode 10.1.

romoore:
@ianegordon You don't need to wait for Travis-CI or any non-Required checks. Some of those are there because they can be helpful or for testing.

PR #6752, opened 2019-03-04
[FlexibleHeader] Fix incorrect insets during animated transitions

fluxminds:
Issue: https://github.com/material-components/material-components-ios/issues/6437

Change:
Allow flexible header to update insets when adjustedContentInset is changed on a tracking view that 
 is currently not being tracked.

Screenshot before change:
fv3

Screenshot after change:
iPhone7+ (10.3.1)
fv2

iPhoneX (12.0)
fv1

### DIFF
--- a/components/AppBar/examples/AppBarAnimatedGlitchExample.swift
+++ b/components/AppBar/examples/AppBarAnimatedGlitchExample.swift
@@ -277,6 +277,8 @@ extension SimpleTableViewController: UITableViewDelegate {
   }
 
   func scrollViewDidChangeAdjustedContentInset(_ scrollView: UIScrollView) {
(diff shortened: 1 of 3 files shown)


jverkoey (changes requested):
Due to the risk profile of behavioral changes to FHV can you break this PR out into a refactoring PR and then a fix? The more we can minimize the amount of behavioral changes the less likely we are to introduce a behavioral regression in the flexible header.

fluxminds:
https://github.com/material-components/material-components-ios/pull/6862 (refactor) and https://github.com/material-components/material-components-ios/pull/6863 (fix) have been created for this issue.

Issue #6765, opened 2019-03-05
[Buttons] Deprecate and remove `accessibilityTraitsIncludesButton`


romoore:
## Feature Request/Suggestion
The migration API `accessibilityTraitsIncludesButton` should eventually be deprecated and deleted once all clients are migrated to the new behavior.

PR #6766, opened 2019-03-05
[Buttons] Allow setting `accessibilityTraits`.

romoore:
Clients should be able to customize the value of `accessibilityTraits`.
However, to migrate any clients currently relying on the old behavior (always
including `.button`) we need a new flag to manage the behavior.

QA=Unit tests!

Part of #6763
Necessitates #6765

### DIFF
--- a/components/Buttons/src/MDCButton.h
+++ b/components/Buttons/src/MDCButton.h
@@ -132,6 +132,13 @@
 @property(nullable, nonatomic, strong) id<MDCShapeGenerating> shapeGenerator;
 
 /**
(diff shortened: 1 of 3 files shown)


jverkoey (approved):
LGTM. To confirm the risk analysis: do you expect that client behavior may break as a result of this change?

romoore:
@jverkoey I don't expect any functional changes as a result of this change. I added unit test that verify that the behavior won't change unless the flag is changed. Do you think more tests are warranted?

PR #6773, opened 2019-03-05
[BottomNavigation] Prevents the client from setting the navigation items directly when using the bottom navigation bar controller.

### LINKED ISSUE #6785: [BottomNavigation] BottomNavigationController should prevent modification of bar's `items` property from outside the controller
Similar to UITabBarController's behavior when a programmer attempts to modify `UITabBar.items` from outside the controller, it should be a programmer error to modify `MDCBottomNavigationBar.items` directly when using the MDCBottomNavigationBarController.

---
#### Internal data
- Associated internal bug: [b/127674927](http://b/127674927)

nokeeo:
Observes the navigation bar's items property and if at anytime the new items do not equal the view controllers' tab bar items an exception will be thrown.

Closes #6785

### DIFF
--- a/components/BottomNavigation/src/MDCBottomNavigationBarController.m
+++ b/components/BottomNavigation/src/MDCBottomNavigationBarController.m
@@ -183,7 +183,6 @@ static void *const kObservationContext = (void *)&kObservationContext;
   // Verify tab bar items correspond with the view controllers tab bar items.
   if (items.count != self.viewControllers.count) {
     [[self unauthorizedItemsChangedException] raise];
-    return;
   }
 
   // Verify each new and the view controller's tab bar items are equal.
@@ -192,7 +191,6 @@ static void *const kObservationContext = (void *)&kObservationContext;
     UITabBarItem *newTabBarItem = [items objectAtIndex:i];
     if (![viewControllerTabBarItem isEqual:newTabBarItem]) {
       [[self unauthorizedItemsChangedException] raise];
-      return;
     }
   }
 }
--- a/components/BottomNavigation/src/MDCBottomNavigationBarController.m
+++ b/components/BottomNavigation/src/MDCBottomNavigationBarController.m
@@ -181,10 +181,8 @@ static void *const kObservationContext = (void *)&kObservationContext;
 
 - (void)didUpdateNavigationBarItemsWithNewValue:(NSArray *)items {
   // Verify tab bar items correspond with the view controllers tab bar items.
-  BOOL isItemViewControllerCountEqual = items.count == self.viewControllers.count;
-  NSAssert(isItemViewControllerCountEqual, [self navigationBarItemsChangedExceptionDescription]);
-  if (!isItemViewControllerCountEqual) {
-    NSLog(@"%@", [self navigationBarItemsChangedExceptionDescription]);
+  if (items.count != self.viewControllers.count) {
+    [[self unauthorizedItemsChangedException] raise];
     return;
   }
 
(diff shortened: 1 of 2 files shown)


romoore:
@nokeeo Please add a tracking Issue for this PR and mark it "Closed" in the PR summary.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
```diff
@@ -30,11 +33,24 @@ - (instancetype)init {
     _content = [[UIView alloc] init];
     _viewControllers = @[];
     _selectedIndex = NSNotFound;
+
+#ifdef DEBUG
```
romoore: I believe this would be MDC-iOS's first introduction of DEBUG-only code. I'm wary of such an introduction since it makes it harder to determine behavior between debug and production builds.

As an alternative, what do you think of using try-catch and NSAsserting on failure + logging a message?  That would make the app crash in debug failures, not crash in production, and have the same code either way.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
```diff
@@ -148,6 +164,38 @@ - (BOOL)bottomNavigationBar:(MDCBottomNavigationBar *)bottomNavigationBar
...
+  if (object == self.navigationBar &&
+      [keyPath isEqualToString:NSStringFromSelector(@selector(items))] &&
+      [newValue isKindOfClass:[NSArray class]]) {
+    [self didUpdateNavigationBarItemsWithNewValue:(NSArray *)newValue];
+  }
+}
+
+- (void)didUpdateNavigationBarItemsWithNewValue:(NSArray *)items {
+  // Verify tab bar items correspond with the view controllers tab bar items.
+  NSAssert(items.count == self.viewControllers.count,
```
romoore: Please make this a conditional (`if`) check and log + return early. If this fails in production, the NSAssert will not fire and the loop below is likely to access an invalid array index.

romoore (changes requested):

romoore:
Looks good overall, I just want to align the debug/release versions as much as possible.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
nokeeo: Done!

components/BottomNavigation/src/MDCBottomNavigationBarController.m
```diff
@@ -148,6 +160,47 @@ - (BOOL)bottomNavigationBar:(MDCBottomNavigationBar *)bottomNavigationBar
...
+    NSLog(@"%@", [self navigationBarItemsChangedExceptionDescription]);
+    return;
+  }
+
+  // Verify each new and the view controller's tab bar items are equal.
+  for (NSUInteger i = 0; i < self.viewControllers.count; i++) {
+    UITabBarItem *viewControllerTabBarItem = [self.viewControllers objectAtIndex:i].tabBarItem;
+    UITabBarItem *newTabBarItem = [items objectAtIndex:i];
+    BOOL tabBarItemsEqual = [viewControllerTabBarItem isEqual:newTabBarItem];
+    NSAssert(tabBarItemsEqual, [self navigationBarItemsChangedExceptionDescription]);
```
romoore: We spoke offline - let's change the `navigationBarItemsChangedExceptionDescription` to return an NSException and have it raised here. This will match UITabBarController's behavior.

romoore (changes requested):

romoore (approved):

PR #6786, opened 2019-03-06
[Tabs] Fix badge text truncation bug in MDCItemBarCell

### LINKED ISSUE #4587: [Tabs] MDCTabBar truncates badge value at 3 characters even though MDCItemBarCell claims it allows 3 characters
## Description
When using a MDCTabBar as the bottom navigation bar, giving a tab bar item a badge value that is 3 characters or greater leads to truncation. So for example, a badge value of '30' shows up fine but a value of '303' would show up as '3...'.

Note that this is odd because MDCItemBarCell https://github.com/material-components/material-components-ios/blob/72045ffd6d303bb99a3e1a71e4f3cdda42fe5ad8/components/Tabs/src/private/MDCItemBarCell.m#L38 specifically calls out that 3 characters should be allowed.

## Reproduction steps

1. Go to BottomNavigationBarExample class where UITabBarItems are created https://github.com/material-components/material-components-ios/blob/3e441f22991368e1ceb3e3e130b0cb666b610b6f/components/Tabs/examples/BottomNavigationBarExample.m#L54
2. Modify the code to look like below so that we give an item a badge value

```
  UIImage *starImage =
      [UIImage imageNamed:@"TabBarDemo_ic_star" inBundle:bundle compatibleWithTraitCollection:nil];
  UITabBarItem *item = [[UITabBarItem alloc] initWithTitle:@"Red" image:infoImage tag:0];
    item.badgeValue = @"303";
  _bottomNavigationBar.items = @[
    item,
    [[UITabBarItem alloc] initWithTitle:@"Blue" image:starImage tag:0]
  ];
```
3. Run the MDCDragons app, go to Tab Bar -> Bottom Navigation

## Expected behavior

The badge value of 303 is shown without truncation

## Actual behavior

The badge value is truncated and shown as '3..'

 - Device: iPhone 7 (11.4) Simulator

screen shot 2018-07-17 at 3 40 18 pm

---
#### Internal data
- Associated internal bug: [b/117177796](http://b/117177796)

andrewoverton:
This PR fixes a bug where badge text 3 characters long was truncating when it shouldn't have been.

Before:
simulator screen shot - iphone x - 2019-03-06 at 12 45 17

After:
simulator screen shot - iphone x - 2019-03-06 at 12 42 51
 
Closes #4587.

### DIFF
--- a/components/Tabs/src/private/MDCItemBarCell.m
+++ b/components/Tabs/src/private/MDCItemBarCell.m
@@ -583,8 +583,7 @@ static const NSTimeInterval kSelectionAnimationDuration = 0.3;
   _titleLabel.text = [[self class] displayedTitleForTitle:_title style:_style];
 }
 
-- (CGSize)badgeLabelSizeWithText:(NSString *)string
-                            font:(UIFont *)font {
+- (CGSize)badgeLabelSizeWithText:(NSString *)string font:(UIFont *)font {
   if (string.length <= 0) {
     return CGSizeZero;
   }
@@ -592,22 +591,21 @@ static const NSTimeInterval kSelectionAnimationDuration = 0.3;
   NSMutableString *longestAllowableBadgeString = [[NSMutableString alloc] init];
   __block NSUInteger composedCharacterIndex = 0;
   [string enumerateSubstringsInRange:NSMakeRange(0, string.length)
-                        options:NSStringEnumerationByComposedCharacterSequences
-                     usingBlock:^(NSString *substring, NSRange substringRange,
-                                  NSRange enclosingRange, BOOL *stop)
-   {
-     [longestAllowableBadgeString appendString:substring];
-     composedCharacterIndex++;
-     if (composedCharacterIndex == kBadgeMaxTextComposedCharacterLength) {
-       *stop = YES;
-     }
-   }];
+                             options:NSStringEnumerationByComposedCharacterSequences
+                          usingBlock:^(NSString *substring, NSRange substringRange,
+                                       NSRange enclosingRange, BOOL *stop) {
+                            [longestAllowableBadgeString appendString:substring];
+                            composedCharacterIndex++;
+                            if (composedCharacterIndex == kBadgeMaxTextComposedCharacterLength) {
+                              *stop = YES;
+                            }
+                          }];
 
   CGRect largestAllowableBadgeRect =
       [[longestAllowableBadgeString copy] boundingRectWithSize:CGSizeMake(CGFLOAT_MAX, CGFLOAT_MAX)
-                                            options:NSStringDrawingUsesLineFragmentOrigin
-                                         attributes:@{NSFontAttributeName : font}
-                                            context:nil];
+                                                       options:NSStringDrawingUsesLineFragmentOrigin
+                                                    attributes:@{NSFontAttributeName : font}
+                                                       context:nil];
   return largestAllowableBadgeRect.size;
 }
 
--- a/components/Tabs/src/private/MDCItemBarCell.m
+++ b/components/Tabs/src/private/MDCItemBarCell.m
@@ -35,7 +35,7 @@ static const CGFloat kBadgeTopPadding = 6;
 
 /// Maximum badge text character length. Badge text longer than this number of characters will
 /// truncate.
-static const NSUInteger kBadgeMaxTextLength = 3;
+static const NSUInteger kBadgeMaxTextComposedCharacterLength = 4;
 
 /// Outer edge padding from spec: https://material.io/go/design-tabs#spec.
 static const UIEdgeInsets kEdgeInsets = {.top = 0, .right = 16, .bottom = 0, .left = 16};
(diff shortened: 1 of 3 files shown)


components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -249,11 +250,17 @@ - (void)layoutSubviews {
   titleCenter.x = CGRectGetMidX(contentBounds);
   titleBounds.size = titleSize;
 
-  // Horizontally align the badge.
-  CGSize badgeSize = [_badgeLabel sizeThatFits:contentBounds.size];
-  badgeSize.width = MIN(kBadgeMaxWidth, badgeSize.width);
+  // Size badge
+  NSString *longestAllowableBadgeString = [_badgeLabel.text substringToIndex:kBadgeMaxTextLength];
```
romoore: It looks like this counts non-printable characters, but I suspect it should not. Is there any reason to limit to 3 when the material.io examples for Bottom Navigation show 4 characters?

https://material.io/design/components/bottom-navigation.html#behavior
screen shot 2019-03-06 at 1 16 15 pm

romoore (changes requested):
I would suggest removing the badge width max and instead set the badge label frame to be within the bounds of the item view.  The current solution may cause problem when rendering non-Latin scripts that use multiple characters per rendered glyph.

components/Tabs/src/private/MDCItemBarCell.m
andrewoverton: The documentation for the constant `kBadgeMaxWidth` that I'm deleting implied that it should be three characters, but this picture makes it look like it's four. I can change it to four.

components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -33,8 +33,9 @@
 /// Padding between top of the cell and the badge.
 static const CGFloat kBadgeTopPadding = 6;
 
-/// Maximum width of a badge. This allows for 3 characters before truncation.
-static const CGFloat kBadgeMaxWidth = 22;
+/// Maximum badge text character length. Badge text longer than this number of characters will
```
romoore, line 36: Tip: To be clear, there is no limit to the number of characters Bottom Navigation badges support.  Here we are imposing an artificial limit of 4. I still recommend limiting the label's frame rather than attempting to count rendered glyphs.  This seems like a lot of extra effort compared to calculating the frame.

components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -582,4 +583,31 @@ - (void)updateDisplayedTitle {
...
+  NSMutableString *longestAllowableBadgeString = [[NSMutableString alloc] init];
+  __block NSUInteger composedCharacterIndex = 0;
+  [string enumerateSubstringsInRange:NSMakeRange(0, string.length)
+                        options:NSStringEnumerationByComposedCharacterSequences
+                     usingBlock:^(NSString *substring, NSRange substringRange,
+                                  NSRange enclosingRange, BOOL *stop)
+   {
+     [longestAllowableBadgeString appendString:substring];
+     composedCharacterIndex++;
+     if (composedCharacterIndex == kBadgeMaxTextComposedCharacterLength) {
```
romoore: Please make this `>=` to make sure it terminates.

romoore (approved):
Please add unit or snapshot tests to verify the behavior.

andrewoverton:
As of a60953280a736513e71363234dea7de3e369cab1, composed character sequences are considered when calculating the substring of the badgeLabel.text that should render. Additionally, the max number of composed characters is now 4, instead of 3. See here how the word "cream" does not render, but the word "baño" does:

baño:
simulator screen shot - iphone x - 2019-03-06 at 14 15 11

cream:
simulator screen shot - iphone x - 2019-03-06 at 14 14 38

components/Tabs/src/private/MDCItemBarCell.m
andrewoverton: The `*stop = YES;` line tells it to terminate, but I can do a `>=` too

components/Tabs/src/private/MDCItemBarCell.m
```diff
@@ -582,4 +583,30 @@ - (void)updateDisplayedTitle {
...
+
+  NSMutableString *longestAllowableBadgeString = [[NSMutableString alloc] init];
+  __block NSUInteger composedCharacterIndex = 0;
+  [string enumerateSubstringsInRange:NSMakeRange(0, string.length)
+                             options:NSStringEnumerationByComposedCharacterSequences
+                          usingBlock:^(NSString *substring, NSRange substringRange,
+                                       NSRange enclosingRange, BOOL *stop) {
+                            [longestAllowableBadgeString appendString:substring];
+                            composedCharacterIndex++;
+                            if (composedCharacterIndex == kBadgeMaxTextComposedCharacterLength) {
```
romoore, line 599: This was supposed to be changed to >=

andrewoverton: I don't think it's necessary, since we're saying `*stop = YES;` a line below

romoore: Yes, but if by some stroke of luck 2 characters are accumulated, it would go on forever. That's why I originally asked for `>=`.  You also said you would change it before submitting.

> The *stop = YES; line tells it to terminate, but I can do a >= too

andrewoverton: I think I'm missing something here--what value of `string` would result in this going on forever?

Whoops! Must have slipped me by. Apologies Rob. I'll put up a PR with the change.

PR #6788, opened 2019-03-06
[Collections] Clean up interface by removing NS_REQUIRES_SUPER

dmaclach:
The delegate calls that return "BOOL" have no reason to *require* subclasses to call them. They set no state in the collection view. Subclasses can certainly call [super] if they want to get the functionality of MDCCollectionViewController, but the NS_REQUIRES_SUPER is too strong a requirement for a lot of subclassing.

### DIFF
--- a/components/Collections/src/MDCCollectionViewController.h
+++ b/components/Collections/src/MDCCollectionViewController.h
@@ -49,20 +49,11 @@
  as well as ink during the highlight/unhighlight states.
  */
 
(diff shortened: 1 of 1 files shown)


ianegordon (approved):
I'll merge this once CI gives a +1.

I would recommend clients away from MDCCollections in favor of MDCCards, UICollectionCells and vanilla UICollectionControllers.

jverkoey (approved):

PR #6789, opened 2019-03-07
{Tests} Fix font comparison in Objective-C.

romoore:
Several unit tests were using identity comparison operations
(`XCTAssertEqual`) instead of equality comparison (`XCTAssertEqualObjects`).
Because UIFont supports `NSCopying`, assigning a UIFont to a `copy` property
can result in a copied object.

Unit tests seemed to flake sometimes, most recently in a Pull Request for
Dynamic Type (#6733).

QA=Unit tests pass


yarneo (approved):
Thank you Rob!!

PR #6839, opened 2019-03-12
[MDCBottomNavigationBar] Update the itemView's selected state after it is updated.

adlerj:
Without this change we were seeing two different itemView's appearing to be selected at the same time. Changing an unselected UITabBarItem's image would present in its selected color. This change ensures proper configuration of the item after images are changed.


components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -437,6 +437,7 @@ - (void)observeValueForKeyPath:(NSString *)keyPath
     } else if ([keyPath isEqualToString:kTitlePositionAdjustment]) {
       itemView.titlePositionAdjustment = [change[kNewString] UIOffsetValue];
     }
+    [itemView setSelected:itemView.selected animated:NO];
```
romoore, line 440: Does this line need to be called for every KVO or can we add it to only those paths that require an update?

romoore (changes requested):
Thanks for the PR! I've opened https://github.com/material-components/material-components-ios/pull/6843 to capture the broken behavior in a snapshot test. Once that's landed, please merge `develop` into this PR (I will add a comment to this thread when it does). 

In order to update the images, you'll need to:

1. [Install git-lfs](https://github.com/material-components/material-components-ios/blob/develop/contributing/tools.md#using-git-lfs) if you haven't already
2. In `MDCBottomNavigationBarSnapshotTests.m`, uncomment `self.recordMode = YES;`
3.  Run the tests in that file using an **iPhone 7 Simulator on iOS 11.2** (very specifically!). This will update the snapshot images.
4. Revert the changes to MDCBottomNavigationBarSnapshotTests.m so the `recordMode` is again behind a comment.
5. Commit updated screenshots (after you check them) to your branch and push them up to the PR.

Marking this as "Changes Requested" so I can take another pass at the PR once it's been updated.

romoore:
The snapshot tests have landed - please merge `develop` into this branch and then follow the steps I outlined to update the images.  Again, thanks for the PR!

components/BottomNavigation/src/MDCBottomNavigationBar.m
adlerj: Originally I only added it to image, and selected image changes but then remembered that title presentation also needs configuration. I chose to do it this way because I don't have that much knowledge of what exactly needs to be configured. If you can give guidance as to which specific changes need configuration I'll go ahead and update the PR accordingly.

romoore: Is the title presentation a separate issue? The PR is only mentioning the image color at the moment.

romoore:
While working with @adlerj offline to update the images, I found the root cause of the bug in MDCBottomNavigationItemView. I've opened https://github.com/material-components/material-components-ios/pull/6847 with the fix.

romoore:
Closing this PR since the bug was fixed in #6847

PR #6847, opened 2019-03-12
[BottomNavigation] Fix image updates in ItemView

### LINKED ISSUE #6846: [BottomNavigation] Update the itemView's selected state after it is updated
Changing an unselected UITabBarItem's image would present in its selected color.

---
#### Internal data
- Associated internal bug: [b/128432654](http://b/128432654)

romoore:
When the `selectedImage` or `image` values were changed on
MDCBottomNavigationItemView, it was accidentally updating the view's
UIImageView regardless of its current selection state. By checking the
current state and only updating as-needed, changes to the icons don't
show up incorrectly.

Fixes #6846


components/BottomNavigation/tests/snapshot/MDCBottomNavigationItemViewSnapshotTests.m
```diff
@@ -316,7 +316,7 @@ - (void)testChangeIconWhenNotSelected {
   [self generateAndVerifySnapshot];
 }
 
-- (void)testChangeIconWhenSelected {
+- (void)testChangeUnselectedImageWhenSelected {
```
romoore: This should set the selectedIcon as a Given

romoore: Also method above.

romoore:
Please add unit tests.

romoore:
Ready for review.

Should not be merged before https://github.com/material-components/material-components-ios/pull/6855 lands and the changes merged into this PR.

codeman7 (approved):

PR #6868, opened 2019-03-13
[FlexibleHeader] Fix incorrect insets during animated transitions

fluxminds:
Closes issue: https://github.com/material-components/material-components-ios/issues/6437

Change:
Allow flexible header to update insets when adjustedContentInset is changed on a tracking view that 
 is currently not being tracked.

Screenshot before change:
fv3

Screenshot after change:
iPhone7+ (10.3.1)
fv2

iPhoneX (12.0)
fv1


jverkoey (approved):
Love it. Confirming that this PR appears to be a low-risk additive change.

PR #6873, opened 2019-03-13
[Typography] Add new typography scheme with scalable fonts

ianegordon:
Adds a new scheme that vends fonts with an associated scaling curve.

This is a BREAKING CHANGE due to a new, required property in the MDCTypographyScheming protocol.
To easily migrate, clients can add the following method to their custom implementations of MDCTypographyScheming.
```
- (BOOL)mdc_adjustsFontForContentSizeCategory {
  return NO;
}
```


romoore:
Looks like a missing import rewrite:

```
Use --sandbox_debug to see verbose messages from the sandbox
components/schemes/Typography/tests/unit/MDCTypographySchemeTests.m:20:9: fatal error: 'MaterialComponents/MaterialTypography.h' file not found
#import <MaterialComponents/MaterialTypography.h>
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

romoore:
Add the following line to the `rewrite_tests` function in `.kokoro`

```
find "${stashed_dir}"components/schemes/*/tests -type f -name '*.m' -exec perl -pi -e "$1" {} + || true
```

ianegordon:
Preemptively assuming this will properly integrate with kokoro and sending out for review.

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
```diff
@@ -0,0 +1,173 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import "TypographyMaterialScalableStylesViewController.h"
+
+#import "MaterialTypography.h"
+#import "MaterialTypographyScheme.h"
+
+@implementation TypographyMaterialScalableStyleViewController {
+  NSArray<NSString *> *_strings;
```
romoore, line 22: Tip: Please use properties rather than iVars where possible.

components/Typography/examples/TypographyMaterialScalableStylesViewController.h
```diff
@@ -0,0 +1,18 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <UIKit/UIKit.h>
+
+@interface TypographyMaterialScalableStyleViewController : UITableViewController
```
romoore: Tip: Please put this interface into the `.m` file. It shouldn't need to be imported to another example.

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
```diff
@@ -0,0 +1,173 @@
...
+- (void)viewDidLoad {
+  [super viewDidLoad];
+
+  self.tableView.separatorStyle = UITableViewCellSeparatorStyleNone;
+
+  self.tableView.rowHeight = UITableViewAutomaticDimension;
+  self.tableView.estimatedRowHeight = 50.0;
+
+  _strings = @[
+    @"Material Design Components", @"A quick brown fox jumped over the lazy dog.",
```
romoore, line 39: Tip: consider using non-Latin characters to show rendering in other scripts. Suggestions: 
* A "CJK" language like Korean
* Cyrillic
* Hindi

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
```diff
@@ -0,0 +1,173 @@
...
+    @"Subtitle2", @"Subtitle2 (Scalable)", @"Headline1", @"Headline1 (Scalable)",
+    @"Headline2", @"Headline2 (Scalable)", @"Headline3", @"Headline3 (Scalable)",
+    @"Headline4", @"Headline4 (Scalable)", @"Headline5", @"Headline5 (Scalable)",
+    @"Headline6", @"Headline6 (Scalable)",
+  ];
+
+  _typography =
+      [[MDCTypographyScheme alloc] initWithDefaults:MDCTypographySchemeDefaultsMaterial201902];
+
+  _styleFonts = @[
```
romoore: Please merge `styleFonts` and `styleNames`  into a dictionary so the pairing isn't so brittle.

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
```diff
@@ -0,0 +1,173 @@
...
+    [_typography.body1 mdc_scaledFontAtDefaultSize],
```
romoore: If I understand correctly, the call to `mdc_scaledFontAtDefaultSize` is unnecessary since the scheme will return *unscaled but scalable* fonts.  Please remove it or add a comment to describe why it's required for clients.

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
```diff
@@ -0,0 +1,173 @@
...
+  return _strings.count;
+}
+
+- (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger)section {
+  return _styleNames.count;
+}
+
+- (UITableViewCell *)tableView:(UITableView *)tableView
+         cellForRowAtIndexPath:(NSIndexPath *)indexPath {
+  UITableViewCell *cell = [self.tableView dequeueReusableCellWithIdentifier:@"cell"];
```
romoore, line 118: Please be sure to [register the correct class in `viewDidLoad`](https://developer.apple.com/documentation/uikit/uitableview/1614878-dequeuereusablecellwithidentifie?language=objc).


> **Important**
You must register a class or nib file using the registerNib:forCellReuseIdentifier: or registerClass:forCellReuseIdentifier: method before calling this method.

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
```diff
@@ -0,0 +1,173 @@
...
+  if (cell == nil) {
+    cell = [[UITableViewCell alloc] initWithStyle:UITableViewCellStyleSubtitle
+                                  reuseIdentifier:@"cell"];
+  }
+  cell.textLabel.text = _strings[indexPath.section];
+  cell.textLabel.font = _styleFonts[indexPath.row];
+  cell.textLabel.numberOfLines = 0;
+  cell.textLabel.lineBreakMode = NSLineBreakByWordWrapping;
+
+  if (cell.textLabel.font.pointSize > 100 && indexPath.section == 0) {
```
romoore, line 124: This looks like you're attempting to approximate cells that are unreasonably large for the device. I would suggest comparing the `intrinsicContentSize` of the label against the cell bounds (or tableView bounds) and changing the text if it's beyond some limit.

On an iPad Pro, I would hope these large fonts would still be renderable, otherwise it implies we're carrying around effectively-useless fonts.

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
```diff
@@ -0,0 +1,173 @@
...
+    cell.textLabel.text = @"MDC";
+  }
+
+  NSString *detail = [NSString
+      stringWithFormat:@"%@ @ %.0f pt", _styleNames[indexPath.row], cell.textLabel.font.pointSize];
+  cell.detailTextLabel.text = detail;
+  cell.detailTextLabel.font = [_typography.caption mdc_scaledFontAtDefaultSize];
```
romoore, line 131: Why wouldn't this scale with Dynamic Type? A user having trouble reading the other text probably has trouble reading `caption` text.

components/schemes/Typography/src/MDCTypographyScheme.h
```diff
@@ -85,6 +85,14 @@
  */
 @property(nonatomic, nonnull, readonly, copy) UIFont *overline;
 
+@optional
+
+/**
+ Whether this typography scheme adjusts the returned fonts based on the device's Dynamic Type
```
romoore: Looks like this description is incorrect. All of the font values appear to use the `default` category size. Perhaps this is mean to read more like, "Whether consumers of this typography scheme should used scaled versions of fonts or not."?

components/schemes/Typography/src/MDCTypographyScheme.h
```diff
@@ -85,6 +85,14 @@
  */
 @property(nonatomic, nonnull, readonly, copy) UIFont *overline;
 
+@optional
```
romoore: Please remove this @optional marker.  My understanding is that subsystem schemes do not have optional properties. Instead, all clients should conform to this API and simply return `NO` if they do not support Material Dynamic Type in the fonts.

components/schemes/Typography/src/MDCTypographyScheme.m
```diff
@@ -15,18 +15,20 @@
...
-  return [self initWithDefaults:MDCTypographySchemeDefaultsMaterial201804];
+  return [self initWithDefaults:MDCTypographySchemeDefaultsMaterial201902];
 }
 
 - (instancetype)initWithDefaults:(MDCTypographySchemeDefaults)defaults {
   self = [super init];
   if (self) {
     switch (defaults) {
-      case MDCTypographySchemeDefaultsMaterial201804:
+      case MDCTypographySchemeDefaultsMaterial201804:  // FALLTHRU
```
romoore: Please create two independent sections within the switch. Having fall-through behavior is more likely than not to introduce bugs in the future.  Each case should fully encapsulate its initialization behavior.

If you prefer, consider creating helper methods or a private subclass as an implementation detail and returning that instead.

components/schemes/Typography/src/MDCTypographyScheme.m
```diff
@@ -63,6 +65,14 @@ - (instancetype)initWithDefaults:(MDCTypographySchemeDefaults)defaults {
 #endif
         break;
     }
+
+    if (defaults == MDCTypographySchemeDefaultsMaterial201902) {
```
romoore: Please move this into the `switch` statement above.  It's less likely to cause problems if we don't spread the logic out into different sections.

components/schemes/Typography/tests/unit/MDCTypographySchemeTests.m
```diff
@@ -68,4 +71,48 @@ - (void)testTypographySchemeCopy {
...
+  XCTAssertNotNil(scheme201902.headline5.mdc_scalingCurve);
+  XCTAssertNotNil(scheme201902.headline6.mdc_scalingCurve);
+  XCTAssertNotNil(scheme201902.subtitle1.mdc_scalingCurve);
+  XCTAssertNotNil(scheme201902.subtitle2.mdc_scalingCurve);
+  XCTAssertNotNil(scheme201902.body1.mdc_scalingCurve);
+  XCTAssertNotNil(scheme201902.body2.mdc_scalingCurve);
+  XCTAssertNotNil(scheme201902.caption.mdc_scalingCurve);
+  XCTAssertNotNil(scheme201902.button.mdc_scalingCurve);
+  XCTAssertNotNil(scheme201902.overline.mdc_scalingCurve);
+}
```
romoore, line 116: Please add a snapshot test to capture the new fonts.

romoore, line 116: Please add a unit test that compares the scheme fonts with the same font in the "default" content size to verify that we don't accidentally return pre-scaled fonts.

romoore (dismissed):
I would suggest splitting the example out from this PR. It would allow the typography scheme changes + tests to be merged more quickly.  The tests would verify the behavior and an example can be done as a follow-up.

components/Typography/examples/TypographyMaterialScalableStylesViewController.h
ianegordon: Respectfully disagree. I'm trying to make this example as vanilla UIKit as possible, and that means a header w/ interface and source w/ implementation.

components/schemes/Typography/src/MDCTypographyScheme.h
ianegordon: Won't that break everyone with an existing implementation?

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
ianegordon: I typically do NOT use properties for internal variables.

components/Typography/examples/TypographyMaterialScalableStylesViewController.h
romoore: I can understand the disagreement, but I believe in this case the project convention should be followed.   Unless otherwise necessary, we generally keep our examples in a single, self-contained `.m` file (or `.swift`). At the moment, the project has 187 `.m` files vs. 91 `.h` files in the `examples/` directories.  That indicates more than half of our examples files have no associated `.h` file.  Since some of those are reusable code, a number of `.h` files do make sense. However, in general we should stick to the overall convention.

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
ianegordon: Good idea, added #6958 to track

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
ianegordon: I'm not sure I follow.  One contains fonts, the other strings.  Are you recommending a tuple?  I don't think a custom object is warranted here.
Documentation added.

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
ianegordon: Removed.

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
romoore: I'm suggesting `NSDictionary<NSString *, UIFont *> *` as the type.

components/schemes/Typography/src/MDCTypographyScheme.h
romoore: Yes, that is expected and one of the reasons we have decided so far to avoid `@optional` properties.  Clients can see the breakage and decide how to proceed.  For internal clients, we can add the property as a method that returns `NO`. For external clients the release will be a Major bump with (simple) migration steps.

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
ianegordon: I want it to be accurately sized so people can see whether it makes sense for them.
This is by design.

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
ianegordon: Intentionally turned off DT in this case to make is easier to compare and contrast various size categories.
If it grows, that is fine, but I *don't* want it to shrink, so it's disabled.

components/schemes/Typography/src/MDCTypographyScheme.h
ianegordon: Comment updated.

components/schemes/Typography/tests/unit/MDCTypographySchemeTests.m
ianegordon: Added.

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
ianegordon: Done, and removed outdated if (cell).

components/Typography/examples/TypographyMaterialScalableStylesViewController.h
ianegordon: That convention has been painful.
This was mentioned at a recent meeting by a member of our team as well as outside clients.
If you feel strongly, I will change it, but with the understanding that it will be worse.

components/schemes/Typography/src/MDCTypographyScheme.m
ianegordon: We should generally try to avoid invoking other methods in an init method.
I intentionally avoid helper methods that only get called from a single place.

ianegordon: I have broken the two cases apart and eliminated the fallthru.

components/Typography/examples/TypographyMaterialScalableStylesViewController.h
romoore: Unfortunately I do not recall that meeting, but it would be good to document the problems it has caused.  I have not personally had any difficulties as a result of this convention, but it's helpful to make sure they're recorded and known by the team at large so we can make improvements.

components/schemes/Typography/tests/unit/MDCTypographySchemeTests.m
ianegordon: Add a tracking issue #6959

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
romoore: Why do we not want it to shrink? The user has explicitly requested smaller fonts to fit more content on screen.  Please add a comment in the code to explain why we should ignore the user's preference for this label.

components/schemes/Typography/src/MDCTypographyScheme.h
ianegordon: Done.

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
romoore: I'm not sure I understand your comment. My suggestion was to only change the label text if the label doesn't fit within the bounds (or is some fraction larger than the bounds).  Why do you think that would not accurately size the text?

components/Typography/examples/TypographyMaterialScalableStylesViewController.h
ianegordon: Additionally, this follows the pattern set by *every* other example in Typography.
Happy to discuss offline, but I'm not currently convinced that we should add a supplemental directory and break apart our example code into multiple files.

components/Typography/examples/TypographyMaterialScalableStylesViewController.m
ianegordon: Ah, now I follow.  Something to note is that _styleFonts actually changes, _styleNames does not.
For context, this code comes from TypographyMaterialStylesViewController.

components/Typography/examples/TypographyMaterialScalableStylesViewController.h
romoore: Please do not create a `supplemental` directory, that's not my request. I'm suggesting that you put your interface in the `.m` file. This matches the majority of our examples throughout MDC-iOS.

romoore: Headers exist for inclusion/import by the build system. In this case, we should not expect other code to import this controller. Since neither the Catalog nor Dragons requires importing headers to identify our examples, there is no need to have an extra file hanging around.

ianegordon:
> I would suggest splitting the example out from this PR. It would allow the typography scheme changes + tests to be merged more quickly. The tests would verify the behavior and an example can be done as a follow-up.

I strongly advocate that tests land with their associated code.  If we have open TODOs, changes we want to happen, I think follow-up issues are completely reasonable.  However, in this case, we need to be able to demonstrate that the new scheme works as expected.  To do that, we need the example.

components/schemes/Typography/src/MDCTypographyScheme.m
```diff
@@ -61,6 +62,51 @@ - (instancetype)initWithDefaults:(MDCTypographySchemeDefaults)defaults {
...
+        _caption = [UIFont systemFontOfSize:12.0];
+        _button = [UIFont systemFontOfSize:14.0];
+        _overline = [UIFont systemFontOfSize:12.0];
+#endif
+
+        _mdc_adjustsFontForContentSizeCategory = YES;
+
+        // We typically avoid calling an instance method in an init method, but in this case the
+        // code would be an exact duplication and we know we are in a state where calling
+        // updateFontsForSizeCategory will behave as expected.
```
romoore: `updateFontsForSizeCategory` doesn't seem to be called in `addScalingCurves`.  Can you clarify why its behavior would impact calling `addScalingCurves`?

components/schemes/Typography/src/MDCTypographyScheme.m
```diff
@@ -61,6 +62,51 @@ - (instancetype)initWithDefaults:(MDCTypographySchemeDefaults)defaults {
...
+        _body2 = [UIFont systemFontOfSize:14.0];
+        _caption = [UIFont systemFontOfSize:12.0];
+        _button = [UIFont systemFontOfSize:14.0];
+        _overline = [UIFont systemFontOfSize:12.0];
+#endif
+
+        _mdc_adjustsFontForContentSizeCategory = YES;
+
+        // We typically avoid calling an instance method in an init method, but in this case the
+        // code would be an exact duplication and we know we are in a state where calling
```
romoore: What duplication is involved? I don't see any other calls to `addScalingCurves` in the change.

components/schemes/Typography/src/MDCTypographyScheme.m
```diff
@@ -84,8 +130,70 @@ - (id)copyWithZone:(NSZone *)zone {
...
+
+  fontScaler = [[MDCFontScaler alloc] initForMaterialTextStyle:MDCTextStyleBody1];
+  _body1 = [fontScaler scaledFontWithFont:_body1];
+  _body1 = [_body1 mdc_scaledFontAtDefaultSize];
+
+  fontScaler = [[MDCFontScaler alloc] initForMaterialTextStyle:MDCTextStyleBody2];
+  _body2 = [fontScaler scaledFontWithFont:_body2];
+  _body2 = [_body2 mdc_scaledFontAtDefaultSize];
+
+  fontScaler = [[MDCFontScaler alloc] initForMaterialTextStyle:MDCTextStyleBody2];
```
romoore: Is this a duplicated line? Copy-paste error?

components/schemes/Typography/src/MDCTypographyScheme.m
ianegordon: Right you are.   Artifact from the previous behavior.
Consolidated.

components/schemes/Typography/src/MDCTypographyScheme.m
ianegordon: Almost certainly C&P.
Removed.

components/schemes/Typography/src/MDCTypographyScheme.m
ianegordon: Old docs.  Removed.

components/Typography/examples/TypographyMaterialScalableStylesViewController.h
ianegordon: Removed.

yarneo (approved):

ianegordon:
Initial Kokoro-CI hung.  Replacement run:
2673fc47-493a-4316-99e5-65e0c177cf3d

UPDATE: Kicking of CI again from the interface fails immediately.

Issue #6840, opened 2019-03-12
[Banner] Add basic Snapshot tests to Banner


wenyuzhang666:
---
#### Internal data
- Associated internal bug: [b/128528525](http://b/128528525)

PR #6895, opened 2019-03-13
[private] Include snapshot tests in examples in Podspec

fluxminds:
Part of work to unblock #6840.

We don't support adding tests under examples directory right now. But adding snapshot tests for Alpha components (Banner, TextField, ChipTextField..) is good for development, and making those tests live under examples is reasonable. 

- Added example tests into test spec source files.
- Added a dummy file into test spec to make sure swift runtime are linked correctly when XCode compiles.

### DIFF
--- a/MaterialComponentsSnapshotTests.podspec
+++ b/MaterialComponentsSnapshotTests.podspec
@@ -61,6 +61,7 @@ Pod::Spec.new do |s|
   s.requires_arc = true
   s.dependency 'MaterialComponents'
   s.dependency 'MaterialComponentsBeta'
+  s.dependency 'MaterialComponentsExamples'
 
   # Top level sources are required. Without them, unit test targets do not show up in Xcode.
   # However, no top level sources can import iOSSnapshotTestCase, otherwise the app will crash on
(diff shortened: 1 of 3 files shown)


fluxminds:
@andrewoverton PR has been updated.

andrewoverton (approved):

MaterialComponentsSnapshotTests.podspec
```diff
@@ -45,7 +45,7 @@ module SnapshotPodspecHelper
 
   def self.components
     return Dir["components/**/tests/snapshot"].map { |dir|
-      dir = Component.new(dir.split(File::SEPARATOR)[-3])
+      dir = Component.new(dir.split(File::SEPARATOR)[1])
```
romoore, line 48: @wenyuzhang666  How does this work for schemes? Why not use -3?

fluxminds: It doesn't work for schemes though scheme currently doesn't have snapshot tests. 1 is used to directly extract the directory name under `components`. Both index 1 and -3 are not good assumptions.

PR #6936, opened 2019-03-19
[Catalog] Add all tests to both schemes.

### LINKED ISSUE #6935: [Catalog] AppBar Theming Unit Tests are in Dragons but Text Fields Theming Unit Tests are in Catalog
`MaterialComponentsBeta-Unit-TextFields+Theming-UnitTests` is in the Catalog `.xcscheme` but  `MaterialComponentsBeta-Unit-AppBar+Theming-UnitTests` is in the Dragons `.xcscheme`. They should probably be in the same scheme.

romoore:
Adding all unit test and snapshot test targets to both Catalog and
Dragons schemes. It seems that at one point they both contained the same
targets, but were not updated consistently over time.

Partially implements #6935


yarneo (commented):
Testing failed:
	"_OBJC_CLASS_$_MDCActionSheetTestHelper", referenced from:
	Linker command failed with exit code 1 (use -v to see invocation)
** TEST FAILED **

romoore:
This PR should exclude that test target for now.  Once https://github.com/material-components/material-components-ios/issues/6957 is fixed, the tracking issue for this work can be picked up again.

jverkoey (approved):

codeman7 (approved):

PR #6947, opened 2019-03-19
[FeatureHighlight] New Dynamic Type curves

ianegordon:
Add support for the new dynamic type scaling curves to feature highlight.

The new curves are enabled by default, but a flag has been added allowing existing users to opt-in to the older behavior.


randallli:
removed reviewers because of WIP status

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
```diff
@@ -182,10 +182,16 @@ - (void)updateTitleFont {
   if (!_titleFont) {
     _titleFont = [MDCFeatureHighlightView defaultTitleFont];
   }
-  if (_mdc_adjustsFontForContentSizeCategory) {
-    _titleLabel.font =
-        [_titleFont mdc_fontSizedForMaterialTextStyle:kTitleTextStyle
-                                 scaledForDynamicType:_mdc_adjustsFontForContentSizeCategory];
+  if (_mdc_adjustsFontForContentSizeCategory && !_mdc_legacyFontScaling) {
+    if (_titleFont.mdc_scalingCurve) {
```
fluxminds: Does `mdc_scalingCurve` need to be checked here?

As a developer, I thought the logic could be 
```
if (__mdc_adjustsFontForContentSizeCategory) {
  if (_mdc_legacyFontScaling) {
      _titleLabel.font =
          [_titleFont mdc_fontSizedForMaterialTextStyle:kTitleTextStyle
  } else {
       _titleLabel.font = [_titleFont mdc_scaledFontForCurrentSizeCategory]; // no matter the curve exists or not
  }
} else {
  _titleLabel.font = _titleFont;
}
```
I might think in the wrong way. Please feel free to correct me.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
```diff
@@ -209,13 +215,20 @@ - (void)updateBodyFont {
   if (!_bodyFont) {
     _bodyFont = [MDCFeatureHighlightView defaultBodyFont];
   }
-  if (_mdc_adjustsFontForContentSizeCategory) {
-    _bodyLabel.font =
-        [_bodyFont mdc_fontSizedForMaterialTextStyle:kBodyTextStyle
-                                scaledForDynamicType:_mdc_adjustsFontForContentSizeCategory];
+  if (_mdc_adjustsFontForContentSizeCategory && !_mdc_legacyFontScaling) {
```
fluxminds: Same question as above.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
```diff
@@ -182,10 +182,16 @@ - (void)updateTitleFont {
   if (!_titleFont) {
     _titleFont = [MDCFeatureHighlightView defaultTitleFont];
```
fluxminds, line 183: Does `defaultTitleFont` have a curve attached? I might miss it.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
ianegordon: You are absolutely correct.
I moved the !_mdc_legacyFontScaling to the proper clause, but I kept the .mdcScalingCurve check because it adds context to the dependent clause.
Fixed.

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
ianegordon: Fixed as above.
Many Thanks!

components/FeatureHighlight/src/private/MDCFeatureHighlightView+Private.m
ianegordon: It does not and that is intentional.  It will have to be migrated after the M1 curves are removed.

fluxminds: Okay. Thanks for the answer.

fluxminds (approved):

PR #6974, opened 2019-03-22
[AppBar] Refactor AppBar example table views

brianjmoore:
This refactors the "simple" example table view controllers used in a bunch of these examples into two types: a "composed" one which creates its own table view and an "inherited" one which subclasses UITableViewController. They behave differently, so this makes that explicit and also reduces some code duplication.


jverkoey (approved):

brianjmoore:
Ready to merge!

PR #6980, opened 2019-03-22
[Tabs] Add Badge view

romoore:
[Tabs] Add Badge view


romoore:
CLA is OK. This commit came from a Googler.

romoore (commented):
This should only land after https://github.com/material-components/material-components-ios/pull/6978 does and after merging those changes and updating screenshots.

andrewoverton (approved):

PR #6985, opened 2019-03-25
[ContainedInputViews] Add API for review

### LINKED ISSUE #6943: [TextFields] Internal issue: b/128910383
This was filed as an internal issue. If you are a Googler, please visit [b/128910383](http://b/128910383) for more details.

---
#### Internal data
- Associated internal bug: [b/128910383](http://b/128910383)

andrewoverton:
This is a proposed API for MDCContainedInputView and its associated protocols.

I put together a small diagram that should hopefully help make clear the ownership relationships and channels of communication between the various elements in the ecosystem:

Screen Shot 2019-04-08 at 10 13 53 AM

Closes #6943.

### DIFF
--- a/components/ContainedInputViews/src/MDCContainedInputView.h
+++ b/components/ContainedInputViews/src/MDCContainedInputView.h
@@ -1,0 +1,319 @@
+// Copyright 2019-present the Material Components for iOS authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+/**
+ A set of Contained Input View states outlined in the Material guidelines. These states overlap with
+ and extend UIControlState.
+ */
+typedef NS_OPTIONS(NSInteger, MDCContainedInputViewState) {
+  /**
+   The default state of the contained input view.
+   */
+  MDCContainedInputViewStateNormal = 1 << 0,
+  /**
+   The state the view is in during normal editing.
+   */
+  MDCContainedInputViewStateFocused = 1 << 1,
+  /**
+   This state most closely resembles the @c selected UIControlState.
+   */
+  MDCContainedInputViewStateActivated = 1 << 2,
+  /**
+   The error state.
+   */
+  MDCContainedInputViewStateErrored = 1 << 3,
+  /**
+   The disabled state.
+   */
+  MDCContainedInputViewStateDisabled = 1 << 4,
+};
+
+/**
+ Dictates the relative importance of the underline labels, and the order in which they are laid out.
+ */
+typedef NS_ENUM(NSUInteger, MDCContainedInputViewUnderlineLabelDrawPriority) {
+  /**
+   When the priority is @c .leading, the @c leadingUnderlineLabel will be laid out first within the
+   horizontal space available for @b both underline labels. Any remaining space will then be given
+   for the @c trailingUnderlineLabel.
+   */
+  MDCContainedInputViewUnderlineLabelDrawPriorityLeading,
+  /**
+   When the priority is @c .trailing, the @c trailingUnderlineLabel will be laid out first within
+   the horizontal space available for @b both underline labels. Any remaining space will then be
+   given for the @c leadingUnderlineLabel.
+   */
+  MDCContainedInputViewUnderlineLabelDrawPriorityTrailing,
+  /**
+   When the priority is @c .custom, the @c customUnderlineLabelDrawPriority property will be used to
+   divide the space available for the two underline labels.
+   */
+  MDCContainedInputViewUnderlineLabelDrawPriorityCustom,
+};
+
+/**
+ This enum represents different states the floating label can be in.
+ */
+typedef NS_ENUM(NSUInteger, MDCContainedInputViewFloatingLabelState) {
+  /**
+   The state where the floating label is not visible.
+   */
+  MDCContainedInputViewFloatingLabelStateNone,
+  /**
+   The state where the floating label is floating.
+   */
+  MDCContainedInputViewFloatingLabelStateFloating,
+  /**
+   The state where the floating label is occupying the normal text area.
+   */
+  MDCContainedInputViewFloatingLabelStateNormal,
+};
+
+@protocol MDCContainedInputViewStyle;
+@protocol MDCContainedInputViewColorScheming;
+
+@protocol MDCContainedInputView <NSObject>
+/**
+ Dictates the @c MDCContainerStyle of the text field. Defaults to an instance of
+ MDCContainerStyleBase.
+ */
+@property(nonatomic, strong, nonnull) id<MDCContainedInputViewStyle> containerStyle;
+
+/**
+ Describes the current @c MDCContainedInputViewState of the view.
+ */
+@property(nonatomic, assign, readonly) MDCContainedInputViewState containedInputViewState;
+
+/**
+ Describes the current @c MDCContainedInputViewFloatingLabelState of the contained input view. This
+ value is affected by things like the view's state, the value for @c canFloatingLabelFloat, and the
+ text of the floating label.
+ */
+@property(nonatomic, assign, readonly) MDCContainedInputViewFloatingLabelState floatingLabelState;
+
+/**
+ The @c floatingLabel is a label that occupies the text area when there is no text and that floats
+ above the text once there is some. It is distinct from a placeholder.
+ */
+@property(strong, nonatomic, readonly, nonnull) UILabel *floatingLabel;
+/**
+ The @c leadingUnderlineLabel can be used to display helper or error text.
+ */
+@property(strong, nonatomic, readonly, nonnull) UILabel *leadingUnderlineLabel;
+
+/**
+ The @c trailingUnderlineLabel can be used to display helper or error text.
+ */
+@property(strong, nonatomic, readonly, nonnull) UILabel *trailingUnderlineLabel;
+
+/**
+ This property is used to determine how much horizontal space to allot for each of the two underline
+ labels.
+
+ @note The default value is MDCContainedInputViewUnderlineLabelDrawPriorityTrailing. The rationale
+ behind this is it is less likely to have long explanatory error text and more likely to have short
+ text, like a character counter. It is better to draw the short text first and use whatever space is
+ leftover for the longer text, which may wrap to new lines.
+ */
+@property(nonatomic, assign)
+    MDCContainedInputViewUnderlineLabelDrawPriority underlineLabelDrawPriority;
+
+/**
+ When @c underlineLabelDrawPriority is set to @c .custom the value of this property helps determine
+ what percentage of the available width each underline label gets. It can be thought of as a
+ divider. A value of @c 0 would result in the trailing underline label getting all the available
+ width. A value of @c 1 would result in the leading underline label getting all the available width.
+ A value of @c .5 would result in each underline label getting 50% of the available width.
+ */
+@property(nonatomic, assign) CGFloat customUnderlineLabelDrawPriority;
+
+/**
+ When set to YES, the floating label floats when the view becomes the first responder. When
+ set to NO it disappears.
+
+ @note The default is YES.
+ */
+@property(nonatomic, assign) BOOL canFloatingLabelFloat;
+
+/**
+ This property toggles the error state (similar to @c isHighlighted, @c isEnabled, @c isSelected,
+ etc.) that is part of a general interpretation of the states outlined in the Material guidelines
+ for Text Fields. See the @c MDCContainedInputViewState enum for more information.
+ */
+@property(nonatomic, assign) BOOL isErrored;
+
+/**
+ This property toggles the activated state (similar to @c isHighlighted, @c isEnabled, @c
+ isSelected, etc.) that is part of a general interpretation of the states outlined in the Material
+ guidelines for Text Fields. See the @c MDCContainedInputViewState enum for more information.
+ */
+@property(nonatomic, assign) BOOL isActivated;
+
+/**
+  This method returns a color scheme for a given state.
+ */
+- (nonnull id<MDCContainedInputViewColorScheming>)containedInputViewColorSchemingForState:
+    (MDCContainedInputViewState)containedInputViewState;
+
+/**
+ This method sets a color scheme for a given state.
+ */
+- (void)setContainedInputViewColorScheming:
+            (nonnull id<MDCContainedInputViewColorScheming>)containedInputViewColorScheming
+                                  forState:(MDCContainedInputViewState)textFieldState;
+
+/**
+ Returns the rect surrounding the main content, i.e. the area that the container should be drawn
+ around.
+ */
+@property(nonatomic, assign, readonly) CGRect containerFrame;
+
+/**
+ This API allows the user to override the default main content area height. The main content area is
+ the part of the view where the where the data input happens. It is located above the underline
+ label area. If this property is set to a value that's lower than the default main content area
+ height the value will be ignored in the calculation of the view's @c intrinsicContentSize.
+ */
+@property(nonatomic, assign) CGFloat preferredMainContentAreaHeight;
+
+/**
+ This API allows the user to override the default underline label area height. The underline label
+ area is the part of the view where the underline labels are. It is located below the main content
+ area. If this property is set to a value that's lower than the default underline label area height
+ the value will be ignored in the calculation of the view's @c intrinsicContentSize.
+ */
+@property(nonatomic, assign) CGFloat preferredUnderlineLabelAreaHeight;
+
+@end
+
+/**
+ This protocol represents a set of colors that are semantically meaningful and specific to
+ MDCContainedInputView. Each property corresponds to the color of one or more views that an
+ MDCContainedInputView manages at a given point of time.
+ */
+@protocol MDCContainedInputViewColorScheming <NSObject>
+/**
+ The color of the contained input view's text.
+ */
+@property(strong, nonatomic, readonly, nonnull) UIColor *textColor;
+/**
+ The color of the contained input view's underline label.
+ */
+@property(strong, nonatomic, readonly, nonnull) UIColor *underlineLabelColor;
+/**
+ The color of the contained input view's floating label.
+ */
+@property(strong, nonatomic, readonly, nonnull) UIColor *floatingLabelColor;
+/**
+ The color of the contained input view's placeholder label.
+ */
+@property(strong, nonatomic, readonly, nonnull) UIColor *placeholderColor;
+/**
+ The tint color of the contained input view's clear button.
+ */
+@property(strong, nonatomic, readonly, nonnull) UIColor *clearButtonTintColor;
+/**
+ The color the contained input view should apply during an error state.
+ */
+@property(strong, nonatomic, readonly, nonnull) UIColor *errorColor;
+@end
+
+/**
+ A base implementation of MDCContainedInputViewColorScheming. Intended to be subclassed for styles
+ that include additional elements that need to be colored.
+ */
+@interface MDCContainedInputViewColorScheme : NSObject <MDCContainedInputViewColorScheming>
+@property(strong, nonatomic, nonnull) UIColor *textColor;
+@property(strong, nonatomic, nonnull) UIColor *underlineLabelColor;
+@property(strong, nonatomic, nonnull) UIColor *floatingLabelColor;
+@property(strong, nonatomic, nonnull) UIColor *placeholderColor;
+@property(strong, nonatomic, nonnull) UIColor *clearButtonTintColor;
+@property(strong, nonatomic, nonnull) UIColor *errorColor;
+@end
+
+@protocol MDCContainedInputViewStyleDensityInforming;
+
+@protocol MDCContainedInputViewStyle <NSObject>
+/**
+ The style's densityInformer. Different implementations of MDCContainedInputView may need different
+ density informers on a per style basis.
+ */
+@property(strong, nonatomic, nonnull) id<MDCContainedInputViewStyleDensityInforming>
+    densityInformer;
+/**
+ This method provides a default object conforming to MDCContainedInputViewColorScheming.
+ */
+- (nonnull id<MDCContainedInputViewColorScheming>)defaultColorSchemeForState:
+    (MDCContainedInputViewState)state;
+/**
+  This method allows objects conforming to MDCContainedInputViewStyle to apply themselves to objects
+  conforming to MDCContainedInputView with a set of colors represented by an object conforming to
+  MDCContainedInputViewColorScheming.
+ */
+- (void)applyStyleToContainedInputView:(nonnull id<MDCContainedInputView>)inputView
+    withContainedInputViewColorScheming:(nonnull id<MDCContainedInputViewColorScheming>)colorScheme;
+/**
+  This method allows objects conforming to MDCContainedInputViewStyle to remove the styling
+  previously applied to objects conforming to MDCContainedInputView.
+ */
+- (void)removeStyleFrom:(nonnull id<MDCContainedInputView>)containedInputView;
+/**
+  The value returned by this method determines how big the font of the floating label should be when
+  it's floating.
+ */
+- (CGFloat)floatingFontSizeScaleFactor;
+@end
+
+/**
+ Objects conforming to this protocol provide information about the vertical positions of views. This
+ helps achieve the variations in floating label position across the filled and outlined styles as
+ well as the general density of the views.
+ */
+@protocol MDCContainedInputViewStyleDensityInforming <NSObject>
+/**
+ This is a value between 0 and 1 that determines the visual vertical density of the view.
+ */
+@property(nonatomic, assign) CGFloat verticalDensity;
+/**
+ This method returns the mininum Y coordinate of the floating label given its height.
+ */
+- (CGFloat)floatingLabelMinYWithFloatingLabelHeight:(CGFloat)floatingLabelHeight;
+/**
+ This method returns the top padding of the main content area (where the text is) when the floating
+ label is floating.
+ */
+- (CGFloat)contentAreaTopPaddingFloatingLabelWithFloatingLabelMaxY:(CGFloat)floatingLabelMaxY;
+/**
+ This method returns the vertical (top and bottom) padding of the main content area (where the text
+ is) when the floating label is in the normal (not floating) state.
+ */
+- (CGFloat)contentAreaVerticalPaddingNormalWithFloatingLabelMaxY:(CGFloat)floatingLabelMaxY;
+@end
+
+/**
+ A base implementation of MDCContainedInputViewStyle.
+ */
+@interface MDCContainerStyleBase : NSObject <MDCContainedInputViewStyle>
+- (nonnull instancetype)init NS_DESIGNATED_INITIALIZER;
+@end
+
+/**
+ A base implementation of MDCContainedInputViewStyleDensityInforming.
+ */
+@interface MDCContainerStyleBaseDensityInformer
+    : NSObject <MDCContainedInputViewStyleDensityInforming>
+@end
--- a/components/ContainedInputViews/src/MDCContainedInputView.h
+++ b/components/ContainedInputViews/src/MDCContainedInputView.h
@@ -83,15 +83,15 @@ typedef NS_ENUM(NSUInteger, MDCContainedInputViewFloatingLabelState) {
   MDCContainedInputViewFloatingLabelStateNormal,
 };
 
-@protocol MDCContainedInputViewStyle;
+@protocol MDCContainedInputViewStyler;
 @protocol MDCContainedInputViewColorScheming;
 
 @protocol MDCContainedInputView <NSObject>
 /**
- Dictates the @c MDCContainerStyle of the text field. Defaults to an instance of
- MDCContainerStyleBase.
+ Dictates the @c MDCContainerStyler of the text field. Defaults to an instance of
+ MDCContainerStylerBase.
  */
-@property(nonatomic, strong, nonnull) id<MDCContainedInputViewStyle> containerStyle;
+@property(nonatomic, strong, nonnull) id<MDCContainedInputViewStyler> containerStyler;
 
 /**
  Describes the current @c MDCContainedInputViewState of the view.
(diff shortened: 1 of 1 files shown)


components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+   */
+  MDCContainedInputViewStateNormal = 1 << 0,
+  /**
+   The state the view is in during normal editing.
+   */
+  MDCContainedInputViewStateFocused = 1 << 1,
+  /**
+   This state most closely resembles the @c selected UIControlState.
+   */
+  MDCContainedInputViewStateActivated = 1 << 2,
```
galiak11, line 34: Clarification:: What is the difference between Focused and Activated?

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+   */
+  MDCContainedInputViewStateErrored = 1 << 3,
+  /**
+   The disabled state.
+   */
+  MDCContainedInputViewStateDisabled = 1 << 4,
+};
+
+/**
+ Dictates the relative importance of the underline labels, and the order in which they are laid out.
```
galiak11, line 46: Tip/Suggestion: "layout priority" instead of "importance"?

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+// limitations under the License.
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+/**
+ A set of Contained Input View states outlined in the Material guidelines. These states overlap with
+ and extend UIControlState.
+ */
+typedef NS_OPTIONS(NSInteger, MDCContainedInputViewState) {
```
galiak11, line 22: Question/Clarification: Why are we using "Contained" in the name? How does "ContainedInput" is different than "Input"?

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+   When the priority is @c .custom, the @c customUnderlineLabelDrawPriority property will be used to
+   divide the space available for the two underline labels.
+   */
+  MDCContainedInputViewUnderlineLabelDrawPriorityCustom,
+};
+
+/**
+ This enum represents different states the floating label can be in.
+ */
+typedef NS_ENUM(NSUInteger, MDCContainedInputViewFloatingLabelState) {
```
galiak11, line 71: Suggestion: Use more specific name instead of "State". For example:  FloatingLabelPosition or FloatingLabelVisibility, etc.

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+- (CGFloat)contentAreaTopPaddingFloatingLabelWithFloatingLabelMaxY:(CGFloat)floatingLabelMaxY;
+/**
+ This method returns the vertical (top and bottom) padding of the main content area (where the text
+ is) when the floating label is in the normal (not floating) state.
+ */
+- (CGFloat)contentAreaVerticalPaddingNormalWithFloatingLabelMaxY:(CGFloat)floatingLabelMaxY;
+@end
+
+/**
+ A base implementation of MDCContainedInputViewStyle.
```
galiak11: Tip: What does the base implementation do? What is it for? Can you elaborate in the comments?

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+ */
+@interface MDCContainerStyleBase : NSObject <MDCContainedInputViewStyle>
+- (nonnull instancetype)init NS_DESIGNATED_INITIALIZER;
+@end
+
+/**
+ A base implementation of MDCContainedInputViewStyleDensityInforming.
```
galiak11: Tip: Same as above.

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+
+@protocol MDCContainedInputViewStyle;
+@protocol MDCContainedInputViewColorScheming;
+
+@protocol MDCContainedInputView <NSObject>
+/**
+ Dictates the @c MDCContainerStyle of the text field. Defaults to an instance of
+ MDCContainerStyleBase.
+ */
+@property(nonatomic, strong, nonnull) id<MDCContainedInputViewStyle> containerStyle;
```
galiak11: Clarification: I'm a little confused about the whole containerInput thing. Why is this a containerStyle and not an InputStyle?

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+  MDCContainedInputViewFloatingLabelStateNormal,
+};
+
+@protocol MDCContainedInputViewStyle;
+@protocol MDCContainedInputViewColorScheming;
+
+@protocol MDCContainedInputView <NSObject>
+/**
+ Dictates the @c MDCContainerStyle of the text field. Defaults to an instance of
+ MDCContainerStyleBase.
```
galiak11: Tip: What kind of style is MDCContainerStyle? Can this comment elaborate more what kind of style this is?

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+ */
+@property(nonatomic, strong, nonnull) id<MDCContainedInputViewStyle> containerStyle;
+
+/**
+ Describes the current @c MDCContainedInputViewState of the view.
+ */
+@property(nonatomic, assign, readonly) MDCContainedInputViewState containedInputViewState;
```
galiak11, line 99: Clarification: 
How does containedInputViewState correlates to UIControl's "state"? 
How do you expect classes that conform to this protocol to make sure the two align?

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+  /**
+   The state where the floating label is occupying the normal text area.
+   */
+  MDCContainedInputViewFloatingLabelStateNormal,
+};
+
+@protocol MDCContainedInputViewStyle;
+@protocol MDCContainedInputViewColorScheming;
+
+@protocol MDCContainedInputView <NSObject>
```
galiak11, line 89: Tip: Ending a protocol name with a "View" is misleading.

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+/**
+ Describes the current @c MDCContainedInputViewFloatingLabelState of the contained input view. This
+ value is affected by things like the view's state, the value for @c canFloatingLabelFloat, and the
+ text of the floating label.
+ */
+@property(nonatomic, assign, readonly) MDCContainedInputViewFloatingLabelState floatingLabelState;
+
+/**
+ The @c floatingLabel is a label that occupies the text area when there is no text and that floats
+ above the text once there is some. It is distinct from a placeholder.
```
galiak11, line 110: Tip: Can you elaborate on how this is different from a placeholder?

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+ */
+@property(strong, nonatomic, readonly, nonnull) UILabel *floatingLabel;
+/**
+ The @c leadingUnderlineLabel can be used to display helper or error text.
```
galiak11, line 114: Tip: If this is its purpose, perhaps it should be part of the name? Like helperLabel? underlineLabel seems a little vague.
Tip: Add a usage examples for the labels, so it's clear to users how they should use them.

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+ */
+@property(nonatomic, assign) CGFloat customUnderlineLabelDrawPriority;
+
+/**
+ When set to YES, the floating label floats when the view becomes the first responder. When
+ set to NO it disappears.
+
+ @note The default is YES.
+ */
+@property(nonatomic, assign) BOOL canFloatingLabelFloat;
```
galiak11, line 150: Why both floatingLabelState and canFloatingLabelFloat are needed? 
Doesn't "canFloatingLabelFloat" already tell you if the label floats or not?

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+ isSelected, etc.) that is part of a general interpretation of the states outlined in the Material
+ guidelines for Text Fields. See the @c MDCContainedInputViewState enum for more information.
+ */
+@property(nonatomic, assign) BOOL isActivated;
+
+/**
+  This method returns a color scheme for a given state.
+ */
+- (nonnull id<MDCContainedInputViewColorScheming>)containedInputViewColorSchemingForState:
+    (MDCContainedInputViewState)containedInputViewState;
```
galiak11, line 170: Not sure how this would work with the new state contrast ratio theming, since states will be affected by both color and elevation. 

It's traditionally the role of themers to determine how states are themed.  Moving it to be the responsibility of the component is contradicting the basic "dumb components" and "mutator themers" architecture of the library, which all other components follow.  

I'm concerned that deviating from the standard will introduce complexity in state theming.
State theming requires that themers have full access to all state-related component attributes and be able to choose values for any attribute in any state.

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+/**
+ This method sets a color scheme for a given state.
+ */
+- (void)setContainedInputViewColorScheming:
+            (nonnull id<MDCContainedInputViewColorScheming>)containedInputViewColorScheming
+                                  forState:(MDCContainedInputViewState)textFieldState;
+
+/**
+ Returns the rect surrounding the main content, i.e. the area that the container should be drawn
+ around.
```
galiak11, line 181: Tip: Explain how clients should use this. What is this for?

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+ */
+@property(nonatomic, assign, readonly) CGRect containerFrame;
+
+/**
+ This API allows the user to override the default main content area height. The main content area is
+ the part of the view where the where the data input happens. It is located above the underline
```
galiak11, line 187: tip: 2 "where"s

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+@property(nonatomic, assign) CGFloat preferredUnderlineLabelAreaHeight;
+
+@end
+
+/**
+ This protocol represents a set of colors that are semantically meaningful and specific to
+ MDCContainedInputView. Each property corresponds to the color of one or more views that an
+ MDCContainedInputView manages at a given point of time.
+ */
+@protocol MDCContainedInputViewColorScheming <NSObject>
```
galiak11, line 208: This name is misleading. Color scheming is used by the color subsystem to group semantic color names. This protocol groups attributes which happen to be colors. Note that Material state theming also use elevation, which is not a color, and may use more visual indicators in the future which many not be colors.

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+ The style's densityInformer. Different implementations of MDCContainedInputView may need different
+ density informers on a per style basis.
+ */
+@property(strong, nonatomic, nonnull) id<MDCContainedInputViewStyleDensityInforming>
+    densityInformer;
+/**
+ This method provides a default object conforming to MDCContainedInputViewColorScheming.
+ */
+- (nonnull id<MDCContainedInputViewColorScheming>)defaultColorSchemeForState:
+    (MDCContainedInputViewState)state;
```
galiak11, line 262: The convention in MDC components is that there's a default theming that's hard coded in the component. Than there's MDC baseline themers (ie: theming extension).  Which of these is this one? Why does it need its own API?

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: From the [material.io guidelines for text fields](https://material.io/design/components/text-fields.html):

Focused:
Screen Shot 2019-03-25 at 4 05 43 PM
Activated:
Screen Shot 2019-03-25 at 4 05 48 PM

Focused is the state during editing. Activated I think is analogous to UIControlStateSelected, but I'm not sure...

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: The notion of "containers" comes from the [material.io guidelines for text fields](https://material.io/design/components/text-fields.html#anatomy). The first item under "Anatomy" talks about containers. It shows one that is filled and one that is outlined. The idea is these views are input views that can adopt these different container styles.

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: The filled and outline styles are only achievable by calling the appropriate methods on a theming extension, but `containerStyle` is a nonnull property on MDCContainedInputView. So the base style is used until it's themed to look filled or outlined.

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: See the below graphic from material.io:
Screen Shot 2019-03-25 at 4 15 31 PM
It shows a filled thing and an outlined thing, and it calls them both containers. I interpret the filled and outlined variants as being different container _styles_, hence MDCContainedInputViewStyle. Maybe it could be MDCContainedInputViewContainerStyle?

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: The base implementation is essentially unstyled. It has two subclasses so far, one that represents a filled style and one that represents an outlined style.

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: It coexists with UIControlState. UIControlState affects the value returned by the computed MDCContainedInputViewState property, but not the other way around. MDCContainedInputViewState has states not accounted for by UIControlState, like error.

The approach I took with the textfield, for example, ensured that UIControlState behavior was unchanged.

If we had an enum or option set that represented these same Material states it might be reasonable to use that instead of introduce a new type. We don't have something like that though.

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: I actually only recently learned that these two were distinct... I had to do some refactoring on the text field to reflect this reality haha. See http://b/127976833 for more context on this particular thing

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: I carried this over from MDCTextField. I think the idea is that if it can be helper text or error text it doesn't make sense to call it either of those. Simply calling it an underline label is less limiting.

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: The floating label state gives information about the state of the floating label--is it in the "normal" position, the "floating" position, or is it not in the picture at all. The canFloatingLabelFloat property determines whether the label can float. Floating labels do not always float. I have verified this with a designer.

The value of canFloatingLabelFloat influences the value for the floatingLabelState.

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: I'm not sure either. FWIW, I haven't done anything with elevation at this point. I don't believe the material.io guidelines for text fields mention elevation, and I'm kind of using that as my source of truth for all this stuff.

Would calling it MDCContainedInputViewColorViewModel be better? We don't have to call it a scheme.

The rationale for having something like this in general: At any given point, the view needs to know what colors its subviews should be. Those colors change depending on what state the view is in. It needs to have some way to know and update the various colored elements when a state change occurs.

Some maybe good news:
- This stuff is in no way tied to MDCColorScheming. MDCColorSchemes are adapted into these schemes within the theming extensions. This may make changes to the color subsystem in the future less difficult to deal with.
- Clients will never have to mess with any of this stuff if they don't want to.

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: This is actually only here because the styles need to know where to draw the containers. I couldn't think of a better way to give the container style objects this information.

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: I wouldn't say it just _happens_ to have colors. The fact that it has colors and only colors is intentional. Maybe we could change it to have other state dependent information as well? I'm not sure what other state dependent information there is. You keep mentioning elevation, but I don't know of any elevation stuff that's relevant here. I've never heard elevation discussed in the context of text fields, which I'm using as my reference for all this stuff. I could be wrong though!

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: The styling itself is not built into any object conforming to MDCContainedInputView, so a contained input view doesn't necessarily know what needs to be colored. The style knows. This API allows the style to tell the contained input view. The result of this is that if someone did something like `containedInputView.containerStyle = aNewContainerStyle` the view would know what needs to be colored and how to color it.

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+ The color the contained input view should apply during an error state.
+ */
+@property(strong, nonatomic, readonly, nonnull) UIColor *errorColor;
+@end
+
+/**
+ A base implementation of MDCContainedInputViewColorScheming. Intended to be subclassed for styles
+ that include additional elements that need to be colored.
+ */
+@interface MDCContainedInputViewColorScheme : NSObject <MDCContainedInputViewColorScheming>
```
jverkoey, line 239: This looks like a pretty significant departure from how our components do color theming (no component defines its own color scheme).

Can we remove this protocol and class?

components/ContainedInputViews/src/MDCContainedInputView.h
```diff
@@ -0,0 +1,319 @@
...
+@property(strong, nonatomic, nonnull) UIColor *underlineLabelColor;
+@property(strong, nonatomic, nonnull) UIColor *floatingLabelColor;
+@property(strong, nonatomic, nonnull) UIColor *placeholderColor;
+@property(strong, nonatomic, nonnull) UIColor *clearButtonTintColor;
+@property(strong, nonatomic, nonnull) UIColor *errorColor;
+@end
+
+@protocol MDCContainedInputViewStyleDensityInforming;
+
+@protocol MDCContainedInputViewStyle <NSObject>
```
jverkoey: This style pattern is pretty novel, and I'm not certain I understand where or what we gain from it. Is this a proposal for an alternative approach to Themers?

Can we flatten the style methods into properties on a single type, similar to how UIKit views are typically constructed?

jverkoey (changes requested):
I have a couple fairly substantial initial comments that I feel need to be addressed before I give more detailed feedback.

components/ContainedInputViews/src/MDCContainedInputView.h
jverkoey: E.g.

```objc
@protocol MDCContainedInputViewCustomization

@property(nonatomic, assign) CGFloat floatingFontSizeScaleFactor;

...

@end
```

components/ContainedInputViews/src/MDCContainedInputView.h
jverkoey, line 89: Do we need this to be a protocol? Or put another way: what benefits are there to this being a protocol vs a concrete container view implementation?

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: Is it the fact that that it has the word "scheme" in it, or is the fact that it's a data structure comprised of colors? Which is the bad thing? If it's just the word "scheme" that's problematic, I'd be happy to change it, perhaps to "view model." That's really what it is. It's less about "primaryColor" and "secondaryColor" and more about "underlineLabelColor" and "floatingLabelColor," if that makes sense. Sort of a subtle difference.

I took great effort to ensure that the color scheming is in line with our conventions from the clients' perspective. They pass container schemes into methods like `-applyOutlinedThemeWithScheme:` and the view is themed.

Currently this stuff is crucial to how subviews and sublayers are colored, so getting rid of it would require that some other mechanism take its place.

Maybe it would be less scary if I explained how it came to be. Initially, as you might recall, the SimpleTextField owned a container scheme. Because it always had it on hand, it could look at it during every layout pass, adapt the color scheme into a set of colors for its subviews, and then apply those colors if necessary. Making the shift away from constant MDCColorScheme access to fleeting exposure during a theming extension method required that I take these state dependent combinations of colors and save them for later. What you see here is the method I chose for doing that. There are currently a lot of methods in our codebase like `-setBorderWidth:forState:`. I considered having a bunch of methods like `-setFloatingLabelColor:forState:` and `-setUnderlineLabelColor:forState:`, but decided it would be better to just have one, like `-setColorViewModel:forState:`.

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: As you might recall, I originally had an enum for TextFieldStyle. Following the initial design doc overview meeting, Randall requested that I move away from enums and instead make the styles objects. I believe the hope was that if they were objects we might be able to apply them to views other than text fields. I made the change, and I have so far been successful reusing the styles, as evidenced by both the text field and the input chip view.

This is not a proposal for an alternative approach to themers.

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: The plan is for there to be several objects (text fields, input chip views, text areas) that all have this shared set of things like floating labels and underline labels etc. If they all conform to this protocol, that makes it easy for the style objects to treat them the same way. The style objects don't have to know that they're dealing with a text field or an input chip view, they only need to know that they are applicable to whatever it is.

The differences between the text field and the input chip view are substantial enough that I don't think we should try to make them the same concrete class. However, I got started on the text areas over the weekend and have definitely not ruled out making the input chip view and text area inherit from a shared superclass. They are both essentially scrollviews inside of containers.

components/ContainedInputViews/src/MDCContainedInputView.h
galiak11: How about:
The @c floatingLabel is a label that occupies the text area when there is no text and that floats
 above the text when editing starts.  It is distinct from a placeholder that remains in place until editing starts.

components/ContainedInputViews/src/MDCContainedInputView.h
galiak11: My concern is that it's not an intuitive name for a text field and can affect overall "discoverability" of the component.  I believe most iOS users will not intuitively expect this to be a "text field".

components/ContainedInputViews/src/MDCContainedInputView.h
galiak11: Is it possible that "activated" is only for web? What type of user interaction do we expect that will make the textfield "activated"?

components/ContainedInputViews/src/MDCContainedInputView.h
galiak11: I think "container" as a concept in describing text fields is not widely known and make API that uses it less intuitive. Also container is used by our library to describe other concepts, like schemes.

components/ContainedInputViews/src/MDCContainedInputView.h
galiak11: If there is no elevation for input text, then this is not an issue. There are elevation values for input chips, but those are themed separately, not as part of this API, do they?

I think also that currently, this protocol seems to be "doing too much".  There is styling/theming API and state management API and so on. Perhaps breaking it to smaller, more focused protocols would make more sense?

And lastly, I think we should reconsider letting themers be in charge of state theming (by exposing a get/set<property>forState API) instead of having the component be in charge of that - which what happens when creating a MDCContainedInputViewColorScheming protocol.

components/ContainedInputViews/src/MDCContainedInputView.h
galiak11: I believe themers should be in charge of state theming (by exposing a get/setPropertyforState: API) instead of having the component be in charge of that - which what happens when creating a MDCContainedInputViewColorScheming protocol.

components/ContainedInputViews/src/MDCContainedInputView.h
jverkoey: > The differences between the text field and the input chip view are substantial enough that I don't think we should try to make them the same concrete class. However, I got started on the text areas over the weekend and have definitely not ruled out making the input chip view and text area inherit from a shared superclass. They are both essentially scrollviews inside of containers.

My question is focused on whether a compositional design pattern has been evaluated where a control container is provided with an arbitrary control.

For example:

```objc
@interface MDCControlContainer: UIView

- (instancetype)initWithControl:(nonnull UIControl *)control;
@property(strong, nonatomic, nonnull) UIControl *control;

@property(strong, nonatomic, readonly, nonnull) UILabel *floatingLabel;
@property(strong, nonatomic, readonly, nonnull) UILabel *leadingUnderlineLabel;
@property(strong, nonatomic, readonly, nonnull) UILabel *trailingUnderlineLabel;

@end

// Example usage

UITextField *textField = [[UITextField alloc] init];
MDCControlContainer *container = [[MDCControlContainer alloc] initWithControl:textField];
container.floatingLabel.text = @"Search";
[self.view addSubview:container];
```

jverkoey: I feel the above API would allow us to invest in a single control container implementation that can be reused by any UIControl type. If needed, there are additionally a variety of compositional ways to enable bi-directional communication between the container and the control.

jverkoey: That being said, I want to better understand the evaluative considerations between a compositional approach like the above and the proposed API.

components/ContainedInputViews/src/MDCContainedInputView.h
jverkoey: https://github.com/material-components/material-components-ios/pull/6985/files/10faae53f1a24686ab6851be47ab7306d696d27f#r269345857 is discussing a more root question, so I'll wait to respond here until that thread is resolved.

components/ContainedInputViews/src/MDCContainedInputView.h
galiak11: +1 for this concern.
I don't see users using this protocol. They will always instantiate concrete classes. And if we only maintain this protocol for ourselves, than perhaps an abstract superclass will be a better solution (exposing the API with no implementation in the super class, so classes can override with very different implementations).

jverkoey: I realize that my proposal might be missing some clarity in terms of how it will work, so I'm going to put together a small prototype to try to better explain what I think is doable with this pattern.

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: MDCInputTextField is the text field that conforms to this protocol. It has "TextField" in the name. If that is not enough to indicate that it is a text field, I imagine the course of action would be to make it very clear in the docs.

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: InputChipView is my proposed solution for input chips. It is not the focus of this API review. With InputChipView, any UIView can be added as a chip. If a client would want "Material Input Chip" style chips to be displayed in the InputChipView, they could use MDCChipViews and theme each chip they add as they add them. 

The protocol intentionally has a lot of things going on. It's supposed to contain everything that these very similar input views with containers (like the textfield and inputchipview) have in common. That was the point. And it's not actually that big. **To put it in perspective, MDCContainedInputView has 16 properties and methods. MDCTextInput has 21 properties and methods, and MDCTextInputController has 44.**

And lastly, I think we as a team are moving away from themers. The new preference is theming extensions. Contained Input Views are intended to have theming extensions. Two already exist. They set appropriate MDCContainedInputViewColorSchemes for given states.

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: MDCContainedInputView provides the state dependent accessors you're describing. Better yet, the theming extension (not a themer!) uses them to achieve theming.

EDIT: I'm re-reading this comment and wondering, when have themers exposed state dependent accessors like this? I thought I'd only seen them on components.

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: The approach you are describing is philosophically similar to MDCTextInputController. From what I can tell, the big difference is that it's also a view. It has the thing it's controlling as a subview. 

I made it very clear at the top of the initial design doc for SimpleTextFields (which I presented and received feedback on from the team) that the perceived overhead of something like an input controller was one of the prime motivators for moving towards self-contained controls. Given my prior experience with UITextField, I saw the need to retain and manage both a text field and a controller as not being idiomatic for iOS. What is so odd is I have memories of you expressing a similar sentiment at one point, around the lunch table one day haha. But anyway, I very intentionally went the route of having the views be their own things. With that being said, I think there are great opportunities for certain shared behaviors to be encapsulated by shared objects internally. We've talked a bit about this. 

I'm very surprised that this foundational decision I made months ago is now so controversial.

components/ContainedInputViews/src/MDCContainedInputView.h
andrewoverton: It is entirely possible. I don't actually know. I was just trying to bring the guidelines to life, and I knew for a fact hover was web specific. We could get rid of `.activated`? 

A user would do something like `inputTextField.isActivated = YES;`

andrewoverton:
In 2ffcf18 I renamed a few things to more closely resemble names we've used in the past. For example, `MDCContainedInputViewDensityInformer` is now `MDCContainedInputViewPositioningDelegate`. Examples of other positioning delegates include our `MDCTextInputPositioningDelegate` and UIKit's `UIBarPositioningDelegate`.

I also renamed `MDCContainedInputViewStyle` to  `MDCContainedInputViewStyler`. I think some people were unclear about whether the "styler" was actively styling things. This should hopefully make it more clear. An example of a view in our library with a name like this is `MDCCollectionViewStyler`.

andrewoverton:
I'm going to close this as obsolete, as it has become clear that if contained input views become a part of our library they will not be public, and therefore don't need an API review.

Issue #6981, opened 2019-03-22
[Tabs][BottomNavigation] Default badge color + text has insufficient contrast ratio


romoore:
## Context
The current set of colors `UIColor.redColor` and `UIColor.whiteColor` only provide 3.95:1 contrast ratio. For the small text of the badge we need 4.5:1.  Although on iOS 10+ users can customize the color using `- UITabBarItem.badgeColor`, that's not the case for iOS 8 or iOS 9.

We should default to `MDCPalette.redPalette.tint700`, which provides 4.97:1.

## Acceptance Criteria
* Default badge background and text color have a contrast ratio of 4.5:1 or higher.

---
#### Internal data
- Associated internal bug: [b/129285783](http://b/129285783)

PR #7012, opened 2019-03-31
[Tabs] Increase badge text contrast.

romoore:
The Tabs badge default text has a contrast ratio of around 4:1. For WCAG
AA-level text, we should have 4.5:1 for small text like this.  Changing
to Material Red #700 increases contrast to 5:1. Material Red #600 only
provides 4.4:1 against white.

Part of #6981


components/Tabs/src/private/MDCItemBarBadge.m
```diff
@@ -36,7 +37,7 @@ - (instancetype)initWithFrame:(CGRect)frame {
 
 - (void)commonMDCTabBarItemBadgeInit {
   if (!_badgeColor) {
-    _badgeColor = [UIColor redColor];
+    _badgeColor = MDCPalette.redPalette.tint700;
```
andrewoverton, line 40: I know it's bad to use MDC color scheme defaults as default colors in components. I guess it's okay to do that with MDCPalette colors?

romoore: Using the Material Palette colors for defaults is very often done in our components. We currently do this in:

* **ActivityIndicator** - Default cycle colors
* **NavigationDrawer** - Default handle color
* **ProgressView** - Default indicator color
* **Slider** - Default track color
* **TextFields** - Default error color

And more.

andrewoverton: I wasn't sure if it was still considered best practice. SGTM.

andrewoverton (approved):
Conflicts :/ Approved

components/Tabs/src/private/MDCItemBarBadge.m
romoore: I believe when we have to choose default colors, we prefer Material palette colors.

PR #7065, opened 2019-04-02
[Banner] Elevate Banner component to Beta.

### LINKED ISSUE #5241: Implement an MVP Banner prototype
---
#### Internal data
- Associated internal bug: [b/117123271](http://b/117123271)
- Blocked by: https://github.com/material-components/material-components-ios/issues/6771
- Blocked by: https://github.com/material-components/material-components-ios/issues/6769
- Blocked by: https://github.com/material-components/material-components-ios/issues/5563
- Blocked by: https://github.com/material-components/material-components-ios/issues/5534
- Blocked by: https://github.com/material-components/material-components-ios/issues/5529
- Blocked by: https://github.com/material-components/material-components-ios/issues/5489

fluxminds:
closes https://github.com/material-components/material-components-ios/issues/5241.

This PR moves BannerView from Alpha to Beta. The movement is following a previous pull request on BottomNavigationBarController as an example (https://github.com/material-components/material-components-ios/pull/6311), because we don't have an instruction on it currently.


jverkoey (changes requested):
Hm. It unfortunately won't be possible to review this PR because GitHub doesn't allow anyone to leave comments on files that have been solely moved. We'll unfortunately need to do this in three steps:

1. Copy the files to the new location, not move them. 
2. Review the PR as usual.
3. Once LGTM'd, do a `git mv` of the old files so that history will ideally be preserved.

fluxminds:
@jverkoey Sure. It makes sense. Addressed in https://github.com/material-components/material-components-ios/pull/7065/commits/59cd52de416644e08513bed7688b297d9a28f474.

I did it in the original way because I was following the PR mentioned in the description to move things from alpha to beta.

components/Banner/src/MDCBannerView.h
```diff
@@ -0,0 +1,48 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <UIKit/UIKit.h>
+
+#import "MaterialButtons.h"
+
+__attribute__((objc_subclassing_restricted)) @interface MDCBannerView : UIView
```
jverkoey, line 25: Please add class documentation. This should typically include:

- What the class is meant to be used for.
- A link to the Material spec for this component.
- Any meta usage considerations that may not be captured in a single API doc.

components/Banner/src/MDCBannerView.h
```diff
@@ -0,0 +1,48 @@
...
+/**
+ A view that displays the text on a @c MDCBannerView
+ The properties of @c textLabel can be used to configure the text shown on @c MDCBannerView.
+ */
+@property(nonatomic, readonly, strong, nonnull) UILabel *textLabel;
+
+/**
+ A view that displays the image on a @c MDCBannerView.
+ The properties of @c imageView can be used to configure the image shown on @c MDCBannerView.
+ If @c imageView 's own property @c image is @c nil, this view won't show on the @c MDCBannerView.
```
jverkoey: Does `won't show` mean hidden or visible, but empty?

jverkoey (dismissed):
Please add screenshot tests that capture a good amount of the expected configurations for this view.

components/Banner/src/MDCBannerView.m
```diff
@@ -0,0 +1,474 @@
...
+  [buttonContainerView addSubview:trailingButton];
+  _trailingButton = trailingButton;
+
+  [self addObserversToSubviews];
+  [self setupConstraints];
+}
+
+#pragma mark - Constraints Helpers
+
+- (void)setupConstraints {
```
jverkoey, line 158: Mark this method with `NS_AVAILABLE_IOS(9_0)` so that we can remove a layer of indentation.

fluxminds:
@jverkoey There are some screenshot testing in https://github.com/material-components/material-components-ios/blob/develop/components/Banner/examples/tests/snapshot/MDCBannerSnapshotTests.m. I expect to add more in a separate PR.

components/Banner/src/MDCBannerView.m
```diff
@@ -0,0 +1,474 @@
...
+  if (@available(iOS 9.0, *)) {
+    self.imageViewConstraintLeading =
```
jverkoey, line 169: Please document each logical group of constraints in this method with a summary of each group of constraints and their purpose.

components/Banner/src/MDCBannerView.m
```diff
@@ -0,0 +1,474 @@
...
+
+- (void)observeValueForKeyPath:(NSString *)keyPath
+                      ofObject:(id)object
+                        change:(NSDictionary *)change
+                       context:(void *)context {
+  if (!context) {
+    if ([keyPath isEqualToString:kMDCBannerViewImageViewImageKeyPath]) {
+      self.imageView.hidden = (change[NSKeyValueChangeNewKey] == [NSNull null]);
+    }
+  }
```
jverkoey: Any KVO code needs to be treated as a last resort, so it's important that you've shown you've exhausted all alternative solutions. Do you feel that hiding the image view like this is necessary behavior, or can we get away with empty image views hanging around as non-hidden empty views?

I believe I recall mention that other UIKit components work in a similar way, please list them here so that if we do go this route we can document which UIKit components are justifying this behavior.

---

Assuming we do go this route, please follow the pattern for KVO used in MDCNavigationBar. Notably it is healthy practice to:

1. Kick KVO events to the main thread to reduce the likelihood of off-main-thread UIKit interactions.
2. Define your context as a `static char *const`.
3. Invoke super when you don't recognize the context.

https://github.com/material-components/material-components-ios/blob/424a02538bd07b293e9c86020143fadf25c223fd/components/NavigationBar/src/MDCNavigationBar.m#L413-L438

components/Banner/src/MDCBannerView.m
```diff
@@ -0,0 +1,474 @@
...
+}
+
+// TODO: KVO trailingButton.hidden and call `setNeedsUpdateConstraints:`.
```
jverkoey: Please either delete, address, or file a bug.

components/Banner/src/MDCBannerView.m
```diff
@@ -0,0 +1,474 @@
...
+    } else {
+      self.textLabelConstraintLeadingWithMargin.active = YES;
+    }
+    self.textLabelConstraintTop.active = YES;
+    self.textLabelConstraintTrailing.active = YES;
+    self.buttonContainerConstraintTopWithTextLabelGreater.active = YES;
+    self.buttonContainerConstraintTopWithTextLabel.active = YES;
+    self.buttonContainerConstraintLeading.active = YES;
+    self.buttonContainerConstraintTrailing.active = YES;
+    self.buttonContainerConstraintBottom.active = YES;
```
jverkoey: `self.buttonContainerConstraintBottom.active = YES;` is in both of these logical paths. Please extract it out to the top level of this method to clearly indicate that it is common. Similarly for any other repeated logical paths.

components/Banner/src/MDCBannerView.h
fluxminds: Addressed in cc88c99719fe310f14bac8e7d4f1b590015c8218.

components/Banner/src/MDCBannerView.h
fluxminds: Changed to `is hidden` in cc88c99719fe310f14bac8e7d4f1b590015c8218 to make it clear.

jverkoey:
> @jverkoey There are some screenshot testing in https://github.com/material-components/material-components-ios/blob/develop/components/Banner/examples/tests/snapshot/MDCBannerSnapshotTests.m. I expect to add more in a separate PR.

Ah! Great! Did none of the testing code paths need to be updated as part of this PR? I expect the tests to be included and running against the latest code here.

components/Banner/src/MDCBannerView.h
```diff
@@ -0,0 +1,55 @@
...
+// limitations under the License.
+
+#import <UIKit/UIKit.h>
+
+#import "MaterialButtons.h"
+
+
+
+/**
+ The MDCBannerView class creates and configures a view to represent a Materail Banner.
```
jverkoey: Material*

components/Banner/src/MDCBannerView.m
jverkoey: Ah; I just noticed the comments near the top of this .m file which are helpful.

This method is still a bit logically complex, so it will benefit from some summarization in terms of what's going on.

jverkoey: Consider breaking this method up into smaller, named methods as a form of self-documentation.

components/Banner/src/MDCBannerView.h
fluxminds: Addressed in 096eec70086c58cbf2dc8c8bf6ef6293021bd131.

components/Banner/src/MDCBannerView.m
fluxminds: Addressed in aeeecd3982520bfa91577f0db37cb0e67005d986.

components/Banner/src/MDCBannerView.m
```diff
@@ -0,0 +1,474 @@
...
+  if (@available(iOS 9_0, *)) {
+    [self setupConstraints];
+  }
+}
+
+#pragma mark - Constraints Helpers
+
+- (void)setupConstraints NS_AVAILABLE_IOS(9_0) {
+  self.imageViewConstraintLeading =
+  [self.imageView.leadingAnchor constraintEqualToAnchor:self.layoutMarginsGuide.leadingAnchor
```
romoore: Please use if(@available(iOS 9.0, *)){ here to protect against running it on iOS 8.

romoore (dismissed):

components/Banner/src/MDCBannerView.m
fluxminds: It is a very good point which I was also having questions on and unsure about the solutions.

I am not a fan of KVO, so I definitely want to minimize the usage of it in this class. 

> Do you feel that hiding the image view like this is necessary behavior, or can we get away with empty image views hanging around as non-hidden empty views?
I believe I recall mention that other UIKit components work in a similar way, please list them here so that if we do go this route we can document which UIKit components are justifying this behavior.

Yes, we can. I just removed it in 8a5cc283a6b098cd5247404766ede46d078d66db. The behavior I wanted to mimic was `setImage:forState:`, which UIKit allows developers to `setImage:nil forState:..` to set image to be empty and re-layout the button. I think one way to do it is provide a new API (`setImageHidden:(BOOL)hidden`) similar to that, so I can directly update status for constraints. I didn't do that in effort of minimizing the number of APIs at the beginning, but you feel it would be a good way. I think we can add that.

And we can also provide API to hide `trailingButton` by `setTrailingButtonHidden:(BOOL)hidden` instead of `// TODO: KVO trailingButton.hidden and call setNeedsUpdateConstraints:`

Adding these APIs would also facilitate to integrate to state system in the future if needed.

How do you feel about it? @jverkoey

components/Banner/src/MDCBannerView.m
fluxminds: Same concerns as mentioned above (https://github.com/material-components/material-components-ios/pull/7065#discussion_r271880532).

components/Banner/src/MDCBannerView.m
fluxminds: Sure. I can do it. One thing I am thinking of is would https://github.com/material-components/material-components-ios/pull/7065/files/aeeecd3982520bfa91577f0db37cb0e67005d986#diff-98d48786cc5ed35541325aca00c9c007R155 be enough to guard it? Add another `@available` will be another layer of runtime of safeguard but does it mean all methods annotated with `NS_AVAILABLE_IOS` need to have this additional runtime safeguard internally? I am not sure if UIKit does it or not.

romoore: If there's an alternative, then I would suggest that the solution above (version-checking before each invocation) is good. However, if there's no alternative, why generate so many `if` blocks? If it's always a no-op, make the method itself a no-op.  It's also much safer that way.

fluxminds: Edited. I got what you were saying.

I feel the complier doesn't give warning when I call `constraintEqualToAnchor:self` in a method annotated `NS_AVAILABLE_IOS`, which means it is probably fine from their perspective. But I will add the check here too.

fluxminds: Addressed in 88e43257d5d16bda0726dac9632fddb2beb9290c.

components/Banner/src/MDCBannerView.m
fluxminds: Sure! Addressed in bdbf5e0e1e0afc268adce436ed34138ebeda821e.

components/Banner/src/MDCBannerView.m
fluxminds: Addressed in 68bf18bd3a26fbf092e5113a063cb4b4903aa25e.

components/Banner/examples/tests/snapshot/MDCBannerSnapshotTests.m
```diff
@@ -161,6 +166,7 @@ - (void)testLongTextWithTwoActionsRTL {
   [button2 setTitleColor:UIColor.blackColor forState:UIControlStateNormal];
   button2.uppercaseTitle = YES;
   [self changeViewToRTL:self.bannerView];
+  self.bannerView.imageView.hidden = YES;
 
   // Then
   [self generateSnapshotAndVerifyForView:self.bannerView];
```
jverkoey, line 172: Please move these tests to the tests/ directory and configure them according to our component testing conventions (BUILD targets namely).

components/Banner/src/MDCBannerView.m
jverkoey: Ah this one wasn't resolved. To arm you with the underlying motivation here:

Any private method that solely consists of logic nested within an `if (@available(` statement and is not a subclass override should be annotated with NS_AVAILABLE_IOS instead. This allows us to remove a layer of logical indentation and more clearly communicates that the method itself is only available for a given platform version.

With the above definition you should also not need to both annotate a method and include the runtime check, because by definition only this class is expected to invoke these methods and the corresponding call-sites can be correctly annotated with the availability checks.

jverkoey: I've filed https://github.com/material-components/material-components-ios/issues/7070 to track this type of feedback so we don't churn on it further.

components/Banner/src/MDCBannerView.m
jverkoey: I don't fully understand the proposed changes here, so I'll try to echo back what I believe the assumptions we're working with here are:

1. If a client manually hides one of the buttons, we won't know that the visible state has changed and layout will not be updated accordingly, resulting in an incorrect layout.
2. We do not expect clients to know to invalidate the layout of the banner view if they hide one of the buttons.
3. In order to encourage good behavior, it is proposed that we add an API that clients can use instead of directly setting the button to hidden. This new API would also call setNeedsUpdateConstraints.

Does this match your chain of assumptions?

jverkoey (dismissed):
Requesting that all tests be moved to the expected directories as part of this migration.

components/Banner/examples/tests/snapshot/MDCBannerSnapshotTests.m
fluxminds: Those were moved to the expected directory. I made these changes first in the original one to make it work and then copied these to the right directory (https://github.com/material-components/material-components-ios/pull/7065/files#diff-62e09c98ab764ee3afe12efbb897870b) for Github review because of the reason you mentioned (https://github.com/material-components/material-components-ios/pull/7065#pullrequestreview-221934468) about Github diff.

components/Banner/src/MDCBannerView.m
fluxminds: > If a client manually hides one of the buttons, we won't know that the visible state has changed and layout will not be updated accordingly, resulting in an incorrect layout.

Yes, unless we add KVO for the `hidden` property and `setNeedUpdateContraints` in the handler of KVO.

> We do not expect clients to know to invalidate the layout of the banner view if they hide one of the buttons.

Right.

> In order to encourage good behavior, it is proposed that we add an API that clients can use instead of directly setting the button to hidden. This new API would also call setNeedsUpdateConstraints.

It is exactly what I proposed:)

components/Banner/src/MDCBannerView.m
fluxminds: This one was addressed and then changed because of another code review comment (https://github.com/material-components/material-components-ios/pull/7065#discussion_r271731598) later. I was kinda of confused on the convention/guideline here, and I felt either way could work.

Thanks for filing the issue.

components/Banner/examples/tests/snapshot/MDCBannerSnapshotTests.m
fluxminds: I will delete the original files after review and before merge.

components/Banner/src/MDCBannerView.m
jverkoey: It sounds then there are two options:

1. add a new API, or
2. implement KVO

both of which are additive changes in terms of API or behavior.

For now, I'd like to encourage a conservative approach where we do neither, and instead document that if someone hides a button then they need to invalidate the layout (assuming this is a requirement). We can additively address this with one or the other approach once we have more data to indicate that this is a problem we need to solve.

components/Banner/src/MDCBannerView.h
```diff
@@ -0,0 +1,53 @@
...
+__attribute__((objc_subclassing_restricted)) @interface MDCBannerView : UIView
+
+/**
+ A view that displays the text on a @c MDCBannerView
+ The properties of @c textLabel can be used to configure the text shown on @c MDCBannerView.
+ */
+@property(nonatomic, readonly, strong, nonnull) UILabel *textLabel;
+
+/**
+ A view that displays the image on a @c MDCBannerView.
```
jverkoey, line 34: Please add a comment explaining that `setNeedsUpdateConstraints` `setNeedsLayout` both need to be called if the visibility of these properties is affected.

jverkoey (approved):
LGTM pending one request for additional documentation.

components/Banner/tests/snapshot/MDCBannerSnapshotTests.m
```diff
@@ -0,0 +1,175 @@
...
+
+#import "MaterialSnapshot.h"
+
+#import "MDCBannerView.h"
+#import "MaterialButtons.h"
+
+static NSString *const kBannerShortText = @"tristique senectus et";
+static NSString *const kBannerLongText =
+    @"Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt.";
+static const CGFloat kBannerContentPadding = 10.0f;
```
jverkoey: Rename these constants from `kBanner` to `k` as a prefix.

components/Banner/src/MDCBannerView.h
fluxminds: Adddressed in 24a98a564e411935da217b55136e14bd40815144

PR #7082, opened 2019-04-05
[TextFields] Outlined Textfield Theming Extension

### LINKED ISSUE #6836: [TextFields] Add a theming extension

andrewoverton:
This PR adds a theming extension for MDCTextInputControllerOutlined and updates the theming docs for TextFields.

Closes #6836.

### DIFF
--- a/components/TextFields/src/Theming/MDCTextInputControllerOutlined+MaterialTheming.h
+++ b/components/TextFields/src/Theming/MDCTextInputControllerOutlined+MaterialTheming.h
@@ -12,5 +12,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#import "MDCTextInputControllerFilled+MaterialTheming.h"
-#import "MDCTextInputControllerOutlined+MaterialTheming.h"
+#import <MaterialComponents/MaterialTextFields.h>
+#import <MaterialComponentsBeta/MaterialContainerScheme.h>
+
+/**
+ This category is used to style MDCTextInputControllerOutlined instances to a specific Material style
+ which can be found within the
+ [Material Guidelines](https://material.io/design/components/text-fields.html).
+ */
+@interface MDCTextInputControllerOutlined (MaterialTheming)
+
+/**
+ Applies a container scheme to this instance.
+
+ @param scheme A container scheme instance containing any desired customizations to the theming
+ system.
+ */
+- (void)applyThemeWithScheme:(nonnull id<MDCContainerScheming>)scheme;
+
+@end
--- a/components/TextFields/src/Theming/MDCTextInputControllerOutlined+MaterialTheming.h
+++ b/components/TextFields/src/Theming/MDCTextInputControllerOutlined+MaterialTheming.h
@@ -16,9 +16,9 @@
 #import <MaterialComponentsBeta/MaterialContainerScheme.h>
 
 /**
- This category is used to style MDCTextInputControllerOutlined instances to a specific Material style
- which can be found within the
- [Material Guidelines](https://material.io/design/components/text-fields.html).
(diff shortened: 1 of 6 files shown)


components/TextFields/src/Theming/MDCTextInputControllerOutlined+MaterialTheming.h
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ This category is used to style MDCTextInputControllerOutlined instances to a specific Material
+ style which can be found within the [Material
+ Guidelines](https://material.io/design/components/text-fields.html).
+ */
+@interface MDCTextInputControllerOutlined (MaterialTheming)
+
+/**
+ Applies a container scheme to this instance.
```
codeman7, line 26: _Nit:_ Applies a Material theme to this instance

components/TextFields/tests/unit/Theming/TextFieldsMaterialThemingTests.swift
```diff
@@ -29,6 +29,11 @@ class TextFieldsMaterialThemingTests: XCTestCase {
   let filledIndicatorLineAlpha: CGFloat = 0.42
   let filledIconAlpha: CGFloat = 0.54
 
+  let outlinedTextFieldActiveAlpha: CGFloat = 0.87
```
codeman7, line 32: Do you need the `CGFloat` won't this work?
```
let outlinedTextFieldActiveAlpha = 0.87
```

codeman7 (approved):

components/TextFields/tests/unit/Theming/TextFieldsMaterialThemingTests.swift
andrewoverton: I was just following what was in the file before the change. FWIW, I think explicit types make Swift compilation faster... Probably not by much, but that makes sense, right?

components/TextFields/src/Theming/MDCTextInputControllerOutlined+MaterialTheming.h
andrewoverton: I just stuck with the language in the MDCTextInputControllerFilled theming extension

romoore: @andrewoverton Please correct this in a follow-up PR. If you see other similar errors in Text Fields, please correct them as well.  We shouldn't propagate mistakes to new code when we catch them.

andrewoverton: @romoore I didn't think there was anything wrong with it :/ But if you feel strongly I'm happy to go back and change it.

Should I have it specify "id\<MDCContainerScheming\>" rather than "container scheme"? Or do you like Cody's suggestion?

romoore: @codeman7  is correct - the method applies a Material theme to the receiver.  It's not applying a scheme object (regardless of its type) to the receiver.  The scheme is a necessary parameter to the behavior of performing the theming.

andrewoverton: Sounds good :)

romoore: I think it's also worth mentioning that non-*Tip:* comments should be addressed to the reviewer's satisfaction before merging. In this case it was a Nit, but a worthwhile change to help  the reader understand the purpose of the API.

PR #7126, opened 2019-04-11
[Dialogs] Layout MDC Dialogs correctly for tall fonts.

### LINKED ISSUE #7062: [Dialogs] The layout of dialogs is bad when using large point sizes or tall fonts
Dialogs are not laid out correctly when using larger point sizes (e.g. Dynamic type) and/or fonts that are tall for their point size (e.g. Noto Nastaliq Urdu)

## Reproduction steps

1. Create a dialog with a message body and two action buttons in Urdu language, with action button text short enough to lay out horizontally. 
2. Set the font to 22pt Noto Nastaliq Urdu
3.  Run on a small screen device like iPod Touch (6th gen)
4.  Display the dialog

## Expected behavior

The dialog should lay out such that the buttons are laid out horizontally, and the entire message and both buttons are fully visible within the dialog.

## Actual behavior

The buttons are are laid out horizontally, but the button of the buttons are cut off.

 - Device: iPod Touch 6th gen
 - OS: iOS 12.1.4

## Additional context

N/A

---
#### Internal data
- Associated internal bug: [b/129781087](http://b/129781087)

fluxminds:
Closes https://github.com/material-components/material-components-ios/issues/7062

Reference: cl/241537224

Layout MDC Dialogs correctly for tall fonts

The MDC Dialog layout code had some hard coded assumptions about the amount of vertical space to allocate for action buttons that were violated when using taller fonts such as Urdu with large Dynamic Type sizes.
There was also a related bug where vertical vs. horizontal action button layout was calculated in two different places.

Snapshot test before change:
snapshot_test_goldens_goldens_64_MDCAlertControllerLocalizationSnapshotTests_testPreferredContentSizeWithNotoNastaliqUrdu_11_2@2x_Before_01dd7fb1a36bc24b27e05c1079c6420994c9f904

Snapshot test result after change:
snapshot_test_goldens_goldens_64_MDCAlertControllerLocalizationSnapshotTests_testPreferredContentSizeWithNotoNastaliqUrdu_11_2@2x_After_78d3fd9172a202641400c0536a9c2e3a33b0455b

Example screenshot before change:
Simulator Screen Shot - iPhone 7 - 2019-04-15 at 17 42 09

Example screenshot after change:
Simulator Screen Shot - iPhone 7 - 2019-04-15 at 17 43 58


jverkoey (commented):
Are there any snapshot tests that were affected by this, or that could be written for this?

romoore (dismissed):
+1 with @jverkoey. This PR requires snapshot tests as mentioned in the internal code reviews that took place earlier.

fluxminds:
Snapshot test was added in https://github.com/material-components/material-components-ios/commit/184ea7380c93b34d48cf879bceb8793e13dc2bee.

jverkoey:
Please add before/after screenshots of the example.

jverkoey (commented):
Is the snapshot test code missing from this PR?

jverkoey:
Ah, nevermind; saw that the snapshot test landed in https://github.com/material-components/material-components-ios/commit/184ea7380c93b34d48cf879bceb8793e13dc2bee

jverkoey (approved):

components/Dialogs/examples/DialogsTallTextAlertExampleViewController.m
```diff
@@ -0,0 +1,99 @@
...
+    dialogBodyFont = [UIFont systemFontOfSize:20.0];
+    dialogButtonFont = [UIFont systemFontOfSize:26.0];
+  }
+  alert.messageFont = dialogBodyFont;
+  alert.buttonFont = dialogButtonFont;
+
+  MDCAlertAction *retryAction = [MDCAlertAction actionWithTitle:@"دوبارہ کوشش کریں" handler:nil];
+  MDCAlertAction *cancelAction = [MDCAlertAction actionWithTitle:@"منسوخ کریں" handler:nil];
+  [alert addAction:retryAction];
+  [alert addAction:cancelAction];
```
galiak11, line 85: Can we theme the alert before presenting it?
I believe the recommended logic (that we should promote here) is to call our themers, and then override required values, like which fonts to use for buttons, after our themers are called.

components/Dialogs/src/private/MDCAlertControllerView+Private.h
```diff
@@ -31,6 +31,8 @@
 
 - (CGSize)calculatePreferredContentSizeForBounds:(CGSize)boundsSize;
 
+- (CGSize)calculateActionsSizeThatFitsWidth:(CGFloat)boundingWidth;
```
galiak11, line 34: Can you please add a comment describing what this method is doing?
Why is it needed in a private header?

components/Dialogs/tests/unit/MDCAlertControllerLayoutTests.m
```diff
@@ -0,0 +1,80 @@
...
+  UIFont *dialogBodyFont;
+  UIFont *dialogButtonFont;
+  if (@available(iOS 11, *)) {
+    // Noto Nastaliq Urdu was added in iOS 11, and is an extremely tall
+    // font for any given nominal point size.
+    dialogBodyFont = [UIFont fontWithName:urduFontName size:20.0];
+    dialogButtonFont = [UIFont fontWithName:urduFontName size:26.0];
+  } else {
+    dialogBodyFont = [UIFont systemFontOfSize:20.0];
+    dialogButtonFont = [UIFont systemFontOfSize:26.0];
```
galiak11, line 51: Shouldn't we test other tall fonts in non ios11 versions as well?

components/Dialogs/tests/unit/MDCAlertControllerLayoutTests.m
```diff
@@ -0,0 +1,80 @@
...
+  [self setupLargePointSizeUrduNetworkRetryDialog];
+  MDCAlertControllerView *view = (MDCAlertControllerView *)_alertController.view;
+
+  CGSize sizeOnFourInchPortrait = [view calculateActionsSizeThatFitsWidth:kFourInchPortraitWidth];
+  CGSize sizeOnFourInchLandscape = [view calculateActionsSizeThatFitsWidth:kFourInchLandscapeWidth];
+
+  // The vertical layout should be approximately twice as high as the horizontal layout, +/-
+  // padding between buttons.
+  XCTAssertEqualWithAccuracy(sizeOnFourInchLandscape.height * 2, sizeOnFourInchPortrait.height, 5);
+  XCTAssertGreaterThan(sizeOnFourInchLandscape.width, sizeOnFourInchPortrait.width);
```
galiak11, line 77: How do these test on tablets? Is this assumption still correct?

components/Dialogs/examples/DialogsTallTextAlertExampleViewController.m
fluxminds: Does it mean `buttonFont` and `messageFont` APIs are deprecated or discouraged to be used?

components/Dialogs/src/private/MDCAlertControllerView+Private.h
fluxminds: I think it is only being used in unit test.

components/Dialogs/tests/unit/MDCAlertControllerLayoutTests.m
fluxminds: I think this test performs specifically with `kFourInchPortraitWidth` and `kFourInchLandscapeWidth` from https://github.com/material-components/material-components-ios/pull/7126/files#diff-868f6100113579fc730dbb66f171761eR71, so it is platform independent.

components/Dialogs/tests/unit/MDCAlertControllerLayoutTests.m
fluxminds: I think that would be ideal. `NotoNastaliqUrdu` is the known tall font that was introduced and broke the functionality, I think it is fine to leave it for pre iOS 11 until we get more information on any tall font pre iOS 11 that breaks the dialog.

components/Dialogs/examples/DialogsTallTextAlertExampleViewController.m
fluxminds: Opened https://github.com/material-components/material-components-ios/issues/7148 to track.

galiak11 (approved):

components/Dialogs/src/private/MDCAlertControllerView+Private.h
romoore: Then it should be named `+Testing.h` and not `+Private.h`

fluxminds: I think `calculatePreferredContentSizeForBounds` also calls `calculateActionsSizeThatFitsWidth` internally, so ideally this method should be exposed in this `+Private.h` and we need to create another API in `+Testing.h` to call this method. I can address that but I am not sure if this is something we  usually do in our codebase.

PR #7128, opened 2019-04-11
[Dialogs] fix dialog snapshot tests not using preferredContentSize.

fluxminds:
When working on https://github.com/material-components/material-components-ios/pull/7126. The view size is not set to be the correct one in snapshot tests. This PR fixes it.


jverkoey (approved):

romoore (commented):
These should have been new tests. The test methods are about the default appearance of the view, and not when the preferredContentSize is used as the bounds.

romoore:
Fixing these in https://github.com/material-components/material-components-ios/pull/7134

PR #7144, opened 2019-04-15
[TextFields] Create concrete MDCFilledTextField and MDCOutlinedTextField classes based on the Contained Input View architecture

### LINKED ISSUE #7111: [TextFields] Internal issue: b/130306935
This was filed as an internal issue. If you are a Googler, please visit [b/130306935](http://b/130306935) for more details.

---
#### Internal data
- Associated internal bug: [b/130306935](http://b/130306935)

andrewoverton:
In this PR I have created MDCFilledTextField and MDCOutlinedTextField concrete classes that inherit from an MDCInputTextField base class. There are no significant changes to how the text fields work under the hood--they are still contained input views. However, that protocol conformance is now private. If we were to include these classes in our library clients would not need to have any idea what a Contained Input View is. For discussion on the rationale for doing this see the ill-fated https://github.com/material-components/material-components-ios/pull/6985 and [b/130306935](http://b/130306935).
 
Closes #7111.


jverkoey (approved):
Not reviewing due to this being experimental code. Lgtm’ing so that we can land the checkpoint.

PR #7146, opened 2019-04-15
[BottomNavigation] Adds a large item dialog view

### LINKED ISSUE #7122: [BottomNavigation] Create pop-up view for .largeContentSizeImage support
This was filed as an internal issue. If you are a Googler, please visit [b/130344111](http://b/130344111) for more details.

---
#### Internal data
- Associated internal bug: [b/130344111](http://b/130344111)

nokeeo:
Adds a view for displaying a large icon and title when the user has dynamic type enabled.

This is part of a larger effort to add dynamic type support to `MDCBottomNavigationBarController`.

Closes #7122


jverkoey:
Because this is a part of a larger effort, please add context to the PR description on the need for a separate `MDCBottomNavigationLargeItemDialogView` + `MDCBottomNavigationSystemDialogView` implementation. Is it possible to consolidate these two types into one?

components/BottomNavigation/src/private/MDCBottomNavigationLargeItemDialogView.h
```diff
@@ -0,0 +1,31 @@
...
+#import "MDCBottomNavigationSystemDialogView.h"
+
+/**
+ * The transient dialog view that displays a UITabBarItem with a larger icon and title font size.
+ */
+@interface MDCBottomNavigationLargeItemDialogView : MDCBottomNavigationSystemDialogView
+
+/**
+ * Updates this view's image and title views with the corresponding properties of the given tab bar
+ * item. If the tab bar item's largeContentSizeImage is not nil that image will be used otherwise
```
romoore: Tip: Please use `@c` in front of variable/property/method names.

components/BottomNavigation/src/private/MDCBottomNavigationSystemDialogView.m
```diff
@@ -0,0 +1,96 @@
...
+- (instancetype)initWithFrame:(CGRect)frame {
+  self = [super initWithFrame:frame];
+  if (self) {
+    [self commonInitMDCBottomNavigationSystemDialogView];
+  }
+
+  return self;
+}
+
+- (void)commonInitMDCBottomNavigationSystemDialogView {
```
romoore: Nit/naming: Please name this `commonMDCBottomNavigationSystemDialogViewInit` to match the general pattern of the code base.

components/BottomNavigation/src/private/MDCBottomNavigationSystemDialogView.m
```diff
@@ -0,0 +1,96 @@
...
+  self.layer.masksToBounds = YES;
+
+  CGFloat margin = kCornerRadius + kMargins;
+  UIEdgeInsets layoutMargins = UIEdgeInsetsMake(margin, margin, margin, margin);
+  if (UIAccessibilityIsReduceTransparencyEnabled()) {
+    self.backgroundColor = [UIColor colorWithWhite:(CGFloat)0.8 alpha:(CGFloat)1];
+    self.layoutMargins = layoutMargins;
+  } else {
+    _blurView = MDCInitializeCompatibleBlurView();
+    _blurView.contentView.layoutMargins = layoutMargins;
```
romoore, line 76: Will the blur view not inherit the layoutMargins from its view hierarchy?  It would be nice if we could just set `self.layoutMargins`. If not, let's document why that seems to be the case.

components/BottomNavigation/src/private/MDCBottomNavigationSystemDialogView.m
```diff
@@ -0,0 +1,96 @@
...
+
+  return self;
+}
+
+- (void)layoutSubviews {
+  [super layoutSubviews];
+  if (self.blurView) {
+    CGFloat height = CGRectGetHeight(self.bounds);
+    CGFloat width = CGRectGetWidth(self.bounds);
+    self.blurView.frame = CGRectMake(0, 0, width, height);
```
romoore: Tip: Please consider setting the blurView bounds + center rather than the frame. Alternatively, consider relying on the autoresizing mask (I see you haven't disabled it).

components/BottomNavigation/src/private/MDCBottomNavigationLargeItemDialogView.m
```diff
@@ -0,0 +1,133 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import "MDCBottomNavigationLargeItemDialogView.h"
+
+#import <CoreGraphics/CoreGraphics.h>
+
+static const CGFloat kTitleFontScaling = (CGFloat)0.5;
+static const CGFloat kTitleFontSize = 35;
+static const CGFloat kImageViewHeightToDialogRatio = (CGFloat)0.35;
```
romoore: Please document this constant. Its purpose is not clear to me from reading the name.  Specifically, I'm not sure how to compare a height (scalar) to a dialog (object) to produce a ratio.

components/BottomNavigation/src/private/MDCBottomNavigationLargeItemDialogView.h
```diff
@@ -0,0 +1,31 @@
...
+ * the item's image property may be scaled up.
+ */
+- (void)updateWithTabBarItem:(UITabBarItem *)item;
```
romoore: Please mark the `item` paramter as `nonnull`

components/BottomNavigation/src/private/MDCBottomNavigationLargeItemDialogView.m
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * Returns an image representing the given UITabBarItem. If the item has a largeContentSizeImage,
+ * this function will return that image, otherwise it returns the value of the item's image
+ * property. This value may be nil.
+ */
+static UIImage *_Nullable MDCImageForItem(UITabBarItem *_Nullable item) {
```
romoore: Please make this API `nonnull` if possible.

components/BottomNavigation/src/private/MDCBottomNavigationLargeItemDialogView.m
```diff
@@ -0,0 +1,133 @@
...
+  [self.contentView addSubview:_imageView];
+
+  _titleLabel = [[UILabel alloc] init];
+  _titleLabel.textAlignment = NSTextAlignmentCenter;
+  _titleLabel.lineBreakMode = NSLineBreakByTruncatingTail;
+  _titleLabel.adjustsFontSizeToFitWidth = YES;
+  _titleLabel.minimumScaleFactor = kTitleFontScaling;
+  _titleLabel.textColor = contentColor;
+  _titleLabel.font = [UIFont systemFontOfSize:kTitleFontSize];
+  _titleLabel.numberOfLines = (_imageView.image) ? 1 : 0;
```
romoore: Is this the behavior of UITabBarController? Since MDCBottomNavigationBar supports multiple lines I want to make sure we're intentionally removing that capability here.

components/BottomNavigation/src/private/MDCBottomNavigationLargeItemDialogView.m
```diff
@@ -0,0 +1,133 @@
...
+}
+
+- (void)layoutSubviews {
+  [super layoutSubviews];
+
+  UIEdgeInsets margins = self.layoutMargins;
+  CGFloat dialogHeight = MAX(0, CGRectGetHeight(self.bounds) - margins.top - margins.bottom);
+  CGFloat dialogWidth = MAX(0, CGRectGetWidth(self.bounds) - margins.left - margins.right);
+  CGFloat imageHeight =
+      self.imageView.image ? (CGFloat)floor(dialogHeight * kImageViewHeightToDialogRatio) : 0;
```
romoore: Tip: I'd prefer to see a fixed height for the image, consistent across devices.  If the view has additional vertical space, I think it's worth dedicating it to the label where possible.

romoore: This also doesn't seem to match the behavior of UITabBarController in my testing.  The image below shows a blown-up version with the vertical midpoint identified with the dotted line.

Screen Shot 2019-04-16 at 8 42 44 AM

romoore (changes requested):
Please add snapshot tests covering the following cases:

1.  Background colors
    1.   Black
    1.   White
    1.   Red
    1.   Green
    1.   Blue
1.  Various view bounds sizes
    1.   Small view bounds (120 x 120)
    1.   Medium view bounds (240 x 240)
    1.   Large view bounds (480 x 480)

components/BottomNavigation/src/private/MDCBottomNavigationLargeItemDialogView.h
nokeeo: Done.

components/BottomNavigation/src/private/MDCBottomNavigationSystemDialogView.m
nokeeo: Done.

components/BottomNavigation/src/private/MDCBottomNavigationSystemDialogView.m
nokeeo: I believe layout margins only propagate to a view's subviews if the subview has `preservesSuperviewLayoutMargins` set to `YES`.  I am hesitant to use this approach, because a change in the view hierarchy of `UIVisualEffectView`  may break the margins functionality.  Thoughts?

components/BottomNavigation/src/private/MDCBottomNavigationSystemDialogView.m
nokeeo: Done.  Just curious why using bounds + center is preferred over setting the frame?

components/BottomNavigation/src/private/MDCBottomNavigationLargeItemDialogView.m
nokeeo: Constant removed with latest update.

components/BottomNavigation/src/private/MDCBottomNavigationLargeItemDialogView.h
nokeeo: Done.

components/BottomNavigation/src/private/MDCBottomNavigationLargeItemDialogView.m
nokeeo: Done.  Made the `item` parameter `_Nonnull`, but the return value if nil by design.

components/BottomNavigation/src/private/MDCBottomNavigationLargeItemDialogView.m
nokeeo: If there is an image, there will be only one line and the font will be sized to fit up.

Because `MDCBottomNavigationBar` supports multi-line I changed this to:
```
_titleLabel.numberOfLines = 0;
```

components/BottomNavigation/src/private/MDCBottomNavigationLargeItemDialogView.m
nokeeo: Done.  Changed the image to a fixed height and positioning to a fixed offset from the top of the dialog.

nokeeo: Screen Shot 2019-04-16 at 17 30 53

nokeeo:
> Because this is a part of a larger effort, please add context to the PR description on the need for a separate `MDCBottomNavigationLargeItemDialogView` + `MDCBottomNavigationSystemDialogView` implementation. Is it possible to consolidate these two types into one?

@jverkoey In the context of the immediate larger effort there is no need for them to be separate and they can be consolidated.  However my intent when writing this code was that that each class should have a single responsibility, `MDCBottomNavigationSystemDialogView`, laying out/styling the container view, and `MDCBottomNavigationLargeItemDialogView`, laying out/styling the content of the container.  It encapsulates each concern and increase reusability of this code if the need arises.

Any thoughts on this @romoore ?

romoore:
@jverkoey @nokeeo Having read the code, I would say that having them separated is easy enough to read and I don't think they need to be combined. However, I can understand that since the separation isn't *necessary*, combining them can be seen as simpler. I'll let @jverkoey decide whether he means it to be a suggestion or a necessity before merging.

components/BottomNavigation/src/private/MDCBottomNavigationSystemDialogView.m
romoore: I think that's a fair reason. You may want to set `translatesAutoresizingMaskIntoConstraints = NO;` to make it clear you're expecting to do manual layout.

components/BottomNavigation/tests/snapshot/MDCBottomNavigationLargeItemDialogViewSnapshotTests.m
```diff
@@ -0,0 +1,215 @@
...
+- (void)setUp {
+  [super setUp];
+  self.dialogView = [[MDCBottomNavigationLargeItemDialogView alloc] initWithFrame:CGRectZero];
+
+  // Uncomment to rerecord golden images.
+  //  self.recordMode = YES;
+}
+
+- (void)tearDown {
+  _dialogView = nil;
```
romoore: Nit: Please use `self.dialogView = nil;` since it's a property. Yay, properties!

components/BottomNavigation/tests/snapshot/MDCBottomNavigationLargeItemDialogViewSnapshotTests.m
```diff
@@ -0,0 +1,215 @@
...
+/** The dialog view undergoing testing. */
+@property(nonatomic, nonnull) MDCBottomNavigationLargeItemDialogView *dialogView;
+
+@end
+
+@implementation MDCBottomNavigationSystemDialogViewSnapshotTests
+
+- (void)setUp {
+  [super setUp];
+  self.dialogView = [[MDCBottomNavigationLargeItemDialogView alloc] initWithFrame:CGRectZero];
```
romoore, line 37: Nit: Please place a blank line between the general method body and `super` calls in `tearDown` and `setUp`.

components/BottomNavigation/tests/snapshot/MDCBottomNavigationLargeItemDialogViewSnapshotTests.m
```diff
@@ -0,0 +1,215 @@
...
+  NSString *title = @"This is a really long tab bar item title";
+  UITabBarItem *item = [[UITabBarItem alloc] initWithTitle:title image:nil tag:0];
+  CGSize size = [[self class] mediumDialogSize];
+  UIColor *backgroundColor = UIColor.whiteColor;
+
+  // When/Then
+  [self verifyDialogWithItem:item size:size backgroundColor:backgroundColor];
+}
+
+#pragma mark - Helper Methods
```
romoore: Tip: Place these helper methods at the top of the file (with `setUp` and `tearDown`) to keep the test methods together in a single block.

romoore (approved):
LGTM. Thanks for the snapshot tests! I'm not 100% excited by the "small" layout, but it's good to capture how it gets laid-out in that case (which should be never).

components/BottomNavigation/src/private/MDCBottomNavigationSystemDialogView.m
romoore: Only because manipulating the frame is undefined if there's a non-identity transform applied to the view.  It's generally safer and since this is basically just a size change (and re-centering) I thought it was easy enough to use center + bounds.

https://developer.apple.com/documentation/uikit/uiview/1622621-frame - Note the severe warning UIKit provides to manipulating `frame` when transformed.

nokeeo: Thanks for the explanation!

components/BottomNavigation/tests/snapshot/MDCBottomNavigationLargeItemDialogViewSnapshotTests.m
nokeeo: Done.

components/BottomNavigation/tests/snapshot/MDCBottomNavigationLargeItemDialogViewSnapshotTests.m
nokeeo: Done.

romoore (approved):
LGTM!

romoore:
@jverkoey is out today. @nokeeo If you're OK with this, let me know and we can merge it in.

nokeeo:
@romoore I am OK with the current state!

PR #7166, opened 2019-04-17
Update to Swift 4.2

### LINKED ISSUE #6874: {Xcode} Update to Swift 4 or 5 or 5.1
Xcode 10.2b4 cannot compile the MaterialComponents project because our Swift code is 3.x.

----

### Unsupported Swift Version

This workspace has projects that contain source code developed with Swift 3.x. This version of Xcode does not support building or migrating Swift 3.x targets.

Use Xcode 10.1 to migrate the code to Swift 4.

----

Screen Shot 2019-03-13 at 2 46 05 PM

---
#### Internal data
- Associated internal bug: [b/128527189](http://b/128527189)
- Blocked by: https://github.com/material-components/material-components-ios/issues/7167

andrewoverton:
This PR updates the Swift version to 4.2.

Partially resolves #6874.

### DIFF
--- a/components/TextFields/examples/TextFieldUnderlineExample.swift
+++ b/components/TextFields/examples/TextFieldUnderlineExample.swift
@@ -195,39 +195,22 @@ final class TextFieldUnderlineSwiftExample: UIViewController {
                                                   metrics: nil,
                                                   views: views)
 
-    #if swift(>=3.2)
-      if #available(iOS 11.0, *) {
-        constraints += [NSLayoutConstraint(item: name,
-                                           attribute: .top,
-                                           relatedBy: .equal,
-                                           toItem: scrollView.contentLayoutGuide,
-                                           attribute: .top,
-                                           multiplier: 1,
-                                           constant: 20),
-                        NSLayoutConstraint(item: message,
-                                           attribute: .bottom,
-                                           relatedBy: .equal,
-                                           toItem: scrollView.contentLayoutGuide,
-                                           attribute: .bottomMargin,
-                                           multiplier: 1,
-                                           constant: -20)]
-      } else {
-        constraints += [NSLayoutConstraint(item: name,
-                                           attribute: .top,
-                                           relatedBy: .equal,
-                                           toItem: scrollView,
-                                           attribute: .top,
-                                           multiplier: 1,
-                                           constant: 20),
-                        NSLayoutConstraint(item: message,
-                                           attribute: .bottom,
-                                           relatedBy: .equal,
-                                           toItem: scrollView,
-                                           attribute: .bottomMargin,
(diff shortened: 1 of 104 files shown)


components/TextFields/examples/TextFieldUnderlineExample.swift
```diff
@@ -151,15 +151,8 @@ final class TextFieldUnderlineSwiftExample: UIViewController {
...
-    #if swift(>=3.2)
-      message.text = """
-      This is where you could put a multi-line message like an email.
-
-      It can even handle new lines.
-      """
-    #else
-      message.text = "This is where you could put a multi-line message like an email. It can even handle new lines./n"
-    #endif
+    message.text = "This is where you could put a multi-line message like an email." + "\n\n" +
```
romoore, line 154: Why did this switch to not using `"""` syntax?

romoore (approved):
Looks like only one change shouldn't have been made. Otherwise LGTM.

romoore:
Looks like we need to migrate our Kokoro jobs to Xcode 10+ before this can land.

andrewoverton:
Yep hehe didn't think of that

romoore:
Opened b/130744704 to track that work.

andrewoverton:
Thanks!

components/TextFields/examples/TextFieldUnderlineExample.swift
andrewoverton: I got tripped up about how it should be formatted and changed it to this way because I felt more sure of it, but I can change it back if you'd like

romoore: I'd prefer to keep the original source and only make the changes required for 4.2 migration.

romoore:
It's sometimes amazing to me how incompatible Swift source code is between versions. Hopefully they calm down a bit after 5.

components/TextFields/examples/TextFieldUnderlineExample.swift
andrewoverton: Fair enough 👍

romoore:
Xcode 10.1 job passed. Looking good. Let's let the others finish to double-check.

romoore:
The existing 10.1.0 presubmit job took 49 minutes, so this may take a little while before it will pass.

romoore:
Let's ignore the ChipTextField for this PR and get the changes in.

components/TextFields/examples/TextFieldUnderlineExample.swift
romoore: Let's leave it. The Presubmits take too long. Sorry!

PR #7182, opened 2019-04-19
[AppBar] Graduate theming extension to ready

codeman7:
## Context
As part of our work to move away from themer objects and onto theming extension we need to graduate all of our components theming extensions to the MaterialComponents pod.

## The fix
This PR updates the AppBar theming extension to the MaterialComponents pod, as well as updating examples and unit test.


fluxminds (approved):

jverkoey (dismissed):
Please update the README as well: https://github.com/material-components/material-components-ios/tree/develop/components/AppBar#theming-extensions-beta

jverkoey (approved):

PR #7208, opened 2019-04-22
[Docs] Fix links within documenation

codeman7:
This change fixes the broken links in `MDCActionSheet` and `MDCAppBar` components. This also moves the _Themer_ documentation to an "unsupported" section.


jverkoey (dismissed):
These changes will be broken the next time the readme generator is ran; can we move these headers to an "Unsupported" section instead?

jverkoey (approved):

PR #7249, opened 2019-04-23
[BottomNavigation] Adds .largeContentSizeImage support

### LINKED ISSUE #7121: [BottomNavigation] Internal issue: b/130344545
This was filed as an internal issue. If you are a Googler, please visit [b/130344545](http://b/130344545) for more details.

---
#### Internal data
- Associated internal bug: [b/130344545](http://b/130344545)
- Blocked by: https://github.com/material-components/material-components-ios/issues/7122
- Blocked by: https://github.com/material-components/material-components-ios/issues/7123

nokeeo:
Adds a flag that is disabled by default that allows clients to enable largeContentSizeImage / dynamic type support to the bottom navigation bar controller.

Fixes #7121

### DIFF
--- a/components/BottomNavigation/src/MDCBottomNavigationBarController.h
+++ b/components/BottomNavigation/src/MDCBottomNavigationBarController.h
@@ -60,6 +60,14 @@
  */
 @property(nonatomic) NSUInteger selectedIndex;
 
+/**
+ * If true, when the user long presses a tab bar item and dynamic type is enabled a large version of
+ * the tab bar item will be shown in a transient dialog that appears centered within the window.
+ * The default of this property is @c NO but will be set to @c YES in the future and then removed
+ * entirely.
+ */
+@property(nonatomic, getter=isDynamicTypeSupportEnabled) BOOL dynamicTypeSupportEnabled;
+
 - (void)viewDidLoad NS_REQUIRES_SUPER;
 
 #pragma mark - MDCBottomNavigationBarDelegate
--- a/components/BottomNavigation/src/MDCBottomNavigationBarController.h
+++ b/components/BottomNavigation/src/MDCBottomNavigationBarController.h
@@ -61,15 +61,11 @@
 @property(nonatomic) NSUInteger selectedIndex;
 
 /**
- * If true, when the user long presses a tab bar item and dynamic type is enabled a large version of
- * the tab bar item will be shown in a transient dialog that appears centered within the window.
- * The default of this property is @c NO but will be set to @c YES in the future and then removed
- * entirely.
- * A view controller may set the @c largeContentSizeImage on its @c tabBarItem. If the tab bar item
- * does not provide a large content size image, its @c image may be scaled.
- * @see UIBarItem#largeContentSizeImage
+ * If enabled and the user has selected a @c UIContentSizeCategory of @c .AccessibilityMedium or
+ * larger, then when the user long-pressed on a tab item a view will be presented with a larger
+ * version of the image and title.
  */
-@property(nonatomic, getter=isDynamicTypeSupportEnabled) BOOL dynamicTypeSupportEnabled;
+@property(nonatomic, getter=isLongPressPopUpViewEnabled) BOOL longPressPopUpViewEnabled;
 
 - (void)viewDidLoad NS_REQUIRES_SUPER;
 
--- a/components/BottomNavigation/src/MDCBottomNavigationBarController.h
+++ b/components/BottomNavigation/src/MDCBottomNavigationBarController.h
@@ -65,6 +65,9 @@
  * the tab bar item will be shown in a transient dialog that appears centered within the window.
  * The default of this property is @c NO but will be set to @c YES in the future and then removed
  * entirely.
+ * A view controller may set the @c largeContentSizeImage on its @c tabBarItem. If the tab bar item
+ * does not provide a large content size image, its @c image may be scaled.
+ * @see UIBarItem#largeContentSizeImage
  */
 @property(nonatomic, getter=isDynamicTypeSupportEnabled) BOOL dynamicTypeSupportEnabled;
 
--- a/components/BottomNavigation/src/MDCBottomNavigationBarController.m
+++ b/components/BottomNavigation/src/MDCBottomNavigationBarController.m
@@ -33,7 +33,7 @@ static const NSUInteger kLongPressNumberOfTouchesRequired = 1;
  * dismissing).
  */
 static CGAffineTransform MDCLargeItemViewAnimationTransitionTransform() {
-  return CGAffineTransformScale(CGAffineTransformIdentity, 0.97, 0.97);
+  return CGAffineTransformScale(CGAffineTransformIdentity, (CGFloat)0.97, (CGFloat)0.97);
 }
 
 @interface MDCBottomNavigationBarController ()
@@ -88,7 +88,7 @@ static CGAffineTransform MDCLargeItemViewAnimationTransitionTransform() {
   [self.view addSubview:self.navigationBar];
   [self loadConstraints];
 
-  if (self.dynamicTypeSupportEnabled) {
+  if ([self isDynamicTypeSupportEnabled] && ![self isNavigationBarLongPressRecognizerRegistered]) {
     [self.navigationBar addGestureRecognizer:self.navigationBarLongPressRecognizer];
   }
 }
@@ -157,11 +157,10 @@ static CGAffineTransform MDCLargeItemViewAnimationTransitionTransform() {
 - (void)setDynamicTypeSupportEnabled:(BOOL)dynamicTypeSupportEnabled {
   _dynamicTypeSupportEnabled = dynamicTypeSupportEnabled;
 
-  BOOL isNavigationBarLongPressEnabled =
-      [self.navigationBar.gestureRecognizers containsObject:self.navigationBarLongPressRecognizer];
-  if (dynamicTypeSupportEnabled && !isNavigationBarLongPressEnabled) {
+  BOOL isNavigationBarLongPressRegistered = [self isNavigationBarLongPressRecognizerRegistered];
+  if (dynamicTypeSupportEnabled && !isNavigationBarLongPressRegistered) {
     [self.navigationBar addGestureRecognizer:self.navigationBarLongPressRecognizer];
-  } else if (!dynamicTypeSupportEnabled && isNavigationBarLongPressEnabled) {
+  } else if (!dynamicTypeSupportEnabled && isNavigationBarLongPressRegistered) {
     [self.navigationBar removeGestureRecognizer:self.navigationBarLongPressRecognizer];
   }
 }
@@ -311,7 +310,7 @@ static CGAffineTransform MDCLargeItemViewAnimationTransitionTransform() {
  * @param point CGPoint The point within @c navigationBar coordinate space.
  */
 - (void)handleNavigationBarLongPressEndedForPoint:(CGPoint)point {
-  if (!self.largeItemDialog || self.isDismissingLargeItemDialog) {
+  if (!self.largeItemDialog || [self isDismissingLargeItemDialog]) {
     return;
   }
 
@@ -591,6 +590,12 @@ static CGAffineTransform MDCLargeItemViewAnimationTransitionTransform() {
         [self.largeItemDialog removeFromSuperview];
         self.dismissingLargeItemView = NO;
       }];
+}
+
+/** Returns if the long press gesture recognizer has been added to the navigation bar. */
+- (BOOL)isNavigationBarLongPressRecognizerRegistered {
+  return
+      [self.navigationBar.gestureRecognizers containsObject:self.navigationBarLongPressRecognizer];
 }
 
 @end
--- a/components/BottomNavigation/src/MDCBottomNavigationBarController.m
+++ b/components/BottomNavigation/src/MDCBottomNavigationBarController.m
@@ -93,13 +93,6 @@ static CGAffineTransform MDCLargeItemViewAnimationTransitionTransform() {
   }
 }
 
-- (void)viewDidLayoutSubviews {
-  [super viewDidLayoutSubviews];
-
-  self.navigationBarLongPressRecognizer.allowableMovement =
-      CGRectGetWidth(self.navigationBar.bounds);
-}
-
 - (void)setSelectedViewController:(nullable UIViewController *)selectedViewController {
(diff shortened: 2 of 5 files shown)


romoore:
Looks like a syntax error in the import statement:

```
components/BottomNavigation/src/MDCBottomNavigationBarController.m:19:9: fatal error: './private/MDCBottomNavigationBar+Private.h' file not found
#import "./private/MDCBottomNavigationBar+Private.h"
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

romoore:
Looks like this can be fixed by removing the first `./` from the imports and exposing the private headers to the `:BottomNavigationBeta` target.

```
diff --git a/components/BottomNavigation/BUILD b/components/BottomNavigation/BUILD
index 94f9ae925f..f75a2d2fb1 100644
--- a/components/BottomNavigation/BUILD
+++ b/components/BottomNavigation/BUILD
@@ -69,6 +69,14 @@ mdc_objc_library(
     ],
 )
 
+# Should only be used by test targets or :BottomNavigationBeta until
+# that can be rolled-up into the main target.
+mdc_objc_library(
+    name = "privateHeaders",
+)
+
 mdc_objc_library(
     name = "BottomNavigationBeta",
     srcs = ["src/MDCBottomNavigationBarController.m"],
@@ -80,7 +88,10 @@ mdc_objc_library(
     includes = ["src"],
     sdk_frameworks = ["CoreGraphics"],
     visibility = ["//:__subpackages__"],
-    deps = [":BottomNavigation"],
+    deps = [
+        ":BottomNavigation",
+        ":privateHeaders",
+    ],
 )
 
 filegroup(
@@ -150,8 +161,8 @@ swift_library(
         "4.2",
     ],
     deps = [
-        ":BottomNavigationBeta",
         ":BottomNavigation",
+        ":BottomNavigationBeta",
         ":ColorThemer",
         ":ObjcExamples",
         ":TypographyThemer",
@@ -200,10 +211,10 @@ mdc_objc_library(
 
 ios_unit_test(
     name = "unit_tests_beta",
-    deps = [":unit_test_beta_sources"],
-    minimum_os_version = "9.0",
     flaky = 1,
+    minimum_os_version = "9.0",
     test_host = "//components/private/Snapshot/TestHost",
+    deps = [":unit_test_beta_sources"],
 )
 
 mdc_unit_test_suite(
diff --git a/components/BottomNavigation/src/MDCBottomNavigationBarController.m b/components/BottomNavigation/src/MDCBottomNavigationBarController.m
index 7257a93d6e..bc32705cd7 100644
--- a/components/BottomNavigation/src/MDCBottomNavigationBarController.m
+++ b/components/BottomNavigation/src/MDCBottomNavigationBarController.m
@@ -16,8 +16,8 @@
 
 #import <CoreGraphics/CoreGraphics.h>
 
-#import "./private/MDCBottomNavigationBar+Private.h"
-#import "./private/MDCBottomNavigationLargeItemDialogView.h"
+#import "private/MDCBottomNavigationBar+Private.h"
+#import "private/MDCBottomNavigationLargeItemDialogView.h"
 
 // A context for Key Value Observing
 static void *const kObservationContext = (void *)&kObservationContext;
```

components/BottomNavigation/src/MDCBottomNavigationBarController.h
```diff
@@ -60,6 +60,17 @@
...
+/**
+ * If true, when the user long presses a tab bar item and dynamic type is enabled a large version of
+ * the tab bar item will be shown in a transient dialog that appears centered within the window.
+ * The default of this property is @c NO but will be set to @c YES in the future and then removed
+ * entirely.
+ * A view controller may set the @c largeContentSizeImage on its @c tabBarItem. If the tab bar item
+ * does not provide a large content size image, its @c image may be scaled.
+ * @see UIBarItem#largeContentSizeImage
+ */
+@property(nonatomic, getter=isDynamicTypeSupportEnabled) BOOL dynamicTypeSupportEnabled;
```
romoore: Nit/naming: I don't want to bike shed, but this isn't support for Dynamic Type.  Please consider something more like `longPressPopUpViewEnabled`.

components/BottomNavigation/src/MDCBottomNavigationBarController.h
```diff
@@ -60,6 +60,17 @@
  */
 @property(nonatomic) NSUInteger selectedIndex;
 
+/**
+ * If true, when the user long presses a tab bar item and dynamic type is enabled a large version of
```
romoore: Tip: Trim this down to the essentials. 

Suggestion:
> If enabled and the user has selected a @c UIContentSizeCategory of @c .AccessibilityMedium or larger, then when the user long-pressed on a tab item a view will be presented with a larger version of the image and title.

components/BottomNavigation/src/MDCBottomNavigationBarController.h
romoore: Let's default this to `YES` for now - it's a Beta component and this property is meant as a quick-fix to turn it off if it causes some severe problem.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
```diff
@@ -14,14 +14,43 @@
...
+static const NSTimeInterval kLargeItemViewAnimationDuration = 0.1;
+static const NSTimeInterval kLongPressMinimumPressDuration = 0.2;
+static const NSUInteger kLongPressNumberOfTouchesRequired = 1;
+
+/**
+ * The transform of the large item view when it is in a transitional state (appearing or
+ * dismissing).
+ */
+static CGAffineTransform MDCLargeItemViewAnimationTransitionTransform() {
+  return CGAffineTransformScale(CGAffineTransformIdentity, (CGFloat)0.97, (CGFloat)0.97);
```
romoore, line 36: Tip: Can we drop the transform in this PR? Does it look completely awful if it's just presented immediately with a transition (or perhaps just a quick fade)?  Transforms are notoriously difficult to get working "just right" and I'd prefer to avoid them where possible.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
```diff
@@ -14,14 +14,43 @@
...
 @property(nonatomic, strong) UIView *content;
 
+/** The gesture recognizer for detecting long presses on tab bar items. */
+@property(nonatomic, strong, nonnull)
+    UILongPressGestureRecognizer *navigationBarLongPressRecognizer;
+
+/** The dialog view to display a large item view. */
+@property(nonatomic, strong, nullable) MDCBottomNavigationLargeItemDialogView *largeItemDialog;
+
+/** Indicates if the large item view is in the process of dismissing. */
```
romoore: Tip: Please expand the comment to describe why this flag is necessary.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
```diff
@@ -56,6 +87,17 @@ - (void)viewDidLoad {
   [self.view addSubview:self.content];
   [self.view addSubview:self.navigationBar];
   [self loadConstraints];
+
+  if ([self isDynamicTypeSupportEnabled] && ![self isNavigationBarLongPressRecognizerRegistered]) {
```
romoore: Please refer to properties with "dot" notation.  

```objc
if (self.isDynamicTypeSupportEnabled && [self isNavigationBarLongPressRecognizerRegistered]) {
```

components/BottomNavigation/src/MDCBottomNavigationBarController.m
```diff
@@ -56,6 +87,17 @@ - (void)viewDidLoad {
...
+    [self.navigationBar addGestureRecognizer:self.navigationBarLongPressRecognizer];
+  }
+}
+
+- (void)viewDidLayoutSubviews {
+  [super viewDidLayoutSubviews];
+
+  self.navigationBarLongPressRecognizer.allowableMovement =
+      CGRectGetWidth(self.navigationBar.bounds);
```
romoore: Is this necessary? It seems that this is just the "fudge factor" when detecting whether the long-press recognizer should trigger *before* the `minimumPressDuration` is met.

I'm also wary of users being able to slide vertically from the Bottom Navigation bar so far (the full width) before the recognizer cancels recognition.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
```diff
@@ -112,6 +154,17 @@ - (void)setViewControllers:(NSArray<UIViewController *> *)viewControllers {
   self.selectedViewController = viewControllersCopy.firstObject;
 }
 
+- (void)setDynamicTypeSupportEnabled:(BOOL)dynamicTypeSupportEnabled {
+  _dynamicTypeSupportEnabled = dynamicTypeSupportEnabled;
+
+  BOOL isNavigationBarLongPressRegistered = [self isNavigationBarLongPressRecognizerRegistered];
```
romoore: Please consider declaring this method as a readonly property. `self.isNavigationBarLongPressRecognizerRegistered` looks so nice. :)

components/BottomNavigation/src/MDCBottomNavigationBarController.m
```diff
@@ -195,6 +260,64 @@ - (void)didUpdateNavigationBarItemsWithNewValue:(NSArray *)items {
...
+  }
+}
+
+/**
+ * Handles when the navigation bar long press gesture recognizer gesture has been initiated or the
+ * touch point was updated.
+ * @param point CGPoint The point within @c navigationBar coordinate space.
+ */
+- (void)handleNavigationBarLongPressUpdatedForPoint:(CGPoint)point {
+  UIFont *preferredFont = [UIFont preferredFontForTextStyle:UIFontTextStyleBody];
```
romoore: Based on the design document I would expect to check the `UIContentSizeCategory` value.  Is there a reason why it wouldn't be possible and we need to check a font? 

For this controller, we should probably read it using the following logic:

```objc
- (BOOL)contentSizeCategoryIsAccessibilityCategory {
  UIContentSizeCategory sizeCategory = UIContentSizeCategoryLarge;
    sizeCategory = self.traitCollection.preferredContentSizeCategory;
  } else {
    sizeCategory = UIApplication.mdc_safeSharedApplication.preferredContentSizeCategory;
  }
    return UIContentSizeCategoryIsAccessibilityCategory(sizeCategory);
  }
  if ([sizeCategory isEqualToString:UIContentSizeCategoryAccessibilityMedium] ||
      [sizeCategory isEqualToString:UIContentSizeCategoryAccessibilityLarge] ||
      [sizeCategory isEqualToString:UIContentSizeCategoryAccessibilityExtraLarge] ||
      [sizeCategory isEqualToString:UIContentSizeCategoryAccessibilityExtraExtraLarge] ||
      [sizeCategory isEqualToString:UIContentSizeCategoryAccessibilityExtraExtraExtraLarge]) {
    return YES;
  }
  return NO;
}
```

romoore (changes requested):
The approach looks good over all, just some smaller things to address.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
nokeeo: Done.

components/BottomNavigation/src/MDCBottomNavigationBarController.h
nokeeo: Done.

components/BottomNavigation/src/MDCBottomNavigationBarController.h
nokeeo: Done.

components/BottomNavigation/src/MDCBottomNavigationBarController.h
nokeeo: Done.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
nokeeo: Done.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
nokeeo: I am following the [Google objective-c style](http://google.github.io/styleguide/objcguide.html#objective-c-method-names) guide here.  If I have misinterpreted these rules or if this project uses a different style, please let me know.

```
BOOL isGood = [object isGlorious];  // GOOD.
BOOL isGood = object.glorious;      // GOOD.
BOOL isGood = object.isGlorious;    // AVOID.
```

components/BottomNavigation/src/MDCBottomNavigationBarController.m
nokeeo: I have attached screen recordings of the three approaches.  The transform is nice but certainly not necessary. 

## Hard cut:
bn_hardCut

## Opacity Only:
bn_opacity

## Opacity and Transform:
bn_transform

romoore: Agreed the transform looks nice! If you'd like to keep it, please just double-check that we never read/write `frame` directly on the accessibility view.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
romoore: Agreed for method names. Since `isDynamicTypeSupportEnabled` is declared as a property, it should be accessed like a property.

```objc
@property(nonatomic, getter=isDynamicTypeSupportEnabled) BOOL dynamicTypeSupportEnabled;
```

Every property is implemented with a method, I believe the style guide refers to more traditional methods (that are not declared as properties).

components/BottomNavigation/src/MDCBottomNavigationBarController.m
nokeeo: Confirmed.  We never read/write `frame`.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
nokeeo: Done.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
nokeeo: Done.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
nokeeo: Done.

romoore (approved):

PR #7255, opened 2019-04-23
[TextFields] Fix character count overflow.

romoore:
Fixes possible overflow when reading out the character count remaining value.

Found while testing #7157


jverkoey (approved):
Is this technically an underflow?

romoore:
@jverkoey Yep, I think it is. It's the reason why [Google's Objective-C Style Guide](http://google.github.io/styleguide/objcguide.html#unsigned-integers) explicitly discourages the use of unsigned values whenever possible.

PR #7256, opened 2019-04-23
[TextFields] Correct UIAccessibility values and behavior.

### LINKED ISSUE #7157: [TextFields] MDCTextField's accessibility is broken for error messages

romoore:
## `accessibilityLabel` is used for all non-user-input values in `MDCTextField`.
Previously, this text was all presented in the `accessibilityValue` of the text field, which resulted in weird effects, including:
*  Changes in pitch to indicate that the value had changed (up to indicate additions, down for deletions).
*  The entire error message being read out as if it were the value.
*  The placeholder and labels being read out as if they were the value.

With this change, anything not input by the user (or programmatically set as if the user had done so) is now moved to `accessibilityLabel`.  This more closely matches the intent of the property.  Specifically, the `accessibilityLabel` property will return the contents of the following (if present):
1.   The placeholder text.
2.  The leading underline label.
3.  The trailing underline label. **This was previously excluded from the `accessibilityValue`, though it's not clear why.**

The `accessibilityLabel` will also introduce commas (slight pauses) between each of these values when concatenating them into a single string.  This alleviates some of the difficulty in mentally separating each of the values when hearing it read by VoiceOver.

## `UIAccessibilityAnnouncementNotification` is now used to announce the error to the user in `MDCTextInputControllerBase`.
Previously, a `UIAccessibilityLayoutChangedNotification` was used and pointed to the `leadingUnderlineLabel`.  However, this could easily result in the user becoming "trapped" in the label because it was an accessibility element outside of a "container" element. This is incorrect behavior - any UIView that returns `isAccessibilityElement = YES` should summarize its subviews.

Because announcements can be interrupted, and because announcements are already used for things like character counters, *error messages can interrupt other announcements*. This is intentional, however, since errors should be corrected immediately and therefore it's reasonable to cut short any other announcements generated by the text field.  A slight delay (not currently customizable) is introduced to allow an opportunity for the text field to read-out any content that was added or removed just before the error is announced.

Fixes #7157


romoore:
## IMPORTANT: IF YOU REVIEW

Please verify the new behavior **on a device using VoiceOver**.  Please do not grant LGTM/Approval until you have performed manual testing. I recommend using the Kitchen Sink or "Outlined Controller" examples.  Specifically pay attention to:

*  Error states - be sure the accessibility values are read-out.
*  Character counters and how they're read-out in different text field states.
*  Helper text and how it's read-out in different states.

codeman7 (dismissed):
Still want to test this on a device when I get in tomorrow morning.

components/TextFields/src/MDCTextField.m
```diff
@@ -792,20 +792,39 @@ - (void)mdc_setAdjustsFontForContentSizeCategory:(BOOL)adjusts {
   [_fundament mdc_setAdjustsFontForContentSizeCategory:adjusts];
 }
 
-- (NSString *)accessibilityValue {
+/*
+ Returns a combination of the following:
+ -  The superclass `accessibilityLabel` value
```
codeman7, line 797: _Question:_ What was the motivation behind using \`accessibilityLabel\` instead of @c accessibilityLabel?

romoore: In non-doc comments? I think it's just easier to read since we're mostly familiar with Markdown.

romoore:
@codeman7 Please feel free to merge this when you finish testing. I'm taking the label off, however, since I don't want a Good Samaritan passing by to merge this by accident.

components/TextFields/src/MDCTextInputControllerBase.m
```diff
@@ -41,6 +41,7 @@
     MDCTextInputControllerBaseDefaultFloatingPlaceholderDownAnimationDuration = (CGFloat)0.266666;
 static const NSTimeInterval
     MDCTextInputControllerBaseDefaultFloatingPlaceholderUpAnimationDuration = (CGFloat)0.3;
+static const NSTimeInterval kDefaultErrorAnnouncementDelay = (CGFloat)0.500;
```
codeman7, line 44: _Question:_ Where is this value coming from?

romoore: I just made it up. It seems sufficient to vocalize most text input. It slightly mitigates the error announcement interruption effect.

codeman7 (approved):

Issue #7318, opened 2019-05-01
Our podspecs don't specify values for the `swift_version ` attribute


andrewoverton:
Our podspecs don't specify values for the `swift_version ` attribute.

## Reproduction steps

Run `pod lib lint MaterialComponents.podspec --skip-tests`

## Expected behavior

There shouldn't be any warnings related to Swift version.

## Actual behavior

There is the following warning related to Swift version:

```
 -> MaterialComponents (84.0.0)
    - WARN  | [MaterialComponents/ActionSheet, MaterialComponents/ActivityIndicator, 
MaterialComponents/AppBar, and more...] swift: The validator used Swift 3.2 by default
because no Swift version was specified. To specify a Swift version during validation,
add the swift_version attribute in your podspec. Note that usage of the --swift-version
parameter or a .swift-version file is now deprecated.

[!] MaterialComponents did not pass validation, due to 1 warning (but you can use
--allow-warnings to ignore it).
You can use the --no-clean option to inspect any issue.
```

Please revert https://github.com/material-components/material-components-ios/pull/7336 when this is fixed

---
#### Internal data
- Associated internal bug: [b/131768882](http://b/131768882)

randallli:
This is forcing our publish to cocoapods command to allow warnings

> pod trunk push MaterialComponents.podspec --skip-tests
> Updating spec repo `master`
> 
> CocoaPods 1.7.0.beta.3 is available.
> To update use: `sudo gem install cocoapods --pre`
> [!] This is a test version we'd love you to try.
> 
> For more information, see https://blog.cocoapods.org and the CHANGELOG for this version at https://github.com/CocoaPods/CocoaPods/releases/tag/1.7.0.beta.3
> 
> Validating podspec
> MaterialComponents (84.0.0)
>     - WARN  | [MaterialComponents/ActionSheet, MaterialComponents/ActivityIndicator, MaterialComponents/AppBar, and more...] swift: The validator used Swift 3.2 by default because no Swift version was specified. To specify a Swift version during validation, add the `swift_version` attribute in your podspec. Note that usage of the `--swift-version` parameter or a `.swift-version` file is now deprecated.
> 
> [!] The spec did not pass validation, due to 1 warning (but you can use `--allow-warnings` to ignore it).

PR #7336, opened 2019-05-02
[MDC release] Publish cocoapods allows warnings

randallli:
Because of a warning in the linter about no swift version specified (https://github.com/material-components/material-components-ios/issues/7318) we need to allow warnings in order to publish the podspec


jverkoey (approved):

andrewoverton:
What if we remove `--allow-warnings` and make a note saying that if it fails we should add `--allow-warnings` and file a bug to address the warning?

Issue #7388, opened 2019-05-14
[Buttons] Internal issue: b/132686929


material-mirror:
This was filed as an internal issue. If you are a Googler, please visit [b/132686929](http://b/132686929) for more details.

---
#### Internal data
- Associated internal bug: [b/132686929](http://b/132686929)
- Blocked by: https://github.com/material-components/material-components-ios/issues/7392
- Blocked by: https://github.com/material-components/material-components-ios/issues/7393

material-mirror:
The internal issue [b/132686929](http://b/132686929) is now closed. This issue is being closed as a result.

PR #7391, opened 2019-05-14
[Buttons] Remove dependency on TypographyThemer in theming extension. 

### LINKED ISSUE #7393: [Buttons] Internal issue: b/132704158
This was filed as an internal issue. If you are a Googler, please visit [b/132704158](http://b/132704158) for more details.

---
#### Internal data
- Associated internal bug: [b/132704158](http://b/132704158)

codeman7:
In order for us to add dynamic type to our theming extension we need to be setting the font directly in the theming extension. We cannot pass the scheme to the themer object because that would require us updating the themer object that we plan to deprecate. This removes that dependency to allow us to set the font directly in the theming extension and gets the Material iOS team off the themer object. This is in greater service to #7388.

Closes #7393

### DIFF
--- a/components/Buttons/src/Theming/MDCButton+MaterialTheming.m
+++ b/components/Buttons/src/Theming/MDCButton+MaterialTheming.m
@@ -157,7 +157,7 @@
 
 - (void)resetTitleForAllStates {
   NSUInteger maximumStateValue = UIControlStateNormal | UIControlStateSelected |
-  UIControlStateHighlighted | UIControlStateDisabled;
+                                 UIControlStateHighlighted | UIControlStateDisabled;
   for (NSUInteger state = 0; state <= maximumStateValue; ++state) {
     [self setTitleFont:nil forState:state];
   }
--- a/components/Buttons/src/Theming/MDCButton+MaterialTheming.m
+++ b/components/Buttons/src/Theming/MDCButton+MaterialTheming.m
@@ -55,7 +55,7 @@
 }
 
 - (void)applyContainedThemeWithTypographyScheme:(id<MDCTypographyScheming>)typographyScheme {
-  [self resetTitleForAllStates];
+  [self resetTitleFontForAllStates];
   [self setTitleFont:typographyScheme.button forState:UIControlStateNormal];
 }
 
@@ -100,7 +100,7 @@
 }
 
 - (void)applyOutlinedThemeWithTypographyScheme:(id<MDCTypographyScheming>)typographyScheme {
-  [self resetTitleForAllStates];
+  [self resetTitleFontForAllStates];
   [self setTitleFont:typographyScheme.button forState:UIControlStateNormal];
 }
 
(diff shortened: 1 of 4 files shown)


components/Buttons/src/Theming/MDCFloatingButton+MaterialTheming.m
```diff
@@ -55,7 +54,18 @@ - (void)applySecondaryThemeWithShapeScheme:(id<MDCShapeScheming>)scheme {
...
 
 - (void)applySecondaryThemeWithTypographyScheme:(id<MDCTypographyScheming>)scheme {
-  [MDCButtonTypographyThemer applyTypographyScheme:scheme toButton:self];
+  [self resetTitleForAllStates];
+  [self setTitleFont:scheme.button forState:UIControlStateNormal];
+}
+
+#pragma mark - General helpers
+
+- (void)resetTitleForAllStates {
```
andrewoverton: Maybe this should be named `resetTitleFontForAllStates`

components/Buttons/src/Theming/MDCButton+MaterialTheming.m
```diff
@@ -144,11 +145,22 @@ - (void)applyTextThemeWithColorScheme:(id<MDCColorScheming>)colorScheme {
...
+  [self setTitleFont:typographyScheme.button forState:UIControlStateNormal];
 }
 
 - (void)applyTextThemeWithShapeScheme:(id<MDCShapeScheming>)shapeScheme {
   [MDCButtonShapeThemer applyShapeScheme:shapeScheme toButton:self];
 }
 
+#pragma mark - General helpers
+
+- (void)resetTitleForAllStates {
```
andrewoverton: Maybe this should be called `resetTitleFontForAllStates` too. It's kind of a bummer we have these two methods doing the same thing :/

codeman7: Yeah that's unfortunate, I removed this method as it's only used once and I don't know how it would play with it's `super` having the same method but in an extension.

components/Buttons/src/Theming/MDCFloatingButton+MaterialTheming.m
codeman7: Yeah that's a much better name, I updated it in the PR.

components/Buttons/src/Theming/MDCFloatingButton+MaterialTheming.m
```diff
@@ -55,7 +54,12 @@ - (void)applySecondaryThemeWithShapeScheme:(id<MDCShapeScheming>)scheme {
 }
 
 - (void)applySecondaryThemeWithTypographyScheme:(id<MDCTypographyScheming>)scheme {
-  [MDCButtonTypographyThemer applyTypographyScheme:scheme toButton:self];
+  NSUInteger maximumStateValue = UIControlStateNormal | UIControlStateSelected |
```
andrewoverton: I'm seeing this logic everywhere! It would be nice to have it in one place

andrewoverton (approved):

PR #7402, opened 2019-05-16
[Buttons] Update titleFont:forState: API

codeman7:
## Motivation
As outlined in #5919 many of our `forState:` APIs don't match what `UIButton` is doing. This addresses that problem for our `titleFont` API. This updates the fallback behavior. Additionally this will help facilitate #7388.

## Test
In other test similar to this one we update a `UIButton` with a `UIColor` to make sure the colors match for a given state in given conditions, since `UIButton` doesn't have any `forState` APIs that set a font we can't test this as easily. Also this follows the same pattern outlined in the other PRs to address this issue.

### DIFF
--- a/components/Buttons/src/MDCButton.m
+++ b/components/Buttons/src/MDCButton.m
@@ -729,13 +729,27 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
 #pragma mark - Title Font
 
 - (nullable UIFont *)titleFontForState:(UIControlState)state {
-  return _fonts[@(state)];
+  // If the `.highlighted` flag is set, turn off the `.disabled` flag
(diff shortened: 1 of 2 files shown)


romoore:
@codeman7  I think this is not a breaking change, but does correct a bug in the APIs. To be clear, the font always rendered correctly, but the API didn't return the rendered value. What do you think?

romoore (approved):

codeman7:
@romoore I thought rendering didn't impact if we said a change was breaking but API did that's why I marked it breaking. I'll leave it up to you, as you are in charge of the release or in charge of helping with the release this week.

codeman7:
@romoore and I spoke offline and came to the conclusion this change is not breaking.

PR #7421, opened 2019-05-20
{WIP}[Buttons] Set dynamic type based on scheme in theming extension

codeman7:
In #6873 we added a flag within the `MDCTypographyScheme` to enable dynamic type via the _theming extension_. This flag was added to the theming extension can then set the same flag on the component when the component is _themed_. This change sets the value on the component within the theming extension.

Related to #7388


romoore (changes requested):
Theming extensions should be reading that flag, but not modifying the component. It's intended as a marker for setting scaled font sizes.

PR #7423, opened 2019-05-20
[Chips] Use umbrella header

codeman7:
When using #7394 as a template for other components I noticed chips wasn't using the umbrella header. This change updates the import statement to be the umbrella header instead of the direct header.

Related to #7394


fluxminds (approved):
Thanks!

Issue #7386, opened 2019-05-14
[Dialogs] Internal issue: b/132688428


material-mirror:
This was filed as an internal issue. If you are a Googler, please visit [b/132688428](http://b/132688428) for more details.

---
#### Internal data
- Associated internal bug: [b/132688428](http://b/132688428)

material-mirror:
The internal issue [b/133215685](http://b/133215685) is now closed. This issue is being closed as a result.

codeman7:
The bot closed this on accident.

material-mirror:
The internal issue [b/132688428](http://b/132688428) is now closed. This issue is being closed as a result.

PR #7446, opened 2019-05-22
[Dialogs] Add flag for legacy behavior for dynamic type

codeman7:
## Motivation
In order to closely match UIKit behavior we need to have a flag for when a component wants to update some fonts for dynamic type but only those with scaling curves. Currently this is not possible to do because if dynamic type is on we automatically scale everything.

## Detailed changes
Add a runtime flag that can turn on scaling fonts no matter what or scales only if scalingCurves are attached to the font.

Related to #7386

### DIFF
--- a/components/Dialogs/src/MDCAlertController.h
+++ b/components/Dialogs/src/MDCAlertController.h
@@ -122,6 +122,18 @@
 @property(nonatomic, readwrite, setter=mdc_setAdjustsFontForContentSizeCategory:)
     BOOL mdc_adjustsFontForContentSizeCategory;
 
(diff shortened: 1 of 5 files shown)


jverkoey (approved):
Will these code paths need to be updated again once we integrate the new scaling curve behavior?

codeman7:
@jverkoey Once we add the scaling curve behavior all of this logic will move to an `else` statement.
```objc
if (font.scalingCurve) {
// Code to be added in a follow up PR
} else {
// Existing logic plus this PR.
}
```

PR #7458, opened 2019-05-22
[List] Snapshot tests for upcoming Scalable Font behavior.

fluxminds:
Adding tests for the upcoming Scalable Fonts Dynamic Type work. This PR focuses on the use case where `mdc_adjustsFontForContentSizeCategory = YES;` and when Scalable Fonts are attached. 

A separate PR will handle the "fall-back" case (when a non-Scalable Font is used) and potentially a third PR for the current (legacy) behavior.

Pre-work of https://github.com/material-components/material-components-ios/issues/7385.


components/List/tests/snapshot/MDCSelfSizingStereoCellSnapshotTests.m
```diff
@@ -42,11 +70,41 @@ - (void)tearDown {
...
+                 contentSizeCategory:(UIContentSizeCategory)sizeCategory
+                              insets:(UIEdgeInsets)insets {
+  MDCSelfSizingStereoCellSnapshotTestsContentSizeCategoryOverrideWindow *backgroundWindow =
+      [[MDCSelfSizingStereoCellSnapshotTestsContentSizeCategoryOverrideWindow alloc]
+          initWithFrame:CGRectMake(0, 0, CGRectGetWidth(view.bounds) + insets.left + insets.right,
+                                   CGRectGetHeight(view.bounds) + insets.top + insets.bottom)];
+  backgroundWindow.contentSizeCategoryOverride = sizeCategory;
+  backgroundWindow.backgroundColor = [UIColor colorWithWhite:(CGFloat)0.8 alpha:1];
+  [backgroundWindow addSubview:view];
+  [backgroundWindow makeKeyAndVisible];
```
romoore: This usually crashes in a headless environment - is it passing for you locally using Bazel?

fluxminds: Yes. It passed the test using Bazel.

fluxminds: I can change it to `.hidden = NO` if you feel more comfortable with that.

romoore: I don't mind it being not hidden, I'm surprised that we need to make it the `keyWindow` for the test to pass.  Feel free to keep it as-is, but a comment here or inline comment about its necessity would be welcome.

components/List/tests/snapshot/MDCSelfSizingStereoCellSnapshotTests.m
```diff
@@ -78,4 +136,30 @@ - (void)testCellWithTitleAndDetailAndImage {
   [self generateSnapshotAndVerifyForView:self.cell];
 }
 
+- (void)testCellWithDynamicTypeForContentSizeCategoryExtraSmallEnabledForTitleAndDetail {
+  if (@available(iOS 10.0, *)) {
+    // When
+    self.cell.titleLabel.text = @"Title";
+    self.cell.detailLabel.text = @"Detail";
+    self.cell.mdc_adjustsFontForContentSizeCategory = YES;
```
romoore, line 154: It looks like you forgot to set a font with attached scaling curves here. Please be sure to add them. If you are intended to test the current/legacy behavior, then I suspect you'll need to change the font to something that's not the default, then verify that the `adjustsFontForContentSizeCategory` replaces it with a Material Typography font.

romoore (dismissed):
Please add/update the tests to either test the new behavior (when it lands) or the legacy behavior. 

-  [x] With scaling curves (new behavior) and multiple contentSizeCategory values
-  [ ] Without scaling curves (legacy + fallback behavior) with a custom font that gets replaced.

components/List/tests/snapshot/MDCSelfSizingStereoCellSnapshotTests.m
fluxminds: Addressed in 7ec0b163eadea7f4f28a86396413e7b4f094b96d.

components/List/tests/snapshot/MDCSelfSizingStereoCellSnapshotTests.m
fluxminds: Addressed in 2a572487eb4ec3015fa52debd41cd3387e10d968. Added fonts with attached scaling curves.

fluxminds:
Two tests were intended. More tests to cover more category sizes (decoupled from helpers added in this PR) will land in a separate incoming PR:)

romoore (approved):
Please just change the test method `generate...` as we discussed. It should reflect the fact that a notification is being posted, windows added, etc.

PR #7514, opened 2019-05-29
[Buttons] Fix fallback behavior for Dynamic Type.

romoore:
The `titleFontForState:` API did not correctly reflect the rendered font
when no explicit fonts were set via the MDCButton APIs. As a result,
calling `updateTitleFont` internally could result in a different font
being returned than what was displayed.

Moving the "fall back" and Dynamic Type behavior into the
`titleFontForState:` method allows more consistent behavior (reduces the
risk of deviations) and corrects the APIs to reflect what is rendered.
This change should result in no visible changes to clients (the rendered
fonts should remain the same), but will result in non-`nil` results
being returned from `titleFontForState:` to correctly reflect the
rendered font for a given state.

Part of #7492


components/Buttons/tests/unit/ButtonsTests.m
```diff
@@ -521,6 +521,70 @@ - (void)testTitleFontForStateFallbackBehavior {
   }
 }
 
+- (void)testTitleFontForStateFallbackBehaviorWithLegacyDynamicType {
+  // Given
```
codeman7, line 525: _Tip:_ Fix this

codeman7 (changes requested):

bryanoltman (approved):

components/Buttons/tests/unit/ButtonsTests.m
romoore: Done.

components/ButtonBar/tests/unit/ButtonBarButtonTitleFontTests.swift
```diff
@@ -32,7 +33,7 @@ class ButtonBarButtonTitleFontTests: XCTestCase {
     // Then
     for view in buttonBar.subviews {
       if let button = view as? MDCButton {
-        XCTAssertNil(button.titleFont(for: .normal))
+        XCTAssertTrue(button.titleFont(for: .normal)!.mdc_isSimplyEqual(MDCTypography.buttonFont()))
```
codeman7: _Tip:_ Safely unwrap this in an `if let` and then in the `else` print out a logical error message.

components/NavigationBar/tests/unit/NavigationBarButtonTitleFontTests.swift
```diff
@@ -52,7 +53,7 @@ class NavigationBarButtonTitleFontTests: XCTestCase {
     // Then
     for view in recursiveSubviews(of: navigationBar) {
       if let button = view as? MDCButton {
-        XCTAssertNil(button.titleFont(for: .normal))
+        XCTAssertTrue(button.titleFont(for: .normal)!.mdc_isSimplyEqual(MDCTypography.buttonFont()))
```
codeman7: _Tip:_ Same as above.

codeman7 (approved):

codeman7:
@romoore Can you add a test where `adjust...Unavailabe = NO` and `mdc_adjust... = YES`.

romoore:
@codeman7 Isn't that `testLegacyDynamicTypeDisabledThenDynamicTypeTurnedOn`? Is there something in that test you think it should verify further?

components/NavigationBar/tests/unit/NavigationBarButtonTitleFontTests.swift
romoore: Done

components/ButtonBar/tests/unit/ButtonBarButtonTitleFontTests.swift
romoore: Done

codeman7:
@romoore Whoops I just saw the asserts, I didn't expand to see the Given, When. LGTM!

Issue #4456, opened 2018-06-29
[Tabs] We don't respect leading/trailing padding for scrollable tabs


codeman7:
We don't respect the leading or trailing padding for scrollable tabs.

## Reproduction steps

1. TabBar within dragons
2. Select 'TextTabs'

## Expected behavior

The guidelines says we should have a 52dp padding before scrollable tabs.

https://material.io/design/assets/1XBNaDj-1y46FsN4UN5Ak5-F2vTolLDIl/tabs-spec-scrollable.png

## Actual behavior

Currently we have padding but it is 8dp 

<img width="375" alt="screen shot 2018-06-29 at 2 17 09 pm" src="https://user-images.githubusercontent.com/7131294/42108018-51323bd4-7ba7-11e8-9b10-b1c8c4ddaf60.png">

## Platform (please complete the following information)

iPhone X/iOS 11

---
#### Internal data
- Associated internal bug: [b/117178106](http://b/117178106)

romoore:
MDCTabBarView has the correct leading inset and allows clients to adjust `contentInsets` if desired.

PR #7515, opened 2019-05-29
[Tabs] Provide a way for subclasses of MDCTabBar to override the trait collection.

mrhappyasthma:
This causes an awkward dependency between MDCItemBar -> MDCTabBar, but
it's my understanding that the long-term goal is to redesign MDCTabBar
to not use this subclass.

Basically this CL adds a helper class method in MDCTabBar (which may be overriden by
its subclasses) to return the size class for a given object (the
MDCItemBar instance). By default, this just returns the item bar's
sizeClass (as before). However subclasses may override this to return
custom sizeClass values as a way to circumvent the hardcoded dependency
on UITraitCollection's horizontalSizeClass.

I wrote this code in a way that none of the interfaces were changed,
such that this code can be dropped as the code of MDCTabBar/MDCItemBar
is refactored, while still allowing clients of this code to have an
avenue to avoid bugs caused in projects that do not yet adopt
UITraitCollection.

### DIFF
--- a/components/Tabs/src/MDCTabBar.m
+++ b/components/Tabs/src/MDCTabBar.m
@@ -445,6 +445,12 @@ static MDCItemBarAlignment MDCItemBarAlignmentForTabBarAlignment(
 
 #pragma mark - Private
 
+// This method may be overriden by subclasses that wish to return a different size
+// class to be used by MDCItemBar.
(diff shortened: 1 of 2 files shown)


mrhappyasthma:
This doesn't necessarily close issue #4456, but it allows clients to override the existing behavior by using the workaround.

romoore:
Closing this because the solution in #7517 was preferred.

PR #7570, opened 2019-06-12
[BottomSheet] Add respondsToSelector check in BottomSheet accessibility escape

### LINKED ISSUE #7569: [BottomSheet] MDCBottomSheetController accessibility escape can crash if delegate doesn't implement optional method
MDCBottomSheetController assumes its delegate implements the optional method `-bottomSheetControllerDidDismissBottomSheet:` in its `-accessibilityPerformEscape` method. 

1. Open the BottomSheetTypicalUseExample in VoiceOver.
2. Perform the two finger "Z" gesture.

## Expected behavior

The BottomSheet dismisses without issue.

## Actual behavior

The app crashes.

---
#### Internal data
- Associated internal bug: [b/135176883](http://b/135176883)

andrewoverton:
This PR fixes a crash by adding a respondsToSelector check to the BottomSheet accessibility escape.

Closes #7569.

### DIFF
--- a/components/BottomSheet/src/MDCBottomSheetController.m
+++ b/components/BottomSheet/src/MDCBottomSheetController.m
@@ -93,8 +93,10 @@
   [self
       dismissViewControllerAnimated:YES
                          completion:^{
-                           if ([weakSelf.delegate respondsToSelector:@selector(bottomSheetControllerDidDismissBottomSheet:)]) {
-                             [weakSelf.delegate bottomSheetControllerDidDismissBottomSheet:weakSelf];
+                           if ([weakSelf.delegate respondsToSelector:@selector
+                                                  (bottomSheetControllerDidDismissBottomSheet:)]) {
+                             [weakSelf.delegate
+                                 bottomSheetControllerDidDismissBottomSheet:weakSelf];
                            }
                          }];
   return YES;
--- a/components/BottomSheet/src/MDCBottomSheetController.m
+++ b/components/BottomSheet/src/MDCBottomSheetController.m
@@ -90,16 +90,16 @@
     return NO;
   }
   __weak MDCBottomSheetController *weakSelf = self;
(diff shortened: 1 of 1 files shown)


components/BottomSheet/src/MDCBottomSheetController.m
```diff
@@ -93,7 +93,11 @@ - (BOOL)accessibilityPerformEscape {
   [self
       dismissViewControllerAnimated:YES
                          completion:^{
-                           [weakSelf.delegate bottomSheetControllerDidDismissBottomSheet:weakSelf];
+                           if ([weakSelf.delegate respondsToSelector:@selector
```
romoore: Please capture `weakSelf.delegate` into a variable so it doesn't [disappear/change between the if and the use](https://github.com/stephanemoore/archives/blob/master/objc/pitfalls/tocttou.md).

romoore (approved):

components/BottomSheet/src/MDCBottomSheetController.m
```diff
@@ -93,7 +93,12 @@ - (BOOL)accessibilityPerformEscape {
   [self
       dismissViewControllerAnimated:YES
                          completion:^{
-                           [weakSelf.delegate bottomSheetControllerDidDismissBottomSheet:weakSelf];
+                           __strong __typeof(weakSelf) strongSelf = weakSelf;
```
romoore: This is a good improvement (switching to `__typeof`), but our team's convention is to avoid `typeof` and  `__typeof` whenever possible. We prefer to use explicitly-typed pointers in these circumstances.  I think in this case it's acceptable to use a pointer of type `MDCBottomSheetController *` explicitly.

If there's a specific reason not to use an explicit pointer type in this case, I think it's worth adding a comment to explain to future maintainers.

PR #7612, opened 2019-06-17
[Docs] Remove Travis CI badge, it is unused and confusing.

ajsecord:


jverkoey (approved):

romoore:
We should probably remove codecov.io as well since we don't actively look at it.

Issue #7640, opened 2019-06-19
{Bazel} Add unit test library rules


romoore:
Most of our components defined Objective-C and Swift unit test source libraries. The rules for them are mostly boilerplate (excepting dependencies).  We should have standardized macros so we don't need to write so much boilerplate.

PR #7642, opened 2019-06-19
[AppBar] Use new unit test Bazel rules

romoore:
Follow-up to #7640

### DIFF
--- a/components/AppBar/BUILD
+++ b/components/AppBar/BUILD
@@ -22,6 +22,8 @@ load(
     "mdc_public_objc_library",
     "mdc_snapshot_objc_library",
     "mdc_snapshot_test",
+    "mdc_unit_test_objc_library",
+    "mdc_unit_test_swift_library",
     "mdc_unit_test_suite",
 )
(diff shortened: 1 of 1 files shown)


components/AppBar/BUILD
```diff
@@ -119,28 +121,19 @@ mdc_examples_swift_library(
     ],
 )
 
-mdc_objc_library(
+mdc_unit_test_objc_library(
```
galiak11, line 124: Why remove the private visibility for these 2?

galiak11 (approved):

components/AppBar/BUILD
romoore: It's not - the new macros have the visibility defaulting to private. We would only override visibility on these targets if it was anything except private.

```
def mdc_unit_test_objc_library(
    name,
    extra_srcs = [],
    deps = [],
    sdk_frameworks = [],
    testonly = 1,
    **kwargs):
    """Declare an mdc_objc_library for unit test sources."""
    mdc_objc_library(
```

PR #7648, opened 2019-06-19
[Banner] Fix horizontal constraint between text label and button.

fluxminds:
Fix the issue that label are not correctly laid out and shrunk because of low compression priority.

An example content view update issue is fixed to show a correct example screenshot.

Note: this issue doesn't happen on iPhone X and only happens on iPhone 7.

| Before | After |
| --- | --- |
| Simulator Screen Shot - iPhone 7 - 2019-06-19 at 13 34 37 | Simulator Screen Shot - iPhone 7 - 2019-06-19 at 13 31 35 |
|Simulator Screen Shot - iPhone 7 - 2019-06-19 at 13 34 38 |   Simulator Screen Shot - iPhone 7 - 2019-06-19 at 13 31 36 |

### DIFF
--- a/components/Banner/src/MDCBannerView.m
+++ b/components/Banner/src/MDCBannerView.m
@@ -60,9 +60,9 @@ static NSString *const kMDCBannerViewImageViewImageKeyPath = @"image";
 // Buttons constraints
 @property(nonatomic, readwrite, strong) NSLayoutConstraint *buttonContainerConstraintLeading;
 @property(nonatomic, readwrite, strong)
-    NSLayoutConstraint *buttonContainerConstraintLeadingWithTextLabel;
+    NSLayoutConstraint *buttonContainerConstraintWidthWithLeadingButton;
 @property(nonatomic, readwrite, strong)
-    NSLayoutConstraint *buttonContainerConstraintLeadingWithTextLabelGreater;
+    NSLayoutConstraint *buttonContainerConstraintLeadingWithTextLabel;
 @property(nonatomic, readwrite, strong) NSLayoutConstraint *buttonContainerConstraintTrailing;
 @property(nonatomic, readwrite, strong) NSLayoutConstraint *buttonContainerConstraintTopWithMargin;
 @property(nonatomic, readwrite, strong)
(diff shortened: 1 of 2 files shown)


components/Banner/src/MDCBannerView.m
```diff
@@ -60,9 +60,9 @@ @interface MDCBannerView ()
 // Buttons constraints
 @property(nonatomic, readwrite, strong) NSLayoutConstraint *buttonContainerConstraintLeading;
 @property(nonatomic, readwrite, strong)
-    NSLayoutConstraint *buttonContainerConstraintLeadingWithTextLabel;
+    NSLayoutConstraint *buttonContainerConstraintWidthWithLeadingButton;
```
codeman7, line 63: _Tip:_ buttonContainerWidthConstraint
Also documentating these properties would help our future selves understand what they are for.

components/Banner/src/MDCBannerView.m
```diff
@@ -60,9 +60,9 @@ @interface MDCBannerView ()
...
 @property(nonatomic, readwrite, strong)
-    NSLayoutConstraint *buttonContainerConstraintLeadingWithTextLabelGreater;
+    NSLayoutConstraint *buttonContainerConstraintLeadingWithTextLabel;
```
codeman7: Please add docs so it's easier to understand what this property does.

bryanoltman:
I'm not seeing this issue in the develop branch. I'm using the iPhone 7 simulator with 11.4

codeman7 (approved):

fluxminds:
@bryanoltman Maybe it only happens for iPhone 7 11.0.1 and some older version. I can understand the reason behind it why it doesn't show correctly, but I am confused why different OSs produce different result. I assume there is some autolayout system change behind the scene.

components/Banner/src/MDCBannerView.m
fluxminds: I felt `..WithLeadingButton` would be more explicit for now when it is a property inside the implementation file. How do you feel if we keep it for now? I am thinking there might be more width constraints to be added later on so I added this suffix.

components/Banner/src/MDCBannerView.m
fluxminds: Addressed in 6e7588f8128db5c72fc411fea383a46b607b31ad.

components/Banner/src/MDCBannerView.m
codeman7: Sounds good, I think docs would make it more clear what this constraint is responsible for.

fluxminds: Sounds like a general doc would be good for these constraints. I will address it as a separate issue :)

PR #7649, opened 2019-06-19
[Banner] fix image constraint when it is hidden and on single row mode.

fluxminds:
After adding support for single row mode on Banner, image constraint needs to be updated. Otherwise it will break the vertical constraints because of the height change.


components/Banner/src/MDCBannerView.m
```diff
@@ -48,9 +48,9 @@ @interface MDCBannerView ()
 
 // Image constraints
 @property(nonatomic, readwrite, strong) NSLayoutConstraint *imageViewConstraintLeading;
-@property(nonatomic, readwrite, strong) NSLayoutConstraint *imageViewConstraintTopSmall;
+@property(nonatomic, readwrite, strong) NSLayoutConstraint *imageViewConstraintCenterY;
```
codeman7, line 51: Please add docs.

components/Banner/src/MDCBannerView.m
```diff
@@ -195,15 +195,11 @@ - (void)setUpImageViewConstraints {
...
-      [self.imageView.topAnchor constraintEqualToAnchor:self.layoutMarginsGuide.topAnchor
-                                               constant:kTopPaddingSmall];
   self.imageViewConstraintTopLarge =
       [self.imageView.topAnchor constraintEqualToAnchor:self.layoutMarginsGuide.topAnchor
                                                constant:kTopPaddingLarge];
-  self.imageViewConstraintBottom =
-      [self.imageView.bottomAnchor constraintEqualToAnchor:self.layoutMarginsGuide.bottomAnchor
-                                                  constant:-kBottomPadding];
+  self.imageViewConstraintCenterY =
+      [self.imageView.centerYAnchor constraintEqualToAnchor:self.buttonContainerView.centerYAnchor];
```
codeman7, line 202: Should this be centered to self.centerYAnchor?

codeman7 (approved):
Please add screenshot to the commit message or snapshot test to capture the improvement so we don't regress later.

components/Banner/src/MDCBannerView.m
fluxminds: Button, text and button are not centered to the view on Y axis according to the guideline. There is a one point off.

components/Banner/src/MDCBannerView.m
fluxminds: Followed up in https://github.com/material-components/material-components-ios/issues/7650. Will be done altogether when finalized.

fluxminds:
Snapshot tests have been added.

PR #7656, opened 2019-06-19
[Tabs2] Add items and selectedItems API to the MDCTabBarView

### LINKED ISSUE #7646: [Tabs] Internal issue: b/135607139
This was filed as an internal issue. If you are a Googler, please visit [b/135607139](http://b/135607139) for more details.

---
#### Internal data
- Associated internal bug: [b/135607139](http://b/135607139)

leonmz:
Implement the items and selectedItems properties in the MDCTabBarView.
Adds test for these two properties.

closes #7646


components/Tabs2/src/MDCTabBarView.h
```diff
@@ -19,4 +19,13 @@
...
 __attribute__((objc_subclassing_restricted)) @interface MDCTabBarView : UIScrollView
 
+/** The set of items displayed in the Tab bar. */
+@property(nonnull, nonatomic, copy) NSArray<__kindof UITabBarItem *> *items;
+
+/** The currently-selected item in the Tab bar. */
+@property(nullable, nonatomic, strong) __kindof UITabBarItem *selectedItem;
+
+/** Selects one of the items with optional animation. */
+- (void)setSelectedItem:(__kindof UITabBarItem *)selectedItem animated:(BOOL)animated;
```
romoore: I think it's better to leave the animation for a later PR and remove the `setSelectedItem:animated:` API for now. This PR can focus on the basics of item management and selection. Once we have an Example VC in-place, we can start adding animation so we can verify the timing and appearance.

components/Tabs2/src/MDCTabBarView.m
```diff
@@ -16,4 +16,51 @@
 
 @implementation MDCTabBarView
 
+#pragma mark - Properties
+
+- (void)setItems:(NSArray<UITabBarItem *> *)items {
+  NSAssert([NSThread isMainThread], @"Item array may only be set on the main thread");
+  NSParameterAssert(items != nil);
```
romoore: Tip: Nit: You can just use `NSParameterAssert(items)`. You don't need to explicitly check against `nil`.

components/Tabs2/src/MDCTabBarView.m
```diff
@@ -16,4 +16,51 @@
...
+    _items = [items copy];
+
+    // Determine new selected item, defaulting to the first item.
+    UITabBarItem *newSelectedItem = _items.firstObject;
+    if (_selectedItem && [_items containsObject:_selectedItem]) {
+      // Previously-selected item still around: Preserve selection.
+      newSelectedItem = _selectedItem;
+    }
+
+    // Update _selectedItem directly so it's available for -reload.
```
romoore: Was this comment copied from MDCTabBar? I don't believe this class has a `-reload` method.

components/Tabs2/src/MDCTabBarView.m
```diff
@@ -16,4 +16,51 @@
...
+
+- (void)setItems:(NSArray<UITabBarItem *> *)items {
+  NSAssert([NSThread isMainThread], @"Item array may only be set on the main thread");
+  NSParameterAssert(items != nil);
+
+  if (_items != items && ![_items isEqual:items]) {
+    _items = [items copy];
+
+    // Determine new selected item, defaulting to the first item.
+    UITabBarItem *newSelectedItem = _items.firstObject;
```
romoore: I think since the selected item can be `nil`, it's better to default to that.  What do you think?

components/Tabs2/src/MDCTabBarView.m
```diff
@@ -16,4 +16,51 @@
...
+      // Previously-selected item still around: Preserve selection.
+      newSelectedItem = _selectedItem;
+    }
+
+    // Update _selectedItem directly so it's available for -reload.
+    _selectedItem = newSelectedItem;
+  }
+}
+
+- (void)setSelectedItem:(nullable UITabBarItem *)selectedItem {
```
romoore: Nit: Let's keep nullability annotations limited to the headers. MDC-iOS doesn't generally annotate method implementations with nullability.

components/Tabs2/src/MDCTabBarView.m
```diff
@@ -16,4 +16,51 @@
...
+  [self setSelectedItem:selectedItem animated:NO];
+}
+
+- (void)setSelectedItem:(nullable UITabBarItem *)selectedItem animated:(BOOL)animated {
+  if (_selectedItem != selectedItem) {
+    NSUInteger itemIndex = [self indexForItem:selectedItem];
+    if (selectedItem && (itemIndex == NSNotFound)) {
+      [[NSException exceptionWithName:NSInvalidArgumentException
+                               reason:@"Invalid item"
```
romoore: I love the exception raising! ❤️ 

Nit: Please add a slightly more verbose message. Something like, 

```
[NSString stringWithFormat:"(%@) is not a member of the tab bar's `items`.", selectedItem]
```

Feel free to go with something else, but I'd prefer a more descriptive exception.

components/Tabs2/src/MDCTabBarView.m
```diff
@@ -16,4 +16,51 @@
...
+    }
+
+    _selectedItem = selectedItem;
+  }
+}
+
+#pragma mark - Private
+
+- (NSInteger)indexForItem:(nullable UITabBarItem *)item {
+  if (item) {
```
romoore: This is the same API and behavior as `-[NSArray indexOfObject]`, so I'd prefer to just use that directly. Is there a reason why you think this method should be used instead?

components/Tabs2/tests/unit/MDCTabBarViewTests.m
```diff
@@ -19,10 +19,103 @@
 @interface MDCTabBarViewTests : XCTestCase
 @end
 
-@implementation MDCTabBarViewTests
+@implementation MDCTabBarViewTests {
+  MDCTabBarView *_tabBarView;
```
romoore: Nit: Please prefer properties over instance variables. 

```objc
@interface MDCTabBarViewTests
@property(nonatomic, strong) MDCTabBarView *tabBarView;
...
```

components/Tabs2/tests/unit/MDCTabBarViewTests.m
```diff
@@ -19,10 +19,103 @@
...
+  [super tearDown];
+}
 
 - (void)testInitCreatesObject {
   XCTAssertNotNil([[MDCTabBarView alloc] init]);
 }
 
+/// Tab bars should by default select the first item in their items array.
+- (void)testSelectsFirstItemByDefault {
+  _tabBarView.items = @[ _itemA, _itemB, _itemC ];
```
romoore: Please use `Given`, `When`, and `Then` comments and test structure.  There are [many good resources discussing Given-When-Then testing](http://www.draconianoverlord.com/2017/11/28/given-when-then.html) available on the web, and our code base uses it extensively.

romoore (commented):
Overall looks good. Some feedback on the public APIs and defaulting behavior, plus a few nits/tips on the internals and tests.

components/Tabs2/src/MDCTabBarView.h
leonmz: Sure! Then I moved all logic inside this function to selctedItem setter :)

components/Tabs2/src/MDCTabBarView.m
leonmz: Fix applied

components/Tabs2/src/MDCTabBarView.m
leonmz: Sorry about that. Removed.

components/Tabs2/src/MDCTabBarView.m
leonmz: I thought this is a little bit tricky. The original behavior of MDCTabBar will select the first item from the items. But yes, the behavior of UITabBar stays nil, I will set this as nil

components/Tabs2/src/MDCTabBarView.m
leonmz: Fix applied.

components/Tabs2/src/MDCTabBarView.m
leonmz: Updated!

components/Tabs2/tests/unit/MDCTabBarViewTests.m
leonmz: Fix applied.

components/Tabs2/src/MDCTabBarView.h
jverkoey: Please remove all of the `__kindof` annotations, unless there is a strong technical reason for providing them (please explain it if so). See http://go/__kindof-scary for a discussion on this annotation, and https://stackoverflow.com/a/33799210 for more details on why this can lead to unexpected crashes.

tl;dr: This annotation hides legitimate warnings that come from down-casting an instance to a different type than what is specified by the generics annotations.

E.g.:

```objc
NSMutableArray<__kindof UIView *> *subviews = [[NSMutableArray alloc] init];

[subviews addObject:[[UIView alloc] init]]; // Works
[subviews addObject:[[UIImageView alloc] init]]; // Also works

subviews[0].image = [UIImage ...]; // Crash, no compile-time warning
```

components/Tabs2/src/MDCTabBarView.m
```diff
@@ -16,4 +16,51 @@
...
+    _selectedItem = newSelectedItem;
+  }
+}
+
+- (void)setSelectedItem:(nullable UITabBarItem *)selectedItem {
+  [self setSelectedItem:selectedItem animated:NO];
+}
+
+- (void)setSelectedItem:(nullable UITabBarItem *)selectedItem animated:(BOOL)animated {
+  if (_selectedItem != selectedItem) {
```
jverkoey: An early exit here will remove a layer of indentation and more closely align with our code base's intended conventions:

```objc
if (_selectedItem == selectedItem) {
  return;
}

// Remaining logic

_selectedItem = selectedItem;
```

components/Tabs2/src/MDCTabBarView.m
jverkoey: UITabBar's behavior when the item is not found is to not adjust the `selectedItem`.

I recommend we follow UIKit's behavior rather than introducing a new one.

```objc
UITabBar *tabBar = [[UITabBar alloc] initWithFrame:CGRectZero];
UITabBarItem * item1 = [[UITabBarItem alloc] initWithTitle:@"Hi" image:nil tag:1];
UITabBarItem * item2 = [[UITabBarItem alloc] initWithTitle:@"Hi" image:nil tag:2];
UITabBarItem * item3 = [[UITabBarItem alloc] initWithTitle:@"Hi" image:nil tag:3];
tabBar.items = @[item1, item2 ];

NSLog(@"%@", @(tabBar.selectedItem.tag)); // 0

tabBar.selectedItem = item1;
NSLog(@"%@", @(tabBar.selectedItem.tag)); // 1

tabBar.selectedItem = item2;
NSLog(@"%@", @(tabBar.selectedItem.tag)); // 2

tabBar.selectedItem = item3;
NSLog(@"%@", @(tabBar.selectedItem.tag)); // 2
```

components/Tabs2/src/MDCTabBarView.m
```diff
@@ -16,4 +16,51 @@
 
 @implementation MDCTabBarView
 
+#pragma mark - Properties
+
+- (void)setItems:(NSArray<UITabBarItem *> *)items {
+  NSAssert([NSThread isMainThread], @"Item array may only be set on the main thread");
```
jverkoey: Material iOS UI components don't explicitly guard properties from being accessed off the main thread. This is because UIKit is generally not thread safe to start with. Is there a strong technical reason for this assertion? If not, please remove it.

components/Tabs2/src/MDCTabBarView.m
jverkoey: UIKit does change the selectedItem to nil when the selected item no longer exists in the new items array:

```objc
UITabBar *tabBar = [[UITabBar alloc] initWithFrame:CGRectZero];
UITabBarItem * item1 = [[UITabBarItem alloc] initWithTitle:@"Hi" image:nil tag:1];
UITabBarItem * item2 = [[UITabBarItem alloc] initWithTitle:@"Hi" image:nil tag:2];
UITabBarItem * item3 = [[UITabBarItem alloc] initWithTitle:@"Hi" image:nil tag:3];
tabBar.items = @[item1, item2 ];
tabBar.selectedItem = item1;
NSLog(@"%@", @(tabBar.selectedItem.tag)); // 1
tabBar.items = @[item1 ];
NSLog(@"%@", @(tabBar.selectedItem.tag)); // 0
```

I recommend following the UIKit behavior and writing a test that validates the UIKit behavior over time in case it changes from one iOS version to another.

components/Tabs2/tests/unit/MDCTabBarViewTests.m
leonmz: Updated.

components/Tabs2/src/MDCTabBarView.m
romoore: This was taken from MDCTabBar, so changing will be an intentional deviation from MDCTabBar.  However, since UIKit just ignores these errors (no messages, no exceptions), I think it's fine if the MDCTabBarView does as well.

components/Tabs2/src/MDCTabBarView.h
romoore: Let's remove it for now. I know the API docs specified using `__kindof` (to follow UIKit's more recent conventions), but we can always add it back in later.  Better to move fast and not debate things in a PR.

components/Tabs2/src/MDCTabBarView.m
romoore: +1 to remove - this was copied from MDCItemBar.  It's a good instinct to copy from existing code that seems to do the same thing, but unfortunately that's old code (pre-MDC) and doesn't really follow our team's current conventions.

components/Tabs2/src/MDCTabBarView.m
leonmz: Sure! I changed it to the same behavior with the UIKit.
I try to understand your meaning of test against But what do you exactly mean when you mention the test against the UIKit?

components/Tabs2/src/MDCTabBarView.h
leonmz: SGTM

components/Tabs2/src/MDCTabBarView.m
jverkoey: Something like so, just for consistency validation in terms of expected behavior:

```objc
- (void)testUITabBarSelectedItemBecomesNilWhenRemovedFromItems {
  // Given
  UITabBar *tabBar = [[UITabBar alloc] initWithFrame:CGRectZero];
  UITabBarItem * item1 = [[UITabBarItem alloc] initWithTitle:@"Hi" image:nil tag:1];
  UITabBarItem * item2 = [[UITabBarItem alloc] initWithTitle:@"Hi" image:nil tag:2];
  tabBar.items = @[item1, item2 ];
  tabBar.selectedItem = item2;

  // When
  tabBar.items = @[item1];

  // Then
  XCTAssertNil(tabBar.selectedItem);
}
```

components/Tabs2/src/MDCTabBarView.m
leonmz: Thank for reminding! Fixed.

components/Tabs2/src/MDCTabBarView.m
romoore: +1 to following UIKit behavior (don't automatically pick a selection). 

Regarding the test, I don't know that it's worth writing a unit test to monitor UIKit's behavior. Even if UITabBar does change in the future, there are only three outcomes that I can think of: rushing to change our API to match (without necessarily deciding it's the best option); turning off the test; or turn off the test, file an Issue to investigate changing the behavior, and then possibly change it.

Without the test, we would have discovered the difference once a client asked us to match UIKit's behavior.  Until a client is dissatisfied with the behavior at that point, I don't think it's worth the canary.

Just my thoughts, and @jverkoey trumps as the TL if he really wants the test.

components/Tabs2/src/MDCTabBarView.m
leonmz: Sure. I will try to understand more on how current stand instead of copying from old existing code. :)

components/Tabs2/tests/unit/MDCTabBarViewTests.m
romoore: Looks like these are still instance variables.  I'm going to unresolve this to track the outcome.

components/Tabs2/src/MDCTabBarView.m
jverkoey: Canaries are helpful when it comes time to validating our behavior against the expectations of new iOS versions, but I don't feel particularly strongly about adding it. The risk of UIKit changing this behavior to something that would be surprising for clients is pretty low.

components/Tabs2/src/MDCTabBarView.m
leonmz: Fixed!

components/Tabs2/src/MDCTabBarView.m
leonmz: I mean:
/// Tab bars should by default select nil in their items array. The behavior should also consistent
/// with the UIKit
- (void)testSelectsNilByDefault {
  // Given
  UITabBar *tabBar = [[UITabBar alloc] initWithFrame:CGRectZero];
  tabBar.items = @[ _itemA, _itemB, _itemC ];
  _tabBarView.items = @[ _itemA, _itemB, _itemC ];

  // Then
  XCTAssertNil(_tabBarView.selectedItem);
  XCTAssertNil(tabBar.selectedItem);
}

Does this looks good to you?

leonmz: So shall I left with no tests in this case?

components/Tabs2/src/MDCTabBarView.m
```diff
@@ -16,4 +16,40 @@
...
+  if (_selectedItem == selectedItem) {
+    return;
+  }
+  NSUInteger itemIndex = [_items indexOfObject:selectedItem];
+  if (selectedItem && (itemIndex == NSNotFound)) {
+    NSString *itemTitle = selectedItem.title;
+    NSString *exceptionMessage =
+    [NSString stringWithFormat:@"%@ is not a member of the tab bar's `items`.", itemTitle];
+    [[NSException exceptionWithName:NSInvalidArgumentException
+                             reason:exceptionMessage
```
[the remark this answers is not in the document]
leonmz: Fixed.

components/Tabs2/src/MDCTabBarView.m
leonmz: Thanks for everyone's input! So I just leave with no tests against UIKit for now...

jverkoey: To give more clear guidance here: I think we'll be ok if we don't write the UIKit canary so it's ok not to add such a test in this PR.

components/Tabs2/tests/unit/MDCTabBarViewTests.m
leonmz: Fixed.

components/Tabs2/src/MDCTabBarView.h
```diff
@@ -19,4 +19,10 @@
  */
 __attribute__((objc_subclassing_restricted)) @interface MDCTabBarView : UIScrollView
 
+/** The set of items displayed in the Tab bar. */
+@property(nonnull, nonatomic, copy) NSArray<UITabBarItem *> *items;
```
jverkoey, line 23: UITabBar's items can be `nullable`; do we want to follow suit here as well?

jverkoey: I think a good argument for making this nonnull is that it makes the Swift code a bit easier to work with, so I'm supportive of this deviation in that regard. I'd love y'all's thoughts on this.

If we do decide to keep this as nonnull, please add a test verifying that `items` is not nil after initialization:

```objc
// Given
MDCTabBarView *tabBarView = [[MDCTabBarView alloc] init];

// Then
XCTAssertNotNil(tabBarView.items);
```

romoore: +1 I think `nullable` is fine to match UITabBar.

romoore: MDCTabBar's property is `nonnull`, so making it `nullable` will certainly make Swift migration more expensive. I'm not sure there's much of a benefit to a `nil` array versus an empty one.  I feel like I'm waffling, but I don't feel strongly one way or another.

jverkoey: I say we skate where the puck is going on this one and make the API more friendly for Swift folks. But please add the test (+ you'll find that you're missing an initializer :) ).

leonmz: Thanks! Tests and initializer added :)

components/Tabs2/src/MDCTabBarView.m
```diff
@@ -16,4 +16,49 @@
 
 @implementation MDCTabBarView
 
+#pragma mark - Initialization
+
+- (instancetype)init {
+  self = [super init];
+  if (self) {
+    self.items = @[];
```
romoore: Please assign to the instance variable directly here. Since `self` is not yet fully initialized, the `setItems:` method may fail.

```
if (self) {
  _items = @[];
}
```

components/Tabs2/src/MDCTabBarView.m
```diff
@@ -16,4 +16,49 @@
...
+  }
+  return self;
+}
+
+#pragma mark - Properties
+
+- (void)setItems:(NSArray<UITabBarItem *> *)items {
+  NSParameterAssert(items);
+
+  if (_items != items && ![_items isEqual:items]) {
```
romoore: Tip: Consider flipping this logic.

```objc
if (self.items == items || [items isEqual:self.items]) {
  return;
}
```

romoore: Please access other properties via the appropriate setters/getters here.  The exception is the actual method you're in (setItems).

components/Tabs2/src/MDCTabBarView.m
```diff
@@ -16,4 +16,49 @@
...
+      // Previously-selected item still around: Preserve selection.
+      newSelectedItem = _selectedItem;
+    }
+
+    _selectedItem = newSelectedItem;
+  }
+}
+
+- (void)setSelectedItem:(UITabBarItem *)selectedItem {
+  if (_selectedItem == selectedItem) {
```
romoore: Please use the property getters and setters here except assigning `_selectedItem` since this is the setter.

components/Tabs2/tests/unit/MDCTabBarViewTests.m
```diff
@@ -17,12 +17,131 @@
...
+  _itemB = nil;
+  _itemC = nil;
+  _tabBarView = nil;
+
+  [super tearDown];
+}
+
 - (void)testInitCreatesObject {
-  XCTAssertNotNil([[MDCTabBarView alloc] init]);
+  // Given
```
romoore: Tip: I believe this would be the "When" of the test.  You're testing what the result of creating an object is.

components/Tabs2/tests/unit/MDCTabBarViewTests.m
```diff
@@ -17,12 +17,131 @@
...
+  MDCTabBarView *tabBarView = [[MDCTabBarView alloc] init];
+
+  // Then
+  XCTAssertNotNil(tabBarView);
+  XCTAssertNotNil(tabBarView.items);
+}
+
+/// Tab bars should by default select nil in their items array. The behavior should also consistent
```
romoore: Nit: s/also consistent/also be consistent/

components/Tabs2/tests/unit/MDCTabBarViewTests.m
```diff
@@ -17,12 +17,131 @@
...
+/// with the UIKit
+- (void)testSelectsNilByDefault {
+  // Given
+  _tabBarView.items = @[ _itemA, _itemB, _itemC ];
```
romoore: Please use property accessors here and throughout.

components/Tabs2/tests/unit/MDCTabBarViewTests.m
```diff
@@ -17,12 +17,131 @@
...
+  [super setUp];
+
+  _tabBarView = [[MDCTabBarView alloc] init];
+  _itemA = [[UITabBarItem alloc] initWithTitle:@"A" image:nil tag:0];
+  _itemB = [[UITabBarItem alloc] initWithTitle:@"B" image:nil tag:0];
+  _itemC = [[UITabBarItem alloc] initWithTitle:@"C" image:nil tag:0];
+}
+
+- (void)tearDown {
+  _itemA = nil;
```
romoore: Please use property setters here and throughout.

components/Tabs2/tests/unit/MDCTabBarViewTests.m
```diff
@@ -17,12 +17,131 @@
...
+
 @end
 
 @implementation MDCTabBarViewTests
 
+- (void)setUp {
+  [super setUp];
+
+  _tabBarView = [[MDCTabBarView alloc] init];
+  _itemA = [[UITabBarItem alloc] initWithTitle:@"A" image:nil tag:0];
```
romoore: Please use property setters here and throughout.

romoore (approved):
Please switch to accessors (or do it in a fast-follow PR). They are generally safer and are Apple's recommended best practice.

components/Tabs2/src/MDCTabBarView.m
leonmz: Thanks fro reminding!

components/Tabs2/src/MDCTabBarView.m
leonmz: Sure!

leonmz: Fix applied.

components/Tabs2/src/MDCTabBarView.m
leonmz: Fix applied.

components/Tabs2/tests/unit/MDCTabBarViewTests.m
leonmz: Sure!

components/Tabs2/tests/unit/MDCTabBarViewTests.m
leonmz: Fix applied.

components/Tabs2/tests/unit/MDCTabBarViewTests.m
leonmz: Fix applied.

components/Tabs2/tests/unit/MDCTabBarViewTests.m
leonmz: Fix applied.

components/Tabs2/tests/unit/MDCTabBarViewTests.m
leonmz: Fix applied.

jverkoey (approved):

PR #7678, opened 2019-06-21
[Tabs] Add initial support for custom view.

### LINKED ISSUE #7676: [Tabs] Internal issue: b/135757221
This was filed as an internal issue. If you are a Googler, please visit [b/135757221](http://b/135757221) for more details.

---
#### Internal data
- Associated internal bug: [b/135757221](http://b/135757221)

romoore:
Creates basic support for a custom view property on UITabBarItem. Although not
a category on UITabBarItem, it can be used in the same places UITabBarItem can
be used. This will allow clients of MDCTabBarView to provide a custom view to
replace the default presentation of the title and image.

Closes #7676


leonmz (approved):

components/Tabs/src/TabBarView/MDCTabBarItemCustomViewing.h
```diff
@@ -0,0 +1,25 @@
...
+
+#import <Foundation/Foundation.h>
+
+#import "MDCTabBarSelectionIndicatorSupporting.h"
+
+@protocol MDCTabBarItemCustomViewing
+
+/** A custom view to be displayed for a tab bar item. */
+@property(nullable, nonatomic, strong)
+    __kindof UIView<MDCTabBarSelectionIndicatorSupporting> *mdc_customView;
```
andrewoverton: Do we prefer to use `__kindof` in situations like this? I forget what our team convention is, or if there is one.

fluxminds (approved):

components/Tabs/src/TabBarView/MDCTabBarItemCustomViewing.h
andrewoverton: I think this is fine, I just remember you mentioning that some people were against generics like this

components/Tabs/src/TabBarView/MDCTabBarSelectionIndicatorSupporting.h
```diff
@@ -0,0 +1,36 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <Foundation/Foundation.h>
+
+/**
+ A simple protocol that indicates the responder can be targeted by an
+ @c MDCTabBarSelectionIndicatorTemplate.
+
+ @seealso MDCTabBarSelectionIndicatorTemplate
```
andrewoverton: Is this type in develop? I don't see it anywhere

components/Tabs/src/TabBarView/MDCTabBarItem.h
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+#import <UIKit/UIKit.h>
+
+#import "MDCTabBarItemCustomViewing.h"
+
+/**
+ A subclass of @c UITabBarItem that supports a custom view property.
+ */
+@interface MDCTabBarItem : UITabBarItem <MDCTabBarItemCustomViewing>
```
andrewoverton: This is intended to be public? Clients should use this? Should it have subclassing restricted?

romoore: It is intended to be public, I can restrict subclassing on it.

andrewoverton: Do we expect other objects to conform to `MDCTabBarItemCustomViewing `? If we just want to add a property to UITabBarItem with a subclass do we really need this protocol to do that?

components/Tabs/src/TabBarView/MDCTabBarSelectionIndicatorSupporting.h
romoore: Yep, good catch.

components/Tabs/src/TabBarView/MDCTabBarItemCustomViewing.h
romoore: I think @jverkoey doesn't like it. I will remove it.

romoore: Done.

components/Tabs/src/TabBarView/MDCTabBarSelectionIndicatorSupporting.h
romoore: Done.

components/Tabs/src/TabBarView/MDCTabBarItem.h
romoore: I suspect most clients don't need more than this subclass. However, I didn't want to do `kindOfClass` checks, so the protocol makes the most sense for the implementation.  It also allows clients to use custom implementations without requiring subclassing of our class.

romoore: Does that make sense? Do you think we shouldn't provide a custom subclass at all and just expect clients to write it (we'll still need subclasses for tests and example VCs).

components/Tabs/src/TabBarView/MDCTabBarItemCustomViewing.h
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <Foundation/Foundation.h>
+
+#import "MDCTabBarSelectionIndicatorSupporting.h"
+
+@protocol MDCTabBarItemCustomViewing
```
andrewoverton, line 25: I saw you documented the property in the protocol, maybe the protocol itself deserves some documentation

romoore: Done.

components/Tabs/src/TabBarView/MDCTabBarItem.h
andrewoverton: I don't quite follow yet. If clients don't use MDCTabBarItem will they use UITabBarItem? Or a different subclass if UITabBarItem? Or something that's not a UITabBarItem at all but conforms to MDCTabBarItemCustomViewing?

romoore: They can use any instance or subclass of UITabBarItem for assigning items to the MDCTabBarView.  They can use ours if it's convenient or write their own subclass if that suits their needs better. Since one of our clients definitely needs a custom view for their own app, I think it made sense to provide a simple subclass out of the box, rather than have them write one.  I doubt many clients will conform to the protocol and not be able to use our subclass. However, since it's easier to work with protocols instead of expected subclasses, I went with the protocol route.

andrewoverton: Ok got it. Thanks for clarifying!

andrewoverton (approved):

PR #7694, opened 2019-06-24
[TextFields] Add MDCBaseTextField as extension

andrewoverton:
https://github.com/material-components/material-components-ios/pull/7410 added MDCBaseTextField as a beta component in a sibling directory of the main TextFields directory. https://github.com/material-components/material-components-ios/pull/7584 added MDCBaseTextField as a beta component within the main TextFields directory. This PR adds MDCBaseTextField as a non-beta component (i.e. not in MaterialComponentsBeta.podspec) in an extension of the TextFields subspec called TextFIelds+Beta.


MaterialComponents.podspec
```diff
@@ -1787,6 +1787,23 @@ Pod::Spec.new do |mdc|
     end
   end
 
+  mdc.subspec "TextFields+Beta" do |extension|
```
romoore: Tip: It would be nice if the extension name were a specific feature/project rather than the umbrella term "Beta."  That way we can have multiple things in Beta at any given time.

components/TextFields/src/Beta/MDCBaseTextField.h
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <UIKit/UIKit.h>
+
+/**
+ A UITextField subclass that provides the foundation for Material TextFields.
+ */
+@interface MDCBaseTextField : UITextField
```
romoore: Tip: It would be great to have a note about this being in Beta and breaking changes will not affect the library version.

romoore (approved):

MaterialComponents.podspec
andrewoverton: Do any alternative extension names come to mind for you? I see your point about "Beta" not being ideal.

romoore: Not being familiar with the project/goal, I'm not sure how to name it. For the Tabs work I'm doing, I'm planning to have multiple extensions for the same project (`+TabBarView` is current, `+NewTabBarViewController` (naming!) is planned).  I'm fine with `+Beta` if it's a "glob" target, otherwise "Naming is hard."

andrewoverton: Huh, I could rename it to `MaterialTextFields+ContainedInputViews`...

components/TextFields/src/Beta/MDCBaseTextField.h
andrewoverton: Done.

PR #7704, opened 2019-06-24
[Tabs] Adds justified and scrollable layout to MDCTabBarView

leonmz:
Adds justified and scrollable layout for MDCTabBarView and corresponding tests.

## Example View Controller

|Before|After (Portrait)|After (Landscape)
|---|---|---|
|60151483-ea246900-97aa-11e9-9c1e-688f1d241512|Simulator Screen Shot - iPhone 7 - 2019-06-26 at 08 27 47|Simulator Screen Shot - iPhone 7 - 2019-06-26 at 08 29 09|

Part of Issue #7645


components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -15,13 +15,15 @@
 #import "MDCTabBarView.h"
 #import "private/MDCTabBarViewItemView.h"
 
+#import "private/MDCTabBarViewItemView.h"
```
mikefan1991: This is duplicate import.

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -30,10 +32,10 @@ @implementation MDCTabBarView
 #pragma mark - Initialization
 
 - (instancetype)init {
-  self = [super init];
+  self = [super initWithFrame:CGRectZero];
```
mikefan1991: This change is not necessary.

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -93,28 +80,62 @@ - (void)setSelectedItem:(UITabBarItem *)selectedItem {
...
+  for (UIView *view in self.stackView.arrangedSubviews) {
+    [view removeFromSuperview];
+    [_stackView removeArrangedSubview:view];
+  }
+
+  for (UITabBarItem *item in self.items) {
+    MDCTabBarViewItemView *itemView = [[MDCTabBarViewItemView alloc] init];
+    itemView.translatesAutoresizingMaskIntoConstraints = NO;
+    itemView.title = item.title;
+    itemView.image = item.image;
```
mikefan1991: I have removed the title and the image in another PR. Please wait for that to be merged and then update this.

components/Tabs/tests/snapshot/TabBarView/MDCTabBarViewSnapshotTests.m
```diff
@@ -48,7 +48,7 @@ - (void)setUp {
 
   // Uncomment below to recreate all the goldens (or add the following line to the specific
   // test you wish to recreate the golden for).
-  //  self.recordMode = YES;
+  self.recordMode = YES;
```
mikefan1991: This should stay commended.

mikefan1991 (changes requested):

romoore:
Gonna take a stab at cleaning this up. Ping me if you plan to work on it as well.

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -30,10 +32,10 @@ @implementation MDCTabBarView
 #pragma mark - Initialization
 
 - (instancetype)init {
-  self = [super init];
+  self = [super initWithFrame:CGRectZero];
   if (self) {
     _items = @[];
-    _itemViews = @[];
+    [self setUpSubviews];
```
romoore: This looks far too complex to be called from `-init`.  Let's try to simplify and add the constraints later.

romoore (changes requested):

components/Tabs/tests/snapshot/TabBarView/MDCTabBarViewSnapshotTests.m
romoore: Done.

components/Tabs/src/TabBarView/MDCTabBarView.m

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: Reverted.

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: Removed.

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -20,8 +20,8 @@
 
 @interface MDCTabBarView ()
 
-/** The views representing the items of this tab bar. */
-@property(nonatomic, strong) NSArray<UIView *> *itemViews;
+/** The stack view that contains all tab items. */
+@property(nonatomic) UIStackView *stackView;
```
romoore: Nit: Please document retention.

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -93,28 +78,62 @@ - (void)setSelectedItem:(UITabBarItem *)selectedItem {
   _selectedItem = selectedItem;
 }
 
-#pragma mark - UIView
+- (void)setUpSubviews {
```
romoore: Please rename this to `commonMDCTabBarViewInit`.

romoore (changes requested):

components/Tabs/tests/snapshot/TabBarView/MDCTabBarViewSnapshotTests.m
```diff
@@ -137,4 +137,63 @@ - (void)testItemsWithMixedTitlesAndImages {
   [self generateSnapshotAndVerifyForView:self.tabBarView];
 }
 
+- (void)testMultipleLineItemViewToFitInTheSingleScreen {
```
romoore: Please add a similar test that uses the bar's `intrinsicContentSize` to set the bounds so we can be sure it will return the correct size to fit the items.

components/Tabs/tests/snapshot/TabBarView/MDCTabBarViewSnapshotTests.m
```diff
@@ -137,4 +137,63 @@ - (void)testItemsWithMixedTitlesAndImages {
...
+  UITabBarItem *item6 = [[UITabBarItem alloc] initWithTitle:@"Six" image:nil tag:8];
+
+  // When
+  self.tabBarView.items = @[ item1, item2, item3, item4, item5, item6 ];
+
+  // Then
+  [self generateSnapshotAndVerifyForView:self.tabBarView];
+}
+
+- (void)testTooManyItemsInScrollView {
```
romoore: Nit: Naming: Please remove "InScollView" from the test since the behavior is automatic and depends on bounds sizes, etc.  Consider the name `testItemsLargerThanBoundsChangesToScrollableLayout`?

components/Tabs/tests/snapshot/TabBarView/MDCTabBarViewSnapshotTests.m
```diff
@@ -137,4 +137,63 @@ - (void)testItemsWithMixedTitlesAndImages {
...
+  UITabBarItem *item8 = [[UITabBarItem alloc] initWithTitle:@"Eight" image:nil tag:8];
+
+  // When
+  self.tabBarView.items = @[ item1, item2, item3, item4, item5, item6, item7, item8 ];
+
+  // Then
+  [self generateSnapshotAndVerifyForView:self.tabBarView];
+}
+
+- (void)testSuperLongItemViewInScrollView {
```
romoore: A few things here: 
*  Please make the bounds larger than the single item's maximum width (I believe it's 360)
*  Please remove "InScrollView" from the name, since the layout depends on content + bounds and isn't guaranteed.  Perhaps `testVeryLongItemLimitsWidth` or something similar?

components/Tabs/src/TabBarView/MDCTabBarView.m
leonmz: Not sure if this change resolve this issue..

components/Tabs/tests/snapshot/TabBarView/MDCTabBarViewSnapshotTests.m
leonmz: Added.

components/Tabs/src/TabBarView/MDCTabBarView.m
leonmz: Fixed.

components/Tabs/src/TabBarView/MDCTabBarView.m
leonmz: Fix applied.

components/Tabs/tests/snapshot/TabBarView/MDCTabBarViewSnapshotTests.m
leonmz: Thanks for the advising!

components/Tabs/tests/snapshot/TabBarView/MDCTabBarViewSnapshotTests.m
leonmz: Added :)

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: I don't see a difference. Perhaps you need to push? I think it's also better to name this for its semantic use (rather than its type).  I think something like this would be better:

```objc
@property(nonnull, nonatomic, strong) UIStackView *containerView;
```

leonmz: Fix applied!

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -99,32 +87,88 @@ - (void)setSelectedItem:(UITabBarItem *)selectedItem {
   _selectedItem = selectedItem;
 }
 
-#pragma mark - UIView
```
mikefan1991: #pragma mark - Private API should be added for the following functions.

leonmz: Fixed.

mikefan1991 (approved):

leonmz:
I refactored the colorFortState function a bit. Deleted the items view and use the arranged view from containerView instead. Please take a look at it. Thanks!

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -20,8 +20,11 @@
 
 @interface MDCTabBarView ()
 
-/** The views representing the items of this tab bar. */
-@property(nonnull, nonatomic, strong) NSArray<UIView *> *itemViews;
+/** The stack view that contains all tab item views. */
+@property(nonnull, nonatomic, strong) UIStackView *containerView;
+
+/** The constraints managing this view. */
+@property(nonatomic, strong) NSArray *viewConstraints;
```
romoore: Please add a nullability annotation to this property. Please also use generics to document that it contains `NSLayoutConstraint *` types.

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -142,30 +129,91 @@ - (UIColor *)titleColorForState:(UIControlState)state {
...
+  NSMutableArray *constraints = [NSMutableArray array];
+  [constraints addObjectsFromArray:@[
+    [_containerView.leadingAnchor constraintEqualToAnchor:self.leadingAnchor],
+    [_containerView.trailingAnchor constraintEqualToAnchor:self.trailingAnchor],
+    [_containerView.widthAnchor constraintGreaterThanOrEqualToAnchor:self.widthAnchor],
+    [_containerView.topAnchor constraintEqualToAnchor:self.topAnchor],
+    [_containerView.bottomAnchor constraintEqualToAnchor:self.bottomAnchor],
+  ]];
+
+  [self addConstraints:constraints];
```
romoore: Please don't call `addConstraints` since it will be deprecated. Instead, activate the constraints directly.

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -142,30 +129,91 @@ - (UIColor *)titleColorForState:(UIControlState)state {
...
+    CGSize contentSize = itemView.intrinsicContentSize;
+    if (contentSize.width > maxWidth) {
+      maxWidth = contentSize.width;
+    }
   }
-  CGSize boundsSize = CGSizeMake(CGRectGetWidth(self.bounds), CGRectGetHeight(self.bounds));
-  CGSize itemSize = CGSizeMake(boundsSize.width / self.items.count, boundsSize.height);
+  CGFloat requiredWidth = maxWidth * self.items.count;
+  BOOL canBeJustified = availableWidth > requiredWidth;
+  self.containerView.distribution = canBeJustified ? UIStackViewDistributionFillEqually
```
romoore, line 275: It seems like `layoutSubviews` shouldn't be required if a constraint-based layout is used. Even then, I would expect that this code execute before `[super layoutSubviews]` since that is expected to use any constraints to layout the subviews.

I suspect a better place to configure `self.containerView.distribution` might be in `setItems` and in `setBounds`.

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -142,30 +129,91 @@ - (UIColor *)titleColorForState:(UIControlState)state {
...
+  _containerView = [[UIStackView alloc] init];
+  _containerView.axis = UILayoutConstraintAxisHorizontal;
+  _containerView.translatesAutoresizingMaskIntoConstraints = NO;
+  [self addSubview:_containerView];
+}
+
+- (void)setUpItemViews {
+  for (UIView *view in self.containerView.arrangedSubviews) {
+    [view removeFromSuperview];
+    [_containerView removeArrangedSubview:view];
```
romoore: This isn't necessary since `removeFromSuperview` automatically removes it from the stack view's `arrangedSubviews` property.

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -142,30 +129,91 @@ - (UIColor *)titleColorForState:(UIControlState)state {
...
 #pragma mark - UIView
 
-// TODO(#7645): Temporary layout until
-// https://github.com/material-components/material-components-ios/issues/7645 lands
 - (void)layoutSubviews {
   [super layoutSubviews];
 
-  if (self.items.count == 0) {
-    return;
+  CGFloat availableWidth = self.bounds.size.width;
```
romoore: Nit: Please use `CGRectGetWidth(self.bounds)` instead of accessing the bounds directly. Especially for UIScrollViews, the bounds origin may not be (0, 0).

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -142,30 +129,91 @@ - (UIColor *)titleColorForState:(UIControlState)state {
...
+  CGFloat maxWidth = 0;
+  for (UIView *itemView in self.containerView.arrangedSubviews) {
+    CGSize contentSize = itemView.intrinsicContentSize;
+    if (contentSize.width > maxWidth) {
+      maxWidth = contentSize.width;
+    }
   }
-  CGSize boundsSize = CGSizeMake(CGRectGetWidth(self.bounds), CGRectGetHeight(self.bounds));
-  CGSize itemSize = CGSizeMake(boundsSize.width / self.items.count, boundsSize.height);
+  CGFloat requiredWidth = maxWidth * self.items.count;
```
romoore: Looks like this can be replaced with a call to `self.intrinsicContentSize`?

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -142,30 +129,91 @@ - (UIColor *)titleColorForState:(UIControlState)state {
...
+      maxHeight = contentSize.height;
+    }
+  }
+  return CGSizeMake(totalWidth, MAX(kMinHeight, maxHeight));
 }
 
 - (CGSize)sizeThatFits:(CGSize)size {
-  return CGSizeMake(size.width, MAX(size.height, kMinHeight));
+  CGSize intrinsicSize = self.intrinsicContentSize;
+  return CGSizeMake(MIN(intrinsicSize.width, size.width), MAX(intrinsicSize.height, size.height));
```
romoore: I'm curious about this logic. If the current bounds is `CGRectZero` and `sizeToFit` is called, then this view would only resize its height. Is that expected?  Should it not attempt to provide a fit size for its contents as a minimum, but allow a larger width?

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -142,30 +129,91 @@ - (UIColor *)titleColorForState:(UIControlState)state {
...
+}
+
+#pragma mark - Private Functions
+
+- (void)commonMDCTabBarViewInit {
```
romoore: Nit: Please move this up near the `-init` methods.

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -142,30 +129,91 @@ - (UIColor *)titleColorForState:(UIControlState)state {
...
+  self.backgroundColor = UIColor.whiteColor;
+
+  [self setUpStackView];
+  [self setUpItemViews];
```
romoore: This call doesn't seem necessary.  I believe the `items` array is empty during initialization.

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -142,30 +129,91 @@ - (UIColor *)titleColorForState:(UIControlState)state {
...
+}
+
+- (void)setUpStackView {
+  _containerView = [[UIStackView alloc] init];
+  _containerView.axis = UILayoutConstraintAxisHorizontal;
+  _containerView.translatesAutoresizingMaskIntoConstraints = NO;
+  [self addSubview:_containerView];
+}
+
+- (void)setUpItemViews {
```
romoore: I believe this logic can be put back in `setItems:` since it's only called from there.

romoore (changes requested):
I'm going to address most of my feedback and push up changes to the branch.

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: Ah, no, that returns the scrollable width, not the justified width.

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: I'll leave this here for now. `setBounds:` is called more times during the initial presentation of a view and not much less often during rotation. It's likely not worth the optimization.

components/Tabs/src/TabBarView/MDCTabBarView.m
leonmz: I still feel it worth a function extract the logic..

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: Done. I was able to remove this property and replace it with a flag. #namingIsHard

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: Done

components/Tabs/src/TabBarView/MDCTabBarView.m

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: Done. Moved into `-init`

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: Removed.

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: Could you help me understand your thinking? We cannot unit test this code and it's only ever called from within `setItems:`. Either testability or duplicate code would be good candidates for a method. In this case I'm trying to understand the value of an extra message send (plus complexity in figuring out what the method does).

components/Tabs/src/TabBarView/MDCTabBarView.m
leonmz: I am thinking about since it is a scrollable view thus this should be fits in any width.

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -142,30 +146,56 @@ - (UIColor *)titleColorForState:(UIControlState)state {
...
+    [super updateConstraints];
+    return;
   }
+
+  [_containerView.leadingAnchor constraintEqualToAnchor:self.leadingAnchor].active = YES;
+  [_containerView.trailingAnchor constraintEqualToAnchor:self.trailingAnchor].active = YES;
+  [_containerView.widthAnchor constraintGreaterThanOrEqualToAnchor:self.widthAnchor].active = YES;
+  [_containerView.topAnchor constraintEqualToAnchor:self.topAnchor].active = YES;
+  [_containerView.bottomAnchor constraintEqualToAnchor:self.bottomAnchor].active = YES;
+  self.containerViewConstraintsActive = YES;
```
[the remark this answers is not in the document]
romoore: Done?

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: Removed.

components/Tabs/src/TabBarView/MDCTabBarView.m
leonmz: CGZero is a extremely case, but if a small bounds is provided, the scrollable view should also fit itself in

components/Tabs/src/TabBarView/MDCTabBarView.m
leonmz: Thanks  a lot for detailed explanation!!

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: That's true, but the fit size would (ideally) be the size of its content. I think the best thing to conform to the documented behavior of `sizeThatFits:` is to return the "fit size".  In this case, it's probably the the size of the justified layout (rather than the intrinsic content size).  Does that make sense?

Whether it's kept like this or changed, I think it's worth adding a unit test that calls `sizeToFit:` with a bounds of `CGRectZero` and make sure it gets back a non-zero width and non-zero height.  Something similar for a starting bounds bigger than necessary?

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -142,30 +147,56 @@ - (UIColor *)titleColorForState:(UIControlState)state {
...
+
+  // Must always be called last according to the documentation.
+  [super updateConstraints];
 }
 
 - (CGSize)intrinsicContentSize {
```
romoore, line 297: Please add unit tests to cover verify the behavior/values of `intrinsicContentSize`.

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -142,30 +147,56 @@ - (UIColor *)titleColorForState:(UIControlState)state {
...
+    CGSize contentSize = itemView.intrinsicContentSize;
+    totalWidth += contentSize.width;
+    if (contentSize.height > maxHeight) {
+      maxHeight = contentSize.height;
+    }
+  }
+  return CGSizeMake(totalWidth, MAX(kMinHeight, maxHeight));
 }
 
 - (CGSize)sizeThatFits:(CGSize)size {
```
romoore, line 309: I also mentioned it in another thread, but please add unit tests for `sizeThatFits:`

romoore (changes requested):
Looks good, but needs unit tests for `intrinsicContentSize` and `sizeThatFits:` before it's ready for merge.

components/Tabs/src/TabBarView/MDCTabBarView.m
leonmz: Added.

components/Tabs/src/TabBarView/MDCTabBarView.m
leonmz: Added.

components/Tabs/src/TabBarView/MDCTabBarView.m
leonmz: Sure. Fixed applied. I also refactored the intrisicSize to justified width instead of scrollable width.

leonmz:
Refactored the intrsicSize to justified size. Also added snapshot tests for case that the bounds is 1000 but only two items exist. The current behavior is that the item view has been stretched to 500 due to its low hugging priority.

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: No, I don't think the TabBarView should be dependent on its contents.  I'd prefer the tests cover the more general behavior of `intrinsicContentSize`. For example, ensuring that it returns the justified/scrollable width consistently (depending on what you determine is more useful). That the height grows if one of its items has a very large image or long text.  It's the relationship/behavior that's important and not the exact values.

Let me add one or two as a commit and we can talk about them.  I think that will be faster and clearer than my trying to explain in a comment. ;-)

romoore (approved):

romoore:
@googlebot All authors are Googlers, but one commit has a misconfigured author email.

Issue #7657, opened 2019-06-20
[Tabs] Internal issue: b/135637901


material-mirror:
This was filed as an internal issue. If you are a Googler, please visit [b/135637901](http://b/135637901) for more details.

---
#### Internal data
- Associated internal bug: [b/135637901](http://b/135637901)

PR #7712, opened 2019-06-25
[Tabs] Add titleColorForState: API.

romoore:
Adds a new API to affect the items' title color for `.selected` and `.normal` states.

Part of #7657


components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -21,7 +21,10 @@
 @interface MDCTabBarView ()
 
 /** The views representing the items of this tab bar. */
-@property(nonatomic, strong) NSArray<UIView *> *itemViews;
+@property(nonnull, nonatomic, strong) NSArray<UIView *> *itemViews;
+
+/** The title colors for bar items. */
+@property(nonnull, nonatomic, strong) NSMutableDictionary<NSNumber *, UIColor *> *titleColors;
```
bryanoltman: It was not immediately clear to me that this dictionary is mapping `UIControlState`s to `UIColor`s. Please update either the comment or the property name to reflect this. I'm not sure what our naming conventions are for dictionaries, but something like `statesToTitleColors` would be a lot clearer.

components/Tabs/tests/unit/TabBarView/MDCTabBarViewTests.m
```diff
@@ -163,4 +163,59 @@ - (void)testSettingBackgroundColorUpdatesBarTintColor {
...
+}
+
+- (void)testTitleColorForStateWithNoValuesReturnsNil {
+  // When
+  [self.tabBarView setTitleColor:nil forState:UIControlStateNormal];
+  [self.tabBarView setTitleColor:nil forState:UIControlStateSelected];
+
+  // Then
+  XCTAssertNil([self.tabBarView titleColorForState:UIControlStateNormal]);
+  XCTAssertNil([self.tabBarView titleColorForState:UIControlStateSelected]);
```
bryanoltman, line 218: What is the expected behavior if a title color is set for a non-normal control state, but the title color for `UIControlStateNormal` is `nil`?

bryanoltman (changes requested):

components/Tabs/tests/unit/TabBarView/MDCTabBarViewTests.m
romoore: "A default color is ~returned~ used." The _returned_ value will be `nil`.

components/Tabs/src/TabBarView/MDCTabBarView.m

bryanoltman (approved):

PR #7727, opened 2019-06-27
[TabBarView] Add accessibility label

codeman7:
This adds the `accessibilityLabel` to MDCTabBarView, this falls back to the titleLabel.text if the accessibility label isn't set. This has been tested on a device.

## Testing
Steps to reproduce
1. Update the [example](https://github.com/material-components/material-components-ios/blob/develop/components/Tabs/examples/MDCTabBarViewTypicalExampleViewController.m) to have a custom `accessibilityLabel` for one of the tabBarItems
```objc
item1.accessibilityLabel = @"Hello world";
```
2. Select the item with Voiceover on
3. Navigate to the other items with titles but no custom value.

### DIFF
--- a/components/Tabs/src/TabBarView/private/MDCTabBarViewItemView.m
+++ b/components/Tabs/src/TabBarView/private/MDCTabBarViewItemView.m
@@ -35,7 +35,9 @@ static const UIEdgeInsets kEdgeInsets = {.top = 12, .right = 16, .bottom = 12, .
 
 @end
 
-@implementation MDCTabBarViewItemView
+@implementation MDCTabBarViewItemView {
+  NSString *_accessibilityLabel;
+}
 
 #pragma mark - Init
 
@@ -130,6 +132,16 @@ static const UIEdgeInsets kEdgeInsets = {.top = 12, .right = 16, .bottom = 12, .
   CGFloat height = (CGFloat)ceil(iconSize.height + labelSize.height + verticalPadding);
   height = MAX(minHeight, height);
   return CGSizeMake(width, height);
+}
+
+#pragma mark - Accessibility
+
+- (NSString *)accessibilityLabel {
+  return _accessibilityLabel ?: self.titleLabel.text;
+}
+
+- (void)setAccessibilityLabel:(NSString *)accessibilityLabel {
+  _accessibilityLabel = accessibilityLabel;
 }
 
 @end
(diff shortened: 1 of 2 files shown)


components/Tabs/src/TabBarView/private/MDCTabBarViewItemView.m
```diff
@@ -132,4 +134,14 @@ - (CGSize)sizeThatFits:(CGSize)size {
...
 }
 
+#pragma mark - Accessibility
+
+- (NSString *)accessibilityLabel {
+  return _accessibilityLabel ?: self.titleLabel.text;
+}
+
+- (void)setAccessibilityLabel:(NSString *)accessibilityLabel {
+  _accessibilityLabel = accessibilityLabel;
```
romoore: I believe this setter is not necessary, since this property is settable on UIView by default.

components/Tabs/src/TabBarView/private/MDCTabBarViewItemView.m
```diff
@@ -132,4 +134,14 @@ - (CGSize)sizeThatFits:(CGSize)size {
   return CGSizeMake(width, height);
 }
 
+#pragma mark - Accessibility
+
+- (NSString *)accessibilityLabel {
+  return _accessibilityLabel ?: self.titleLabel.text;
```
romoore: I think this would work.  What do you think?

```objc
  return [super accessibilityLabel] ?: self.titleLabel.text;
```

components/Tabs/src/TabBarView/private/MDCTabBarViewItemView.m
```diff
@@ -132,4 +134,14 @@ - (CGSize)sizeThatFits:(CGSize)size {
   return CGSizeMake(width, height);
 }
 
+#pragma mark - Accessibility
```
romoore: Nit: "UIAccessibility" instead of "Accessibility"

romoore (approved):

components/Tabs/src/TabBarView/private/MDCTabBarViewItemView.m
codeman7: Updated.

components/Tabs/src/TabBarView/private/MDCTabBarViewItemView.m
codeman7: Done

components/Tabs/src/TabBarView/private/MDCTabBarViewItemView.m
codeman7: done

PR #7729, opened 2019-06-27
[TabBarView] Add accessibility hint and identifier

codeman7:
This adds the `accessibilityHint` and `accessibilityIdentifer` to MDCTabBarView. This has been tested on a device and simulator.

### `accessibilityHint`
Steps to reproduce
1. Update the [example](https://github.com/material-components/material-components-ios/blob/develop/components/Tabs/examples/MDCTabBarViewTypicalExampleViewController.m) to have a custom `accessibilityHint` for one of the tabBarItems
```objc
item1.accessibilityHint = @"Hello world";
```
2. Launch on a device
3. Select the item with Voiceover on
4. Navigate to the other items with titles but no custom value.

### `accessibilityHint`
Steps to reproduce
1. Update the [example](https://github.com/material-components/material-components-ios/blob/develop/components/Tabs/examples/MDCTabBarViewTypicalExampleViewController.m) to have a custom `accessibilityIdentifier` for one of the tabBarItems
```objc
item1.accessibilityIdentifier = @"Goodbye";
```
2. Launch MDCDragons on a simulator
3. Open up the accessibility inspector to make sure the value is set correctly

### DIFF
--- a/components/Tabs/src/TabBarView/private/MDCTabBarViewItemView.m
+++ b/components/Tabs/src/TabBarView/private/MDCTabBarViewItemView.m
@@ -132,14 +132,4 @@ static const UIEdgeInsets kEdgeInsets = {.top = 12, .right = 16, .bottom = 12, .
   return CGSizeMake(width, height);
 }
 
-#pragma mark - UIAccessibility
-
-- (NSString *)accessibilityHint {
-  return [super accessibilityLabel];
-}
-
-- (NSString *)accessibilityIdentifier {
-  return [super accessibilityIdentifier];
-}
-
 @end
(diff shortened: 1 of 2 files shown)


components/Tabs/src/TabBarView/private/MDCTabBarViewItemView.m
```diff
@@ -132,4 +132,14 @@ - (CGSize)sizeThatFits:(CGSize)size {
   return CGSizeMake(width, height);
 }
 
+#pragma mark - UIAccessibility
+
+- (NSString *)accessibilityHint {
+  return [super accessibilityLabel];
```
romoore: Looks like we don't need these getters.

romoore (approved):

PR #7736, opened 2019-06-27
[Tabs]Adds EdgeInsets

leonmz:
Adds custom edge insets for MDCTabbarView.
The scrollable view will now have 52 padding at the leading edge.
Simulator Screen Shot - iPhone 7 - 2019-06-27 at 16 57 49


components/Tabs/src/TabBarView/MDCTabBarView.h
```diff
@@ -27,6 +27,9 @@ __attribute__((objc_subclassing_restricted)) @interface MDCTabBarView : UIScroll
 /** The currently-selected item in the Tab bar. */
 @property(nullable, nonatomic, strong) UITabBarItem *selectedItem;
 
+/** The content edge insets of the UITabBar. */
+@property(nonatomic) UIEdgeInsets edgeInsets;
```
romoore: Please name `contentEdgeInsets` and have them flip for RTL.  It will prevent clients from having to do the flip each time.

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -311,8 +314,15 @@ - (void)layoutSubviews {
   CGFloat availableWidth = CGRectGetWidth(self.bounds);
   CGFloat requiredWidth = [self justifiedWidth];
   BOOL canBeJustified = availableWidth >= requiredWidth;
-  self.containerView.distribution = canBeJustified ? UIStackViewDistributionFillEqually
-                                                   : UIStackViewDistributionFillProportionally;
+  if (canBeJustified) {
```
romoore, line 353: Please flip for RTL here based on `effectiveUserInterfaceLayoutDirection`

romoore (changes requested):
Looks good overall. Please add snapshot tests (feel free to skip RTL tests for the moment).

components/Tabs/examples/MDCTabBarViewTypicalExampleViewController.m
```diff
@@ -84,6 +84,7 @@ - (void)viewDidLoad {
                         forState:UIControlStateSelected];
   self.tabBar.selectedItem = item1;
   self.tabBar.translatesAutoresizingMaskIntoConstraints = NO;
+  self.tabBar.edgeInsets = UIEdgeInsetsMake(0, 120, 0, 20);
```
romoore: Please pare these down to something a bit smaller. Perhaps `{0, 10, 0, 10}`?

romoore (changes requested):

components/Tabs/src/TabBarView/MDCTabBarView.h
leonmz: Done.

components/Tabs/src/TabBarView/MDCTabBarView.m
leonmz: Done

components/Tabs/examples/MDCTabBarViewTypicalExampleViewController.m
leonmz: Removed.

leonmz:
Hey Rob, I am not sure if the snapshot tests were enough.. Could you please add more according to your knowledge? Thanks a lot!

romoore (changes requested):
Taking a look at this. Please don't push commits without reaching out to me first. :)

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -153,6 +162,17 @@ - (void)setSelectedItem:(UITabBarItem *)selectedItem {
   [self updateImageTintColorForAllViews];
 }
 
+- (void)setContentEdgeInsets:(UIEdgeInsets)contentEdgeInsets {
+  UIEdgeInsets adjustedInset = contentEdgeInsets;
+  if (self.mdf_effectiveUserInterfaceLayoutDirection == UIUserInterfaceLayoutDirectionRightToLeft) {
```
romoore: I think this flipping should be done in an internal getter. Otherwise you have to respond to changes in the semantic content attribute and keep track of the original + flipped versions.

components/Tabs/tests/snapshot/TabBarView/MDCTabBarViewSnapshotTests.m
```diff
@@ -103,6 +103,67 @@ - (void)testItemsWithOnlyTitles {
   [self generateSnapshotAndVerifyForView:self.tabBarView];
 }
 
+- (void)testItemsWithOnlyTitlesWithInsets {
```
romoore: Needs tests for safe area insets

romoore (changes requested):
Looks to be currently broken. The UIStackView is being repositioned but is not switching to a Scrollable layout style.  This results in Justified tabs that can scroll.

Here's a screenshot from an iPhone XR in landscape being debugged in Reveal. It shows that the UIStackView is being positioned 44 points (the safe area inset) from the leading edge but the content didn't get switched to a scrollable configuration.

Screen Shot 2019-06-28 at 12 10 46 AM

romoore:
Got it working by adjusting the `widthConstraint` for safe area.

|Before (`self.widthAnchor`)|After (`self.safeAreaLayoutGuide.widthAnchor`)|
|---|---|
|Simulator Screen Shot - iPhone XR - 2019-06-28 at 00 11 56|Simulator Screen Shot - iPhone XR - 2019-06-28 at 00 27 25|

romoore (approved):
@leonmz All set if you want to make changes. Looks good to me in general!

romoore (changes requested):
With the recent changes for scrolling to the selected item, this PR is not ready. There are bugs when first presenting the view if the safeAreaInsets are non-zero.

romoore (changes requested):
Let's remove the `contentEdgeInsets` and focus only on the built-in padding in this PR.  We can do a second PR to make fixes for the safe area.

leonmz:
This PR was too heavy. Split into small different PRs.

PR #7739, opened 2019-06-27
[Tabs] Add accessibility traits support

mikefan1991:
Add UIAccessibility traits to item view.

**Testing**
accessibilityTraits
Steps to reproduce

1. Update the example to have a custom accessibilityTraits for one of the tabBarItems
    1. Optionally have it change those traits in the `tabBarView:didSelectItem:` delegate method.
2. item1.accessibilityTraits = @"Hello world";
3. Launch on a device
4. Select the item with Voiceover on
5. Navigate to the other items with titles but no custom value.


components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -103,6 +104,7 @@ - (void)setItems:(NSArray<UITabBarItem *> *)items {
     itemView.accessibilityLabel = item.accessibilityLabel;
     itemView.accessibilityHint = item.accessibilityHint;
     itemView.accessibilityIdentifier = item.accessibilityIdentifier;
+    itemView.accessibilityTraits = item.accessibilityTraits;
```
romoore: I think this should default to `.button`:

```objc
itemView.accessibilityTraits = (item.accessibilityTraits == UIAccessibilityTraitNone) : UIAccessibilityTraitButton : item.accessibilityTraits;
```

romoore (changes requested):
Looks like it needs to be updated to toggle `UIAccessibilityTraitSelected` trait during selection/unselection.  Should also be sure to default to `.button` either in the view or during assignment of the initial traits.

components/Tabs/src/TabBarView/MDCTabBarView.m
mikefan1991: Done.

leonmz (commented):
I thought we can add a unit tests to verify the behavior of the accessibility traits.

mikefan1991:
The unit test is not feasible because the item views are private in the tab bar view.

romoore:
The private nature is not the problem.  UIAccessibility does not function correctly outside of an application + view hierarchy. The PR should list manual testing steps that are taken (and can be reproduced) to verify the behavior. 

https://github.com/material-components/material-components-ios/pull/7729 is a good example of how to write manual testing for these attributes.

romoore (approved):

PR #7745, opened 2019-06-28
[Tabs] Add Selection Indicator support.

### LINKED ISSUE #7744: [Tabs] Internal issue: b/135609974
This was filed as an internal issue. If you are a Googler, please visit [b/135609974](http://b/135609974) for more details.

---
#### Internal data
- Associated internal bug: [b/135609974](http://b/135609974)
- Blocked by: https://github.com/material-components/material-components-ios/issues/7644
- Blocked by: https://github.com/material-components/material-components-ios/issues/7643

romoore:
Adds basic support for a selection indicator. Uses simple animation for showing/hiding the indicator as well as repositioning it when the `selectedItem` changes.

tabbar-selector-portrait

Closes #7744


components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -352,18 +395,30 @@ - (void)observeValueForKeyPath:(NSString *)keyPath
...
+  CGFloat availableWidth = CGRectGetWidth(self.bounds);
+  CGFloat requiredWidth = [self justifiedWidth];
+  UIStackViewDistribution distribution = availableWidth >= requiredWidth
+                                             ? UIStackViewDistributionFillEqually
+                                             : UIStackViewDistributionFillProportionally;
+  if (self.containerView.distribution != distribution) {
+    self.containerView.distribution = distribution;
+    [self.containerView setNeedsLayout];
+  }
+  [self.containerView layoutIfNeeded];
```
mikefan1991: Is this really needed? I remember it's not good to call layoutIfNeeded in layoutSubviews.

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -374,8 +429,12 @@ - (void)setBounds:(CGRect)bounds {
       !CGSizeEqualToSize(CGSizeMake(CGRectGetWidth(self.bounds), CGRectGetHeight(self.bounds)),
                          CGSizeMake(CGRectGetWidth(bounds), CGRectGetHeight(bounds)));
   [super setBounds:bounds];
+  [self updateStackviewLayout];
   if (shouldScroll) {
     [self scrollUntilSelectedItemIsVisibleWithoutAnimation];
+    if (self.selectedItem) {
+      [self updateSelectionIndicatorToIndex:[self.items indexOfObject:self.selectedItem]];
```
mikefan1991: Do we also want to call this if shouldScroll is false?

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -457,6 +547,95 @@ - (CGSize)expectedSizeForView:(UIView *)view {
   return expectedItemSize;
 }
 
+/// Sets _selectionIndicator's bounds and center to display under the item at the given index with
+/// no animation. May be called from an animation block to animate the transition.
+- (void)updateSelectionIndicatorToIndex:(NSUInteger)index {
+  if (index == NSNotFound || index >= self.items.count) {
+    // Hide selection indicator.
+    self.selectionIndicatorView.bounds = CGRectZero;
```
mikefan1991: Maybe use the hidden property to hide the indicator view?

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -457,6 +547,95 @@ - (CGSize)expectedSizeForView:(UIView *)view {
...
+  // Construct a context object describing the selected tab.
+  UITabBarItem *item = self.items[index];
+  MDCTabBarViewPrivateIndicatorContext *context =
+      [[MDCTabBarViewPrivateIndicatorContext alloc] initWithItem:item
+                                                          bounds:selectionIndicatorBounds
+                                                    contentFrame:contentFrame];
+
+  // Ask the template for attributes.
+  id<MDCTabBarViewIndicatorTemplate> template = self.selectionIndicatorTemplate;
+  MDCTabBarViewIndicatorAttributes *indicatorAttributes =
```
mikefan1991: Why do we need an attributes when we already have the context? It seems increasing the complication.

romoore: I agree that in the future it may be an opportunity to reduce the number of classes used to handle the indicator.  However, it hasn't caused problems so far. I'd rather keep what works (in this case) rather than try to refactor/rewrite it. Given our time limitations, it's probably not the best use of our effort.

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: This is the current logic - I believe it allows animation of the indicator as a "shrinking" effect rather than a fade-out. In either case, it hasn't caused bugs to be filed yet, so I'd prefer to keep it as-is and [the selection indicator animation/layout code can be cleaned-up in the future](https://github.com/material-components/material-components-ios/issues/7748).

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: I will double-check, but I believe that causes problems with scrolling. Scrolling changes the bounds (the `origin`, not the `size`) and I thought animating the selection indicator for each event caused scrolling to break.

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: It's not the best idea, but seems not to cause any errors during testing. Unless we switch to fully manual layout, I'm not sure how else I can get the stack view to reposition its subviews.  I know it's possible to "estimate" the position of those views, but I prefer not to call that code if we can avoid it.

Let me see what breaks when I remove this and I will post back here.

romoore:
Fails to correctly scroll to the indicator after a rotation.

components/Tabs/src/TabBarView/MDCTabBarView.m
leonmz: According to apple, this method will call LayoutSubviews which could create a calling cycle. The reason that the tests not broken probably because the 

> If no layout updates are pending, this method exits without modifying the layout or calling any layout-related callbacks.

leonmz (approved):

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: Calling it from the same view is definitely bad. Calling it from another view is not great.  However, it's not returning while doing nothing - it fixes the layout bugs when I test on the simulator.

I definitely need a hand figuring-out how to get the indicator to appear at the correct location. Forcing a subview to layout isn't the worse option, but it's far from ideal.

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: I can't move it out of that check since it breaks layout entirely.

romoore:
@jverkoey I could really use some help on this one if you have time this morning.

codeman7 (approved):

PR #7749, opened 2019-06-28
[Tabs] Allow setting selected item without animation.

romoore:
The property setter will automatically animate setting the selected item. For
some use cases, like tests or configuring a view for the first time, it's
useful to set the selected item without animation.

Part of #7744

### DIFF
--- a/components/Tabs/src/TabBarView/MDCTabBarView.m
+++ b/components/Tabs/src/TabBarView/MDCTabBarView.m
@@ -16,6 +16,8 @@
 #import "MDCTabBarViewDelegate.h"
 #import "private/MDCTabBarViewItemView.h"
 
+#import <CoreGraphics/CoreGraphics.h>
+
 // KVO contexts
 static char *const kKVOContextMDCTabBarView = "kKVOContextMDCTabBarView";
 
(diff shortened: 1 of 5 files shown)


components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -16,6 +16,8 @@
 #import "MDCTabBarViewDelegate.h"
 #import "private/MDCTabBarViewItemView.h"
 
+#import <CoreGraphics/CoreGraphics.h>
```
bryanoltman, line 19: Why do we need to import CoreGraphics here?

bryanoltman (approved):

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: for CGRect things.

bryanoltman: Which CGRect things? I'm not seeing any CGRect manipulation in this PR

romoore: For example, `layoutSubviews` uses `CGFloat` and `CGRectGetWidth`. This PR did not add those calls but I noticed the missing import statement while writing this one.

PR #7751, opened 2019-06-28
[Tabs] Fork default selection indicator template.

romoore:
Adds the default (underline) selection indicator from Material.io. This is the same one as `MDCTabBar` and has been copied and renamed.

Part of #7744

### DIFF
--- a/components/Tabs/src/TabBarView/MDCTabBarViewUnderlineIndicatorTemplate.h
+++ b/components/Tabs/src/TabBarView/MDCTabBarViewUnderlineIndicatorTemplate.h
@@ -1,0 +1,23 @@
+// Copyright 2017-present the Material Components for iOS authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


components/Tabs/src/TabBarView/MDCTabBarViewUnderlineIndicatorTemplate.h
```diff
@@ -0,0 +1,23 @@
+// Copyright 2017-present the Material Components for iOS authors. All Rights Reserved.
```
codeman7, line 1: Please fix year

components/Tabs/src/TabBarView/MDCTabBarViewUnderlineIndicatorTemplate.m
```diff
@@ -0,0 +1,36 @@
+// Copyright 2017-present the Material Components for iOS authors. All Rights Reserved.
```
codeman7, line 1: Same

codeman7 (approved):
Should we tests?

components/Tabs/src/TabBarView/MDCTabBarViewUnderlineIndicatorTemplate.h
romoore: This was a copy-paste, so @jverkoey says don't change copyright.

components/Tabs/src/TabBarView/MDCTabBarViewUnderlineIndicatorTemplate.m
romoore: Same reply. it's a fork.

romoore:
Tests I would love, I'll be handling that when I integrate it with snapshots.

Issue #7731, opened 2019-06-27
[Tabs] Internal issue: b/136175682


material-mirror:
This was filed as an internal issue. If you are a Googler, please visit [b/136175682](http://b/136175682) for more details.

---
#### Internal data
- Associated internal bug: [b/136175682](http://b/136175682)

PR #7754, opened 2019-06-29
[Tabs] Fix scroll-to-item for RTL. 

romoore:
Corrects the calculation of item position in RTL layouts.

Part of #7731


components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -431,23 +432,53 @@ - (CGFloat)justifiedWidth {
...
+  }
+  CGRect itemFrameInContainerView = CGRectMake(
+      viewOriginX, CGRectGetMinY(self.containerView.bounds), viewSize.width, viewSize.height);
+  CGRect convertedRect = [self convertRect:itemFrameInContainerView fromView:self.containerView];
+  return convertedRect;
 }
 
 - (CGSize)expectedSizeForView:(UIView *)view {
+  if (self.containerView.distribution == UIStackViewDistributionFillEqually) {
+    if (CGRectGetWidth(self.containerView.bounds) < 0) {
```
mikefan1991: What is the case that the with can be less than 0?

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -431,23 +432,53 @@ - (CGFloat)justifiedWidth {
 - (void)scrollUntilSelectedItemIsVisibleWithoutAnimation {
   NSUInteger index = [self.items indexOfObject:self.selectedItem];
   if (index == NSNotFound || index >= self.containerView.arrangedSubviews.count) {
+    index = 0;
+  }
+  if (self.containerView.arrangedSubviews.count == 0U) {
```
galiak11, line 531: Nit: Is there a reason why you use 0U here and no format for CGFloats below (in: (CGRectGetWidth(self.containerView.bounds) < 0)...) ?

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -431,23 +432,53 @@ - (CGFloat)justifiedWidth {
...
 }
 
 - (CGSize)expectedSizeForView:(UIView *)view {
+  if (self.containerView.distribution == UIStackViewDistributionFillEqually) {
+    if (CGRectGetWidth(self.containerView.bounds) < 0) {
+      return CGSizeMake(
+          CGRectGetWidth(self.containerView.bounds) / self.containerView.arrangedSubviews.count,
+          CGRectGetHeight(self.containerView.bounds));
+    }
+  }
```
galiak11, line 579: Nit: Should the 2nd part of this method be inside an "else" to make it more clear that there are 2 cases for this method?

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -431,23 +432,53 @@ - (CGFloat)justifiedWidth {
 - (void)scrollUntilSelectedItemIsVisibleWithoutAnimation {
```
galiak11, line 526: How is scrolling with animation handled for RTL?

galiak11 (approved):

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: Ah, that should be `>`. I'll fix it, thanks!

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: `count` is an unsigned value. So unsigned comparison is required.  Below the return type of `CGRectGet...` functions is `CGFloat`.

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: It's no different from scrolling in LTR - it's handled above in `setSelected:animated:`.  The fix in this PR is that the position of the `itemView` for RTL was miscalculated.  Here's scroll-to-selected in RTL. The Catalog has a bug with layoutMargins, so I had to hack the top constraint:

tabbar-rtl-selection-scroll

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: Do you mean you'd like me to combine the two `if` conditions into a single expression?  I think early returns are preferred over indentation. @jverkoey has commented on multiple occasions (and I agree) that unnecessary indentation tends to make the code less readable.  At least it generally does.

romoore: Combined the two conditions into a single. That removed one layer of indentation. I still feel that the early return is easier to read than keeping an `else` block. The `else` requires the reader to mentally check its start when reaching the end.

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: Add a TODO because I couldn't reach this code (with either condition) during simulator testing. However, I'm going to leave it in for now.

PR #7756, opened 2019-06-29
[Tabs] Fix image tint color when unselecting an item.

romoore:
When the selected item became `nil`, the image tint color for the previously-selected item was not updated.

Follow-up to #7714


mikefan1991 (approved):

leonmz (approved):
Thanks for fixing!

codeman7 (approved):

PR #7763, opened 2019-07-01
[Tabs] Fix the height bug of MDCTabBarView

mikefan1991:
If the tab bar view is added as the MDCAppBar's bottom bar, the height should be its intrinsic height. Or the size of the app bar's stack view is calculated incorrectly.


leonmz:
Let's wait for a bit to see how Rob's manual layout work out

PR #7767, opened 2019-07-02
[Tabs] Add Selection Indicator

### LINKED ISSUE #7744: [Tabs] Internal issue: b/135609974

romoore:
Adds selection indicator support and rendering to MDCTabBarView. The indicator animates between items and is sized to the full width of the item. Theming (color) of the indicator will be done in a follow-up change.

|Scrollable without Safe Area Insets|Justified with Safe Area Insets|
|---|---|
|tabs-si-nosafe|tabs-si-safe|

Closes #7744


components/Tabs/src/TabBarView/MDCTabBarView.h
```diff
@@ -36,6 +37,11 @@ __attribute__((objc_subclassing_restricted)) @interface MDCTabBarView : UIScroll
 /** The tab bar view delegate. */
 @property(nullable, nonatomic, weak) id<MDCTabBarViewDelegate> tabBarDelegate;
 
+/**
+ The template for the selection indicator.
+ */
+@property(nonnull, nonatomic, strong) id<MDCTabBarViewIndicatorTemplate> selectionIndicatorTemplate;
```
galiak11, line 44: Nit: Could be easier to understand if had a short explanation of what template means in this context (unless already explained elsewhere?)

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -181,6 +214,7 @@ - (void)setSelectedItem:(UITabBarItem *)selectedItem animated:(BOOL)animated {
     [self updateTitleColorForAllViews];
     [self updateImageTintColorForAllViews];
     [self updateTitleFontForAllViews];
+    [self didSelectItemAtIndex:NSNotFound animateTransition:animated];
```
galiak11, line 206: Nit: What does selecting an NSNotFound item do?

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -603,4 +647,114 @@ - (void)didTapItemView:(UITapGestureRecognizer *)tap {
   }
 }
 
+- (void)applySelectionTemplateToSelectionViewForItemView:(UIView *)itemView {
+  if (!itemView) {
+    return;
```
galiak11, line 641: When would this be called with a nil view? 
Is that a likely scenario? 
Nit - If not, did you consider using a parameter assert?

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -603,4 +647,114 @@ - (void)didTapItemView:(UITapGestureRecognizer *)tap {
...
+  // Ask the template for attributes.
+  id<MDCTabBarViewIndicatorTemplate> template = self.selectionIndicatorTemplate;
+  MDCTabBarViewIndicatorAttributes *indicatorAttributes =
+      [template indicatorAttributesForContext:context];
+
+  // Update the selection indicator.
+  [self.selectionIndicatorView applySelectionIndicatorAttributes:indicatorAttributes];
+}
+
+- (void)didSelectItemAtIndex:(NSUInteger)index animateTransition:(BOOL)animate {
```
galiak11, line 721: Nit - Can you add a comment explaining this method? Specifically what happens when the index is NotFound...

fluxminds (approved):

galiak11 (approved):

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: It makes nothing selected.

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: At the moment, I don't believe it is. It's just a bit of defensive coding to keep the code safer for the future. I don't think it should be an error if the user passes `nil` (the parameter is "null unspecified").  There may be a legitimate case of unselecting or "animating out" the selection indicator of an item is removed.

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: Done.

components/Tabs/src/TabBarView/MDCTabBarView.h
romoore: Added a bit more. If we need further details, it should probably be in a follow-up PR.

PR #7784, opened 2019-07-03
[Tabs] Allow setting ripple color.

### LINKED ISSUE #7657: [Tabs] Internal issue: b/135637901

romoore:
Adds a property that customizes the ripple color.

## Example View Controller

|Before|After|
|---|---|
|Simulator Screen Shot - iPhone X - 2019-07-03 at 00 14 24|Simulator Screen Shot - iPhone X - 2019-07-03 at 00 12 49|

Closes #7657


leonmz (approved):

mikefan1991 (approved):

components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -82,6 +85,7 @@ @implementation MDCTabBarView
 - (instancetype)init {
   self = [super init];
   if (self) {
+    _defaultRippleColor = [[UIColor alloc] initWithWhite:0 alpha:(CGFloat)0.16];
```
codeman7: _Tip:_ I believe this is supposed to be 0.12 not 0.16.

components/Tabs/src/TabBarView/MDCTabBarView.h
```diff
@@ -34,6 +34,11 @@ __attribute__((objc_subclassing_restricted)) @interface MDCTabBarView : UIScroll
 /** The color of the Tab bar's background. */
 @property(nullable, nonatomic, copy) UIColor *barTintColor;
 
+/**
+ The color for the Ripple effect for touch feedback. If no value is provided, uses a default value.
+ */
+@property(nullable, nonatomic, copy) UIColor *rippleColor;
```
codeman7: How do you feel about rippleColorForState?

codeman7: How do you feel about this being _null_resettable_?

romoore: It doesn't support stateful ripple right now.  We can add that API later (and bind this to `.Normal` until deprecation/deletion).

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: Oh, it's not matching anything from our design guidelines. I can make it `0.12` if you like

components/Tabs/src/TabBarView/MDCTabBarView.h
romoore: I don't like exposing the default value. None of the other APIs on this class expose their defaults, but simply do "something". In this case I'm defaulting to the RippleView default value.

romoore: Is that OK with you? If we keep it `nullable` for now and leave the default as an implementation detail?

romoore: I spoke to @yarneo about it and it wasn't immediately obvious how it would work with the work @galiak11 is doing.  I think the simplest API is the best for our MVP.

romoore: Curious to hear your feedback if you know better, though.

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: FYI: This came from MDCRippleView's implementation.

codeman7: I'm fairly certain its supposed to be `0.12` I'll confirm with @yarneo today.

components/Tabs/src/TabBarView/MDCTabBarView.h
codeman7: That seems reasonable for now, please open a bug for this to track that we don't lose this.

codeman7: For me it's just confusing that if a client sets `rippleColor = nil` then a ripple still occurs and the ask what the `rippleColor` is we return `nil`. I can be convinced otherwise if you feel strongly about it.

codeman7: https://github.com/material-components/material-components-ios/issues/7786

romoore: Ok. I'll open an issue to change all the other APIs.

romoore: Opened #7787 to change all APIs.

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: Ok. I'm going to leave it for now since it matches Ripple's implementation. If it needs to change, then I suspect many components will have to be updated.

components/Tabs/src/TabBarView/MDCTabBarView.h
romoore: Changed this to `nonnull`.

codeman7 (approved):

Issue #7752, opened 2019-06-28
Internal issue: b/136277957


material-mirror:
This was filed as an internal issue. If you are a Googler, please visit [b/136277957](http://b/136277957) for more details.

---
#### Internal data
- Associated internal bug: [b/136277957](http://b/136277957)

PR #7804, opened 2019-07-08
[Math] Add MDCEdgeInsetsEqualToEdgeInsets to compare two UIEdgeInsets.

fluxminds:
Created to prepare for a fix for https://github.com/material-components/material-components-ios/issues/7752.

`UIEdgeInsetsEqualToEdgeInsets` directly compares `CGFloat` number within two . The `MDCEdgeInsetsEqualToEdgeInsets` added in this PR compares UIEdgeInsets using `MDCCGFloatEqual` to take epsilon into consideration.


romoore (dismissed):
Please add an Issue for this or otherwise explain why it's necessary.

fluxminds:
Added some context in the description. It wasn't added because another PR will be created to use this new method to close that issue.

andrewoverton (approved):

PR #7820, opened 2019-07-09
[Banner] fix dynamic type support for attributed string.

### LINKED ISSUE #7819: [Banner] dynamic type doesn't work when attributed string is set in textLabel.
BannerView's text doesn't respond to dynamic type changes when attributed string is set.

fluxminds:
close #7819 

Previously, Banner only looks at the font on the textLabel to make dynamic type changes. After the change, an attributed string displayed on a banner is able to auto adjust its size based on content sizing category.

|Before|After|
|---|---|
|Simulator Screen Shot - iPhone 8 - 2019-07-09 at 00 27 58| Simulator Screen Shot - iPhone 8 - 2019-07-09 at 00 19 02|


romoore (dismissed):
Please add snapshot tests to cover this use case.

components/Banner/src/MDCBannerView.m
```diff
@@ -558,11 +558,40 @@ - (void)updateBannerFont {
...
+        enumerateAttribute:NSFontAttributeName
+                   inRange:NSMakeRange(0, mutableAttributedText.length)
+                   options:0
+                usingBlock:^(id value, NSRange range, BOOL *stop) {
+                  if (value) {
+                    UIFont *previousFont = (UIFont *)value;
+                    if (previousFont.mdc_scalingCurve) {
+                      hasScalableFont = YES;
+                      UIFont *scaledFont =
+                          [previousFont mdc_scaledFontForTraitEnvironment:weakSelf];
```
romoore: Please capture `weakSelf` in a strong pointer so it isn't lost while enumerating.

components/Banner/tests/snapshot/MDCBannerSnapshotTests.m
```diff
@@ -303,4 +303,40 @@ - (void)testDynamicTypeForContentSizeCategoryExtraExtraLarge {
...
+                         range:NSMakeRange(10, 8)];
+    [bannerString addAttribute:NSForegroundColorAttributeName
+                         value:UIColor.redColor
+                         range:NSMakeRange(0, 9)];
+    [bannerString addAttribute:NSLinkAttributeName
+                         value:@"http://www.google.com"
+                         range:NSMakeRange([kBannerShortText length] - 2, 2)];
+    self.bannerView.textLabel.font = self.typographyScheme.body2;
+    self.bannerView.textLabel.attributedText = bannerString;
+    MDCButton *button = self.bannerView.leadingButton;
```
romoore: Please move the non-attributed string code into the "Given" section, since it appears the purpose of this test is to verify the layout when attributed text is provided.

romoore (approved):
Are banners allowed to have no buttons? If yes, please add a snapshot test for a no-button banner.

components/Banner/src/MDCBannerView.m
fluxminds: This is the only place `self` is referenced in this block. Capturing this weak pointer is not needed in this case.

components/Banner/tests/snapshot/MDCBannerSnapshotTests.m
fluxminds: Addressed in d5953cafef899e0d094e3319e91bd446d0f7fc43.

fluxminds:
Banner doesn't allow to have no buttons from the spec.

components/Banner/src/MDCBannerView.m
fluxminds: Discussed offline (`mdc_scaledFontForTraitEnvironment` only takes nonnull). Addressed in 5ed3632d0ba4658c7ba55f653cbd2b38dee6d895.

PR #7814, opened 2019-07-09
[Tabs] Add support for `selectedImage` from UITabBarItem.

romoore:
MDCTabBarView needs to support `selectedImage` from UITabBarItem. Some clients will want to use different images for selected/unselected states.

Closes #7798

PR #7825, opened 2019-07-09
[Tabs] Minor refactoring of KVO code.

romoore:
The KVO code for title text and images requires invalidation of the intrinsic
content size and layout for both the item view and the tab bar. The code was
repeated and it was easy to miss one or more steps for any of the use cases.

Follow-up to #7814

### DIFF
--- a/components/Tabs/src/TabBarView/MDCTabBarView.m
+++ b/components/Tabs/src/TabBarView/MDCTabBarView.m
@@ -489,22 +489,13 @@ static NSString *const kAccessibilityTraitsKeyPath = @"accessibilityTraits";
     }
     if ([keyPath isEqualToString:kImageKeyPath]) {
       tabBarItemView.image = newValue;
-      [tabBarItemView invalidateIntrinsicContentSize];
-      [tabBarItemView setNeedsLayout];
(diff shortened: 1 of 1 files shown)


components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -526,6 +517,13 @@ - (void)observeValueForKeyPath:(NSString *)keyPath
   }
 }
 
+- (void)markIntrinsicContentSizeAndLayoutNeedingUpdateForSelfAndItemView:(UIView *)itemView {
```
codeman7, line 520: Nit: Can this be `invalidateLayoutForSelfAndItemView:`

codeman7 (approved):

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: It could, but that would only be `setNeedsLayout`, yeah? I know naming is hard but in this case I feel it's important that the method capture that `intrinsicContentSize` is no longer valid.  That said, if you feel strongly I can certainly change it.

PR #7834, opened 2019-07-09
[Tabs] Fix the sizeThatFits bug in MDCTabBarView

mikefan1991:
With this change, the calculated width will not be larger than the provided width. So the justified and scrollable behavior can work properly.


romoore:
This change will break the behavior of `sizeToFit` (it will no longer fit its content).  Could you please file a bug (or let's chat offline) to discuss what the problematic behavior is? A snapshot test, unit test, or example View Controller to reproduce the issue would be the best.

romoore (changes requested):
Marking Changes Requested until we can better understand the problematic scenario and reproduce it in a test environment.

romoore:
I took a further look at this with MDCHeaderStackView and agree that the change is necessary. I've opened https://github.com/material-components/material-components-ios/pull/7846 to make the same change and add further tests.  Closing this PR as the other is closer to merging.

PR #7837, opened 2019-07-10
[Tabs] Add tests for `sizeToFit`.

romoore:
Adds tests to explicitly verify the behavior of `sizeToFit` to ensure that the tab bar can resize itself to fit its content in the expected behavior.


romoore:
No longer necessary, replaced with tests in https://github.com/material-components/material-components-ios/pull/7846

PR #7838, opened 2019-07-10
[Catalog] Test pod lib lint in kokoro

andrewoverton:
Giving this renewed attention after after https://github.com/material-components/material-components-ios/pull/7836

### DIFF
--- a/.kokoro
+++ b/.kokoro
@@ -393,9 +393,10 @@ run_cocoapods() {
     cd github/repo
 
     gem_install xcpretty 
-    # TODO(https://github.com/material-components/material-components-ios/issues/6356 ): Move to 1.6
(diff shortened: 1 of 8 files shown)


andrewoverton:
I _think_ this change fixes it, but it failed because it didn't include the FlexibleHeader Math dependency 😛 . Continuing these efforts with https://github.com/material-components/material-components-ios/pull/7847

andrewoverton:
This is only passing because I deleted the line that enforces 1.6.0 😑. If we can't fix it any other way, maybe we should just rely on 1.5.3 for the podspec job. It might be better than nothing 🤷‍♂

andrewoverton:
Closing this for now, may come back to it in the future.

PR #7841, opened 2019-07-10
[Tabs] Allow clients to calculate when a view appears/disappears.

### LINKED ISSUE #7839: [Tabs] Internal issue: b/137142790
This was filed as an internal issue. If you are a Googler, please visit [b/137142790](http://b/137142790) for more details.

---
#### Internal data
- Associated internal bug: [b/137142790](http://b/137142790)

romoore:
A new API, `- rectForItem:inCoordinateSpace:` provides the item's corresponding
view's frame in the provided coordinate space. Clients can use this API to
calculate which views are "shown" at any given time. A simple use of this API
for that purpose is added to the example view controller.

Alternative solution to #6275
Closes #7839


components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -410,6 +410,21 @@ - (id)accessibilityElementForItem:(UITabBarItem *)item {
...
+  if (item == nil) {
+    return CGRectNull;
+  }
+  NSUInteger index = [self.items indexOfObject:item];
+  if (index == NSNotFound || index >= self.itemViews.count) {
+    return CGRectNull;
+  }
+  CGRect frame = self.itemViews[index].frame;
+  frame = CGRectMake(CGRectGetMinX(frame), CGRectGetMinY(frame), CGRectGetWidth(frame),
+                     CGRectGetHeight(frame));
```
mrhappyasthma: Could you explain the use of this line? To me it looks like you're using CGRectMake() to create the exact same frame as before.

Am I missing something?

components/Tabs/examples/MDCTabBarViewTypicalExampleViewController.m
```diff
@@ -265,6 +267,47 @@ - (void)changeItemsToTextAndImage {
...
+  [super viewDidLayoutSubviews];
+  [self logVisibleItems];
+}
+
+- (void)logVisibleItems {
+  NSMutableArray<NSString *> *visibleItemTitles = [NSMutableArray array];
+  for (UITabBarItem *item in self.tabBar.items) {
+    CGRect itemViewInWindow = [self.tabBar rectForItem:item inCoordinateSpace:self.view.window];
+    CGRect overlapRect = CGRectIntersection(self.view.window.bounds, itemViewInWindow);
+    if (CGRectIsNull(overlapRect) || CGRectGetWidth(itemViewInWindow) < 1) {
```
mrhappyasthma: Is the `CGRectGetWidth() < 1` check necessary? Is this just to ensure the item is minimally visible of a full point?

components/Tabs/tests/unit/TabBarView/MDCTabBarViewTests.m
```diff
@@ -880,4 +880,58 @@ - (void)testAccessibilityElementForItemInItemsArrayReturnsItemViewWithMatchingTi
   }
 }
 
+- (void)testRectForItemNotFoundReturnsNullRectangle {
+  // Given
+  self.tabBarView.items = @[];
```
mrhappyasthma, line 885: Do you technically need to use `sizetoFit` and `layoutIfNeeded` to ensure that the view is up-to-date with the 0 items before trying to get the frame?

That might make this a slightly more robust.

WDYT?

components/Tabs/tests/unit/TabBarView/MDCTabBarViewTests.m
```diff
@@ -880,4 +880,58 @@ - (void)testAccessibilityElementForItemInItemsArrayReturnsItemViewWithMatchingTi
...
+                NSStringFromCGRect(itemFrame), NSStringFromCGRect(self.tabBarView.bounds));
+}
+
+- (void)testRectForItemConvertedToSuperView {
+  // Given
+  UIOffset tabBarOffsetWithinSuperview = UIOffsetMake(30, 40);
+  self.tabBarView.items = @[ self.itemA ];
+  [self.tabBarView sizeToFit];
+  [self.tabBarView layoutIfNeeded];
+  UIView *superView =
```
mrhappyasthma: nit: should `superview` be one word here to match the capitalization common in UIView.

http://go/appledoc/uikit/uiview/1622474-superview?language=objc

components/Tabs/tests/unit/TabBarView/MDCTabBarViewTests.m
```diff
@@ -880,4 +880,58 @@ - (void)testAccessibilityElementForItemInItemsArrayReturnsItemViewWithMatchingTi
...
+      [[UIView alloc] initWithFrame:CGRectMake(0, 0,
+                                               CGRectGetWidth(self.tabBarView.bounds) +
+                                                   tabBarOffsetWithinSuperview.horizontal,
+                                               CGRectGetHeight(self.tabBarView.bounds) +
+                                                   tabBarOffsetWithinSuperview.vertical)];
+  [superView addSubview:self.tabBarView];
+  self.tabBarView.center =
```
mrhappyasthma: Rather than this, could we just use CGRectOffset? That might be more clear rather than trying to mess with the center of the element in order to offset it.

mrhappyasthma (changes requested):
Generally looks pretty good.

components/Tabs/examples/MDCTabBarViewTypicalExampleViewController.m
romoore: It's to avoid division by zero. This gets called before the subviews are laid-out during initial presentation. Since this was a hacky little demo, it doesn't have all the sharp edges worked-out yet.

components/Tabs/tests/unit/TabBarView/MDCTabBarViewTests.m
romoore: I could if you prefer. It's not testing that code path but it wouldn't hurt.

components/Tabs/tests/unit/TabBarView/MDCTabBarViewTests.m
romoore: I was trying to avoid that exact capitalization. I'll change it to something like `tabBarViewSupeview`

components/Tabs/tests/unit/TabBarView/MDCTabBarViewTests.m
romoore: I'm probably missing something. My understanding is that in general it's better to manipulate `bounds` and `center` rather than `frame` on a view. That's because the `frame` is undefined when the transform is anything other than the identity.  I was using `center` here because it was rather simple and didn't require me to standardize the `frame` as an initial step.

components/Tabs/src/TabBarView/MDCTabBarView.m
romoore: The frame may not be standardized.  I can change it to `CGRectStandardize()`, but I was writing this rather hastily as a Proof of Concept.

mrhappyasthma: Ah, I see. Yeah I'd prefer CGRectStandardize() as I think it will be more clear.

components/Tabs/examples/MDCTabBarViewTypicalExampleViewController.m
mrhappyasthma: Could we rewrite it as `width == 0` to make it a bit more clear then? Or is this because it's a float and the <1 is for approximately zero?

components/Tabs/tests/unit/TabBarView/MDCTabBarViewTests.m
mrhappyasthma: SGTM

components/Tabs/tests/unit/TabBarView/MDCTabBarViewTests.m
mrhappyasthma: `bounds` and `center` manipulate the frame. they are syntactic sugar/convenience.

I imagine you should probably make a local variable to copy the CGRect of the frame, manipulate it, and then assign it back. (Multiple accesses of the `frame` are also suboptimal)

components/Tabs/tests/unit/TabBarView/MDCTabBarViewTests.m
mrhappyasthma: Oh, I see. On second look the function short-circuits on invalid items before it tries to check any frames.

Probably not necessary.

components/Tabs/examples/MDCTabBarViewTypicalExampleViewController.m
romoore: I originally wrote it as `< (CGFloat)0.001` but that seemed like it would generate questions about the specific choice of precision.  As long as this example doesn't crash, I don't have much opinion. I would expect that production code would be significantly more robust (and more efficient).

components/Tabs/tests/unit/TabBarView/MDCTabBarViewTests.m
romoore: I agree that multiple accesses are suboptimal and could be improved. I'll pull the frame into a local variable.

It is my understanding that `bounds` and `center` are not syntactic sugar for `frame`, nor convenience. I believe that the opposite is actually true.  Based on [Apple's documentation for `bounds`](https://developer.apple.com/documentation/uikit/uiview/1622580-bounds) it seems that it and `center` are intended as the source of the value of `frame` (potentially along with `transform`).  The [documentation for the `center` property](https://developer.apple.com/documentation/uikit/uiview/1622627-center) explicitly calls-out that property as the correct way to move a view within its superview's bounds.

>  Use this property, **instead of the frame property**, when you want to change the position of a view. The center point is always valid, even when scaling or rotation factors are applied to the view's transform (emphasis added). 

I think [the UIKit documentation for `frame`](https://developer.apple.com/documentation/uikit/uiview/1622621-frame) makes it pretty clear that using `center` and `bounds` can be preferred over `frame` when the caller is unsure if the `transform` has been manipulated.

> . . . if the transform property contains a non-identity transform, the value of the frame property is undefined and should not be modified. In that case, **reposition the view using the center property and adjust the size using the bounds property instead** (emphasis added).

In general, I do personally prefer manipulating `center` and `bounds` rather than `frame`, but it often involves arithmetic "gymnastics" that make the code harder to read.  Since `transform` manipulations are rare, it's often safe to go with `frame` as a shortcut.

mrhappyasthma: Interesting. I hadn't read all the way through the documentation. I stopped after the line that read: 

> Setting this property updates the origin of the rectangle in the frame property appropriately.

If Apple recommends using those fields instead, I have no objection to it.

components/Tabs/src/TabBarView/MDCTabBarView.m

components/Tabs/examples/MDCTabBarViewTypicalExampleViewController.m
romoore: Documented a bit more helpfully.

components/Tabs/tests/unit/TabBarView/MDCTabBarViewTests.m
romoore: Acknowledged.

components/Tabs/tests/unit/TabBarView/MDCTabBarViewTests.m
romoore: Done.

components/Tabs/tests/unit/TabBarView/MDCTabBarViewTests.m
romoore: Leaving it for now.

components/Tabs/examples/MDCTabBarViewTypicalExampleViewController.m
```diff
@@ -265,6 +271,65 @@ - (void)changeItemsToTextAndImage {
...
+
+  NSMutableSet<UITabBarItem *> *noLongerVisibleItems = [NSMutableSet set];
+  NSMutableSet<UITabBarItem *> *visibleItemsCopy = [visibleItems copy];
+  for (UITabBarItem *item in self.visibleItems) {
+    if (![visibleItems containsObject:item]) {
+      [noLongerVisibleItems addObject:item];
+    }
+    [visibleItems removeObject:item];
+  }
+  self.visibleItems = visibleItemsCopy;
```
mrhappyasthma: I got a little lost on these loops.

Why do you assign the visibleItemsCopy back to self.visibleItems on line 319? Wouldn't you want to assign `visibleItems` since it now contains only the still-visible items.

It seems `visibleItemsCopy` would still contain the elements in `noLongerVisibleItems`.

components/Tabs/tests/snapshot/TabBarView/MDCTabBarViewSnapshotTests.m
```diff
@@ -1212,4 +1212,82 @@ - (void)testCustomView {
...
+
+- (void)testRectForItemInCoordinateSpaceWithinBounds {
+  // Given
+  self.tabBarView.items = @[ self.item1, self.item2, self.item3 ];
+  CGSize intrinsicContentSize = self.tabBarView.intrinsicContentSize;
+  self.tabBarView.bounds =
+      CGRectMake(0, 0, intrinsicContentSize.width, intrinsicContentSize.height);
+  UIView *itemRectOverlayView = [[UIView alloc] init];
+  itemRectOverlayView.backgroundColor =
+      [UIColor.magentaColor colorWithAlphaComponent:(CGFloat)0.25];
```
mrhappyasthma, line 1228: nit: Is there a reason you cast to CGFloat rather than using `0.25f`?

components/Tabs/tests/snapshot/TabBarView/MDCTabBarViewSnapshotTests.m
```diff
@@ -1212,4 +1212,82 @@ - (void)testCustomView {
...
+
+  // Then
+  [self generateSnapshotAndVerifyForView:self.tabBarView];
+}
+
+- (void)testRectForItemInCoordinateSpaceWhenPartiallyOutsideBarBounds {
+  // Given
+  self.tabBarView.items = @[ self.item1, self.item2 ];
+  CGSize intrinsicContentSize = self.tabBarView.intrinsicContentSize;
+  self.tabBarView.bounds = CGRectMake(0, 0, 170, intrinsicContentSize.height);
```
mrhappyasthma: suggestion: this `170` seems like a magic number. Perhaps use a constant variable with a meaningful name.

components/Tabs/tests/snapshot/TabBarView/MDCTabBarViewSnapshotTests.m
mrhappyasthma: Same for the others below.

romoore: Yep, we don't use `0.123f` in our codebase. We decided to use double-precision and stick with it. There are a few stragglers after a massive change we did, but we even do things like

```
float f = (float)0.25;
```

Forcing the downcast makes it easier to remember that we're retaining full-width precision on 64-bit hosts.  Occasionally it has made a difference in computed values.

components/Tabs/examples/MDCTabBarViewTypicalExampleViewController.m
romoore: I rewrote the logic and renamed the sets. Let me know what you think.

components/Tabs/tests/snapshot/TabBarView/MDCTabBarViewSnapshotTests.m
romoore: Done.

codeman7 (approved):

bryanoltman (approved):

mrhappyasthma (approved):

PR #7843, opened 2019-07-10
[Snackbar] Use MDCShadowLayer instead of UIKit shadows

### LINKED ISSUE #7829: [Snackbar] Internal issue: b/137124095
This was filed as an internal issue. If you are a Googler, please visit [b/137124095](http://b/137124095) for more details.

---
#### Internal data
- Associated internal bug: [b/137124095](http://b/137124095)

codeman7:
MDCSnackbarMessageView was previously using UIKit to display its shadows. This is a deviation from the rest of our components. This change updates the implementation so that we can later expose an API for `elevation` so that clients can customize the elevation of the snackbar and get its elevation accordingly. No tests were added because this changes the underlying layer class and snapshot tests were already in place to catch the visual diffs.

Closes #7829


components/Snackbar/src/MDCSnackbarMessageView.m
```diff
@@ -409,6 +401,15 @@ - (void)dismissWithAction:(MDCSnackbarMessageAction *)action userInitiated:(BOOL
...
++ (Class)layerClass {
+  return [MDCShadowLayer class];
+}
+
+- (void)setElevation:(MDCShadowElevation)elevation {
+  _elevation = elevation;
+  [(MDCShadowLayer *)self.layer setElevation:_elevation];
+}
+
 #pragma mark - Subclass overrides
```
romoore, line 413: (For Snapshots): Please increase the spacing of the background view around the Snackbar View so the shadow rendering can be fully captured. The easiest way to see if you've fully-captured the shadow is to use the "Digital Color Meter" tool and verify that all 4 edges of the snapshot image are rendered in the background color (0xCCC)

romoore (approved):
LGTM. Please address clang feedback (or resolve it if it's fixed). Please also attach an issue number for this PR.

components/Snackbar/src/MDCSnackbarMessageView.m
codeman7: Done

codeman7: Done

PR #7844, opened 2019-07-10
[Snackbar] Allow clients to customize elevation

codeman7:
This allows clients to customize the elevation of their snackbar and additionally get the elevation of their snackbar. Unit test have been added for custom elevation and default elevation as well as a custom elevation snapshot test.


components/Snackbar/src/MDCSnackbarManager.h
```diff
@@ -165,6 +167,9 @@
  */
 @property(nonatomic, strong, nullable) UIColor *snackbarMessageViewShadowColor;
 
+/** The elevation for the Snackbar message view. */
+@property(nonatomic, assign) MDCShadowElevation elevation;
```
romoore: Nit: Naming: Please call this something like `messageElevation`. That will make it clearer the elevation is for the snackbar messages and not the manager (or something else).

components/Snackbar/tests/snapshot/MDCSnackbarMessageViewSnapshotTests.m
```diff
@@ -186,4 +186,17 @@ - (void)testWithLongMessageLongActionRTLArabic {
...
 }
 
+- (void)testWithCustomElevation {
+  // Given
+  MDCSnackbarMessageView *messageView = [self snackbarMessageViewWithText:kItemTitleShort1Latin
+                                                              actionTitle:kItemTitleShort2Latin];
+  messageView.frame = CGRectMake(0, 0, kWidth, kHeightSingleLineText);
+
+  // When
+  messageView.elevation = 0;
```
romoore, line 196: Please either add a second test for custom elevation with a non-zero value or change this to something other than zero.  If there were a bug in setting the elevation that caused the shadow elevation to become invalid (and perhaps reset to `0`), it would not be possible to detect in this test.

romoore (approved):
Please attach an Issue for this change. It's not clear why custom elevations for Snackbars are allowed since the guidelines didn't previously allow it.

LGTM with naming nits.

components/Snackbar/src/MDCSnackbarManager.h
codeman7: Done

components/Snackbar/tests/snapshot/MDCSnackbarMessageViewSnapshotTests.m
codeman7: Done

codeman7:
@romoore The setter was added per @randallli request to have a setter anytime there is a getter.

PR #7846, opened 2019-07-10
[Tabs] Adjust sizeThatFits: behavior for TabBarView.

### LINKED ISSUE #7845: [Tabs] Internal issue: b/137186862
This was filed as an internal issue. If you are a Googler, please visit [b/137186862](http://b/137186862) for more details.

---
#### Internal data
- Associated internal bug: [b/137186862](http://b/137186862)

romoore:
The `sizeThatFits:` beahvior of `MDCTabBarView` previously would return a size
large enough for Fixed tabs in a justified layout. If the provided `size` was
wider than necessary, that value would be returned.

A solution that works better for inclusion in MDCHeaderStackView, which uses
`sizeThatFits:` to determine the view's frame size.

Closes #7845


codeman7 (commented):
This change makes it so that clients must set an `intrinsicContentSize` for their MDCTabBarView now, they can no longer call `sizeToFit` is that correct?

romoore:
@codeman7 Intrinsic content size is not set on views, but reflects the view's contents' natural sizing. Clients won't need to set anything, but you are correct they cannot blindly call `sizeToFit` to make it adopt a sufficient width for a Justified layout.

 They can call `sizeToFit`, but like UIView it will no longer change its width (it will continue to update its height). The new behavior probably makes more sense given that it is a scroll view and so it doesn't "need" to be any specific width in order to show its content correctly.

components/Tabs/BUILD
```diff
@@ -225,6 +225,8 @@ mdc_snapshot_objc_library(
         ":Tabs",
         ":Theming",
         ":privateTabBarView",
+	"//components/AppBar",
```
codeman7: Please run `buildifier`

codeman7 (approved):

components/Tabs/BUILD
romoore: Done.

PR #7850, opened 2019-07-10
Add documentation to install using Carthage

abdulhaq-e:
Not a fix to #75 but at least a workaround.

### DIFF
--- a/docs/component-installation.md
+++ b/docs/component-installation.md
@@ -28,4 +28,47 @@ import MaterialComponents.Material<#ComponentName#>
 ```objc
 #import "Material<#ComponentName#>.h"
 ```
+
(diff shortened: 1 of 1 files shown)


abdulhaq-e:
I didn't know where to place something like this. Please note afaik, the edited mark down file is not linked to anywhere in the docs.

romoore (changes requested):
Thanks for the PR. However, given our official lack of support for Carthage I don't think it's appropriate to place these instructions in the project. 

My recommended alternative would be to post this same information (as formatted) in the conversation of #75. That will allow clients to see the work-around when they come looking for support.

abdulhaq-e:
Done

PR #7852, opened 2019-07-10
v85.6.0

romoore:


andrewoverton (approved):
LGTM if breaking beta changes don't require major releases

romoore:
They haven't so far. I believe the idea of Beta is that they are under active (heavy) development and breaking changes will be expected. It's why they don't ship to CocoaPods.

Issue #7855, opened 2019-07-10
[Collections] Internal issue: b/137184244


material-mirror:
This was filed as an internal issue. If you are a Googler, please visit [b/137184244](http://b/137184244) for more details.

---
#### Internal data
- Associated internal bug: [b/137184244](http://b/137184244)

material-mirror:
The internal issue [b/137184244](http://b/137184244) is now closed. This issue is being closed as a result.

PR #7856, opened 2019-07-10
[Collections] Make MDCCollectionInfoBarView border color dynamic

andrewoverton:
I'm pretty sure this is very wrong. This is separate from the work that Bryan is doing in https://github.com/material-components/material-components-ios/pull/7857.

Related to #7855.

### DIFF
--- a/components/Collections/src/private/MDCCollectionInfoBarView.m
+++ b/components/Collections/src/private/MDCCollectionInfoBarView.m
@@ -187,15 +187,15 @@ static inline UIColor *CollectionInfoBarRedColor(void) {
 }
 
 - (void)updateBorderColor {
-  CGFloat white = 0;
+  CGFloat white = 0.0f;
   if (@available(iOS 12.0, *)) {
     if (self.traitCollection.userInterfaceStyle == UIUserInterfaceStyleDark) {
-      white = 1;
+      white = 1.0f;
     }
   }
   CGFloat alpha = 0;
   if (_style == MDCCollectionInfoBarViewStyleActionable) {
-    alpha = 0.1;
+    alpha = 0.1f;
   }
   _backgroundBorderLayer.borderColor = [UIColor colorWithWhite:white alpha:alpha].CGColor;
 }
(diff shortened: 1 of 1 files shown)


components/Collections/src/private/MDCCollectionInfoBarView.m
```diff
@@ -175,13 +182,22 @@ - (void)setStyle:(MDCCollectionInfoBarViewStyle)style {
...
-      _backgroundBorderLayer = [CALayer layer];
-      _backgroundBorderLayer.borderColor = [UIColor colorWithWhite:0 alpha:(CGFloat)0.1].CGColor;
-      _backgroundBorderLayer.borderWidth = 1 / [[UIScreen mainScreen] scale];
-      [self.backgroundView.layer addSublayer:_backgroundBorderLayer];
+  }
+  [self updateBorderColor];
+}
+
+- (void)updateBorderColor {
+  CGFloat white = 0.0f;
```
romoore, line 190: Please use either an explicit cast to `(CGFloat)` or use integral values (my preferred option).

```objc
CGFloat white = 0;
white = 1;
CGFloat alpha = 0;
alpha = (CGFloat)0.1;

// Alternatively
CGFloat white = (CGFloat)0.0;
white = (CGFloat)1.0;
CGFloat alpha = (CGFloat)0.0;
alpha = (CGFloat)0.1;
```

components/Collections/src/private/MDCCollectionInfoBarView.m
```diff
@@ -71,6 +71,11 @@ - (void)commonMDCCollectionInfoBarViewInit {
       (UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight);
   _backgroundView.hidden = YES;
 
+  _backgroundBorderLayer = [CALayer layer];
+  _backgroundBorderLayer.borderWidth = 1 / [[UIScreen mainScreen] scale];
```
romoore, line 75: Please use `self.window.screen.scale`. Do not rely on the main screen's scale.

romoore (changes requested):
Small nits in the main code. However, requesting changes because there seem to be no affected tests. At a minimum, please add snapshot tests for this view for this property change.  It looks like unit tests might be possible, but perhaps harder to read/write.

romoore:
Please feel free to dismiss my Changes Requested once the tests have been added.

yarneo (commented):
Andrew and I talked about this offline; the component implementation shouldn't be the one setting a color that reacts to dark mode, but rather it should expose this property so that the color scheme can correctly be applied to this element as well.

However, because Collections don't have any sort of themer/theming/etc, and is a "to be deprecated" component, there isn't much value to start doing all this leg work to make Collections compatible with new features at this time.

andrewoverton:
Closing this after discussion in today's standup

PR #7888, opened 2019-07-11
[AppBar] Provide app bar parameter for traitCollectionDidChange block.

romoore:
Because `MDCAppBarViewController` is often injected into other view controllers, setting these blocks on the injector (*e.g.*, `MDCAppBarNavigationController`) means that the eventual app bar view controller object isn't actually available when the block is written. Instead of assuming access to the object receiving a trait collection change, the block can receive that object directly as a parameter.

Part of #7849
Follow-up to #7851


bryanoltman (approved):

yarneo (approved):
Thanks for thinking about this use case and figuring out a good solution for it

PR #7911, opened 2019-07-14
[Dialogs] Add shadow color API

### LINKED ISSUE #7875: [Dialogs] Internal issue: b/137184245
This was filed as an internal issue. If you are a Googler, please visit [b/137184245](http://b/137184245) for more details.

---
#### Internal data
- Associated internal bug: [b/137184245](http://b/137184245)

codeman7:
This PR adds a shadowColor API to MDCAlertController and presentation controller. Snapshot test have been added for both MDCAlertController and the trackingView that the presentation controller uses since the `MDCAlertController.view` is a different view than the view that represents the shadow. This new API supports dynamic color.

Closes #7875


components/Dialogs/src/MDCDialogPresentationController.h
```diff
@@ -95,6 +95,13 @@
  */
 @property(nonatomic, assign) MDCShadowElevation dialogElevation;
 
+/**
+ Customize the shadow color of dialog.
```
romoore: Tip: Nit: Consider copy editing to match the better docs on the AlertController API.

components/Dialogs/tests/snapshot/MDCAlertControllerCustomTraitCollectionTests.m
```diff
@@ -17,29 +17,45 @@
...
 
 - (UITraitCollection *)traitCollection {
   return self.traitCollectionOverride ?: [super traitCollection];
 }
 
 @end
 
-@interface MDCAlertControllerDynamicTypeTests : MDCSnapshotTestCase
-@property(nonatomic, strong, nullable) AlertControllerDynamicTypeSnapshotTestFake *alertController;
+/** A @c MDCDialogShadowedView test fake to override the @c traitCollection to test. */
```
romoore: Tip: Nit: s/A/An/

romoore (commented):
Getting on a plane. Some feedback.

components/Dialogs/tests/unit/MDCAlertControllerTests.m
```diff
@@ -301,6 +302,32 @@ - (void)testCustomDialogPresentationElevation {
...
 
+- (void)testCustomShadowColor {
+  // Given
+  UIColor *fakeColor = UIColor.orangeColor;
+
+  // When
+  self.alert.shadowColor = fakeColor;
+
+  // Then
+  XCTAssertEqualObjects(self.alert.mdc_dialogPresentationController.trackingView.shadowColor,
```
randallli, line 313: lets test the "copy" part too by making sure the pointers are not the same.

randallli (approved):

components/Dialogs/tests/snapshot/MDCAlertControllerCustomTraitCollectionTests.m
codeman7: Done

components/Dialogs/src/MDCDialogPresentationController.h
codeman7: Done

components/Dialogs/tests/unit/MDCAlertControllerTests.m
codeman7: This does tests to make sure that the copy is the same, since `[UIColor copy]` returns `self` they are the same pointer. `copy` in the header is to match `UIKit`please see [UIView documenation](https://developer.apple.com/documentation/uikit/uiview/1622591-backgroundcolor?language=objc). 

```objc
UIColor *fakeColor = UIColor.whiteColor;
XCTAssertNotEqual([fakeColor copy], fakeColor); // Fail, because they are the same pointer
XCTAssertEqualObjects([fakeColor copy], fakeColor); // Pass, because they are equal objects.
```

romoore: We should not be testing that `copy` semantics do not return the same object. Returning the same immutable object is a very efficient implementation of `copy`.

PR #7917, opened 2019-07-15
[FlexibleHeader] Add traitCollectionDidChange block

codeman7:
The Flexible needs an API so clients can hook-in to trait collection changes. This additionally passes the flexibleHeader as a parameter so clients can modify the flexible header within the block.


randallli (commented):
This is AppBar's block?

This title needs to be different from
https://github.com/material-components/material-components-ios/pull/7915

components/FlexibleHeader/tests/unit/FlexibleHeaderTraitColletionDidChangeTests.m
```diff
@@ -0,0 +1,117 @@
...
+        flexibleHeaderViewController.traitCollectionDidChangeBlock = nil;
+      };
+
+  // When
+  UITraitCollection *testCollection = [UITraitCollection traitCollectionWithDisplayScale:77];
+  [flexibleHeaderController traitCollectionDidChange:testCollection];
+
+  // Then
+  [self waitForExpectations:@[ expectation ] timeout:1];
+  XCTAssertEqual(passedTraitCollection, testCollection);
```
randallli: does this pass? passedTraitCollection doesn't get set until the block gets called. that happens after this line runs...

If you want to assert things are passed in correctly then I would put this assert inside the block

randallli (dismissed):

codeman7:
This is MDCFlexibleHeaderViewController block, but MDCAppBarViewController subclasses MDCFlexibleHeaderViewController so this block is needed for both. We may be able to remove it from MDCAppBarViewController.

components/FlexibleHeader/tests/unit/FlexibleHeaderTraitColletionDidChangeTests.m
codeman7: I believe the assertions wait for the expectation to fulfill before they are ran.

romoore:
If it's subclassed, doesn't it result in duplicated symbols when compiling `MDCAppBarViewController`?

romoore (dismissed):
Please remove the call to the block in AppBarViewController's `traitCollectionDidChange:` method.

components/FlexibleHeader/tests/unit/FlexibleHeaderTraitColletionDidChangeTests.m
```diff
@@ -0,0 +1,115 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <XCTest/XCTest.h>
+
+#import "MaterialFlexibleHeader.h"
+
+@interface FlexibleHeaderTraitCollectionDidChangeTests : XCTestCase
```
romoore: Please name this "MDCFlexibleHeaderViewTraitCollectionDidChangeTests" so it can be identified by Xcode as associated with `MDCFlexibleHeaderView`

components/AppBar/src/MDCAppBarNavigationController.h
```diff
@@ -17,6 +17,7 @@
 @class MDCAppBar;
 @class MDCAppBarViewController;
 @class MDCAppBarNavigationController;
+@class MDCFlexibleHeaderViewController;
```
romoore, line 20: Looks like this is mixing FHV and FHVC in the same PR.  Please split them apart.

components/FlexibleHeader/tests/unit/FlexibleHeaderTraitColletionDidChangeTests.m
```diff
@@ -0,0 +1,115 @@
...
+  [flexibleHeader traitCollectionDidChange:testCollection];
+
+  // Then
+  [self waitForExpectations:@[ expectation ] timeout:1];
+  XCTAssertEqual(passedTraitCollection, testCollection);
+  XCTAssertEqual(passedFlexibleHeader, flexibleHeader);
+}
+
+- (void)
+    testTraitCollectionDidChangeBlockCalledWhenTraitCollectionChangesForFlexibleHeaderViewController {
```
romoore: Please put this in a separate file, "MDCFlexibleHeaderViewControllerTraitCollectionDidChangeTests". I wouldn't name them so specifically, but that seems to be the pattern.  In any case, please name the tests to match the class under test so Xcode can associate them.

romoore (changes requested):
Please dismiss my review when changes are made.

components/AppBar/src/MDCAppBarNavigationController.h
romoore: Obsolete.

components/FlexibleHeader/src/MDCFlexibleHeaderViewController.h
```diff
@@ -40,6 +40,14 @@
 /** The flexible header view instance that this controller manages. */
 @property(nonatomic, strong, nonnull, readonly) MDCFlexibleHeaderView *headerView;
 
+/**
+ A block that is invoked when the FlexibleHeaderViewController receives a call to @c
```
romoore: Nit: Naming: ...when the MDCFlexibleHeaderViewController receives...

components/FlexibleHeader/tests/unit/FlexibleHeaderTraitColletionDidChangeTests.m
```diff
@@ -22,7 +22,7 @@ @interface FlexibleHeaderTraitCollectionDidChangeTests : XCTestCase
 
 @implementation FlexibleHeaderTraitCollectionDidChangeTests
 
-- (void)testTraitCollectionDidChangeBlockCalledWhenTraitCollectionChanges {
+- (void)testTraitCollectionDidChangeBlockCalledWhenTraitCollectionChangesForFlexibleHeaderView {
```
romoore: Tip: (Follow-up PR) Please move this to a file reflecting the class's name. Also "ColletionDidChange" is misspelled in this file.

Suggestion: `MDCFlexibleHeaderTests` (or less preferred `MDCFlexibleHeaderTraitCollectionDidChangeTests`).

romoore (approved):
LGTM but please put the tests in files appropriate for the class names.

components/FlexibleHeader/src/MDCFlexibleHeaderViewController.h
codeman7: Done

components/FlexibleHeader/tests/unit/FlexibleHeaderTraitColletionDidChangeTests.m
codeman7: Updated.

components/FlexibleHeader/tests/unit/FlexibleHeaderTraitColletionDidChangeTests.m
codeman7: Put these in `FlexibleHeaderViewTraitCollectionTests`, and the controller tests in `FHViewControllerTraitCollectionTests`. Since they test different objects.

PR #7956, opened 2019-07-18
[Banner] fix layout margin not being considered for all layout styles.

### LINKED ISSUE #7959: [Banner] Banner height is not correct because of layout margin not being used.
image

The button should be fully wrapped in this snapshot test. 

---
#### Internal data
- Associated internal bug: [b/137871026](http://b/137871026)

fluxminds:
closes https://github.com/material-components/material-components-ios/issues/7959

There was a misalignment issue  on the horizontal axis for non single line style banner. `widthLimit` should be used in all layout styles.


romoore (dismissed):
Please add tests for the bug/fix.  Please also attach the associated Issue that is being addressed by this PR.

components/Banner/src/MDCBannerView.m
```diff
@@ -352,15 +353,15 @@ - (CGSize)sizeThatFits:(CGSize)size {
     }
     case MDCBannerViewLayoutStyleMultiRowAlignedButton: {
       frameHeight += kTopPaddingLarge + kBottomPadding;
-      frameHeight += [self getFrameHeightOfImageViewAndTextLabelWithSizeToFit:size];
+      frameHeight += [self getFrameHeightOfImageViewAndTextLabelWithSizeToFit:adjustedSizeToFit];
```
romoore: Tip: Should this change for multi-row aligned button layout result in a visible difference? If not, please revert this change. If it should be visible, please add a test for it.

components/Banner/src/MDCBannerView.m
```diff
@@ -327,13 +327,14 @@ - (void)deactivateAllConstraints {
 - (CGSize)sizeThatFits:(CGSize)size {
   MDCBannerViewLayoutStyle layoutStyle = [self layoutStyleForSizeToFit:size];
   CGFloat frameHeight = 0.0f;
+  CGFloat widthLimit = size.width;
```
romoore: Tip: This code looks like it's duplicated in `layoutStyleForSizeToFit:`. Can you refactor a bit to eliminate the duplication and provide a reusable method?

```objc
- (CGSize)contentSizeForLayoutSize:(CGSize)layoutSize {
  CGFloat remainingWidth = layoutSize.width;
  CGFloat marginsPadding = self.layoutMargins.left + self.layoutMargins.right;
  remainingWidth -= marginsPadding;
  remainingWidth -= (kLeadingPadding + kTrailingPadding);
  return CGSizeMake(remainingWidth, layoutSize.height);
}
```

romoore (approved):

components/Banner/src/MDCBannerView.m
fluxminds: Good suggestion. Addressed in 0f55b0f4ad62e9feb898b5334df7eb7aeea65da9.

components/Banner/src/MDCBannerView.m
fluxminds: This is aligned style but only single button. This one will be kept in this PR for snapshot comparison. I added a separate snapshot for stacked style in 68abaf82d01c639d7a0403f674f36db4ce5f5ed6.

PR #7957, opened 2019-07-18
[Banner] add snapshot test for middle length text with large layout margin.

fluxminds:
Prepares a snapshot test to solve https://github.com/material-components/material-components-ios/issues/7959.

The snapshot shows that the banner is not generated with a correct height as the button layout is beyond the banner region slightly.:
image


components/Banner/tests/snapshot/MDCBannerSnapshotTests.m
```diff
@@ -19,9 +19,12 @@
 #import "MaterialTypographyScheme.h"
 
 static NSString *const kBannerShortText = @"tristique senectus et";
+static NSString *const kBannerMiddleLengthText =
```
romoore, line 22: Tip: What makes these "middle" versus "long"? It looks like the difference is one word?

components/Banner/tests/snapshot/MDCBannerSnapshotTests.m
```diff
@@ -19,9 +19,12 @@
...
+    @"Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor.";
 static NSString *const kBannerLongText =
     @"Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt.";
 static const CGFloat kBannerContentPadding = 10.0f;
+static const CGFloat kBannerLargeContentPadding = 30.0f;
```
romoore, line 29: Nit: Please add comments to explain what this is used for.

romoore (approved):
Please add an Issue for this work.

components/Banner/tests/snapshot/MDCBannerSnapshotTests.m
fluxminds: The difference was one word. The bug happens on an edge case, and I didn't want to change these two contants long and extraLong for diff concerns.

components/Banner/tests/snapshot/MDCBannerSnapshotTests.m
fluxminds: Addressed in 09ce09c952e7ca194c3dcee1a4816819aef2cbbe.

Issue #7866, opened 2019-07-11
[Slider] Internal issue: b/137183548


material-mirror:
This was filed as an internal issue. If you are a Googler, please visit [b/137183548](http://b/137183548) for more details.

---
#### Internal data
- Associated internal bug: [b/137183548](http://b/137183548)

PR #7960, opened 2019-07-18
[Slider] Add snapshot test for dynamic color with elevation.

fluxminds:
Prepare for https://github.com/material-components/material-components-ios/issues/7866.


codeman7 (approved):
Should we make the `trackColor` something other than gray like yellow?

fluxminds:
@codeman7 I saw your comment after merging it :( Sorry about it. I will open another PR to change color from gray to yellow :)

PR #7962, opened 2019-07-18
[Elevation] Add elevationDidChangeBlock to MDCElevation protocol

codeman7:
This change adds a block that to the MDCElevation protocol so that clients can react to elevation changes within this new block.

Related to #7964

### DIFF
--- a/components/Elevation/src/MDCElevation.h
+++ b/components/Elevation/src/MDCElevation.h
@@ -25,4 +25,14 @@
  */
 @property(nonatomic, assign, readonly) CGFloat mdc_currentElevation;
 
+/**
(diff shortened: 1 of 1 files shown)


components/Elevation/src/MDCElevation.h
```diff
@@ -25,4 +25,15 @@
  */
 @property(nonatomic, assign, readonly) CGFloat mdc_currentElevation;
 
+/**
+ This block is called when the elevation changes for the conforming @c UIView or @c UIViewController
+ reciever or one of it's direct ancestors in the view hierarchy.
```
bryanoltman: NB: s/it's/its/

bryanoltman (approved):

PR #7992, opened 2019-07-23
[BottomNavigation] Store child VCs in `childViewControllers`.

### LINKED ISSUE #7984: [BottomNavigation] Use `childViewControllers` instead of a custom property
`MDCBottomNavigationBarController` should keep all item view controllers in its `childViewControllers` property so that state restoration works as expected. Instead of keeping a custom property, `viewControllers`, it should rely on the superclass `childViewControllers` APIs.

romoore:
The `MDCBottomNavigationBarController` was incorrectly only adding its child
view controllers as "proper" child view controllers when they became the
selected view controller. This resulted in difficulties when attempting to
perform state restoration within apps with the controller as the root view
controller.

Closes #7984


components/BottomNavigation/src/MDCBottomNavigationBarController.m
```diff
@@ -119,8 +118,10 @@ - (void)setSelectedViewController:(nullable UIViewController *)selectedViewContr
   }
 
   // Remove current VC and add new one.
-  [self removeContentViewController:self.selectedViewController];
-  [self addContentViewController:selectedViewController];
+  [self.selectedViewController.view removeFromSuperview];
+  [self.content addSubview:selectedViewController.view];
+  [selectedViewController didMoveToParentViewController:self];
```
nokeeo: I think `didMoveToParentViewController:` should be called when the view controller is added as a child.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
```diff
@@ -152,13 +153,32 @@ - (void)setSelectedIndex:(NSUInteger)selectedIndex {
   [self updateViewsForSelectedIndex:selectedIndex];
 }
 
+- (void)removeExistingViewControllers {
```
nokeeo: Nit: Please move private methods under private methods section.

nokeeo (approved):
Thanks for making this change!

components/BottomNavigation/src/MDCBottomNavigationBarController.m
romoore: The docs suggest that it's called immediately after calling `addSubview:`. So it's called every time the view is added as a subview. I have some honest concerns about how to test VC/view lifecycle for this situation.  If you have any ideas, I'd welcome the suggestions.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
romoore: Done.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
nokeeo: I think the docs suggest calling it immediately after `addChildViewController:`.

It also appears that this is the behavior of `UITabBarController`.

What are your concerns regarding testing?

> If you are implementing your own container view controller, it must call the didMoveToParentViewController: method of the child view controller after the transition to the new controller is complete or, if there is no transition, immediately after calling the addChildViewController: method.

romoore: Ok. I might have been too tired reading the docs and UIVC programming guides last night. >_<

nokeeo: :) no worries!

components/BottomNavigation/src/MDCBottomNavigationBarController.m
```diff
@@ -325,33 +351,12 @@ - (void)handleNavigationBarLongPressEndedForPoint:(CGPoint)point {
...
-  BOOL doesNotContainViewController = ![self.childViewControllers containsObject:viewController];
-  if (viewController && doesNotContainViewController) {
-    [self addChildViewController:viewController];
-    [self.content addSubview:viewController.view];
-    [self addConstraintsForChildViewControllerView:viewController.view];
-    [viewController didMoveToParentViewController:self];
+- (void)removeExistingViewControllers {
+  NSArray<UIViewController *> *childViewControllers = self.childViewControllers;
+  for (UIViewController *childViewController in childViewControllers) {
+    [childViewController.view removeFromSuperview];
```
bryanoltman, line 358: As per [the docs](https://developer.apple.com/library/archive/featuredarticles/ViewControllerPGforiPhoneOS/ImplementingaContainerViewController.html), a call to `willMoveToParentViewController:` with a nil argument is required before `removeFromSuperview`.

bryanoltman (changes requested):

components/BottomNavigation/src/MDCBottomNavigationBarController.m
nokeeo, line 358: I think we also need to call `[childViewController willMoveToParentViewController:nil];` before we remove the view controller as a child and remove its view from the hierarchy.

nokeeo (changes requested):

components/BottomNavigation/src/MDCBottomNavigationBarController.m
nokeeo: Sorry didn't see the previous review.

nokeeo (approved):
Looks good after @bryanoltman recommended changes.  Thanks making these!

components/BottomNavigation/src/MDCBottomNavigationBarController.m
romoore: Done.

components/BottomNavigation/src/MDCBottomNavigationBarController.m
romoore: Done.

bryanoltman (approved):

PR #7993, opened 2019-07-23
[Banner] use UITextView for text instead of UILabel.

### LINKED ISSUE #7990: [Banner] Internal issue: b/137048479
This was filed as an internal issue. If you are a Googler, please visit [b/137048479](http://b/137048479) for more details.

---
#### Internal data
- Associated internal bug: [b/137048479](http://b/137048479)

fluxminds:
closes #7990 

This API change enables more flexibility on the text configuration on banner. Minor visual difference was generated by UITextView's implementation.


romoore (dismissed):
Looks like the location of periods changed with this. The new locations don't look right.

romoore (dismissed):
https://github.com/material-components/material-components-ios/pull/7993/files?short_path=7221bdf#diff-7221bdfe9f773c384a948aa555959ba7 Doesn't look like an existing bug - it looks like this change is somehow rendering the text incorrectly. Since this is changing the text-rendering view from a UILabel to a UITextView, it makes sense that this change may be the cause. 

You suggest that it's an existing issue with the snapshot class - do you think this class is flaky?

fluxminds:
I don't think the snapshot class is flaky. I am not sure about what is causing the difference (which I need to investigate) between the example screenshot and the snapshot screenshot. The period definitely doesn't look right in the snapshot ones.

romoore:
If you like, we can work together to identify the issue. It may be a Bi-Di issue.  I saw the new Snapshot change forces text writing direction right-to-left even when the text is an LTR script like Latin or English.

fluxminds:
Updated the test with RTL script (Arabic). The issue with Bi-Di will be followed up separately:)

randallli:
why are we switching away from a label?

"UITextView defines a text view, which displays multiple lines of editable text."
from https://developer.apple.com/library/archive/documentation/StringsTextFonts/Conceptual/TextAndWebiPhoneOS/UsingTextClasses/UsingTextClasses.html

fluxminds:
@randallli It is more configurable in terms of clickable content on this view. #7990 provides more context on it.

romoore (changes requested):
Looks like there are some remaining visual improvements/adjustments needed.

I see the hyperlink formatting changed. Is this intentional that clients will "lose" underline?  I'm concerned that it may make Banner less accessible.

Looks like the new TextView needs a baseline alignment with the button.

|Before|After|
|---|---|
|banner_baseline_before|banner_baseline_after|

romoore:
It also looks like the Arabic text is being clipped.

|Before|After|
|---|---|
|banner_arabic_diacritic_before|banner_arabic_diacritic_after|

fluxminds:
@romoore Thanks for catching these two. I think the Arabic clipping is expected, the original implementation allows text showing outside of bounds, which was unexpected. It has been corrected because of the UITextView change. 

For the alignment one, the button text doesn't need to be aligned with baseline. The button bound itself is centered together with the text view which is the requirement. So I think the change happening there is acceptable. I suspect the line spacing issue is something caused by UITextView. I tried to remove the font leading but didn't affect the display (because this font doesn't have leading that could affect line spacing) Since attributed text is allowed, I think users can adjust the spacing by themselves.

What's your thoughts?

romoore:
If the alignment change is expected, that's fine. I'm more worried about clipping the font glyphs. Can you please just double-check that the UITextView's `intrinsicContentSize` or `contentSize` match its bounds? If UITextView has a bug, that's unfortunate but not part of this PR. If our layout has a bug I'd like to make sure it's corrected.

romoore:
What about the loss of underline for the hyperlink? I know this is already a breaking change, but aside from cautioning clients in the release notes is there anything the component code should be doing?

fluxminds:
@romoore I confirmed the bounds (which are returned by sizeThatFits/intrinsicContentSize) are the same for UILabel and UITextView. The difference was caused by the `clipsToBounds`. 

For the loss of underline, I prefer not to change the snapshot test here to align with the original appearance. Since it is under Beta, users can still use `NSUnderlineStyleAttributeName` in their code to keep the original appearance (underline for the hyperlink) if they would like, which is expected. How do you feel about it?

romoore (approved):

Issue #7995, opened 2019-07-23
[Tabs] Internal issue: b/138151252


material-mirror:
This was filed as an internal issue. If you are a Googler, please visit [b/138151252](http://b/138151252) for more details.

---
#### Internal data
- Associated internal bug: [b/138151252](http://b/138151252)

material-mirror:
The internal issue [b/138151252](http://b/138151252) is now closed. This issue is being closed as a result.

PR #8003, opened 2019-07-23
[Tabs] Allow clients to set preferred layout style.

romoore:
Clients can select the preferred layout style with a new public API. Whenever
possible, the preferred layout style will be used. When it is not possible, for
example if the bounds is too narrow, a reasonable default value will be chosen
instead.

Part of #7995


components/Tabs/src/TabBarView/MDCTabBarView.h
```diff
@@ -56,6 +75,13 @@ __attribute__((objc_subclassing_restricted)) @interface MDCTabBarView : UIScroll
  */
 @property(nullable, nonatomic, copy) UIColor *selectionIndicatorStrokeColor;
 
+/**
+ The preferred layout style of the bar.  If possible, this layout style will be used. If not,
+ another style will be used. For example, if the bar is too narrow for a Fixed layout style, then a
+ Scrollable layout style may be used instead.
```
bryanoltman: Please document which layout style the Tab Bar View defaults to if this isn't set, _Rob_.

bryanoltman (changes requested):

components/Tabs/src/TabBarView/MDCTabBarView.h
romoore: Done.

bryanoltman (approved):

PR #8053, opened 2019-07-23
[Elevation] Add MDCElevationOverride header to umbrella header

codeman7:
Follow up from #8034 where I missed added the new header to the umbrella header.


codeman7:
@romoore beat me to it.

PR #8069, opened 2019-07-24
[Buttons] Move MDCButton's defaultContentInset configuration to later in init

### LINKED ISSUE #7983: [NavigationBar] Internal issue: b/136088498
This was filed as an internal issue. If you are a Googler, please visit [b/136088498](http://b/136088498) for more details.

---
#### Internal data
- Associated internal bug: [b/136088498](http://b/136088498)

bryanoltman:
After digging in a bit, I've found that this is an issue with UIButton in iOS 13 (as of Xcode 11 beta 4). To reproduce:

1) Create a `UIButton`.
2) Create a `UIImage` and flip it using `imageWithHorizontallyFlippedOrientation`.
3) *Before setting the button's image*, give the button non-zero contentEdgeInsets (I used `{20, 10, 20, 10}`).
4) Set the image on the button.

You should be able to see that the image is not visible, and that the button's internal UIImageView's bounds.origin is `{ -(i.left + i.right), -(i.top + i.bottom) }` (where `i` is the button's contentEdgeInsets). In iOS 12, the button's imageView's bounds.origin is `{0, 0}`.

Closes #7983


romoore (approved):
Please add a unit test that fails on iOS 13 simulators without this fix.

romoore (commented):
Can you please verify that this is also a defect with Xcode 11 beta 1 or beta 2?

bryanoltman:
Confirmed this is an issue in Xcode 11 Beta 1, will add a snapshot test

PR #8077, opened 2019-07-24
[Color] add method for regular color blending.

fluxminds:
Prepare for b/137091265.

This method is copied from the one in MDCSemanticColorScheme. 

We need to add it here for a clear structure and to avoid depending on the scheme target for low level elevation classes. (This method intuitively doesn't need to exist under scheme). 

The original method in MDCSemanticColorScheme will need to redirect/call this method internally if we don't want to deprecate it as a public available API, which will be addressed in a follow-up PR.

### DIFF
--- a/components/private/Color/src/UIColor+MaterialBlending.h
+++ b/components/private/Color/src/UIColor+MaterialBlending.h
@@ -25,8 +25,8 @@ NS_ASSUME_NONNULL_BEGIN
  @param color UIColor value that sits on top.
  @param backgroundColor UIColor on the background.
  */
-+ (nonnull UIColor *)blendColor:(nonnull UIColor *)color
-            withBackgroundColor:(nonnull UIColor *)backgroundColor;
++ (nonnull UIColor *)mdc_blendColor:(nonnull UIColor *)color
+                withBackgroundColor:(nonnull UIColor *)backgroundColor;
 
 @end
 
--- a/components/private/Color/src/UIColor+MaterialBlending.h
+++ b/components/private/Color/src/UIColor+MaterialBlending.h
@@ -1,0 +1,33 @@
+// Copyright 2019-present the Material Components for iOS authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


components/private/Color/src/UIColor+MaterialBlending.h
```diff
@@ -0,0 +1,33 @@
...
+@interface UIColor (MaterialBlending)
+
+/**
+ Blending a color over a background color using Alpha compositing technique.
+ More info about Alpha compositing: https://en.wikipedia.org/wiki/Alpha_compositing
+
+ @param color UIColor value that sits on top.
+ @param backgroundColor UIColor on the background.
+ */
++ (nonnull UIColor *)blendColor:(nonnull UIColor *)color
```
romoore: Please prefix category methods with `mdc_` when the original class is in another library.

romoore (dismissed):

components/private/Color/tests/unit/MaterialColorTests.m
```diff
@@ -77,4 +79,112 @@ - (void)testDynamicColorWhenUserInterfaceStyleIsLightForPreiOS13 {
...
+  UIColor *expectedColor = [UIColor colorWithRed:(CGFloat)0.32857142857142863
+                                           green:(CGFloat)0.64761904761904765
+                                            blue:(CGFloat)0.87619047619047618
+                                           alpha:(CGFloat)0.84000000000000008];
+  UIColor *resultColor = [UIColor blendColor:blendColor withBackgroundColor:backgroundColor];
+  XCTAssertTrue([self compareColorsWithFloatPrecisionFirstColor:resultColor
+                                                    secondColor:expectedColor]);
+}
+
+- (void)testHBSColorMergeTest {
```
romoore, line 156: Please also add tests for the P3 color space.

https://developer.apple.com/documentation/uikit/uicolor/1648568-init

components/private/Color/src/UIColor+MaterialBlending.h
fluxminds: Addressed in 3056d46034ed83dc3bf8ab68897ba95c0551e4bc.

components/private/Color/tests/unit/MaterialColorTests.m
fluxminds: Addressed in 5b40c949c97f1940aa6bc79f4118b5a019fe3e3c

fluxminds:
Addressed in the pattern color in f34797c4bca5276b86c54a61b7f00ab636cd0b4c.

yarneo (approved):

PR #8087, opened 2019-07-25
[Typography] Internal comments to explain font scaling.

romoore:
[Typography] Internal comments to explain font scaling.


romoore:
The original commit came from a Googler and has been exported to GitHub. I don't know why our export account isn't working correctly.

yarneo (approved):

Issue #8086, opened 2019-07-25
Internal issue: b/137091265


material-mirror:
This was filed as an internal issue. If you are a Googler, please visit [b/137091265](http://b/137091265) for more details.

---
#### Internal data
- Associated internal bug: [b/137091265](http://b/137091265)

PR #8091, opened 2019-07-25
[Color] Create umbrella header.

romoore:
All components should have a top-level umbrella header for their includes.
This allows easier refactoring of classes and files within the component.
Creating an umbrella for Color and using it outside the component.

Part of #8086


fluxminds (approved):
Thanks!

Issue #8020, opened 2019-07-23
[Dialogs] Internal issue: b/138204422


material-mirror:
This was filed as an internal issue. If you are a Googler, please visit [b/138204422](http://b/138204422) for more details.

---
#### Internal data
- Associated internal bug: [b/138204422](http://b/138204422)

PR #8098, opened 2019-07-26
[Dialogs] Add MaterialElevation support to AlertController.

romoore:
Adds conformance to `MDCElevatable` and `MDCElevationOverriding` to
`MDCAlertController`.

Part of #8020


components/Dialogs/src/MDCAlertController.m
```diff
@@ -282,8 +286,20 @@ - (void)setCornerRadius:(CGFloat)cornerRadius {
...
+  if (shouldNotifyChanges) {
+    [self.view mdc_elevationDidChange];
+  }
+}
+
+- (CGFloat)mdc_currentElevation {
+  return self.elevation;
+}
+
+- (CGFloat)mdc_overrideBaseElevation {
```
bryanoltman: This function should be removed as it's autogenerated by `@synthesize`

bryanoltman (approved):

components/Dialogs/src/MDCAlertController.m
romoore: Ah, yes, sorry. I put that there for debugging.

PR #8108, opened 2019-07-29
[BottomSheet] Mark init unavailable

### LINKED ISSUE #8109: [BottomSheet] Initializing with `init` causes a crash
MDCBottomSheetController does not allow me to initialize with the `init` method but that method is not marked unavailable. If I initialize with that method in `viewDidLoad` my app crashes.

Screen Shot 2019-07-29 at 7 00 45 AM

codeman7:
Currently the MDCBottomSheet controller does not mark its `init` method as unavailable but the contentViewController is marked as `readonly` so clients would have to _initialize_ with a contentViewController. If the contentViewController is `nil` then when `addChildViewController` is called in `viewDidLoad` this causes a crash as you cannot pass `nil` to an array.

Closes #8109

### DIFF
--- a/components/BottomSheet/src/MDCBottomSheetController.h
+++ b/components/BottomSheet/src/MDCBottomSheetController.h
@@ -110,6 +110,8 @@
  */
 @property(nonatomic, assign) MDCShadowElevation elevation;
 
+- (nonnull instancetype)init NS_DEPRECATED_IOS(9_0, 9_0, "Use initWithContentViewController:");
+
 /**
  Sets the shape generator for state that is used to define the bottom sheet's shape for that state.
 
--- a/components/BottomSheet/src/MDCBottomSheetController.h
+++ b/components/BottomSheet/src/MDCBottomSheetController.h
@@ -111,7 +111,10 @@
  */
 @property(nonatomic, assign) MDCShadowElevation elevation;
 
-- (nonnull instancetype)init NS_DEPRECATED_IOS(9_0, 9_0, "Use initWithContentViewController:");
+/**
+ Bottom sheet controllers must be created with @c initWithContentViewController:.
+ */
+- (nonnull instancetype)init NS_UNAVAILABLE;
 
 /**
  Initializes the controller with a content view controller.
--- a/components/BottomSheet/src/MDCBottomSheetController.m
+++ b/components/BottomSheet/src/MDCBottomSheetController.m
@@ -61,12 +61,14 @@
   [super viewDidLoad];
 
   self.view.preservesSuperviewLayoutMargins = YES;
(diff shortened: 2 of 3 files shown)


components/BottomSheet/src/MDCBottomSheetController.h
```diff
@@ -110,6 +110,8 @@
  */
 @property(nonatomic, assign) MDCShadowElevation elevation;
 
+- (nonnull instancetype)init NS_DEPRECATED_IOS(9_0, 9_0, "Use initWithContentViewController:");
```
romoore: Shouldn't this be `NS_UNAVAILABLE`?

romoore: You mention in the PR summary that there's a potential crash. Unless you mark `init` is unavailable, please add guards to `viewDidLoad` to prevent the crash.

romoore (changes requested):

romoore:
Please file an Issue for this crash/bug and link to it from the PR summary.

components/BottomSheet/src/MDCBottomSheetController.h
codeman7: After checking internally nobody is using this. But, there is a chance clients are using it via CocoaPods, I think it is better to do deprecated this week and unavailable next what do you think?

codeman7: Updated.

romoore: My understanding is that it always caused a crash in any client that used it.  In that case this is a bug fix and not a breaking change. 

If I've misunderstood the issue, please try to make the referenced Issue and/or PR summary a bit more explicit about the crash conditions.

codeman7: If a client never calls `viewDidLoad` then the crash wouldn't occur. For instance if they were using the `init` in a test file.

```objc
- (void)testFoo {
  MDCBottomSheetController *bottomSheet = [[MDCBottomSheetController alloc] init];

  bottomSheet.elevation = 5;

  XCTAssertEqualWithAccuracy(bottomSheet.elevation, 5, 0.001);
}
```

That would not cause a crash but if they were using it in production to present a bottom sheet then the crash would occur.

romoore: I think it's fair to assume that being unable to use `init` in production meant that nobody was using it.  I think we should mark `-init` unavailable and consider this to be a non-breaking bug fix PR.  Is that OK with you?

codeman7: Done

romoore (approved):

PR #8146, opened 2019-07-31
[Ink, Ripple] Use static default colors

### LINKED ISSUE #8141: [Ink, Ripple] Default colors are reallocated

codeman7:
Instead of reallocating new colors every time a default is required, using a static color can reduce memory use and allocations.

Impact: “memory watermark would be dropped in terms of KB. Memory fragmentation, cache performance, malloc contention will also be reduced slightly.”

Closes #8141


jverkoey (approved):
LGTM; I've asked the original author for more details about the impact of this change so that we may include it in our release notes. Will post back here once we hear back.

yarneo (approved):

Issue #8150, opened 2019-08-01
{*} Use Skylark macros as much as possible


romoore:
## Acceptance Criteria

All component BUILD files should make use of our custom macros  as possible. This will enable easier BUILD file transform in the future.

---
#### Internal data
- Associated internal bug: [b/138788911](http://b/138788911)

PR #8152, opened 2019-08-01
[ActionSheet] BUILD file uses Skylark macros.

romoore:
Updates the BUILD file to use as many Skylark macros as possible to reduce
boilerplate and make it easier to transform the BUILD files when releasing each
week.

Part of #8150


bryanoltman (approved):

jverkoey (approved):
Love it!

PR #8177, opened 2019-08-02
[NavigationDrawer] Use more Starlark macros.

romoore:
Uses additional Starlark macros in the BUILD files and cleans-up the file as
well. Makes releasing easier.

Part of #8150

### DIFF
--- a/components/NavigationDrawer/BUILD
+++ b/components/NavigationDrawer/BUILD
@@ -21,6 +21,7 @@ load(
     "mdc_public_objc_library",
     "mdc_snapshot_objc_library",
     "mdc_snapshot_test",
+    "mdc_unit_test_objc_library",
     "mdc_unit_test_suite",
 )
 load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library")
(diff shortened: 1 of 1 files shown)


components/NavigationDrawer/BUILD
```diff
@@ -52,20 +49,11 @@ mdc_extension_objc_library(
...
 mdc_objc_library(
     name = "private",
+    testonly = 1,
     hdrs = native.glob(["src/private/*.h"]),
-    includes = ["src/private"],
     visibility = ["//visibility:private"],
 )
 
-mdc_objc_library(
-    name = "TestingHeader",
```
fluxminds, line 61: I am not sure if this target is being used somewhere, maybe put some context in the PR description?

romoore: It couldn't have been used since removing it caused no problems.

romoore: What context would you like in the description? This target wasn't used and was part of the "clean up".

fluxminds (approved):

PR #8190, opened 2019-08-02
[Cards] Update snapshot golden images.

### LINKED ISSUE #8189: [Cards] MDCCardSnapshotTest is failing on Xcode 11.
MDCCardSnapshotTest is failing on Xcode 11.

---
#### Internal data
- Associated internal bug: [b/138927546](http://b/138927546)

fluxminds:
closes https://github.com/material-components/material-components-ios/issues/8189.


romoore (dismissed):
Why are the image references so much larger than before? It looks like you might have committed the actual image assets rather than using Git LFS.  Can you please check that these files are still just LFS links and not actual images?

If you do a clean clone of the repository without Git LFS, then check out your PR branch, you should be able to edit the files as text.

romoore (commented):
Disregard my previous comment - I guess GitHub shows the diffs of the LFS files. However, they don't seem to be loading for me. Can you please verify that they are OK and perhaps it's just a bug in GitHub's web UI?

fluxminds:
@romoore Good catch. I found the image contents are the same after I compared them. It doesn't show because it was changed from a LFS file to real asset. And I noticed that was caused by the turned off LFS. Card was one of the earliest ones so it was not affected. I sent https://github.com/material-components/material-components-ios/pull/8221 for that issue.

Closing this because I think Card is fine.

PR #8221, opened 2019-08-05
Fix snapshots that were not stored on LFS.

### LINKED ISSUE #8189: [Cards] MDCCardSnapshotTest is failing on Xcode 11.

fluxminds:
After fixing the lfs attribute in https://github.com/material-components/material-components-ios/pull/6350. We should turn git LFS back on to make sure new snapshots are stored on LFS. Snapshots committed after the change were not stored on LFS as a result. Because of it, some of our snapshots are stored on LFS, the rest are stored in binary file, which causes our build system is not able to run snapshot tests on those LFS snapshots.

This PR deletes all snapshot image binary files and adds them back in Git LFS file format. After turning the LFS back, Git is able to use LFS for our build/test system.

It also fixes https://github.com/material-components/material-components-ios/issues/8189.


jverkoey (changes requested):
The PR description indicates that this PR should fix our LFS storage, but the PR itself appears to be deleting a large number of images from the repo. Please expand in the PR description on what was broken, why this PR fixes it, and why the PR appears to be deleting a large number of images.

fluxminds:
I updated the PR description to clarify the problem :)

jverkoey (approved):
PR description is much improved, thank you!

PR #8230, opened 2019-08-07
[ActionSheet] Add ability to add a divider between header and table

### LINKED ISSUE #7395: [ActionSheet] Add Optional Separator in Action Sheet
This was filed as an internal issue. If you are a Googler, please visit [b/132725818](http://b/132725818) for more details.

---
#### Internal data
- Associated internal bug: [b/132725818](http://b/132725818)
- Blocked by: https://github.com/material-components/material-components-ios/issues/7281

codeman7:
This PR adds an API 

*new property* `*headerDividerColor` on _MDCActionSheetController_
*new property* `*showHeaderDivider` on _MDCActionSheetController_

that will allow clients to add a divider to the action sheet between the header and the table. If a client wants to show a header divider they would need to set `showHeaderDivider` and then set an appropriate color for their use case, by default it is `clearColor`. By default the property `showHeaderDivider` view is `NO` so that we do not break clients. This matches a pattern outlined in [`MDCBannerView`](https://github.com/material-components/material-components-ios/blob/ac114b4bda9dbad328ae445c4529a43a94ccd97d/components/Banner/src/MDCBannerView.h#L76-L88).

Closes #7395


fluxminds:
Two properties (`showDivider`, `dividerColor`) were used in Banner for clarity. Some context on the discussion can be found in https://github.com/material-components/material-components-ios/pull/7567. Maybe it would be helpful to keep the consistence for users?

components/ActionSheet/src/MDCActionSheetController.m
```diff
@@ -175,7 +179,14 @@ - (void)viewWillLayoutSubviews {
   }
   CGSize size = [self.header sizeThatFits:CGRectStandardize(self.view.bounds).size];
   self.header.frame = CGRectMake(0, 0, self.view.bounds.size.width, size.height);
-  UIEdgeInsets insets = UIEdgeInsetsMake(self.header.frame.size.height, 0, 0, 0);
+  if (self.title != nil || self.message != nil) {
```
romoore: Shouldn't this be checking the `headerDividerColor` to see if it's `nil`? As I understand the header documentation, it's the color that determines if the view is added.

components/ActionSheet/src/MDCActionSheetController.m
```diff
@@ -175,7 +179,14 @@ - (void)viewWillLayoutSubviews {
...
+    [self.view addSubview:self.headerDividerView];
```
romoore: Won't this re-add the divider view on every layout pass? Can you check to see if the divider view's `superview` is `self.view`? Or maybe to the `addSubview:` and `removeFromSuperview` in the `setHeaderDividerColor:` method?

components/ActionSheet/tests/unit/MDCActionSheetTest.m
```diff
@@ -457,4 +463,28 @@ - (void)testSettingBaseOverrideBaseElevationReturnsSetValue {
   XCTAssertEqualWithAccuracy(self.actionSheet.mdc_overrideBaseElevation, fakeElevation, 0.001);
 }
 
+- (void)testSetHeaderDividerColor {
+  // Given
+  UIColor *fakeColor = UIColor.orangeColor;
```
romoore: Tip: Nit: Why is this a fake color?  It looks like a real color.

components/ActionSheet/tests/unit/MDCActionSheetTest.m
```diff
@@ -457,4 +463,28 @@ - (void)testSettingBaseOverrideBaseElevationReturnsSetValue {
   XCTAssertEqualWithAccuracy(self.actionSheet.mdc_overrideBaseElevation, fakeElevation, 0.001);
 }
 
+- (void)testSetHeaderDividerColor {
```
romoore, line 469: Please add a test that setting the `headerDividerColor` causes the bounds/frame to increase by exactly 1 point.

components/ActionSheet/src/MDCActionSheetController.m
romoore: `-[UIView addSubview:]` requires a `nonnull` argument and it looks like `headerDividerView` can be `nil` at this point if the divider color is `nil`.

romoore (commented):
The documentation and implementation don't seem to match. The docs suggest that the divider view is fully controlled by the color, but the implementation seems to rely on a title or message to exist.  It also appears like the code is potentially passing `nil` to the `nonnull` `addSubview:` API.

romoore (dismissed):
Oops, meant to Request Changes. Please see my previous review comments.

components/ActionSheet/tests/unit/MDCActionSheetTest.m
codeman7: It only changes the contentOffset of the tableView and I have added test for that.

components/ActionSheet/src/MDCActionSheetController.m
codeman7: This has been updated the headerDividerView is never nil.

components/ActionSheet/src/MDCActionSheetController.m
codeman7: This has been removed.

components/ActionSheet/tests/unit/MDCActionSheetTest.m
codeman7: It was just to make sure the color we set is the same color in the expectation, I have changed the name to be `expectedColor` to clarify this more.

components/ActionSheet/src/MDCActionSheetController.m
codeman7: The headerDividerView is no nonnull.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -203,6 +203,21 @@ __attribute__((objc_subclassing_restricted)) @interface MDCActionSheetController
...
+ Defaults to NO.
+ */
+@property(nonatomic, assign) BOOL showHeaderDivider;
+
+/**
+ The color of the divider between the header and actions.
+
+ Defaults to UIColor.clearColor.
+ */
+@property(nonatomic, copy, nonnull) UIColor *headerDividerColor;
```
romoore, line 219: I don't really understand why there are two separate properties controlling the visibility for this view. Could you help me understand why the two properties are necessary?

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -203,6 +203,21 @@ __attribute__((objc_subclassing_restricted)) @interface MDCActionSheetController
  */
 @property(nonatomic) UIImageRenderingMode imageRenderingMode;
 
+/**
+ Determines if a divider should be shown between the header and actions. To customize the divider
+ color see @c headerDividerColor.
+
+ Defaults to NO.
+ */
+@property(nonatomic, assign) BOOL showHeaderDivider;
```
romoore: Based on the code, this only has any effect if there is either a title or message (or both).  Please update the documentation (or implementation) to reflect the correct behavior.

romoore: Nit/naming: I believe `showsHeaderDivider` or `headerDividerVisible` would better align with Apple's naming conventions.
*  `setShowsHeaderDivider` (setter) and `showsHeaderDivider` (getter)
*  `setHeaderDividerVisible` (setter) and `isHeaderDividerVisible` (getter)

From [Apple's Naming guidelines](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingMethods.html), "If the property is expressed as a verb, …the verb should be in the simple present tense."

components/ActionSheet/src/MDCActionSheetController.m
```diff
@@ -393,6 +408,12 @@ - (UIColor *)messageTextColor {
   return self.header.messageTextColor;
 }
 
+- (void)setHeaderDividerColor:(UIColor *)headerDividerColor {
+  _headerDividerColor = [headerDividerColor copy];
```
romoore: Tip: If the `headerDividerView` is `nonnull`, can this just be a pass-through property?

components/ActionSheet/tests/snapshot/MDCActionSheetControllerSnapshotTests.m
```diff
@@ -871,4 +871,55 @@ - (void)testThreeActionsSufficientSizeShortTextLTRWithDefaultPresentationStyleOn
   [self snapshotVerifyViewForIOS13:window];
 }
 
+- (void)testActionSheetWithHeaderShown {
```
romoore, line 874: Should there be an iOS 13 dynamic color test? Will that be done in a follow-up for the theming extension?

components/ActionSheet/src/MDCActionSheetController.h
codeman7: I didn't really like the _magic_ I had added in that situation. I think its better just to have clients have full control over showing the headerDivider or not regardless of the title and message being set. This has been cleaned up.

components/ActionSheet/src/MDCActionSheetController.m
codeman7: Whoops, forgot to clean this up after addressing other things, thanks for catching this!

components/ActionSheet/src/MDCActionSheetController.h
codeman7: This has been updated.

components/ActionSheet/tests/snapshot/MDCActionSheetControllerSnapshotTests.m
codeman7: I don't believe so. I think we only test that dynamic colors are supported for `CGColor` instances but this is setting a UIColor on a UIView so I think we are okay.

components/ActionSheet/src/MDCActionSheetController.h
codeman7: This is following the convention set in #7567, I think this is what @wenyuzhang666 and @jverkoey would prefer.

romoore: While I don't believe the boolean is strictly necessary, in #7567 the default value is visible. It's documented as "light grey". In this case it appears the default color is clear.  In order to use the property on Banner the user only needs to toggle the boolean flag. However here, they need to toggle the boolean *and* set a non-clear color. I think there needs to be a visible default color for the divider so users don't have to take two steps to enable the divider.

I'd also like to note that in [the relevant conversation thread for #7567](https://github.com/material-components/material-components-ios/pull/7567/files#r293004510), having both properties is referred to as an "acceptable path forward". This would seem like there was a compromise made at some point and the final decision was a result of that compromise.  Unfortunately the reasons for the decision aren't entirely clear, except a reference to theming extensions. I would be cautious about using that decision as the only basis for this one.

components/ActionSheet/src/MDCActionSheetController.m
```diff
@@ -393,6 +407,23 @@ - (UIColor *)messageTextColor {
...
+  self.headerDividerView.backgroundColor = headerDividerColor;
+  [self.view setNeedsDisplay];
+}
+
+- (UIColor *)headerDividerColor {
+  return self.headerDividerView.backgroundColor;
+}
+
+- (void)setShowsHeaderDivider:(BOOL)showsHeaderDivider {
+  if (_showsHeaderDivider == showsHeaderDivider) {
```
romoore: Tip: Can this be bound to the [`hidden` property on the view](https://developer.apple.com/documentation/uikit/uiview/1622585-hidden)? I believe that property will take care of any layout/display updates required.

components/ActionSheet/src/MDCActionSheetController.m
```diff
@@ -175,7 +181,15 @@ - (void)viewWillLayoutSubviews {
   }
   CGSize size = [self.header sizeThatFits:CGRectStandardize(self.view.bounds).size];
   self.header.frame = CGRectMake(0, 0, self.view.bounds.size.width, size.height);
-  UIEdgeInsets insets = UIEdgeInsetsMake(self.header.frame.size.height, 0, 0, 0);
+  CGFloat dividerHeight = 1;
```
romoore: Tip: I'm curious about this block of logic.  Is there a reason the same frame cannot be used in both cases?  If the height is zero, then it seems it will have the same effect (not resizing the view) of setting the frame to `CGRectZero`.

```objc
CGFloat dividerHeight = self.showsHeaderDivider ? 1 : 0;
self.headerDividerView.frame =
        CGRectMake(0, size.height, CGRectGetWidth(self.view.bounds), dividerHeight);
UIEdgeInsets insets = UIEdgeInsetsMake(size.height + dividerHeight, 0, 0, 0);
```

components/ActionSheet/src/MDCActionSheetController.m
```diff
@@ -393,6 +407,23 @@ - (UIColor *)messageTextColor {
   return self.header.messageTextColor;
 }
 
+- (void)setHeaderDividerColor:(UIColor *)headerDividerColor {
+  self.headerDividerView.backgroundColor = headerDividerColor;
+  [self.view setNeedsDisplay];
```
romoore: Tip: Is this call necessary? I'm surprised that setting the `backgroundColor` isn't causing a layout pass. When I attempted to reproduce the need for `setNeedsDisplay`, the subview's updated `backgroundColor` is shown immediately.

At the least, please limit forced display updates to the headerDividerView's frame:

```objc
// Since the frame is a computed property, rely on the divider's bounds.
[self.view setNeedsDisplayInRect:[self.view convertRect:self.headerDividerView.bounds 
                                               fromView:self.headerDividerView]];
```

components/ActionSheet/tests/snapshot/MDCActionSheetControllerSnapshotTests.m
romoore: Acknowledged.

components/ActionSheet/tests/snapshot/MDCActionSheetControllerSnapshotTests.m
```diff
@@ -871,4 +871,55 @@ - (void)testThreeActionsSufficientSizeShortTextLTRWithDefaultPresentationStyleOn
...
+  self.actionSheetController.showsHeaderDivider = YES;
+  self.actionSheetController.headerDividerColor = UIColor.blackColor;
+  [self.actionSheetController.view setNeedsLayout];
+  [self.actionSheetController.view layoutIfNeeded];
+
+  // Then
+  [self generateSnapshotAndVerifyForView:self.actionSheetController.view];
+}
+
+- (void)testActionSheetWithHeaderShownButNoTitleOrMessage {
```
romoore, line 899: Looks like the golden image for this test needs updating.

components/ActionSheet/src/MDCActionSheetController.m
codeman7: Ah much cleaner, I didn't even consider that, thanks!

components/ActionSheet/src/MDCActionSheetController.m
codeman7: This has been updated.

components/ActionSheet/tests/snapshot/MDCActionSheetControllerSnapshotTests.m
codeman7: done

components/ActionSheet/src/MDCActionSheetController.m
codeman7: This is no longer needed since we don't need `setNeedsDisplay`

components/ActionSheet/src/MDCActionSheetController.h
codeman7: I've added a default color to match Banner.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -203,6 +203,22 @@ __attribute__((objc_subclassing_restricted)) @interface MDCActionSheetController
...
+/**
+ Determines if a divider should be shown between the header and actions. To customize the divider
+ color see @c headerDividerColor.
+
+ Defaults to NO.
+ */
+@property(nonatomic, assign) BOOL showsHeaderDivider;
+
+/**
+ The color of the divider between the header and actions. This API should be used with
```
romoore: Nit/spelling: s/visable/visible/ s/ligh/light/

romoore: Tip: You can probably remove the sentence, "This API should be used..." since I believe it's no longer required that updating this property is necessary to make the divider visible.

components/ActionSheet/src/MDCActionSheetController.m
```diff
@@ -23,6 +23,7 @@
 static NSString *const kReuseIdentifier = @"BaseCell";
 static const CGFloat kActionImageAlpha = (CGFloat)0.6;
 static const CGFloat kActionTextAlpha = (CGFloat)0.87;
+static const CGFloat kDividerDefaultOpacity = (CGFloat)0.12;
```
romoore: Nit: Please name this `kDividerDefaultAlpha` to match the names of similar constants above.

components/ActionSheet/src/MDCActionSheetController.h
```diff
@@ -203,6 +203,22 @@ __attribute__((objc_subclassing_restricted)) @interface MDCActionSheetController
...
+ @c showsHeaderDivider to make the divider visable.
+
+ Defaults to a ligh grey color.
```
romoore: Either this comment or the code below is incorrect. The color is not a light grey color, but a partially-transparent black color.  "Light grey" generally means an opaque white color with a white value greater than 0.5.

Perhaps `[UIColor colorWithWhite:0.88 alpha:1]` or something similar?  As an added benefit, opaque colors often render faster because no blending has to take place.

components/ActionSheet/src/MDCActionSheetController.m
```diff
@@ -131,6 +135,10 @@ - (instancetype)initWithTitle:(NSString *)title message:(NSString *)message {
     _actionTextColor = [UIColor.blackColor colorWithAlphaComponent:kActionTextAlpha];
     _actionTintColor = [UIColor.blackColor colorWithAlphaComponent:kActionImageAlpha];
     _imageRenderingMode = UIImageRenderingModeAlwaysTemplate;
+    _headerDividerView = [[UIView alloc] init];
+    _headerDividerView.backgroundColor =
+        [UIColor.blackColor colorWithAlphaComponent:kDividerDefaultOpacity];
+    ;
```
romoore: Nit: Extra semicolon here.

romoore (approved):
LGTM, but please correct the default color to be opaque or else update the documentation to reflect the documentation that it's actually a partially-transparent black.

components/ActionSheet/src/MDCActionSheetController.m

components/ActionSheet/src/MDCActionSheetController.h
codeman7: Updated the docs.

components/ActionSheet/src/MDCActionSheetController.m
codeman7: Updated, good catch!

components/ActionSheet/src/MDCActionSheetController.h
codeman7: Removed.

codeman7: Updated

PR #8241, opened 2019-08-08
[BottomNavigation] Propagate `accessibilityHint` to the accessibility element.

### LINKED ISSUE #7290: [BottomNavigation] UITabBarItem.accessibilityHint not assigned to item view
The `accessibilityHint` of the UITabBarItem is not being correctly passed through to the (button in) the item view. Users should be able to set the `accessibilityHint` of the item and VoiceOver should read it out.  Similarly, the value should be Key-Value Observed so changes are propagated to the item even after assignment.

---
#### Internal data
- Associated internal bug: [b/132166866](http://b/132166866)

romoore:
The UIButton subview of the Bottom Navigation bar should be the
one receiving the `accessibilityHint` value from the UITabBarItem. There's no
ill effect if the itemView also receives the hint, and preserving that
behavior is less likely to break clients.
    
In the long term, it's reasonable to remove the `accessibilityHint` from the
item view directly, since it's not an accessibility element and doesn't need
the hint to be set.
    
Closes #7290

### DIFF
--- a/components/BottomNavigation/src/MDCBottomNavigationBar.m
+++ b/components/BottomNavigation/src/MDCBottomNavigationBar.m
@@ -547,7 +547,13 @@ static NSString *const kOfAnnouncement = @"of";
           key, kMaterialBottomNavigationStringsTableName, [[self class] bundle], kOfString);
       NSString *localizedPosition =
           [NSString localizedStringWithFormat:itemOfTotalString, (i + 1), (int)items.count];
-      itemView.button.accessibilityHint = localizedPosition;
+      // Allow a custom `accessibilityHint` to be assigned even if "faking" a tab bar is enabled.
+      if (itemView.button.accessibilityHint.length) {
+        itemView.button.accessibilityHint =
+            [NSString stringWithFormat:@"%@. %@", localizedPosition, itemView.accessibilityHint];
+      } else {
+        itemView.button.accessibilityHint = localizedPosition;
+      }
     }
     if (item.image) {
       itemView.image = item.image;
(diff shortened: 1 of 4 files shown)


components/BottomNavigation/src/MDCBottomNavigationBar.m
```diff
@@ -547,7 +547,13 @@ - (void)setItems:(NSArray<UITabBarItem *> *)items {
           key, kMaterialBottomNavigationStringsTableName, [[self class] bundle], kOfString);
       NSString *localizedPosition =
           [NSString localizedStringWithFormat:itemOfTotalString, (i + 1), (int)items.count];
-      itemView.button.accessibilityHint = localizedPosition;
+      // Allow a custom `accessibilityHint` to be assigned even if "faking" a tab bar is enabled.
+      if (itemView.button.accessibilityHint.length) {
+        itemView.button.accessibilityHint =
+            [NSString stringWithFormat:@"%@. %@", localizedPosition, itemView.accessibilityHint];
```
bryanoltman, line 553: Should this be localized?

bryanoltman (approved):

components/BottomNavigation/src/MDCBottomNavigationBar.m
romoore: No, I don't think so.  The `accessibilityHint` could be first in ordering, but it's not a matter of localization. In other words, Hebrew speakers wouldn't expect the `accessibilityHint` to come at a different position than the "faked" positional data.

romoore: I chose the `accessibilityHint` to come last in this case since it typically comes last on accessibility elements.

PR #8303, opened 2019-08-13
[Snackbar] Add dynamic type M2 support.

fluxminds:
part of https://github.com/material-components/material-components-ios/issues/7428.

The message text and then button text need to follow the M2 typography scale (https://material.io/design/typography/the-type-system.html#type-scale). Adding a `adjustsFontForContentSizeCategoryWhenScaledFontIsUnavailable` flag to make the component compatible with the old behavior. This flag is YES by default.


romoore (approved):
Please update the PR summary to include a description of the changes/motivation and anything else you think its warranted. It may be useful to link to the Material Dynamic Type documentation as well.

PR #8315, opened 2019-08-14
[Color+Elevation] Add a convenience method to resolve dynamic color only when traitCollection's color appearance has changed.

fluxminds:
Add this convenience method would help supporting theming extension in an easier way that we don't need to check color difference on traitCollection every time like https://github.com/material-components/material-components-ios/pull/8285/files#diff-48ca693ee2870d02b06a42900a1dc69dR73.

### DIFF
--- a/components/Elevation/src/UIColor+MaterialElevation.h
+++ b/components/Elevation/src/UIColor+MaterialElevation.h
@@ -31,10 +31,9 @@
 - (nonnull UIColor *)mdc_resolvedColorWithElevation:(CGFloat)elevation;
 
 /**
- Returns a color that takes the specified elevation value and traits into account when there is a color
- appearance difference between current traits and previous traits.
- When userInterfaceStyle is UIUserInterfaceStyleDark in currentTraitCollection, elevation will be used
- to resolve the color.
+ Returns a color that takes the specified elevation value and traits into account when there is a
+ color appearance difference between current traits and previous traits. When userInterfaceStyle is
+ UIUserInterfaceStyleDark in currentTraitCollection, elevation will be used to resolve the color.
  Negative elevation is treated as 0.
  Pattern-based UIColor is not supported.
  UIColor in UIExtendedGrayColorSpace will be resolved to UIExtendedSRGBColorSpace.
@@ -43,9 +42,10 @@
  @param previousTraitCollection The previous traits to use when comparing color appearance.
  @param elevation The @c mdc_absoluteElevation to use when resolving the color.
  */
-- (nonnull UIColor *)mdc_resolvedColorWithTraitCollection:(nonnull UITraitCollection *)currentTraitCollection
-                                  previousTraitCollection:(nonnull UITraitCollection *)previousTraitCollection
-                                                elevation:(CGFloat)elevation;
+- (nonnull UIColor *)
+    mdc_resolvedColorWithTraitCollection:(nonnull UITraitCollection *)currentTraitCollection
+                 previousTraitCollection:(nonnull UITraitCollection *)previousTraitCollection
+                               elevation:(CGFloat)elevation;
 
 /**
  Returns a color that takes the specified elevation value and traits into account.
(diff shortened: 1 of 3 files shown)


components/Elevation/src/UIColor+MaterialElevation.h
```diff
@@ -30,6 +30,23 @@
  */
 - (nonnull UIColor *)mdc_resolvedColorWithElevation:(CGFloat)elevation;
 
+/**
+ Returns a color that takes the specified elevation value and traits into account when there is a
+ color appearance difference between current traits and previous traits. When userInterfaceStyle is
+ UIUserInterfaceStyleDark in currentTraitCollection, elevation will be used to resolve the color.
+ Negative elevation is treated as 0.
```
codeman7, line 38: Should we separate these on new lines to make it a little more readable?

components/Elevation/src/UIColor+MaterialElevation.h
```diff
@@ -30,6 +30,23 @@
...
+ Pattern-based UIColor is not supported.
+ UIColor in UIExtendedGrayColorSpace will be resolved to UIExtendedSRGBColorSpace.
+
+ @param currentTraitCollection The traits to use when resolving the color.
+ @param previousTraitCollection The previous traits to use when comparing color appearance.
+ @param elevation The @c mdc_absoluteElevation to use when resolving the color.
+ */
+- (nonnull UIColor *)
+    mdc_resolvedColorWithTraitCollection:(nonnull UITraitCollection *)currentTraitCollection
```
codeman7: Can we make this `withTraitCollection:(UITraitCollection *)traitCollection` to match the existing API and UIKit?

components/Elevation/tests/unit/MaterialElevationColorTests.m
```diff
@@ -296,4 +296,50 @@ - (void)testSmoothJumpBetweenElevationToAlphaForValuesCloseToOne {
...
+        [dynamicColor mdc_resolvedColorWithTraitCollection:currentTraitCollection
+                                   previousTraitCollection:previousTraitCollection
+                                                 elevation:elevation];
+    // Then
+    XCTAssertEqualObjects(resolvedColor, dynamicColor);
+  }
+}
+
+- (void)testResolvingColorWithDifferenceCurrentTraitCollectionAndPreviousTraitCollection {
+  if (@available(iOS 13.0, *)) {
```
codeman7, line 324: Should this have the `if defined`?

components/Elevation/tests/unit/MaterialElevationColorTests.m
```diff
@@ -296,4 +296,50 @@ - (void)testSmoothJumpBetweenElevationToAlphaForValuesCloseToOne {
...
+    UITraitCollection *previousTraitCollection =
+        [UITraitCollection traitCollectionWithUserInterfaceStyle:UIUserInterfaceStyleDark];
+    UITraitCollection *currentTraitCollection =
+        [UITraitCollection traitCollectionWithUserInterfaceStyle:UIUserInterfaceStyleLight];
+    UIColor *resolvedColor =
+        [dynamicColor mdc_resolvedColorWithTraitCollection:currentTraitCollection
+                                   previousTraitCollection:previousTraitCollection
+                                                 elevation:elevation];
+    // Then
+    XCTAssertNotEqual(resolvedColor, dynamicColor);
```
codeman7, line 342: Can we test that the trait collection = what UIKit would return?

codeman7 (approved):

components/Elevation/src/UIColor+MaterialElevation.h
fluxminds: Addressed in 75d9d11dfb3fd470687a570978baba739a411c59.

components/Elevation/src/UIColor+MaterialElevation.h
fluxminds: Addressed in 75d9d11dfb3fd470687a570978baba739a411c59.

components/Elevation/tests/unit/MaterialElevationColorTests.m
fluxminds: Good point! Addressed in 75d9d11dfb3fd470687a570978baba739a411c59.

components/Elevation/tests/unit/MaterialElevationColorTests.m
fluxminds: Addressed in 75d9d11dfb3fd470687a570978baba739a411c59.

yarneo (approved):
nice!

PR #8319, opened 2019-08-14
[Banner] use same bottom anchor for leading and trailing buttons.

fluxminds:
It is a minor internal cleanup that both leading and trailing buttons are using the same anchor on the vertical axis. 

No visual difference was generated from the snapshot tests.

### DIFF
--- a/components/Banner/src/MDCBannerView.m
+++ b/components/Banner/src/MDCBannerView.m
@@ -76,6 +76,7 @@ static NSString *const kMDCBannerViewImageViewImageKeyPath = @"image";
 
 @property(nonatomic, readwrite, strong) NSLayoutConstraint *leadingButtonConstraintLeading;
 @property(nonatomic, readwrite, strong) NSLayoutConstraint *leadingButtonConstraintTop;
+@property(nonatomic, readwrite, strong) NSLayoutConstraint *leadingButtonConstraintBottom;
 @property(nonatomic, readwrite, strong) NSLayoutConstraint *leadingButtonConstraintTrailing;
 @property(nonatomic, readwrite, strong) NSLayoutConstraint *leadingButtonConstraintCenterY;
 @property(nonatomic, readwrite, strong)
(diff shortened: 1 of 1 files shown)


components/Banner/src/MDCBannerView.m
```diff
@@ -271,6 +272,8 @@ - (void)setUpButtonsConstraints {
       constraintGreaterThanOrEqualToAnchor:self.buttonContainerView.leadingAnchor];
   self.leadingButtonConstraintTop =
       [self.leadingButton.topAnchor constraintEqualToAnchor:self.buttonContainerView.topAnchor];
+  self.leadingButtonConstraintBottom = [self.leadingButton.bottomAnchor
```
romoore, line 275: Tip: Can you use `firstBaselineAnchor` or `lastBaselineAnchor` anchors instead to ensure the text is aligned?  I'm concerned that two different button styles may result in different text alignment if only the bottoms are aligned.

romoore (commented):
I don't see any changes to tests. Is there any type of test that can be added to catch a regression of this behavior in the future?

components/Banner/src/MDCBannerView.m
fluxminds: I think it is a good idea! while I just doubled checked the spec, it requires two buttons to be 8pt above the bottom and required two button to be the same style. I think it is safer to align them by the bottom anchor according to the spec for now?

romoore: I would still recommend baseline alignment, unless you prefer the text to be unaligned. Since the button could present two languages (e.g., a translation option), the text may render differently within the buttons.

fluxminds: Talked offline. Opened an [issue](https://github.com/material-components/material-components-ios/issues/8326) to track. Keeping the buttons aligned with bottom meets the design guidelines at the current stage.

romoore: Ack.

romoore (commented):
Is there any test that can be written to identify the original bug and demonstrate that it's now fixed?

fluxminds:
@romoore It is not fixing a bug by itself. The original issue turned out to be not relevant. Now it is more like an improvement or explicit code guard. The visual appearance will stay the same.

romoore:
@wenyuzhang666 I'm a little confused about this PR. If there's no effect to the component (no visual change, no API behavioral change), I'm trying to understand why this change should be made.  If the original issue is no longer relevant, does it make sense to close this PR?

fluxminds:
@romoore Sorry for the confusion. I agree it may be confusing. Though there is no visual and API behavioral change, it makes more sense to align two buttons using the same bottom anchor rather than the original top/bottom logic-wise. The original one was made more from a code reuse perspective (single line). If you feel it is not convincing, I can also close it.

romoore:
I may not understand correctly, but from what I'm reading in your comments, I would recommend closing this PR. If there is no effect from the change and it requires additional properties not currently necessary, then it seems like it's not improving the code.

fluxminds:
@romoore It's reasonable. Will close this PR.

PR #8321, opened 2019-08-15
Fix .gitattributes.

romoore:
The merge from `stable` accidentally made it into `develop`. Once this is
fixed, and now that `stable` ignores merges, there should be no more
conflicts.

Follow-up to #8233


bryanoltman (approved):

romoore:
Presubmits failing due to https://github.com/material-components/material-components-ios/pull/8322

PR #8327, opened 2019-08-16
[Buttons] Changing logic for MDCButton accessibilityTraitsIncludesButton

m-cheung:
`MDCButton` contains a property (`accessibilityTraitsIncludesButton`), which if true, is supposed to include `UIAccessibilityTraitButton` as part of the elements `accessibilityTraits`. However, as documented by Apple, "The default value for this property is `UIAccessibilityTraitNone` unless the receiver is a UIKit control, in which case the value is the standard set of traits associated with the control."

Additionally, it is noted by Apple that "Buttons are accessible by default. The default accessibility traits for a button are Button and User Interaction Enabled."

As a result, `MDCButton` already has `UIAccessibilityTraitButton` (since it subclasses `UIButton`). Instead, if users would like the element to not be identified as a button, we have to remove the trait.

**Apple documentation**
https://developer.apple.com/documentation/objectivec/nsobject/1615202-accessibilitytraits?language=objc
https://developer.apple.com/documentation/uikit/uibutton?language=objc

### DIFF
--- a/components/Buttons/src/MDCButton.m
+++ b/components/Buttons/src/MDCButton.m
@@ -535,8 +535,8 @@ static NSAttributedString *uppercaseAttributedString(NSAttributedString *string)
 }
 
 - (UIAccessibilityTraits)accessibilityTraits {
-  if (self.accessibilityTraitsIncludesButton) {
-    return [super accessibilityTraits] | UIAccessibilityTraitButton;
+  if (!self.accessibilityTraitsIncludesButton) {
+    return [super accessibilityTraits] & ~UIAccessibilityTraitButton;
   }
   return [super accessibilityTraits];
 }


romoore:
Thanks for the PR, @m-cheung .  I agree that `MDCButton` should not be forcing `UIAccessibilityTraitButton` in its `accessibilityTraits` property.  The [`accessibilityTraitsIncludesButton` was introduced to eventually turn-down this behavior entirely](https://github.com/material-components/material-components-ios/pull/6766).

Can you help me better understand the reason for this PR? Do you have a use case that's broken with the current implementation?  I was able to set `accessibilityTraitsIncludesButton = NO` and it produced the default behavior for `UIButton`.

m-cheung:
We have a use case internally where an element sometimes acts as a button, and others it just acts as a label. When the element acts as a label, `accessibilityTraitsIncludesButton` is set to `NO`. However, VoiceOver continued to announced it as a button (i.e. it still had the `UIAccessibilityTraitButton` trait), and I ended up tracing the issue to this.

This PR merely proposes a change which fixes the intended behaviour of `accessibilityTraitsIncludesButton`.

Please feel free to ping me internally if you require more info with regards to our use case.

romoore:
Thanks, @m-cheung. I tried this snippet:

```objc
containedButton.accessibilityTraitsIncludesButton = NO;
containedButton.accessibilityTraits = UIAccessibilityTraitHeader | UIAccessibilityTraitButton; 
```

The result was the button no longer having the `.button` trait. Can you please post a snippet that's breaking for you so I could reproduce it?

m-cheung:
I was expecting that if we had `containedButton.accessibilityTraitsIncludesButton = NO;`
then `containedButton.accessibilityTraits` would not have `UIAccessibilityTraitButton` (and VoiceOver would not identify the element as a button).

In the snippet provided in your comment, I would have assumed that 
```
containedButton.accessibilityTraits = UIAccessibilityTraitHeader | UIAccessibilityTraitButton;
```

negates the intended effects of 
```
containedButton.accessibilityTraitsIncludesButton = NO;
``` 
as the former statement adds back the `UIAccessibilityTraitButton` that should have been removed by the property's effect.

I apologize if I am misunderstanding the intended effect of the property.

romoore:
Ah, I think I understand now. You read the API docs to mean that setting `accessibilityTraitsIncludesButton = NO` would prevent the button from ever acquiring the `.button` trait.  

The intention of `accessibilityTraitsIncludesButton = YES` is that the button **always** returns `.button` as one of its traits, regardless of what is set. That was the original behavior of `MDCButton` (it's not clear to me why that is the case).  By setting `accessibilityTraitsIncludesButton = NO`, callers can use an `MDCButton` as if it were a normal `UIButton` because no "magic" will take place.  

The `accessibilityTraitsIncludesButton` flag is intended as a means of restoring `UIButton` behavior, and it will eventually default to `NO`, become deprecated, and then be deleted. Given the relative obscurity of the behavior and scarcity of setting explicit `accessibilityTraits` values, it's probably going to hang around for a very long time.

Let me know if I've missed the mark. Given this thread, [I've pushed up a documentation change for the property to make its purpose more explicit](https://github.com/material-components/material-components-ios/pull/8328).

romoore:
Also, you're correct that `containedButton.accessibilityTraits = UIAccessibilityTraitHeader | UIAccessibilityTraitButton;` would still include `.button`. If I only set `.header`, it also works as expected and no longer includes `.button`.

m-cheung:
Thanks @romoore for the clarification! Given the context of MDCButton's previous behaviour, the existing code makes a lot more sense now. I will close out this PR and update my internal changes to reflect this behaviour.

romoore:
Fantastic, glad we could find a solution for you quickly. :)  Sorry for the confusion!

PR #8341, opened 2019-08-19
[ActionSheet] Update cells when the background is set.

### LINKED ISSUE #8340: [ActionSheet] Internal issue: b/139645855
This was filed as an internal issue. If you are a Googler, please visit [b/139645855](http://b/139645855) for more details.

---
#### Internal data
- Associated internal bug: [b/139645855](http://b/139645855)

codeman7:
In certain instances when the traitCollection changes on an MDCActionSheet from light/dark the table doesn't update its cells. This causes the cells to be white while the rest of the action sheet is dark. By reloading the cells when the traitCollection changes we can make sure the color are always in sync.

Steps to reproduce.
1. Open MDCCatalog Action Sheet example
2. Present the ActionSheet
3. Toggle light/dark in Xcode settings
4. Dismiss the action sheet
5. Repeat steps 2-3.

| Before | After |
| --- | --- |
|before|after|

Closes #8340

### DIFF
--- a/components/ActionSheet/src/MDCActionSheetController.m
+++ b/components/ActionSheet/src/MDCActionSheetController.m
@@ -340,6 +340,14 @@ static const CGFloat kActionTextAlpha = (CGFloat)0.87;
 - (void)traitCollectionDidChange:(UITraitCollection *)previousTraitCollection {
   [super traitCollectionDidChange:previousTraitCollection];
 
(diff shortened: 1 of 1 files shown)


romoore (commented):
Is it possible to add a snapshot test for this behavior?  If not, can you please provide manual testing steps in the PR description?

codeman7:
I don't believe we can execute snapshot test on iOS 13, I've update the PR description for manual testing.

romoore (approved):

PR #8361, opened 2019-08-23
[Snapshot] Add a window into the app delegate of host application.

fluxminds:
Right now, the behavior of the host application from CocoaPods and Bazel is different because the host application we built for Bazel didn't have a window and a root view controller. It causes https://github.com/material-components/material-components-ios/pull/8360 's failure on Bazel.

### DIFF
--- a/components/private/Snapshot/TestHost/src/TestHostMinimalDelegate.m
+++ b/components/private/Snapshot/TestHost/src/TestHostMinimalDelegate.m
@@ -18,6 +18,7 @@
 - (BOOL)application:(UIApplication *)application
     didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
   self.window = [[UIWindow alloc] initWithFrame:UIScreen.mainScreen.bounds];
+  // Set up the rootViewController to support tests that require root level objects.
   UIViewController *rootViewController = [[UIViewController alloc] init];
   self.window.rootViewController = rootViewController;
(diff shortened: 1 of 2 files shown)


randallli (approved):

components/private/Snapshot/TestHost/src/TestHostMinimalDelegate.m
```diff
@@ -14,4 +14,14 @@
 #import "TestHostMinimalDelegate.h"
 
 @implementation TestHostMinimalDelegate
+
+- (BOOL)application:(UIApplication *)application
+    didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
+  self.window = [[UIWindow alloc] initWithFrame:UIScreen.mainScreen.bounds];
+  UIViewController *rootViewController = [[UIViewController alloc] init];
```
romoore, line 22: If the VC is also required, please add a comment to remark on the reason.

romoore (approved):

components/private/Snapshot/TestHost/src/TestHostMinimalDelegate.m
fluxminds: Addressed in https://github.com/material-components/material-components-ios/pull/8361/commits/370fb2933718647415f0c377f071e6554aa94215.

PR #8362, opened 2019-08-26
v88.1.0

romoore:


randallli (approved):

romoore:
Changed the version number from 89.0.0 (major change) to 88.1.0 (minor change) because the MDCSnackbarMessage change only results in a potential runtime crash and not a breaking API (compiler) change.

randallli (approved):

PR #8369, opened 2019-08-26
[Banner] Add a missing button title in example.

fluxminds:
Closes https://github.com/material-components/material-components-ios/issues/8368.

A button is missing from the example. Note: these two buttons are not aligned well. To be fixed in https://github.com/material-components/material-components-ios/pull/8370.

| Before | After |
| --- | --- |
| Simulator Screen Shot - iPhone Xʀ - 2019-08-26 at 15 18 32 | Simulator Screen Shot - iPhone Xʀ - 2019-08-26 at 15 24 05 |

### DIFF
--- a/components/Banner/examples/BannerTypicalUseExampleViewController.m
+++ b/components/Banner/examples/BannerTypicalUseExampleViewController.m
@@ -285,7 +285,7 @@ static NSString *const exampleExtraLongText =
 
   MDCButton *dismissButton = bannerView.leadingButton;
   [dismissButton applyTextThemeWithScheme:self.containerScheme];
-  [dismissButton sizeToFit];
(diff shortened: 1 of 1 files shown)


randallli (commented):
why are the buttons not aligned in your snapshots after? "Long dismiss" is lower than "dismiss"

fluxminds:
@randallli It is caused by MDCButton overwriting `sizeThatFits`, which causes it returns a different value than `instrinsicContentSize`. It will be fixed in separate PR as mentioned in the description. This PR mainly focuses on adding this button back.

randallli (approved):
whoops sorry didn't see that

fluxminds:
Thanks for the review!

PR #8370, opened 2019-08-26
[Buttons] overwrite intrinsicContentSize method.

fluxminds:
The `instrinsicContentSize` of `MDCButton` was not aware of `minimumSize` property, which is wrong.

Open to use `super.instrinsicContentSize` if reviewer(s) prefer.

### DIFF
--- a/components/Buttons/tests/unit/MDCButtonTests.m
+++ b/components/Buttons/tests/unit/MDCButtonTests.m
@@ -1291,11 +1291,10 @@ static NSString *controlStateDescription(UIControlState controlState) {
 - (void)testIntrinsicContentSizeWithMaximum {
   // Given
   CGSize originalSize = self.button.intrinsicContentSize;
-  CGSize expectedSize = CGSizeMake(originalSize.width + 5, originalSize.height - 5);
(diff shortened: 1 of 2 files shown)


romoore (changes requested):
Please add tests for the new behavior.

fluxminds:
Added tests and removed WIP from title.

randallli (approved):

components/Buttons/tests/unit/MDCButtonTests.m
```diff
@@ -1276,6 +1276,43 @@ - (void)testSizeThatFitsWithMinimumAndMaximum {
                 NSStringFromCGRect(expectedFrame), NSStringFromCGRect(self.button.frame));
 }
 
+- (void)testIntrinsicContentSizeWithMinimum {
```
romoore: Tip: Please describe the "outcome" in the test name.  Here and in the others.

romoore (approved):

components/Buttons/tests/unit/MDCButtonTests.m
fluxminds: Addressed in df99798

PR #8373, opened 2019-08-27
[Banner] Refactor typical example view controller.

### LINKED ISSUE #6842: [Banner] Refactor Banner example view controller.
Per comment in https://github.com/material-components/material-components-ios/pull/6830#discussion_r264480423, we should evaluate the current banner example VC implementation and do necessary refactoring.

fluxminds:
closes https://github.com/material-components/material-components-ios/issues/6842.

There is a minor behavior change here that the background of content view changes from blue to grey.

### DIFF
--- a/components/Banner/examples/BannerTypicalUseExampleViewController.m
+++ b/components/Banner/examples/BannerTypicalUseExampleViewController.m
@@ -32,11 +32,6 @@ static NSString *const exampleExtraLongText =
 @interface BannerExampleContentView : UIView
 
 @property(nonatomic, readwrite, strong) UILabel *contentLabel;
-@property(nonatomic, weak) MDCBannerView *bannerView;
-
-@property(nonatomic, strong) MDCSemanticColorScheme *colorScheme;
-@property(nonatomic, strong) MDCTypographyScheme *typographyScheme;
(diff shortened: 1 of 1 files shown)


components/Banner/examples/BannerTypicalUseExampleViewController.m
```diff
@@ -29,6 +29,34 @@
     @"Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut "
     @"labore et dolore.";
 
+@interface BannerExampleContentView : UIView
```
codeman7, line 32: _Tip:_ Should we prefix this?

codeman7 (approved):

components/Banner/examples/BannerTypicalUseExampleViewController.m
fluxminds: Since this example was not prefixed (as other examples), I feel we probably don't need to prefix this one with `MDC`.

fluxminds:
@codeman7 I added some more refactors as a follow-up to move more logics into the custom view.

fluxminds:
I will put the last commit into a separate PR :)

PR #8391, opened 2019-09-04
[TextFields] Added check on textinput.font set to avoid sideeffects

ggdiez:
This PR should be fix this issue: https://github.com/material-components/material-components-ios/issues/8390
#8390 

This PR is part of code of closed PR hadn't merge anymore. #5810

This code only change the font in this method when is different. Doing that we avoid that font size change while the user is typing on textfields.

### DIFF
--- a/components/TextFields/src/MDCTextInputControllerBase.m
+++ b/components/TextFields/src/MDCTextInputControllerBase.m
@@ -397,16 +397,16 @@ static UITextFieldViewMode _underlineViewModeDefault = UITextFieldViewModeWhileE
 #pragma mark - TextInput Customization
 
 - (void)updateTextInput {
-	UIFont *font = self.textInputFont;
-	if (self.mdc_adjustsFontForContentSizeCategory) {
-		font = [font mdc_fontSizedForMaterialTextStyle:MDCFontTextStyleBody1
+  UIFont *font = self.textInputFont;
+    if (self.mdc_adjustsFontForContentSizeCategory) {
+      font = [font mdc_fontSizedForMaterialTextStyle:MDCFontTextStyleBody1
 								  scaledForDynamicType:_mdc_adjustsFontForContentSizeCategory];
-		// TODO: (#4331) This needs to be converted to the new text scheme.
-	}
+        // TODO: (#4331) This needs to be converted to the new text scheme.
+    }
 	
-	if (![self.textInput.font mdc_isSimplyEqual:font]){
-		self.textInput.font = font;
-	}
+    if (![self.textInput.font mdc_isSimplyEqual:font]){
+      self.textInput.font = font;
+    }
 }
 
 #pragma mark - Placeholder Customization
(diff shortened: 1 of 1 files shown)


components/TextFields/src/MDCTextInputControllerBase.m
```diff
@@ -397,13 +397,16 @@ - (void)updateLeadingUnderlineLabel {
...
-  self.textInput.font = font;
+	UIFont *font = self.textInputFont;
+	if (self.mdc_adjustsFontForContentSizeCategory) {
+		font = [font mdc_fontSizedForMaterialTextStyle:MDCFontTextStyleBody1
+								  scaledForDynamicType:_mdc_adjustsFontForContentSizeCategory];
+		// TODO: (#4331) This needs to be converted to the new text scheme.
+	}
+	
+	if (![self.textInput.font mdc_isSimplyEqual:font]){
+		self.textInput.font = font;
```
romoore: From this change, it looks like there is a bug when setting the MDCTextField's `font` property. Could you please describe in the PR summary why the fix needs to be made here and not where the side-effect takes place?

components/TextFields/src/MDCTextInputControllerBase.m
```diff
@@ -397,13 +397,16 @@ - (void)updateLeadingUnderlineLabel {
...
 
 - (void)updateTextInput {
-  UIFont *font = self.textInputFont;
-  if (self.mdc_adjustsFontForContentSizeCategory) {
-    font = [font mdc_fontSizedForMaterialTextStyle:MDCFontTextStyleBody1
-                              scaledForDynamicType:_mdc_adjustsFontForContentSizeCategory];
-    // TODO: (#4331) This needs to be converted to the new text scheme.
-  }
-  self.textInput.font = font;
+	UIFont *font = self.textInputFont;
```
romoore: Nit: Please correct the indentation for this method. It looks like it accidentally changed. It should be 2 spaces from the start of the line for method bodies.

romoore (changes requested):
Please provide greater detail about why assigning an identical font can cause unexpected side-effects. It  seems like the correct fix might actually be in MDCTextField and not the controller.

components/TextFields/src/MDCTextInputControllerBase.m
ggdiez: Changed

components/TextFields/src/MDCTextInputControllerBase.m
ggdiez: This PR is related with bug #8390 
The problem is not in the MCDTextField font is in the flotating placeholder text. 
When use start to type in MCDTextField flotating placeholder text change the size. 
It seems when in the controller the textInput font changes produces size changes in flotating placeholder. 
This code prevent changes in font except when font changes really.

romoore: @ggdiez Can you see if https://github.com/material-components/material-components-ios/pull/8444 will resolve your issues?  I believe the `didSetFont` behavior was the root cause of the problem.

ggdiez: Yes, I agree with you. I just test your code and works. I can not reproduce the issue.
Thank you.

romoore: Awesome, thanks for confirming!

romoore:
Closing. I believe this will be obsoleted by #8444.

PR #8406, opened 2019-09-06
[TextFields] Add initial stateful coloring stuff

andrewoverton:
This PR adds stateful coloring functionality to MDCBaseTextField. It ended up being a little bigger than I would have hoped, but it's pretty straightforward.

I know there were concerns about grouping colors in the same object. I don't know where we currently stand on that, but the way I have it here it shouldn't be too hard to change later if it turns out this approach doesn't work.

NOTE: The recent design doc on custom states specifies that new state constants should take the form of `MDC<#Class#>State<#State#`. However, I didn't want to call it MDCBaseTextFieldStateEditing because it's going to be used by not only its subclasses (MDCFilledTextField and MDCOutlinedTextField), but also other contained input views, like the input chip views and text areas. For that reason, I called it MDCTextControlStateEditing. This way the name makes sense for all of these text related control classes. More can be read about this at go/thoughts-on-custom-states.

Related to #6942.


components/TextFields/src/ContainedInputView/UIControlState+TextControl.h
```diff
@@ -0,0 +1,24 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+typedef UIControlState MDCTextControlState;
```
andrewoverton: As noted in the PR description, I'm calling it MDCTextControlState instead of MDCBaseTextFieldState because things like MDCFilledTextFields, MDCOutlinedTextFields, and all the input chip view and text area classes will use it too.

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
```diff
@@ -55,4 +55,50 @@
  */
 @property(nonatomic, assign) UITextFieldViewMode trailingViewMode;
 
+/**
+ Sets the floating label color for a given state. Floating label color refers to the color of the
```
randallli: nit: Second sentence would read nicer if it started on a new line.

Lets do this for the comments below too.

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
```diff
@@ -55,4 +55,50 @@
...
+ label when it's in its "floating position," i.e. when it's above the text area.
+ @param floatingLabelColor The UIColor for the given state.
+ @param state The UIControlState. The accepted values are UIControlStateNormal,
+ UIControlStateDisabled, and MDCTextControlStateEditing, which is a custom MDC
+ UIControlState value.
+ */
+- (void)setFloatingLabelColor:(nonnull UIColor *)floatingLabelColor forState:(UIControlState)state;
+/**
```
randallli, line 69: nit: new line.

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
```diff
@@ -55,4 +55,50 @@
...
+- (void)setNormalLabelColor:(nonnull UIColor *)normalLabelColor forState:(UIControlState)state;
+/**
+ Returns the normal label color for a given state. Normal label color refers to the color of the
+ label when it's in its "normal position," i.e. when it's not floating.
+ @param state The UIControlState.
+ */
+- (nonnull UIColor *)normalLabelColorForState:(UIControlState)state;
+
+/**
+ Sets the text color for a given state.
```
randallli, line 95: There are multiple labels on screen for a text field. Is this for all of them? Lets make that more clear in this property doc.

components/TextFields/src/ContainedInputView/MaterialTextFields+ContainedInputView.h
```diff
@@ -13,3 +13,5 @@
 // limitations under the License.
 
 #import "MDCBaseTextField.h"
+#import "MDCTextControlLabelBehavior.h"
```
randallli, line 16: Did we start to use this in this PR? Im having difficulty finding the new usage. Can you point me at it?

components/TextFields/tests/unit/ContainedInputView/UIControlState+TextControlUnitTests.m
```diff
@@ -0,0 +1,29 @@
...
+
+#import "../../../src/ContainedInputView/UIControlState+TextControl.h"
+
+@interface UIControlStateTextControlUnitTests : XCTestCase
+@end
+
+@implementation UIControlStateTextControlUnitTests
+
+- (void)testMDCTextControlStateEditing {
+  XCTAssertTrue((MDCTextControlStateEditing & UIControlStateApplication) ==
```
randallli: Nice! I was thinking that we should double check our maths here :)

components/TextFields/tests/unit/ContainedInputView/MDCBaseTextFieldTests.m
```diff
@@ -184,4 +184,67 @@ - (void)testClearButtonRectForBounds {
...
+  // Given
+  CGRect textFieldFrame = CGRectMake(0, 0, 130, 40);
+  MDCBaseTextField *textField = [[MDCBaseTextField alloc] initWithFrame:textFieldFrame];
+  UIColor *floatingLabelColor = [UIColor redColor];
+
+  // When
+  [textField setFloatingLabelColor:floatingLabelColor forState:UIControlStateNormal];
+  [textField setFloatingLabelColor:floatingLabelColor forState:MDCTextControlStateEditing];
+  [textField setFloatingLabelColor:floatingLabelColor forState:UIControlStateDisabled];
+  [textField setNeedsLayout];
```
randallli: Remove calls to setNeedsLayout nor layoutIfNeeded.
If this is a test for setter/getters then we should not need to call them.

components/TextFields/tests/unit/ContainedInputView/MDCBaseTextFieldTests.m
```diff
@@ -184,4 +184,67 @@ - (void)testClearButtonRectForBounds {
   XCTAssertTrue(CGRectEqualToRect(actualClearButtonFrame, expectedClearButtonFrame));
 }
 
+- (void)testFloatingLabelColorAccessors {
+  // Given
+  CGRect textFieldFrame = CGRectMake(0, 0, 130, 40);
+  MDCBaseTextField *textField = [[MDCBaseTextField alloc] initWithFrame:textFieldFrame];
+  UIColor *floatingLabelColor = [UIColor redColor];
```
randallli: We should not use the same color for all setters. If there is a bug where we have wired them up incorrectly this would mask that issue. 

I was expecting something similar to:
https://github.com/material-components/material-components-ios/blob/e3784afe259055679479f09b2e92324b61cf7fe0/components/Buttons/tests/unit/MDCButtonTests.m#L383

components/TextFields/tests/unit/ContainedInputView/MDCBaseTextFieldTests.m
```diff
@@ -184,4 +184,67 @@ - (void)testClearButtonRectForBounds {
...
+
+  // When
+  [textField setFloatingLabelColor:floatingLabelColor forState:UIControlStateNormal];
```
randallli: we are only testing this specific combination of these. we should touch all 8 of them.

https://github.com/material-components/material-components-ios/blob/e3784afe259055679479f09b2e92324b61cf7fe0/components/Buttons/tests/unit/MDCButtonTests.m#L383

Since its a discontinuous range we could setup an array of all the bits we care about and loop through them:
controlValues = @[@(Normal),@(Normal |  Editing), ...]

https://github.com/material-components/material-components-ios/blob/e3784afe259055679479f09b2e92324b61cf7fe0/components/Buttons/tests/unit/MDCButtonTests.m#L383

components/TextFields/tests/unit/ContainedInputView/MDCBaseTextFieldTests.m
```diff
@@ -184,4 +184,67 @@ - (void)testClearButtonRectForBounds {
   XCTAssertTrue(CGRectEqualToRect(actualClearButtonFrame, expectedClearButtonFrame));
 }
 
+- (void)testFloatingLabelColorAccessors {
```
randallli, line 187: Are we going to add some tests to make sure that setting the labelColor results in visual changes? This could be a snapshot test.

randallli (changes requested):
I looked at the public headers and tests and left comments. I will do another pass on the implementation after sending these.

components/TextFields/tests/unit/ContainedInputView/MDCBaseTextFieldTests.m
```diff
@@ -184,4 +184,67 @@ - (void)testClearButtonRectForBounds {
...
+  [textField layoutIfNeeded];
+
+  // Then
+  XCTAssertEqualObjects(floatingLabelColor,
+                        [textField floatingLabelColorForState:UIControlStateNormal]);
+  XCTAssertEqualObjects(floatingLabelColor,
+                        [textField floatingLabelColorForState:MDCTextControlStateEditing]);
+  XCTAssertEqualObjects(floatingLabelColor,
+                        [textField floatingLabelColorForState:UIControlStateDisabled]);
+}
```
randallli, line 207: I was also expecting a test for the fallback logic of these getters. If normal is set and you ask for the others it uses normal. 

Reference test for MDCButton does this here:

https://github.com/material-components/material-components-ios/blob/e3784afe259055679479f09b2e92324b61cf7fe0/components/Buttons/tests/unit/MDCButtonTests.m#L482

components/TextFields/src/ContainedInputView/MaterialTextFields+ContainedInputView.h
andrewoverton: Ah, sorry, I just realized it wasn't here so I added it. It's a public type but I hadn't added it previously.

components/TextFields/tests/unit/ContainedInputView/UIControlState+TextControlUnitTests.m
andrewoverton: Testing this was the design doc's idea heheh

components/TextFields/tests/unit/ContainedInputView/MDCBaseTextFieldTests.m
randallli, line 207: For the most complicated use case we want to make sure that we do the same thing that UIButton does for all the states we support.

Here I would create a local UIButton and expect that the same combinations work equal for our custom stateful object. (we only care about the 3 explained in the header so an array would be useful here too)

Reference test in MDCButton:
https://github.com/material-components/material-components-ios/blob/e3784afe259055679479f09b2e92324b61cf7fe0/components/Buttons/tests/unit/MDCButtonTests.m#L461

randallli (changes requested):

components/TextFields/src/ContainedInputView/UIControlState+TextControl.h
randallli: lgtm

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
andrewoverton: Done.

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
andrewoverton: Done.

components/TextFields/tests/unit/ContainedInputView/MDCBaseTextFieldTests.m
andrewoverton: Added a snapshot test for the floatingLabelColor accessor here: https://github.com/material-components/material-components-ios/pull/8406/files?short_path=bbaeb1f#diff-bbaeb1f9da4e514834f3ac6b04a1b234

components/TextFields/tests/unit/ContainedInputView/MDCBaseTextFieldTests.m
andrewoverton: Got rid of them--Done.

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
andrewoverton: Added some newlines

components/TextFields/tests/unit/ContainedInputView/MDCBaseTextFieldTests.m
andrewoverton: I made it so that each accessor got its own color

components/TextFields/tests/unit/ContainedInputView/MDCBaseTextFieldTests.m
andrewoverton: Sorry, what are the 8 combinations you're referring to?

`@(Normal | Editing)` isn't a valid state for this textfield. The MDCContainedInputView states are actually mutually exclusive. A contained input view is either normal, disabled, or editing. I actually just made it an NS_ENUM instead of an NS_OPTIONS in a recent commit for this reason.

Honestly, I still think that using UIControlState for these accessors feels forced. I don't think it's the right abstraction for what we're trying to accomplish here, even if we're subclassing a UITextField, which subclasses UIControl. I still think it would be less of a headache down the road to have clients deal directly with MDContainedInputViewState (even if we renamed it to MDCTextControlState, since MDCContainedInputView is private), the same way we have them deal with MDCTabBarItemState, MDCRippleState, MDCBottomDrawerState, MDCCardCellState, and MDCSheetState.

components/TextFields/tests/unit/ContainedInputView/MDCBaseTextFieldTests.m
andrewoverton: Hmmm, so what kind of fallback behavior do we want to support exactly?

In general, I don't think we should try to treat these TextFields like they're UIButtons. The weird thing about UIControlState is that every UIControl treats it differently. UITextFields treat UIControlState totally differently than UIButtons do. Where a UIButton might actually look and behave differently when it's highlighted or selected, a UITextField won't. This screenshot of a bunch of UITextFields in different states shows that only disabled makes a visual difference:
Screen Shot 2019-09-05 at 10 00 49 AM
With the way it is right now there are default values for the three supported UIControlStates (i.e. the three that are adapted to underlying MDCContainedInputViewStates), and the stateful accessors for those states only accept nonnull colors, so clients will not be able to set nil for a given state. All this makes me wonder how much we really need to think about fallback behavior.

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
```diff
@@ -55,4 +55,58 @@
...
+/**
+ Sets the floating label color for a given state.
+ Floating label color refers to the color of the label when it's in its "floating position," i.e.
+ when it's above the text area.
+ @param floatingLabelColor The UIColor for the given state.
+ @param state The UIControlState. The accepted values are UIControlStateNormal,
+ UIControlStateDisabled, and MDCTextControlStateEditing, which is a custom MDC
+ UIControlState value.
+ */
+- (void)setFloatingLabelColor:(nonnull UIColor *)floatingLabelColor forState:(UIControlState)state;
```
randallli: In UIButton these colors are nullable

https://developer.apple.com/documentation/uikit/uibutton/1623993-settitlecolor?language=objc

We spoke in person about this and I dont want to forget to chat about this more.

randallli:
After much offline discussion this PR has pivoted to a different approach where we are no longer using the UIControlState enum. @andrewoverton will be morphing this PR or starting another one to that effect.

components/TextFields/src/ContainedInputView/MDCBaseTextField.m
```diff
@@ -392,4 +435,75 @@ - (BOOL)isRTL {
...
+  return colorViewModel;
+}
+
+#pragma mark Color Accessors
+
+- (void)setNormalLabelColor:(nonnull UIColor *)labelColor forState:(MDCTextControlState)state {
+  MDCContainedInputViewColorViewModel *colorViewModel =
+      [self containedInputViewColorViewModelForState:state];
+  colorViewModel.normalLabelColor = labelColor;
+  [self setNeedsLayout];
```
randallli, line 474: I was expecting more of these calls when the state changes. IE for setEnabled

components/TextFields/tests/unit/ContainedInputView/MDCBaseTextFieldTests.m
randallli: Cool we should add some for each state to make sure that setting the various colors for various states results in visual changes of the component.

randallli (approved):

randallli:
Nit: Your PR title and first comment will go into the commit history of the merged code. Please copy edit them for the intended user: Our future selves reading it to make sense of the change.

components/TextFields/src/ContainedInputView/MDCBaseTextField.m
andrewoverton: I forgot to bring over the call to setNeedsLayout in setIsEnabled. Will address later.

PR #8407, opened 2019-09-06
[TextFields] Try new approach to snapshot testing textfields

andrewoverton:
In this PR I'm trying a new approach for snapshot testing MDCBaseTextField. Getting an accurate snapshot of a textfield that's editing is not very straightforward. In order to do it here I had to do a few things:
1. Add it to the snapshot test's app host's key window.
2. Take a UIView snapshot of it from the main thread.
3. Verify the golden using that UIView snapshot, as opposed to the view itself.

This PR is kind of a test. Someone who knows a lot more about snapshot testing might see something wrong with it that I'm missing 😅

Related to #6942.


components/TextFields/tests/snapshot/MDCBaseTextFieldSnapshotTests.m
```diff
@@ -18,6 +18,8 @@
 
 #import "MaterialTextFields+ContainedInputView.h"
 
+static const NSTimeInterval kTextFieldValidationAnimationTimeout = 1.5;
```
randallli: 1.5 second timeout seems like a large value. Did lower values not work? something closer to .1 or even lower is what i would expect. It only seems to need to pump the run loop once to render the view.
Are you expecting an animation to occur?

components/TextFields/tests/snapshot/MDCBaseTextFieldSnapshotTests.m
```diff
@@ -35,6 +37,29 @@ - (void)tearDown {
   [super tearDown];
 }
 
+- (MDCBaseTextField *)createBaseTextFieldInKeyWindow {
+  MDCBaseTextField *textField = [[MDCBaseTextField alloc] initWithFrame:CGRectMake(0, 0, 200, 60)];
+  textField.borderStyle = UITextBorderStyleRoundedRect;
+  UIWindow *keyWindow = [[UIApplication sharedApplication] keyWindow];
+  [keyWindow addSubview:textField];
```
randallli, line 48: I expect that we would want to tear this down so that it does not affect other tests.
In the tear down remove it from the window.

andrewoverton: In 35b1fb59ea6828b2fbb9ebb17a9b95894a7da2d0 I implement `setUp` and `tearDown` so textfields aren't left just sitting in the window 👍

components/TextFields/tests/snapshot/MDCBaseTextFieldSnapshotTests.m
andrewoverton: I made it 1 just now in 35b1fb59ea6828b2fbb9ebb17a9b95894a7da2d0, is that low enough? To be clear, this is just how long the expectation waits before failing, the tests run much much faster than that. I picked a high-ish number out of fear of random delays leading to flaky tests. I'm sort of afraid of flaky tests 😅

components/TextFields/tests/snapshot/MDCBaseTextFieldSnapshotTests.m
```diff
@@ -18,21 +18,50 @@
...
+
+- (MDCBaseTextField *)createBaseTextFieldInKeyWindow {
+  MDCBaseTextField *textField = [[MDCBaseTextField alloc] initWithFrame:CGRectMake(0, 0, 200, 60)];
+  textField.borderStyle = UITextBorderStyleRoundedRect;
+  UIWindow *keyWindow = [[UIApplication sharedApplication] keyWindow];
+  [keyWindow addSubview:textField];
+  return textField;
+}
+
+- (void)validateTextField:(MDCBaseTextField *)textField {
```
randallli, line 52: FYI: It seems to be very similar to what you have going on here. 

https://github.com/material-components/material-components-ios/blob/5bee8bc3b6be6745b777fbc091131489da4fb4d0/components/TextFields/tests/snapshot/supplemental/MDCTextFieldSnapshotTestCase.m

We may want to copy some of the explanation for why this is needed.

components/TextFields/tests/snapshot/MDCBaseTextFieldSnapshotTests.m
```diff
@@ -18,21 +18,50 @@
...
+  XCTestExpectation *expectation =
+      [[XCTestExpectation alloc] initWithDescription:@"textfield_validation_expectation"];
+  dispatch_async(dispatch_get_main_queue(), ^{
```
randallli, line 55: Nit: Im worried this dispatch async will cause collisions with other textfield snapshot tests that are also waiting to be first responder. I suppose we just have to keep an eye out for them.

FYI: Here is another textfield run loop oddity
https://github.com/material-components/material-components-ios/blob/b38372192ff5f7d3273c87e901a2eab402cabef3/components/TextFields/tests/unit/TextFieldControllerLegacyTests.swift#L25

waiting for 1 second in a unit test may be expensive though.

randallli:
Dont forget to update the PR description (if it was your intent to do so)

randallli (approved):

components/TextFields/tests/snapshot/MDCBaseTextFieldSnapshotTests.m
andrewoverton: Which part do you want copied? I do have some explanation in there, I can add more in a follow up 👍

PR #8448, opened 2019-09-12
[TextFields] Add some placeholder logic

andrewoverton:
This PR adds some placeholder support to MDCBaseTextField.

MDCBaseTextField has some special considerations related to placeholder visibility tthat UITextField does not have due to the fact that the label can sometimes occupy the space where the placeholder goes when there is no text.

Here's a screenshot of an MDCBaseTextField with a placeholder:
Screen Shot 2019-09-26 at 11 03 09 AM

Related to #6942.


components/TextFields/src/ContainedInputView/MDCBaseTextField.h
```diff
@@ -29,6 +29,11 @@
  */
 @property(strong, nonatomic, readonly, nonnull) UILabel *label;
 
+/**
+ The color of the placeholder. The default is @c .lightGrayColor .
+ */
+@property(nonatomic, strong, nonnull) UIColor *placeholderColor;
```
randallli: Putting down thoughts discussed in person:
As written this is a magic property that changes another property: attributedPlaceholder. Properties should not change other properties. 

Instead we should consider implementing this as a category on the UITextField as a convenience method following the mutator pattern.
If we are use the public API of the UIKit object let us try to accomplish it as a category on the UIKit object.

randallli (changes requested):

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
andrewoverton: @randallli, the current solution does rely on side effects, but it does so in a way that I think is unlikely to get in the way of users' intentions. With the current solution, setting `placeholder` or `placeholderColor` will result in a call to `-setAttributedPlaceholder:`, but it won't affect clients' ability to make use of that API on its own. It will behave like this:

```objc
NSDictionary *redAttribute =
    @{NSForegroundColorAttributeName: [UIColor redColor]};
NSAttributedString *attributedString =
    [[NSAttributedString alloc] initWithString:@"C" attributes:redAttribute];

self.textField.placeholder = @"A";
// placeholder will be a gray "A"
self.textField.placeholderColor = [UIColor blueColor];
// placeholder will be a blue "A"
self.textField.placeholder = @"B";
// placeholder will be a blue "B"
self.textField.attributedPlaceholder = attributedString;
// placeholder will be a red "C"
self.textField.placeholderColor = [UIColor blueColor];
// placeholder will be a blue "C"
```
The solution we talked about yesterday involved a UITextField category class method like `+setPlaceholderColor:onTextField:`. That would behave like this: 

```objc
self.textField.placeholder = @"A";
// placeholder will be a gray "A"
[UITextField setPlaceholderColor:[UIColor blueColor] onTextField:self.textField];
// placeholder will be a blue "A"
self.textField.placeholder = @"B";
// placeholder will be a gray "B"
```

Do you definitely think the second one is better?

randallli: After looking at your examples I think a better signature would be a convenience setter that documents that it is changing the placeholderAttributed property.
it would look something like this:
[self.textField setPlaceholder:@"A" withColor:[UIColor redColor]]

it would be a UITextField category without a getter.

what do you think?

randallli (dismissed):

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
andrewoverton: I downscoped the PR to get it landed. There's no longer a placeholderColor property.
Also, something I said about UITextField behavior in the last comment was inaccurate.

This is how UITextField really behaves:

```
textField.attributedPlaceholder = [self attributedStringWithString:@"C" color:[UIColor redColor]];
// placeholder will be a red "C"
textField.placeholder = @"Q";
// placeholder will be a red "Q"
textField.attributedPlaceholder = nil;
// placeholder doesn't exist
textField.placeholder = @"B";
// placeholder will be a red "B"
textField.attributedPlaceholder = [self attributedStringWithString:@"Z" color:[UIColor greenColor]];
// placeholder will be a green "Z"
```

components/TextFields/src/ContainedInputView/MDCBaseTextField.m
```diff
@@ -226,6 +226,19 @@ - (CGRect)clearButtonRectForBounds:(CGRect)bounds {
...
 }
 
+- (CGRect)placeholderRectForBounds:(CGRect)bounds {
+  if (self.shouldPlaceholderBeVisible) {
+    return [super placeholderRectForBounds:bounds];
+  }
+  return CGRectZero;
+}
+
+- (void)drawPlaceholderInRect:(CGRect)rect {
```
randallli: please add a comment to the apple docs so that our future selves know that this is an override of an app API
https://developer.apple.com/documentation/uikit/uitextfield/1619637-drawplaceholderinrect?language=objc

components/TextFields/src/ContainedInputView/MDCBaseTextField.m
```diff
@@ -226,6 +226,19 @@ - (CGRect)clearButtonRectForBounds:(CGRect)bounds {
   return self.layout.clearButtonFrameNormal;
 }
 
+- (CGRect)placeholderRectForBounds:(CGRect)bounds {
```
randallli: Please add a comment to the apple docs so that our future selves know that this is an override of an app API
https://developer.apple.com/documentation/uikit/uitextfield/1619615-placeholderrectforbounds?language=objc

components/TextFields/src/ContainedInputView/MDCBaseTextField.m
```diff
@@ -369,6 +382,37 @@ - (UIFont *)uiTextFieldDefaultFont {
...
+                                             text:(NSString *)text
+                                        isEditing:(BOOL)isEditing {
+  BOOL hasPlaceholder = placeholder.length > 0;
+  BOOL hasText = text.length > 0;
+
+  if (hasPlaceholder) {
+    if (hasText) {
+      return NO;
+    } else {
+      if (labelState == MDCContainedInputViewLabelStateNormal) {
```
randallli, line 462: Yes, and let us add some snapshots for all the branching logic we have put here.
Looks like there are 4 cases to cover here.

randallli (commented):
Required: change the description to be more accurate. 

"I also tried by implementing my own drawing logic in -drawPlaceholderInRect:, but couldn't quite get it looking right with that." makes it sounds like the code changes in the PR did not work.

components/TextFields/src/ContainedInputView/MDCBaseTextField.m
andrewoverton: How about a unit test instead? This looks to me like a good candidate for a unit test.

components/TextFields/src/ContainedInputView/MDCBaseTextField.m
andrewoverton: I can do that. Do we typically do that though? The fact that it calls `super` is usually enough to indicate that we're overriding something right?

andrewoverton: Do I need to add a link to Apple docs for every method that we override? Or just UITextField methods? Or just this one and `-placeholderRectForBounds:`?

components/TextFields/src/ContainedInputView/MDCBaseTextField.m
andrewoverton: Added a unit test.

components/TextFields/src/ContainedInputView/MDCBaseTextField.m
randallli: If we are not calling super on the first line it is helpful to give a hint to ourselves.

Or we can use a #pragma mark.
I see we have one for `UITextField Accessor Overrides`. I didn't see that. We can add one for non accessors.

components/TextFields/src/ContainedInputView/MDCBaseTextField.m
randallli: Sharing testing best practices: If we have to expose the a private method in our tests it is a code smell. It becomes a brittle test if we ever refactor the code. This is why I suggested a snapshot test.

randallli (approved):
Im having difficulty visualizing the delta from this PR can you add a screenshot?

components/TextFields/src/ContainedInputView/MDCBaseTextField.m
andrewoverton: I just put `-drawPlaceholderInRect:` under a `UITextField Drawing Overrides` pragma mark 👍

components/TextFields/src/ContainedInputView/MDCBaseTextField.m
andrewoverton: This should have been under the `UITextField Layout Overrides` pragma but wasn't. I just put it there 👍

andrewoverton:
Added a screenshot to the PR description.

components/TextFields/src/ContainedInputView/MDCBaseTextField.m
andrewoverton: Yes, this unit test will probably have to change if we change the conditions under which the placeholder should be visible. I added a snapshot test for a textfield with a placeholder but I didn't add one for every possible code path in the method because I think placeholder visibility behavior can generally be captured by other snapshot tests not necessarily devoted to it.

andrewoverton:
One thing I noticed was that the snapshot tests were occasionally reaching their timeout of 1 second and failing. I wanted a better sense of what the upper bound of execution time was so I increased the timeout to 30 seconds in this PR. I understand that that is not a good permanent solution, and I am going to try to figure out how to make that number go down. I just don't want any flakiness until that happens. Oof! Here are some metrics from four recent runs:

With 1 second timeout:
```
    . testFloatingLabelWithCustomColorWhileEditing (0.598 seconds)
    . testTextFieldWithLeadingView (0.288 seconds)
    . testTextFieldWithLeadingViewAndTrailingView (0.285 seconds)
    . testTextFieldWithLeadingViewWhileEditing (0.373 seconds)
    . testTextFieldWithText (0.276 seconds)
    . testTextFieldWithTrailingView (0.276 seconds)
    . testTextFieldWithVisibleClearButton (0.286 seconds)
```
With 1 second timeout: 
```
    . testDisabledTextField (0.309 seconds)
    . testEditingTextFieldWithVisiblePlaceholder (0.720 seconds)
    . testFloatingLabelWithCustomColorWhileEditing (0.355 seconds)
    . testTextFieldWithLeadingView (0.281 seconds)
    . testTextFieldWithLeadingViewAndTrailingView (0.269 seconds)
    . testTextFieldWithLeadingViewWhileEditing (0.356 seconds)
    . testTextFieldWithText (0.274 seconds)
    . testTextFieldWithTrailingView (0.270 seconds)
    . testTextFieldWithVisibleClearButton (0.291 seconds)
```
With 5 second timeout: 
```
    . testDisabledTextField (0.302 seconds)
    x testEditingTextFieldWithVisiblePlaceholder, Asynchronous wait failed: Exceeded timeout of 5 seconds, with unfulfilled expectations: "textfield_validation_expectation".
    . testFloatingLabelWithCustomColorWhileEditing (0.354 seconds)
    . testTextFieldWithLeadingView (0.290 seconds)
    . testTextFieldWithLeadingViewAndTrailingView (3.763 seconds)
    . testTextFieldWithLeadingViewWhileEditing (0.592 seconds)
    . testTextFieldWithText (0.290 seconds)
    . testTextFieldWithTrailingView (0.293 seconds)
    . testTextFieldWithVisibleClearButton (2.058 seconds)
```
With 30 second timeout:
```
    . testDisabledTextField (2.622 seconds)
    . testEditingTextFieldWithVisiblePlaceholder (12.499 seconds)
    . testFloatingLabelWithCustomColorWhileEditing (0.859 seconds)
    . testTextFieldWithLeadingView (0.356 seconds)
    . testTextFieldWithLeadingViewAndTrailingView (0.278 seconds)
    . testTextFieldWithLeadingViewWhileEditing (0.688 seconds)
    . testTextFieldWithText (0.298 seconds)
    . testTextFieldWithTrailingView (0.286 seconds)
    . testTextFieldWithVisibleClearButton (0.646 seconds)
```

PR #8460, opened 2019-09-18
[MDCTypography] Separate MDCTypopgraphyScheming from MDCTypography

### LINKED ISSUE #8445: [TypographyScheme] Separate protocol from classes
## Feature Request/Suggestion

MDCTypgoraphyScheming should be in its own target without dependency on the MDCTypographyScheme implementation.

## Context
Clients implementing their own conforming class don't need to import the default implementation.

---
#### Internal data
- Associated internal bug: [b/140931246](http://b/140931246)

bryanoltman:
Clients who want to conform to `MDCTypographyScheming` should not need to import the headers of `MDCTypographyScheme`. This should decrease the size of the `MDCTypographyScheming` target.

Closes #8445.


bryanoltman:
@googlebot I fixed it.

romoore (approved):
Not necessarily required in this PR, but it would be better to move this protocol into an extension target  instead. The main target could then depend on it. That way clients can use the protocol directly through Cocoapods or Bazel without having to pull-in the whole scheme.

Issue #8392, opened 2019-09-04
[Git LFS] Find a way to keep LFS images out of `stable`.


romoore:
The git merge strategy for `.gitattributes` isn't working. Instead, we should find a process- or tool-based solution to ensure that we aren't merging `.gitattributes` across branches. Alternatively, we should find another mechanism to keep LFS images out of the `stable` branch.

## Acceptance Criteria
1. There is a process to follow or tool to use that ensures Git LFS images are not merged into `stable`.
2. Checking out the `stable` branch does not require Git LFS to be installed on the client.
3. `develop` branch retains Git LFS support.

---
#### Internal data
- Associated internal bug: [b/140723402](http://b/140723402)

PR #8478, opened 2019-09-23
[Git LFS] Fix .gitattributes in `develop`.

romoore:
`.gitattributes` was accidentally merged back from `stable` to `develop`.

Prework for #8392


jverkoey (approved):
Would love it if there was a way to fix this problem :)

PR #8510, opened 2019-09-26
[TextFields] Adding border stroke color

galiak11:
Adding a borderStrokeColor property to input text field to allow better contrast ratio when theming.

See also: cl/270937632.

Example usage of new API in theming:

Before (self.normalColor was used to theme the border):
image

After (self.borderStrokeColor is used to theme the border):
image

### DIFF
--- a/components/TextFields/src/MDCTextInputControllerBase.h
+++ b/components/TextFields/src/MDCTextInputControllerBase.h
@@ -72,6 +72,13 @@ extern const CGFloat MDCTextInputControllerBaseDefaultBorderRadius;
 @property(class, nonatomic, null_resettable, strong) UIColor *borderFillColorDefault;
 
 /**
+ The color the input field's border in the resting state. A nil value yields a clear border.
+
+ Default is clear.
+ */
+@property(nonatomic, nullable, strong) UIColor *borderStrokeColor;
+
+/**
  Should the controller's .textInput grow vertically as new lines are added.
 
  If the text input does not conform to MDCMultilineTextInput, this parameter has no effect.
--- a/components/TextFields/src/MDCTextInputControllerBase.h
+++ b/components/TextFields/src/MDCTextInputControllerBase.h
@@ -74,7 +74,7 @@ extern const CGFloat MDCTextInputControllerBaseDefaultBorderRadius;
 /**
  The color the input field's border in the resting state. A nil value yields a clear border.
 
- Default is clear.
+ Default is nil.
  */
 @property(nonatomic, nullable, strong) UIColor *borderStrokeColor;
 
--- a/components/TextFields/tests/unit/OutlinedTextFieldColorThemerTests.m
+++ b/components/TextFields/tests/unit/OutlinedTextFieldColorThemerTests.m
@@ -160,13 +160,10 @@
   MDCTextField *textField = [[MDCTextField alloc] init];
   MDCTextInputControllerOutlined *controller =
       [[MDCTextInputControllerOutlined alloc] initWithTextInput:textField];
-  MDCSemanticColorScheme *colorScheme =
-      [[MDCSemanticColorScheme alloc] initWithDefaults:MDCColorSchemeDefaultsMaterial201804];
 
   // When
   controller.borderStrokeColor = UIColor.redColor;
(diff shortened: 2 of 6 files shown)


components/TextFields/src/MDCTextInputControllerBase.h
```diff
@@ -71,6 +71,13 @@ extern const CGFloat MDCTextInputControllerBaseDefaultBorderRadius;
  */
 @property(class, nonatomic, null_resettable, strong) UIColor *borderFillColorDefault;
 
+/**
+ The color the input field's border in the resting state. A nil value yields a clear border.
+
+ Default is clear.
```
randallli: Default is nil

components/TextFields/src/MDCTextInputControllerBase.h
```diff
@@ -71,6 +71,13 @@ extern const CGFloat MDCTextInputControllerBaseDefaultBorderRadius;
  */
 @property(class, nonatomic, null_resettable, strong) UIColor *borderFillColorDefault;
 
+/**
+ The color the input field's border in the resting state. A nil value yields a clear border.
```
randallli, line 75: A nil value uses the normal color.

at least that is my read of the code.

components/TextFields/tests/unit/OutlinedTextFieldColorThemerTests.m
```diff
@@ -154,4 +155,39 @@ - (void)_disabled_testThemerWithBaselineColorSchemeAppliesToMultilineTextFieldDi
...
+  // Given
+  MDCTextField *textField = [[MDCTextField alloc] init];
+  MDCTextInputControllerOutlined *controller =
+      [[MDCTextInputControllerOutlined alloc] initWithTextInput:textField];
+  MDCSemanticColorScheme *colorScheme =
+      [[MDCSemanticColorScheme alloc] initWithDefaults:MDCColorSchemeDefaultsMaterial201804];
+
+  // When
+  controller.borderStrokeColor = UIColor.redColor;
+  [MDCOutlinedTextFieldColorThemer applySemanticColorScheme:colorScheme
```
randallli: Calling the themer is not relevant and is confusing for this test. let us remove it

components/TextFields/tests/unit/OutlinedTextFieldColorThemerTests.m
```diff
@@ -154,4 +155,39 @@ - (void)_disabled_testThemerWithBaselineColorSchemeAppliesToMultilineTextFieldDi
...
+  // Given
+  MDCTextField *textField = [[MDCTextField alloc] init];
+  MDCTextInputControllerOutlined *controller =
+      [[MDCTextInputControllerOutlined alloc] initWithTextInput:textField];
+  MDCSemanticColorScheme *colorScheme =
+      [[MDCSemanticColorScheme alloc] initWithDefaults:MDCColorSchemeDefaultsMaterial201804];
+
+  // When
+  controller.borderStrokeColor = nil;
+  [MDCOutlinedTextFieldColorThemer applySemanticColorScheme:colorScheme
```
randallli: let us just set the normal color and check that logic rather than assume that the color themer will do so. By calling the colorThemer we are also assuming that it will not touch the boarderStrokeColor in the future.

randallli (approved):
How did you make your snapshots? i didn't see anything like this in 
https://github.com/material-components/material-components-ios/tree/develop/snapshot_test_goldens/goldens_64/MDCTextFieldOutlinedControllerBaselineSnapshotTests

It would be awesome if we could lean in here and have snapshot coverage with like we did here:
https://github.com/material-components/material-components-ios/blob/develop/components/TextFields/tests/snapshot/MDCTextFieldOutlinedControllerBaselineLeadingImageSnapshotTests.m

I took the snapshots from here (cl/270937632):
https://scuba.corp.google.com/#/report/78c74ca7-40a3-4cdf-973f-47adee4415ad?query=&groupBy=Golden+Location&pageSize=25&pageIndex=0&selected_node=78c74ca7-40a3-4cdf-973f-47adee4415ad&status=2

components/TextFields/src/MDCTextInputControllerBase.h
galiak11: The base class is shared with both Filled and outlined buttons, so the default is a clear border.

PR #8512, opened 2019-09-26
[Tabs] Correct MDCTabBarView behavior for contentInset.

### LINKED ISSUE #8236: [Tabs] Layout is not correctly handling content insets
When setting a custom value for `contentInsets`, the `MDCTabBarView` is not laying-out the `FixedJustified` internal layout style.

iphone x - inset 50 points left and right - rightmost item selected

## Steps to reproduce
1. Set `preferredLayoutStyle` so that it will render in `Fixed` or `ClusteredTrailing`.
2. Set `contentInset` with a non-zero left value.

---
#### Internal data
- Associated internal bug: [b/139064657](http://b/139064657)

romoore:
When `contentInset` is non-zero, MDCTabBarView was misplacing the tabs views.
This PR corrects the issue by making the following changes:

*  Clustered Fixed tabs now set the `contentSize` to fill the entire safe area
   bounds.
*  Justified tabs tabs now expand the `contentSize` to fill at least the entire safe
   area bounds.
*  Right-aligned tabs position tabs views only within the content area and no
   longer depend on the min-X position of the bounds.
*  Rendering the divider line at the end of `layoutSubviews` to handle cases
   where scrolling adjusted the X-offset of the bounds.

Fixes #8236

### DIFF
--- a/components/Tabs/src/TabBarView/MDCTabBarView.m
+++ b/components/Tabs/src/TabBarView/MDCTabBarView.m
@@ -613,9 +613,6 @@ typedef NS_ENUM(NSUInteger, MDCTabBarViewInternalLayoutStyle) {
     }
   }
 
-  self.bottomDividerView.frame =
-      CGRectMake(CGRectGetMinX(self.bounds), CGRectGetMaxY(self.bounds) - kBottomDividerHeight,
-                 CGRectGetWidth(self.bounds), kBottomDividerHeight);
   self.contentSize = [self calculatedContentSize];
   [self updateSelectionIndicatorToIndex:[self.items indexOfObject:self.selectedItem]];
 
@@ -631,6 +628,11 @@ typedef NS_ENUM(NSUInteger, MDCTabBarViewInternalLayoutStyle) {
     }
     [self scrollUntilSelectedItemIsVisibleWithoutAnimation];
   }
+  // It's possible that after scrolling the minX of bounds could have changed. Positioning it last
+  // ensures that its frame matches the displayed content bounds.
+  self.bottomDividerView.frame =
+      CGRectMake(CGRectGetMinX(self.bounds), CGRectGetMaxY(self.bounds) - kBottomDividerHeight,
+                 CGRectGetWidth(self.bounds), kBottomDividerHeight);
 }
 
 - (void)traitCollectionDidChange:(UITraitCollection *)previousTraitCollection {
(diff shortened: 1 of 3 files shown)


components/Tabs/src/TabBarView/MDCTabBarView.m
```diff
@@ -789,15 +790,28 @@ - (CGSize)intrinsicContentSize {
...
+      CGSize boundsSize = CGSizeMake(CGRectGetWidth(self.bounds), CGRectGetHeight(self.bounds));
+
+      return CGSizeMake(MAX(boundsSize.width, intrinsicContentSize.width),
+                        MAX(boundsSize.height, intrinsicContentSize.height));
     }
     case MDCTabBarViewInternalLayoutStyleFixedClusteredCentered:
     case MDCTabBarViewInternalLayoutStyleFixedClusteredTrailing:
     case MDCTabBarViewInternalLayoutStyleFixedClusteredLeading: {
-      return [self intrinsicContentSizeForClusteredLayout];
+      return [self availableSizeForSubviewLayout];
```
randallli, line 785: Im not sure if im following this. can you elaborate more on why this is different from the other cases?

also why does intrinsicContentSize also have a similar change?

romoore: Are you asking why Clustered layouts use the full width of the view rather than the intrinsic content size of that layout? 

I didn't make any changes to `intrinsicContentSize` in this PR. Could you please provide more detail about what you mean?

randallli: > Are you asking why Clustered layouts use the full width of the view rather than the intrinsic content size of that layout?

yes

> I didn't make any changes to intrinsicContentSize in this PR. Could you please provide more detail about what you mean?

I saw that intrinsicContentSize and calculatedContentSize do very similar things but now they are drifting apart and im trying to understand why one is changing and not the other.

romoore: Clustered tabs fill the entire space available (as noted in the summary). This is so that they can be positioned accurately while allowing `contentInset` to behave normally for UIScrollView. This is only really relevant to right-aligned and center-clustered Tabs. Right-aligned Tabs' behavior is noted in the summary, but I didn't call-out center-clustered Tabs. Did you want me to do so?

UIView's `intrinsicContentSize` and UIScrollView's `contentSize` are very different concepts. That they might have been similar was an artifact of the implementation (and partly from this bug). Is there a specific reason why you feel they should not be very different?  I'm sorry, but I don't understand why they might be related.

randallli: ok sgtm

randallli (approved):

PR #8514, opened 2019-09-26
[NavigationDrawer] Fix broken header when content expands

codeman7:
In #8503 this line was removed but after further testing when the contentViewController expands it's preferredContentSize this causes a regression in the header behavior. With this line added back into the check this behavior is fixed.


yarneo (approved):
LGTM % clang-format

PR #8524, opened 2019-09-27
[TextFields] Rename "contained input view" to "text control"

andrewoverton:
In this PR I am beginning the naming transition from "contained input view" to "text control." This is just a renaming PR, there are no other changes.

At some point after this I want to make it its own component, rather than an extension on TextFields, in the spirit of Ripple and Ink.

Related to #6942.


yarneo (commented):
Doesn't the directory name of ContainedInputView along with the BUILD file also need to be changed?

andrewoverton:
I'm going to do all that later. Like separating the tests and changing the podfiles and BUILD files and whatnot.

yarneo (approved):

PR #8526, opened 2019-09-27
[TextFields] Adding floatingPlaceholder Error Active color API to cont…

galiak11:
Adding an error color for the active state of the floating placeholder label (the label that floats up when the TextField is in editing mode).

Also note cl/271208733.

Example usage of new API in theming:

**Before** (the floating placeholder label defaults to the error color when editing): 
image

**After** (when assigning UIColor.orange to floatingPlaceholderErrorActiveColor):
image

### DIFF
--- a/components/TextFields/src/MDCTextInputControllerBase.m
+++ b/components/TextFields/src/MDCTextInputControllerBase.m
@@ -103,6 +103,7 @@ static UITextFieldViewMode _underlineViewModeDefault = UITextFieldViewModeWhileE
   UIColor *_errorColor;
   UIColor *_floatingPlaceholderActiveColor;
   UIColor *_floatingPlaceholderNormalColor;
+  UIColor *_floatingPlaceholderErrorActiveColor;
   UIColor *_inlinePlaceholderColor;
   UIColor *_leadingUnderlineLabelTextColor;
   UIColor *_normalColor;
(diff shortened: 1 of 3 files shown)


components/TextFields/src/MDCTextInputControllerBase.m
```diff
@@ -956,6 +961,17 @@ + (void)setFloatingPlaceholderNormalColorDefault:(UIColor *)floatingPlaceholderN
                                                : [self class].inlinePlaceholderColorDefault;
 }
 
+- (UIColor *)floatingPlaceholderErrorActiveColor {
+  return _floatingPlaceholderErrorActiveColor;
+}
+
+- (void)setFloatingPlaceholderErrorActiveColor:(UIColor *)floatingPlaceholderErrorActiveColor {
+  if (![_floatingPlaceholderErrorActiveColor isEqual:floatingPlaceholderErrorActiveColor]) {
```
andrewoverton, line 969: Color equality is really weird on iOS (hence all of our new test helpers for comparing colors) so I would suggest not bothering with a call to `-isEqual:`. I would just set it :D

andrewoverton (approved):

components/TextFields/src/MDCTextInputControllerBase.m
galiak11: That makes sense. However I saw other places in this file use this pattern, so I think it's better to keep it for consistency sake.

PR #8530, opened 2019-09-30
[TextFields] Bring more of base positioning reference over

andrewoverton:
This PR brings more of the MDCTextControlVerticalPositioningReferenceBase over. The positioning reference objects provide information about the vertical positioning of subviews in MDCTextControls. Each MDCTextControlStyle vends a different positioning reference object to be used by the layout object.

It is important to remember that this PR does not bring over everything the positioning reference object does. Things like MDCTextControl density (dense textfields are textfields that are squished vertically so more textfields can be shown using less space) are made possible with this object.

Related to #6942.


randallli:
ns/goldens_64/MDCBaseTextFieldTestsSnapshotTests/testTextFieldWithLeadingView_11_2@2x.png 
looks not quite correct. The text use to be center aligned but now it seems to be bottom aligned and looks odd. is this intentional? which guidlines are we fixing for this body of work?

andrewoverton:
Eventually there will be behavior that will make it so that when there is no label text the text area won't go down so much. What you see in the snapshot is temporary.

andrewoverton:
I don't think I understand the "which guidlines are we fixing for this body of work" question.

components/TextFields/src/ContainedInputView/private/MDCTextControlStyleBase.m
```diff
@@ -40,4 +40,16 @@ - (void)removeStyleFrom:(id<MDCTextControl>)textControl {
   return [[MDCTextControlVerticalPositioningReferenceBase alloc] init];
 }
 
+- (id<MDCTextControlVerticalPositioningReference>)
+    positioningReferenceWithFloatingFontLineHeight:(CGFloat)floatingLabelHeight
+                              normalFontLineHeight:(CGFloat)normalFontLineHeight
+                                     textRowHeight:(CGFloat)textRowHeight
+                                  numberOfTextRows:(CGFloat)numberOfTextRows {
+  return [[MDCTextControlVerticalPositioningReferenceBase alloc]
```
randallli, line 48: It is a little odd to be spinning up a new object every time that we call this.  We want to explain this in a comment since it is a bit unexpected.

Is this called often? every layoutSubviews? should it be a class method? Are we changing this at runtime? im trying to understand the architectural benefits of using the class.

randallli: I spoke offline with andrew. A comment explaining the protocol's intended use is in the works.

randallli (approved):

randallli:
> I don't think I understand the "which guidlines are we fixing for this body of work" question.

I was trying to ask for a mock of what we are trying to accomplish. But it sounds like this is an intermediate step backward to make forward steps in the future?

components/TextFields/src/ContainedInputView/private/MDCTextControlStyleBase.m
andrewoverton: A layout object is calculated on every layout pass, and positioning reference objects are needed for layout calculation, so a positioning reference is created on every layout pass. I don't think it should be a class method--it's useful to have an object with the various padding values as properties. It makes it portable.

PR #8536, opened 2019-09-30
v92.1.0

randallli:
#### Don't forget:
- [ ] Please add unit tests, snapshot tests, or manual testing steps depending on the change.
- [ ] Sign the CLA bot. You can do this once the pull request is opened.

[Contributing](https://github.com/material-components/material-components-ios/tree/develop/contributing) has more information and tips for a great
pull request.


components/Tabs/examples/MDCTabBarViewTypicalExampleViewController.m
```diff
@@ -14,13 +14,23 @@
 
 #import <UIKit/UIKit.h>
 
+#import <MaterialComponents/MaterialActionSheet+Theming.h>
+#import <MaterialComponents/MaterialActionSheet.h>
 #import <MaterialComponents/MaterialAnimationTiming.h>
 #import <MaterialComponents/MaterialContainerScheme.h>
+#import <MaterialComponents/MaterialIcons+ic_check.h>
```
randallli, line 21: this is causing issues with the import script. it is expecting it in the format found in 
https://github.com/material-components/material-components-ios/blob/800188a64994c68d089af849ea7cd080b1a4d4ab/components/ButtonBar/examples/ButtonBarIconExample.m#L18

.gitattributes
```diff
@@ -1,25 +1,3 @@
...
-# to be propagated from the `develop` branch to the `stable` branch. In nearly
-# all cases, it should not be propagated from `develop` to `stable`.
-#
-# If you are a releaser and see this file change and you're not sure why, you
-# might have accidentally skipped [setting the correct
-# driver in your cloned
-# repository](https://github.com/material-components/material-components-ios/blob/develop/contributing/releasing.md#configure-the-merge-strategy-for-gitattributes).
-# If that's the case, please either revert the accidental change manually or
-# restart the release with a fresh clone and the correct driver.
+# Do not merge this version into `stable`.
```
randallli: @romoore am i suppose to revert this change for the release?

components/Tabs/examples/MDCTabBarViewTypicalExampleViewController.m
randallli: @romoore can you catch me up on why <> notation was used here?

romoore: `MDCTabBarView` is in beta, so we use the <> brackets to reference other podspecs. If this isn't necessary any longer, then I believe we can update many of our Beta component source files.

romoore: You will probably have to update the import script to handle these Beta components until we are allowed to mark them "Ready".

.gitattributes
romoore: If possible, please try to remove that change from the merge commit. If not possible, then yes a revert is necessary. If you perform a revert on `release-candidate` you'll have to revert that revert when you merge back to `develop`. 

I believe our current process that branches from `develop` rather than branching from `stable` makes it more difficult to manage these differences.  It may be simpler to handle the two versions if we instead branch from `stable`, create a merge commit from `develop` (and omit these changes before committing), then fast-forward stable from there.

PR #8539, opened 2019-10-02
[TextFields] Add assistive label view

andrewoverton:
This PR adds assistive labels to MDCBaseTextField. It does so using a private class called MDCTextControlAssistiveLabelView. The textfield's layout object determines the frame of the MDCTextControlAssistiveLabelView just like it does with any other subview. What makes the MDCTextControlAssistiveLabelView different from other subviews of MDCBaseTextField is that it also has its own layout object. The MDCBaseTextFieldLayout calculates an MDCTextControlAssistiveLabelViewLayout which the MDCTextField applies to the MDCTextControlAssistiveLabelView at the same time it applies the rest of the layout.

The code in MDCBaseTextFieldLayout's layout calculation method ends up looking like this:

```Objective-c
  self.assistiveLabelViewLayout = [[MDCTextControlAssistiveLabelViewLayout alloc]
                         initWithWidth:textFieldWidth
                    leftAssistiveLabel:leftAssistiveLabel
                   rightAssistiveLabel:rightAssistiveLabel
            assistiveLabelDrawPriority:assistiveLabelDrawPriority
      customAssistiveLabelDrawPriority:customAssistiveLabelDrawPriority
                     horizontalPadding:kHorizontalPadding
           paddingAboveAssistiveLabels:positioningReference.paddingAboveAssistiveLabels
           paddingBelowAssistiveLabels:positioningReference.paddingBelowAssistiveLabels
                                 isRTL:isRTL];
  self.assistiveLabelViewFrame = CGRectMake(0, positioningReference.containerHeight, textFieldWidth,
                                            self.assistiveLabelViewLayout.calculatedHeight);
```

And the code in MDCBaseTextField's `-postLayoutSubviews` method ends up looking like this:
```
  self.assistiveLabelView.frame = self.layout.assistiveLabelViewFrame;
  self.assistiveLabelView.layout = self.layout.assistiveLabelViewLayout;
  [self.assistiveLabelView setNeedsLayout];
```
All the other MDCTextControls I plan on building will use this exact same pattern.

Related to #6942.


components/TextFields/src/ContainedInputView/MDCBaseTextField.h
```diff
@@ -36,6 +36,18 @@
...
+ The @c leadingAssistiveLabel is a label below the text on the leading edge of the view. It can be
+ used to display helper or error text.
+ */
+@property(strong, nonatomic, readonly, nonnull) UILabel *leadingAssistiveLabel;
+
+/**
+ The @c trailingAssistiveLabel is a label below the text on the trailing edge of the view. It can be
+ used to display helper or error text.
+ */
+@property(strong, nonatomic, readonly, nonnull) UILabel *trailingAssistiveLabel;
```
randallli, line 49: Future recommendation: Split this into multiple smaller PRs. Adding the assistive labels is a different "idea" than setting the colors of those labels.

components/TextFields/tests/snapshot/MDCBaseTextFieldSnapshotTests.m
```diff
@@ -220,4 +220,18 @@ - (void)testEditingTextFieldWithVisiblePlaceholder {
   [self validateTextField:textField];
 }
 
+- (void)testTextFieldWithAssistiveLabelText {
```
randallli, line 223: Since we are adding color: forState: APIs in this PR I was expecting additional snapshots showing what it looks like when we set these values to colors to ensure they continue to affect the visual appearance of our text control. One per state to show off that it works.

components/TextFields/src/ContainedInputView/private/MDCBaseTextFieldLayout.m
```diff
@@ -53,6 +58,10 @@ - (instancetype)initWithTextFieldSize:(CGSize)textFieldSize
                              rightViewMode:rightViewMode
                      clearButtonSideLength:clearButtonSideLength
                            clearButtonMode:clearButtonMode
+                        leftAssistiveLabel:leftAssistiveLabel
+                       rightAssistiveLabel:rightAssistiveLabel
+                assistiveLabelDrawPriority:assistiveLabelDrawPriority
+          customAssistiveLabelDrawPriority:customAssistiveLabelDrawPriority
```
randallli, line 64: nit: The number of arguments passed in here is rather large. It would be helpful to those stepping through this logic to understand why we are doing so. I recommend a comment  explaining some of this inline next to the definition.

components/TextFields/src/ContainedInputView/private/MDCTextControlAssistiveLabelViewLayout.h
```diff
@@ -0,0 +1,38 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+#import "MDCTextControl.h"
+#import "MDCTextControlAssistiveLabelDrawPriority.h"
+
+@interface MDCTextControlAssistiveLabelViewLayout : NSObject
```
randallli, line 27: Lets add a class doc here. This is my understanding of the purpose of this object:
This functional object will compute layout and hold the results of that computation in its readonly properties. 

Other ways that we can do something like this:
https://stackoverflow.com/questions/1692005/returning-multiple-values-from-a-method-in-objective-c

components/TextFields/src/ContainedInputView/private/MDCTextControlAssistiveLabelViewLayout.m
```diff
@@ -0,0 +1,175 @@
...
+
+@implementation MDCTextControlAssistiveLabelViewLayout
+
+#pragma mark Object Lifecycle
+
+- (instancetype)initWithWidth:(CGFloat)superviewWidth
+                  leftAssistiveLabel:(UILabel *)leftAssistiveLabel
+                 rightAssistiveLabel:(UILabel *)rightAssistiveLabel
+          assistiveLabelDrawPriority:
+              (MDCTextControlAssistiveLabelDrawPriority)assistiveLabelDrawPriority
```
randallli, line 33: trying to follow along here: How does a client choose a different draw priority? We should have a test turning that knob. (we might already have this, I didn't see anything in the PR)

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
```diff
@@ -105,4 +117,18 @@
  */
 - (nonnull UIColor *)textColorForState:(MDCTextControlState)state;
 
+/**
+ Sets the assistive label color for a given state.
```
randallli: copy edit suggestion: Sets the leading and trailing assistive label colors.

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
```diff
@@ -105,4 +117,18 @@
...
+ @param assistiveLabelColor The UIColor for the given state.
+ @param state The MDCTextControlState.
+ */
+- (void)setAssistiveLabelColor:(nonnull UIColor *)assistiveLabelColor
```
randallli: trying to understand: 
We are setting both leading and trailing colors here. It seems like a convenience method for more specific leading and trailing APIs. Why did we not create the setLeadingAssistiveLabelColor and setTrailingAssistiveLabelColor?

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
```diff
@@ -105,4 +117,18 @@
...
+                      forState:(MDCTextControlState)state;
+
+/**
+ Returns the assistive label color for a given state.
+ @param state The MDCTextControlState.
+ */
+- (nonnull UIColor *)assistiveLabelColorForState:(MDCTextControlState)state;
```
randallli: In the future if clients want different values for the leading and trailing the value gotten from this getter will not be obvious.

components/TextFields/src/ContainedInputView/MDCBaseTextField.m
```diff
@@ -205,6 +231,16 @@ - (MDCBaseTextFieldLayout *)calculateLayoutWithTextFieldSize:(CGSize)textFieldSi
                                     numberOfTextRows:1];
 }
 
+- (CGFloat)normalizedCustomAssistiveLabelDrawPriority:(CGFloat)customPriority {
```
randallli: s/nomralized/clamp/

This is more often called clamping
https://en.m.wikipedia.org/wiki/Clamping_(graphics)

randallli (changes requested):
I was expecting more snapshot tests given the public API additions. Specifically for the color setters.

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
andrewoverton: Ah, I see. Ok.

components/TextFields/src/ContainedInputView/private/MDCTextControlAssistiveLabelViewLayout.m
andrewoverton: No API associated with it ever got approved so my plan was going to be to just land it without exposing anything and then expose it if/when a client wanted something other than a priority of `.trailing`.

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
andrewoverton: I guess just because they always seem to be the same color in the guidelines. MDCTextField didn't allow you to set them separately either FWIW. Do you think we should have separate APIs for the leading and trailing assistive labels?

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
andrewoverton: It sort of sounds like you want separate APIs for the leading and trailing assistive label colors

components/TextFields/src/ContainedInputView/MDCBaseTextField.m
andrewoverton: Ahhhh nice, thank you!

components/TextFields/src/ContainedInputView/private/MDCTextControlAssistiveLabelViewLayout.m
andrewoverton: There are actually a few things that I built solutions for that do not currently have any public APIs associated with them. This is one of them. These are the other two:
1. `density`. TextFields are dense if there is very little vertical padding between the various subviews. With my solution for density there is a CGFloat property on the view that clients would be able to specify a value for between 0 and 1 inclusive where 0 is not dense (this is the default) and 1 is dense.
2. `preferredContainerHeight`. The container is the region in the outline for outlined textfields and the area of the filled background in filled textfields. One concern that I thought clients might have about these textfields was that their heights were affected by things like font, and there wasn't really a way to specify exactly how tall they should be, which iOS developers are pretty used to being able to do. To address that potential concern I came up with this notion of `preferredContainerHeight`. It's also a CGFloat. When it's set to a non-zero value the layout object tries to come up with a layout that results in the container of the textfield being exactly that many points tall. This property would probably take precedence over density if both were set.

What should we do about these things that we don't immediately plan to make public? Should we throw them away? Or maybe we can bring them in and watch clients' behavior for a while and then later make the decision to delete them (this wouldn't be a breaking change in any way) or expose them? I think we should do the latter. I don't think there is any harm in bringing them in, even if it's temporary. It's also actually just easier for me and will probably result in things landing more quickly.

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
randallli: yes please. I think it will be more scale able if clients want to add them.

components/TextFields/src/ContainedInputView/private/MDCTextControlAssistiveLabelViewLayout.m
randallli: Ok, lets leave as is. It may be worth adding a comment to explain this private feature not yet asked for by clinets. (you may have already done so and I missed this)

components/TextFields/tests/snapshot/MDCBaseTextFieldSnapshotTests.m
andrewoverton: I updated the existing assistive label snapshot test to have different colors, but I kind of think having a snapshot test for each state is overkill... I did update the unit test to test the leading/trailing accessors on a per state basis though.

components/TextFields/src/ContainedInputView/private/MDCTextControlAssistiveLabelViewLayout.h
andrewoverton: I added a class doc. I'd rather not do the "return an NSDictionary" thing in that stack overflow thread.

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
andrewoverton: Done.

components/TextFields/src/ContainedInputView/MDCBaseTextField.h
andrewoverton: I separated it into two accessors.

components/TextFields/src/ContainedInputView/private/MDCBaseTextFieldLayout.m
andrewoverton: I added an explanation in d277b461bc2aa7d9095c90e4c57ee923cb0663e6

components/TextFields/tests/snapshot/MDCBaseTextFieldSnapshotTests.m
randallli: I dont want to add a feature without testing it. The feature: setting the color of the label for different states.

A snapshot test for confirming that the color of the leadingAssitiveLabel changes colors in the editing state is more scalable than gif's demoing it every time we adjust this code.

The other option is to make unit tests that inspects the color of the leadingAssistiveLabel while in the normal, editing and disabled states.

randallli: We are already reducing our code coverage with the hidden features (MDCTextControlAssistiveLabelDrawPriority) in this PR. Lets cover as much as possible in our publicly available APIs so our numbers stay up.

randallli (approved):
Im assuming that we will add more tests for the states.

components/TextFields/tests/snapshot/MDCBaseTextFieldSnapshotTests.m
andrewoverton: The correct coloring of subviews on a per state basis is dependent on three methods: `-setTextControlColorViewModel:forState:`, `-textControlColorViewModelForState:`, and `-applyColorViewModel:withLabelState:`. Maybe if I throughly unit test all three that will be sufficient. I can also add snapshot tests for all three states. I'm happy to do either in the next PR.

At some point, when more of the textfields have landed, I do want to go through and rethink a lot of the snapshot tests I've already added. They're going to change a lot.

andrewoverton: Ah, it seems like you want more added in this PR rather than in the next one, I can do that 👍

PR #8543, opened 2019-10-06
[TextFields] Minor cleanup

andrewoverton:
This PR contains some minor cleanup for TextFields.

Related to #6942.

### DIFF
--- a/components/TextFields/src/ContainedInputView/MDCBaseTextField.m
+++ b/components/TextFields/src/ContainedInputView/MDCBaseTextField.m
@@ -56,7 +56,7 @@
 - (instancetype)initWithFrame:(CGRect)frame {
   self = [super initWithFrame:frame];
   if (self) {
-    [self commonMDCInputTextFieldInit];
+    [self commonMDCBaseTextFieldInit];
   }
   return self;
 }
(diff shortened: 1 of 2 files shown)


components/TextFields/src/ContainedInputView/MDCBaseTextField.m
```diff
@@ -490,7 +492,7 @@ - (UIFont *)normalFont {
 }
 
 - (UIFont *)floatingFont {
-  return [self.normalFont fontWithSize:(self.normalFont.pointSize * (CGFloat)0.5)];
+  return [self.containerStyle floatingFontWithNormalFont:self.normalFont];
```
randallli, line 495: This seems like it would result in a snapshot delta. If we dont already I would like to make sure that we have one in place.

randallli (approved):
Please double check our testing coverage here.

components/TextFields/src/ContainedInputView/MDCBaseTextField.m
andrewoverton: It doesn't result in a snapshot delta because I changed the floating font scale factor. This stuff is still in flux a bit.

Issue #8595, opened 2019-10-16
All components should support UIContentSizeCategoryAdjusting behavior


romoore:
## Feature Request/Suggestion
All components should support `adjustsFontForContentSizeCategory` APIs appropriate for the implementation. Clients should be able to set a `-[UIFont preferredFont...]` API font, set `adjustsFontForContentSizeCategory = YES` on the appropriate view or view controller, and see the fonts respond automatically for Dynamic Type.

PR #8606, opened 2019-10-18
[Buttons] Add API to disable state-based fonts.

romoore:
Introduces a new API to disable the `titleFont:forState:` APIs. This is the
first step toward removing those APIs.

State-based fonts make it impossible to support UIContentSizeCategoryAdjusting
behavior on the button's `titleLabel`. Providing support for automatic font
resizing for Dynamic Type means removing state-based font APIs.

Part of #8595

### DIFF
--- a/components/Buttons/tests/unit/MDCButtonTests.m
+++ b/components/Buttons/tests/unit/MDCButtonTests.m
@@ -590,6 +590,77 @@ static NSString *controlStateDescription(UIControlState controlState) {
                 self.button.titleLabel.font);
 }
 
+- (void)testTitleFontForStateDisabledAfterSettingFontsPreventsFontChange {
+  // Given
+  UIFont *normalFont = [UIFont systemFontOfSize:10];
+  UIFont *selectedFont = [UIFont systemFontOfSize:40];
+  UIFont *directlyAssignedFont = [UIFont systemFontOfSize:25];
+  [self.button setTitleFont:normalFont forState:UIControlStateNormal];
+  [self.button setTitleFont:selectedFont forState:UIControlStateSelected];
+
+  // When
+  self.button.enableTitleFontForState = NO;
+  self.button.titleLabel.font = directlyAssignedFont;
+  self.button.selected = YES;
+
+  // Then
+  XCTAssertEqualObjects(self.button.titleLabel.font, directlyAssignedFont);
+}
+
+- (void)testTitleFontForStateDisabledBeforeSettingFontsPreventsFontChange {
+  // Given
+  UIFont *normalFont = [UIFont systemFontOfSize:10];
+  UIFont *selectedFont = [UIFont systemFontOfSize:40];
+  UIFont *directlyAssignedFont = [UIFont systemFontOfSize:25];
+  self.button.enableTitleFontForState = NO;
+  self.button.titleLabel.font = directlyAssignedFont;
+
+  // When
+  [self.button setTitleFont:normalFont forState:UIControlStateNormal];
+  [self.button setTitleFont:selectedFont forState:UIControlStateSelected];
+  self.button.selected = YES;
+
+  // Then
+  XCTAssertEqualObjects(self.button.titleLabel.font, directlyAssignedFont);
+}
+
+- (void)testTitleFontForStateReenabledDoesNotImmediatelyUpdateFont {
+  // Given
+  UIFont *normalFont = [UIFont systemFontOfSize:10];
+  UIFont *selectedFont = [UIFont systemFontOfSize:40];
+  UIFont *directlyAssignedFont = [UIFont systemFontOfSize:25];
+  self.button.enableTitleFontForState = NO;
+  [self.button setTitleFont:normalFont forState:UIControlStateNormal];
+  [self.button setTitleFont:selectedFont forState:UIControlStateSelected];
+  self.button.titleLabel.font = directlyAssignedFont;
+  self.button.selected = YES;
+
+  // When
+  self.button.enableTitleFontForState = YES;
+
+  // Then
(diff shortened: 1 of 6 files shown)


components/Buttons/tests/unit/MDCButtonTests.m
```diff
@@ -590,6 +590,77 @@ - (void)testTitleFontForStateWithLegacyDynamicTypeRestoresUserFonts {
...
+  // When
+  [self.button setTitleFont:normalFont forState:UIControlStateNormal];
+  [self.button setTitleFont:selectedFont forState:UIControlStateSelected];
+  self.button.selected = YES;
+
+  // Then
+  XCTAssertEqualObjects(self.button.titleLabel.font, directlyAssignedFont);
+}
+
+- (void)testTitleFontForStateReenabledDoesNotImmediatelyUpdateFont {
```
randallli: can you elaborate why its important that we do not update the font immediately once we set this flag?

components/Buttons/tests/unit/MDCButtonTests.m
```diff
@@ -590,6 +590,77 @@ - (void)testTitleFontForStateWithLegacyDynamicTypeRestoresUserFonts {
...
+  // When
+  self.button.enableTitleFontForState = NO;
+  self.button.titleLabel.font = directlyAssignedFont;
+  self.button.selected = YES;
+
+  // Then
+  XCTAssertEqualObjects(self.button.titleLabel.font, directlyAssignedFont);
+}
+
+- (void)testTitleFontForStateDisabledBeforeSettingFontsPreventsFontChange {
```
randallli, line 618: s/disabled/selected/
looks like we are setting the selected font not the disabled font.

and above.

components/Buttons/tests/unit/MDCButtonTests.m
```diff
@@ -590,6 +590,77 @@ - (void)testTitleFontForStateWithLegacyDynamicTypeRestoresUserFonts {
...
+  [self.button setTitleFont:normalFont forState:UIControlStateNormal];
+  [self.button setTitleFont:selectedFont forState:UIControlStateSelected];
+
+  // When
+  self.button.enableTitleFontForState = NO;
+  self.button.titleLabel.font = directlyAssignedFont;
+  self.button.selected = YES;
+
+  // Then
+  XCTAssertEqualObjects(self.button.titleLabel.font, directlyAssignedFont);
```
randallli, line 611: I think a comment needs to be added to this test to explain the reason we dont want the font to change.
A vanilla read of the test just seems to be confirming incorrect behavior: that the font does not change to the selectedFont specified by setTitleFont:forState:Selected.

In person you mentioned that its for dynamic type reasons but we see no dynamic type explanation here.

randallli (changes requested):

components/Buttons/tests/unit/MDCButtonTests.m
romoore: It's not important, but it seems unexpected. If you want, I can change the behavior.

components/Buttons/tests/unit/MDCButtonTests.m
romoore: The name of the test reflects disabling `titleFontForState`, not the button state. Can you recommend a clearer name?

components/Buttons/tests/unit/MDCButtonTests.m
romoore: We don't want the font to change because the API to enable that behavior was set to `NO`: `self.button.enableTitleFontForState = NO;`.   Can you help me understand what is hard to read?

## When
1. `titleFontForState` is disabled
2. and `titleLabel.font` is assigned
3. and the button's state changes

## Then
The button's title label's font does not get changed.

components/Buttons/tests/unit/MDCButtonTests.m
randallli: Adding a test for behavior makes me think it is something we want to preserve. I would change the behavior.

components/Buttons/tests/unit/MDCButtonTests.m
randallli: hrm struggling here now that I understand what you are testing...
best I could come up with:
testDisabledTitleFontForStateFlagBeforeSettingFontForStateShouldNotChangeFonts

might be worth a comment to explain that it is a test for the flag.

components/Buttons/tests/unit/MDCButtonTests.m
randallli: oh right. i got confused I forgot that your testing the enableTitleFontForSate = NO

Perhaps a comment:
// This tests checks enableTitleFontForState = NO

randallli (approved):

components/Buttons/tests/unit/MDCButtonTests.m
romoore: Ok, thanks. I'll remove the test. I wouldn't expect the font to change immediately.

Updated to clarify: I don't think it's something we should preserve (either way), it was just a canary test if we ever changed the behavior.  This is a migration flag, so I don't expect clients to be toggling it back and forth in the middle of its use.

components/Buttons/tests/unit/MDCButtonTests.m
romoore: Naming is hard. I just added comments to all of the new tests.

components/Buttons/tests/unit/MDCButtonTests.m
romoore: Done. Added a comment.

randallli (approved):

PR #8607, opened 2019-10-18
[TextFields] Add density and preferredContainerHeight as private functionality

andrewoverton:
This PR brings over notions of `density` and `preferredContainerHeight`. Density refers to how vertically "squished" the textfields appear. `preferredContainerHeight` is a desired height for the container of the textfield. The container of the textfield is the rect within the outline in outlined textfields and the filled rect in filled textfields.

This functionality will remain private for now. In the near-ish future we will either expose it or delete it. Deleting it will not result in behavioral changes. The reason for bringing it over now is that it will make it significantly easier to bring everything else over.

Related to #6942.


components/TextFields/src/ContainedInputView/private/MDCTextControlVerticalPositioningReferenceBase.m
```diff
@@ -15,89 +15,187 @@
...
+               _paddingBetweenEditingTextAndContainerBottom];
+    CGFloat halfOfHeightWithOneRow = (CGFloat)0.5 * heightWithOneRow;
+    _paddingBetweenContainerTopAndNormalLabel = halfOfHeightWithOneRow - halfOfNormalFontLineHeight;
+  } else {
+    CGFloat halfOfContainerHeight = (CGFloat)0.5 * _containerHeight;
+    _paddingBetweenContainerTopAndNormalLabel = halfOfContainerHeight - halfOfNormalFontLineHeight;
+  }
+}
+
+- (CGFloat)standardizeDensity:(CGFloat)density {
```
randallli: lets use normalizeDensity

"The terms normalization and standardization are sometimes used interchangeably, but they usually refer to different things. Normalization usually means to scale a variable to have a values between 0 and 1, while standardization transforms data to have a mean of zero and a standard deviation of 1. " source: https://www.statisticshowto.datasciencecentral.com/normalized/

components/TextFields/src/ContainedInputView/private/MDCTextControlVerticalPositioningReferenceBase.m
```diff
@@ -15,89 +15,187 @@
 #import "MDCTextControlVerticalPositioningReferenceBase.h"
 #import "MaterialMath.h"
 
-static const CGFloat kPaddingBetweenTopAndFloatingLabel = (CGFloat)10.0;
-static const CGFloat kPaddingBetweenFloatingLabelAndText = (CGFloat)6.0;
-static const CGFloat kPaddingBetweenTextAndBottom = (CGFloat)10.0;
-static const CGFloat kPaddingAboveAssistiveLabels = (CGFloat)0.0;
-static const CGFloat kPaddingBelowAssistiveLabels = (CGFloat)6.0;
+static const CGFloat kMinPaddingBetweenContainerTopAndFloatingLabel = (CGFloat)6.0;
```
randallli, line 18: Let us add an in line comment to share the source for where we are taking these constants from. If we are trying to achieve a mock then share that.

components/TextFields/src/ContainedInputView/private/MDCTextControlVerticalPositioningReferenceBase.m
```diff
@@ -15,89 +15,187 @@
...
-
-  _paddingBetweenTopAndNormalLabel = _paddingBetweenTopAndFloatingLabel + floatingLabelHeight +
-                                     _paddingBetweenFloatingLabelAndText;
+                                       numberOfTextRows:(CGFloat)numberOfTextRows
+                                                density:(CGFloat)density
+                               preferredContainerHeight:(CGFloat)preferredContainerHeight {
+  BOOL isMultiline = numberOfTextRows > 1 || numberOfTextRows == 0;
+  CGFloat standardizedDensity = [self standardizeDensity:density];
+
+  CGFloat paddingBetweenContainerTopAndFloatingLabelRange =
```
randallli: rangePaddingBetweenContainerTopAndFloatingLabel
We should be consistent with the variable names here.

Weaker suggestion is to rename range to minMaxDelta as range implies multiple data points.
Range vs delta

components/TextFields/src/ContainedInputView/private/MDCTextControlVerticalPositioningReferenceBase.m
```diff
@@ -15,89 +15,187 @@
...
+      kMaxPaddingBetweenContainerTopAndFloatingLabel -
+      kMinPaddingBetweenContainerTopAndFloatingLabel;
+  CGFloat paddingBetweenContainerTopAndFloatingLabelAddition =
```
randallli: We appear to be repeating ourselves here. Perhaps we should make a function:
value(min,max,percent)

components/TextFields/src/ContainedInputView/private/MDCTextControlVerticalPositioningReferenceBase.m
```diff
@@ -15,89 +15,187 @@
...
+      kMinPaddingBelowAssistiveLabels + paddingBelowAssistiveLabelsAddition;
+
+  CGFloat containerHeightWithPaddingsDeterminedByDensity = [self
+      calculateContainerHeightWithFoatingLabelHeight:floatingLabelHeight
+                                       textRowHeight:textRowHeight
+                                    numberOfTextRows:numberOfTextRows
+          paddingBetweenContainerTopAndFloatingLabel:_paddingBetweenContainerTopAndFloatingLabel
+           paddingBetweenFloatingLabelAndEditingText:_paddingBetweenFloatingLabelAndEditingText
+         paddingBetweenEditingTextAndContainerBottom:_paddingBetweenEditingTextAndContainerBottom];
+  if (preferredContainerHeight > 0) {
```
randallli: Lets add a comment explaining what you mentioned in person. "if the user has not specified a preferred container height then we can ...."

components/TextFields/src/ContainedInputView/private/MDCTextControlVerticalPositioningReferenceBase.m
```diff
@@ -15,89 +15,187 @@
...
+    if (preferredContainerHeight > containerHeightWithPaddingsDeterminedByDensity) {
```
randallli: Echoing back my read of this code: this is for when the user has specified a taller height than what the density system's scaled height produces.

components/TextFields/src/ContainedInputView/private/MDCTextControlVerticalPositioningReferenceBase.m
```diff
@@ -15,89 +15,187 @@
...
+            _paddingBetweenFloatingLabelAndEditingText +
+            ((_paddingBetweenFloatingLabelAndEditingText / sumOfPaddingValues) * difference);
+        _paddingBetweenEditingTextAndContainerBottom =
+            _paddingBetweenEditingTextAndContainerBottom +
+            ((_paddingBetweenEditingTextAndContainerBottom / sumOfPaddingValues) * difference);
+      }
+    }
+  }
+
+  _containerHeight = containerHeightWithPaddingsDeterminedByDensity;
```
randallli: Should this be in the else clause in the next line's if clause?

components/TextFields/src/ContainedInputView/private/MDCTextControlVerticalPositioningReferenceBase.m
```diff
@@ -15,89 +15,187 @@
...
+    CGFloat heightWithOneRow = [self
+        calculateContainerHeightWithFoatingLabelHeight:floatingLabelHeight
+                                         textRowHeight:textRowHeight
+                                      numberOfTextRows:1
+            paddingBetweenContainerTopAndFloatingLabel:_paddingBetweenContainerTopAndFloatingLabel
+             paddingBetweenFloatingLabelAndEditingText:_paddingBetweenFloatingLabelAndEditingText
+           paddingBetweenEditingTextAndContainerBottom:
+               _paddingBetweenEditingTextAndContainerBottom];
+    CGFloat halfOfHeightWithOneRow = (CGFloat)0.5 * heightWithOneRow;
+    _paddingBetweenContainerTopAndNormalLabel = halfOfHeightWithOneRow - halfOfNormalFontLineHeight;
```
randallli, line 144: why is this not scaling with the others? using the normalized multiplier

randallli (approved):

components/TextFields/src/ContainedInputView/private/MDCTextControlVerticalPositioningReferenceBase.m
andrewoverton: Nice, thanks 👍

components/TextFields/src/ContainedInputView/private/MDCTextControlVerticalPositioningReferenceBase.m
andrewoverton: This chunk of code is determining where the label should go when it's not floating, i.e. when it's "normal." If it's single-line, it should be vertically centered within the container height. That's in the else clause. If it's multiline, it should go where it would go if it were single lined, even if the container height is taller. Does that make sense? I'm adding some comments and stuff that hopefully make it more clear.

components/TextFields/src/ContainedInputView/private/MDCTextControlVerticalPositioningReferenceBase.m
andrewoverton: Sure, that's more clear 👍

components/TextFields/src/ContainedInputView/private/MDCTextControlVerticalPositioningReferenceBase.m
andrewoverton: Since the base style is supposed to be the absence of a style these values don't really come from anywhere. I'll add a comment like that for the other styles though!

components/TextFields/src/ContainedInputView/private/MDCTextControlVerticalPositioningReferenceBase.m
andrewoverton: I think this has been resolved with edbcdaf

components/TextFields/src/ContainedInputView/private/MDCTextControlVerticalPositioningReferenceBase.m
andrewoverton: Done in edbcdaf

andrewoverton: I don't know why I didn't do that before...

components/TextFields/src/ContainedInputView/private/MDCTextControlVerticalPositioningReferenceBase.m
andrewoverton: I think this is resolved by edbcdaf

components/TextFields/src/ContainedInputView/private/MDCTextControlVerticalPositioningReferenceBase.m
andrewoverton: I think this is resolved by edbcdaf

PR #8612, opened 2019-10-18
[NavigationDrawer] Hide scrim when scrolled to bottom with tracking scroll view

### LINKED ISSUE #5782: [NavigationDrawer] Dragging the drawer further up when already at the bottom shows the scrim and that looks odd

andrewoverton:
This is not a great fix but if we don't have any better ideas we could possibly use it to resolve https://github.com/material-components/material-components-ios/issues/5782.

With this change the scrim is visible when it's shown at the top of the screen and at the top of the content. It's hidden when it's shown at the bottom of the screen and at the bottom of the content. In other words, the scrim is visible when it should be when hidden when it shouldn't be.

Someone who knows NavigationDrawer better would be better able to say whether this fix could cause problems.

# Before:
Simulator Screen Shot - iPhone 7 - 2019-10-18 at 16 54 24

# After:
Simulator Screen Shot - iPhone 7 - 2019-10-18 at 16 51 48

### DIFF
--- a/components/NavigationDrawer/src/MDCBottomDrawerPresentationController.m
+++ b/components/NavigationDrawer/src/MDCBottomDrawerPresentationController.m
@@ -377,4 +377,11 @@ static CGFloat kTopHandleTopMargin = (CGFloat)5.0;
   }
 }
 
+- (void)bottomDrawerContainerViewControllerDidUpdateValueForScrollViewIsScrolledToBottom:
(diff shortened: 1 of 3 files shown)


randallli:
I have taken over this work and continued with https://github.com/material-components/material-components-ios/pull/8618

PR #8613, opened 2019-10-21
[Tabs] Optimization to avoid creating unnecessary custom views.

mrhappyasthma:
The code, as written, was invoking the `mdc_customView` method twice per iteration. Once to check within the if-statement, and once to actually assign the view. If that method allocated a new UIView, then each of these invocations will cause new UIViews to be created.

Instead, we should call the method once and store it in a local to be used in both the if-statement and the assignment. This avoids unnecessary view creation if the implementation of `mdc_customView` allocates a new view on each call.


romoore (approved):
Looks fine to me, but more as a guard against it changing on another thread.

Creating a new object every time a property is accessed would make it very hard to customize that object.  It also would seem to go against [Cocoa's naming conventions that constructors should start with the class being constructed](https://developer.apple.com/library/archive/documentation/General/Conceptual/CocoaEncyclopedia/ClassFactoryMethods/ClassFactoryMethods.html).  I would strongly recommend that any current implementation of this method that allocates a new object should be updated to be a simple property, which is the intention of declaring it both `nullable` and `strong`.

mrhappyasthma:
I'm also going to make changes on our end to synchronize 1 view to each item.

PR #8627, opened 2019-10-23
[TextFields] Add filled and outlined positioning references

andrewoverton:
Yesterday my plan was to bring over the filled textfields in one PR and the outlined textfields in another PR. However, the filled textfield PR ended up being about 1000 lines, which felt too big, so I decided to take a different approach. I'm going to start bringing the filled and outlined textfields by bringing over their positioning reference objects. The two classes in this PR are extremely similar to the MDCTextControlVerticalPositioningReferenceBase brought over in previous PRs. They are so similar, in fact, that as I was looking over them I started to think that all three should probably share a superclass so there isn't so much duplicate code. At some point I will look into that. However, in the interest of time, I don't this work should be done now. It should be fine to bring them over as separate entities for now.

Here's a comparison link demonstrating the code diff between the base positioning reference and the filled positioning reference:
https://github.com/material-components/material-components-ios/compare/develop...andrewoverton:textfield-pr-filled-positioning-reference-comparison?expand=1
And here is an image of an MDCBaseTextField with the filled positioning reference:
Screen Shot 2019-10-24 at 10 55 12 AM

Here's a comparison link demonstrating the code diff between the base positioning reference and the outlined positioning reference:
https://github.com/material-components/material-components-ios/compare/develop...andrewoverton:textfield-pr-outlined-positioning-reference-comparison?expand=1
And here is an image of an MDCBaseTextField with the outlined positioning reference:
Screen Shot 2019-10-24 at 10 54 11 AM

Related to #6942.


randallli (approved):
Let us add a todo to hold ourselves accountable to the snapshot tests that we are committing to in the description of the PR.

andrewoverton:
I didn't realize you were going to be out Monday and Tuesday. That increases the urgency of this landing. So that I can get the next PRs up as soon as possible I'm going to merge this and then add the TODOs you requested in an immediate follow up.

PR #8628, opened 2019-10-23
[TextFields] Add path drawing utils for style objects

andrewoverton:
The MDCTextControlStyleFilled and MDCTextControlStyleOutlined objects make use of this path drawing util class to draw paths for things like the outlines, filled backgrounds, and underlines.

Here are some screenshots of TextFields whose styles make use of these methods:
Screen Shot 2019-10-25 at 12 22 25 PM
Screen Shot 2019-10-25 at 12 22 18 PM
The filled background and the underline in the filled textfield are layers whose paths were drawn using these methods. The outline in the outlined textfield is a layer whose path was drawn with these methods. Once the filled and outlined styles and the MDCBaseTextField subclasses that make use of them have landed there will be snapshot tests validating these methods.

Related to #6942.


randallli:
If nothing is using these APIs how do we know they are working? I was expecting a snapshot to show these curves in action.

Would adding the style that uses this as a non styled first and then adding and using this make sense?

components/TextFields/src/ContainedInputView/private/MDCTextControlStylePathDrawingUtils.h
```diff
@@ -0,0 +1,37 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+@interface MDCTextControlStylePathDrawingUtils : NSObject
+
++ (void)addTopRightCornerToPath:(UIBezierPath *)path
```
randallli: I think this looks similar enough to the UIBezierPath APIs that we should make a category
https://developer.apple.com/documentation/uikit/uibezierpath/1624367-addarcwithcenter?language=objc

Then we wont be incurring the cost of creating another object.

components/TextFields/src/ContainedInputView/private/MDCTextControlStylePathDrawingUtils.h
```diff
@@ -0,0 +1,37 @@
...
+                      fromPoint:(CGPoint)point1
+                        toPoint:(CGPoint)point2
+                     withRadius:(CGFloat)radius;
++ (void)addBottomRightCornerToPath:(UIBezierPath *)path
```
randallli: how are we expected to use this? Are all 4 corners suppose to be used together?

would a native API call work instead?
https://developer.apple.com/documentation/uikit/uibezierpath/1624356-bezierpathwithroundedrect?language=objc

components/TextFields/src/ContainedInputView/private/MDCTextControlStylePathDrawingUtils.h
```diff
@@ -0,0 +1,37 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+@interface MDCTextControlStylePathDrawingUtils : NSObject
```
randallli: Are we building something already offered by
https://github.com/material-components/material-components-ios/tree/develop/components/Shapes ?

which features were missing?

randallli (changes requested):

components/TextFields/src/ContainedInputView/private/MDCTextControlStylePathDrawingUtils.h
andrewoverton: I'm fine with moving these methods to a category if you prefer that.

You can think of these methods as slightly less esoteric wrappers around the `-addArcWithCenter:radius:startAngle:endAngle:clockwise:` you linked to. I think it's preferable for the caller to only have to say "add the bottom left corner to the path" rather than "add a curve to the path with a start angle of `-(CGFloat)((M_PI * 3) / 2)` and an end angle of `-(CGFloat)M_PI`." Do you see what I mean?

components/TextFields/src/ContainedInputView/private/MDCTextControlStylePathDrawingUtils.h
andrewoverton: `+bezierPathWithRoundedRect:cornerRadius:` doesn't really meet our needs here. It wouldn't meet our needs for outlined textfields because the path for an outline in an outlined textfield has a break in it where the label goes. This method doesn't help us achieve that. It doesn't work for filled textfields because not all the corner radii in a filled background path, for example, are equal. The bottom corners are right angles. Given all this, are you okay with not using this method?

components/TextFields/src/ContainedInputView/private/MDCTextControlStylePathDrawingUtils.h
andrewoverton: I don't know, I've never looked at the Shapes library. If Shapes has similar helpers for adding specific corners to paths with corner radii I'm down to use those. Do such helpers exist?

components/TextFields/src/ContainedInputView/private/MDCTextControlStylePathDrawingUtils.h
randallli: ah right forgot about that. thanks for the extra info.

components/TextFields/src/ContainedInputView/private/MDCTextControlStylePathDrawingUtils.h
randallli: hmm im not sure. @yarneo may know more. 
For now Im ok with us adding a todo to look into shape integration.

components/TextFields/src/ContainedInputView/private/MDCTextControlStylePathDrawingUtils.h
randallli: yep i see the value of the simpler API we are adding. 

Is there a unit/snapshot comparison test we could do that confirms our results are the same as a bezierPathWithRoundedRect?

randallli: test not required. i just got excited.

We should add a image of the end result in the description though.

randallli (approved):

randallli (approved):
approval after 
- bezier path category
- image in description of result when used. (with path cut out for outline style)

PR #8657, opened 2019-10-25
[TextFields] Add filled style object

andrewoverton:
This PR adds the filled style object. Once this is PR is merged we can open one for the MDCFilledTextField. At that point, we can snapshot test the textfield that results from this style object, as well as the filled positioning reference object that landed in https://github.com/material-components/material-components-ios/pull/8627.

This is what the filled textfield ultimately looks like:
Screen Shot 2019-10-25 at 12 22 18 PM

Related to #6942.

### DIFF
--- a/components/TextFields/src/ContainedInputView/private/MDCTextControlStyleFilled.h
+++ b/components/TextFields/src/ContainedInputView/private/MDCTextControlStyleFilled.h
@@ -18,7 +18,8 @@
 #import "MDCTextControlLabelBehavior.h"
 #import "MDCTextControlStyleBase.h"
 
-//TODO: When the MDCBaseTextField subclass that makes use of this style (and the path drawing logic inside it) lands there should be snapshot tests for it.
+// TODO: When the MDCBaseTextField subclass that makes use of this style (and the path drawing logic
+// inside it) lands there should be snapshot tests for it.
 @interface MDCTextControlStyleFilled : NSObject <MDCTextControlStyle>
 - (nonnull UIColor *)underlineColorForState:(MDCTextControlState)state;
 - (void)setUnderlineColor:(nonnull UIColor *)underlineColor forState:(MDCTextControlState)state;
--- a/components/TextFields/src/ContainedInputView/private/MDCTextControlStyleFilled.h
+++ b/components/TextFields/src/ContainedInputView/private/MDCTextControlStyleFilled.h
@@ -18,6 +18,7 @@
 #import "MDCTextControlLabelBehavior.h"
 #import "MDCTextControlStyleBase.h"
 
+//TODO: When the MDCBaseTextField subclass that makes use of this style (and the path drawing logic inside it) lands there should be snapshot tests for it.
 @interface MDCTextControlStyleFilled : NSObject <MDCTextControlStyle>
 - (nonnull UIColor *)underlineColorForState:(MDCTextControlState)state;
 - (void)setUnderlineColor:(nonnull UIColor *)underlineColor forState:(MDCTextControlState)state;
--- a/components/TextFields/src/ContainedInputView/private/MDCTextControlStyleFilled.h
+++ b/components/TextFields/src/ContainedInputView/private/MDCTextControlStyleFilled.h
@@ -20,10 +20,36 @@
 
 // TODO: When the MDCBaseTextField subclass that makes use of this style (and the path drawing logic
 // inside it) lands there should be snapshot tests for it.
+/**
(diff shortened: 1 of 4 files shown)


components/TextFields/src/ContainedInputView/private/MDCTextControlStyleFilled.m
```diff
@@ -0,0 +1,435 @@
...
+
+- (void)setFilledBackgroundColor:(nonnull UIColor *)filledBackgroundColor
+                        forState:(MDCTextControlState)state {
+  self.filledBackgroundColors[@(state)] = filledBackgroundColor;
+}
+
+- (void)applyStyleToTextControl:(id<MDCTextControl>)textControl {
+  if (![textControl isKindOfClass:[UIView class]]) {
+    [self removeStyleFrom:textControl];
+    return;
```
randallli: should this return NO to indicate an unsuccessful operation?

components/TextFields/src/ContainedInputView/private/MDCTextControlStyleFilled.m
```diff
@@ -0,0 +1,435 @@
...
+  return self.filledBackgroundColors[@(state)];
+}
+
+- (void)setFilledBackgroundColor:(nonnull UIColor *)filledBackgroundColor
+                        forState:(MDCTextControlState)state {
+  self.filledBackgroundColors[@(state)] = filledBackgroundColor;
+}
+
+- (void)applyStyleToTextControl:(id<MDCTextControl>)textControl {
+  if (![textControl isKindOfClass:[UIView class]]) {
```
randallli: Is this required by all of our textControls?
perhaps we should enforce that at the API / compiler level rather than at run time.

Proposed change to the protocol to
`- (void)applyStyleToTextControl:(UIView<MDCTextControl> *)textControl;`

randallli (approved):
check out my compile time check (instead of a runtime)

components/TextFields/src/ContainedInputView/private/MDCTextControlStyleFilled.m
andrewoverton: Good idea 👍

components/TextFields/src/ContainedInputView/private/MDCTextControlStyleFilled.m
andrewoverton: I don't think it needs to return a BOOL now that I've made the change suggested in https://github.com/material-components/material-components-ios/pull/8657#discussion_r339178977

components/TextFields/src/ContainedInputView/private/MDCTextControlStyleFilled.h
```diff
@@ -20,10 +20,36 @@
...
 // TODO: When the MDCBaseTextField subclass that makes use of this style (and the path drawing logic
 // inside it) lands there should be snapshot tests for it.
+/**
+This style object is used by MDCTextControls adopting the Material Filled style.
+*/
 @interface MDCTextControlStyleFilled : NSObject <MDCTextControlStyle>
-- (nonnull UIColor *)underlineColorForState:(MDCTextControlState)state;
+
+/**
+Sets the underline color color for a given state.
```
randallli, line 29: Yay method comments!

components/TextFields/src/ContainedInputView/private/MDCTextControlStyleFilled.m
```diff
@@ -0,0 +1,442 @@
...
+
+static const CGFloat kFilledContainerStyleTopCornerRadius = (CGFloat)4.0;
+static const CGFloat kFilledContainerStyleUnderlineWidthThin = (CGFloat)1.0;
+static const CGFloat kFilledContainerStyleUnderlineWidthThick = (CGFloat)2.0;
+
+static const CGFloat kFilledFloatingLabelScaleFactor = 0.75;
+
+@interface MDCTextControlStyleFilled () <CAAnimationDelegate>
+
+@property(strong, nonatomic) CAShapeLayer *filledSublayer;
```
randallli, line 31: It might be worth commenting/ pragma mark grouping the ones required for MDCTextControl protocol.

randallli (approved):

components/TextFields/src/ContainedInputView/private/MDCTextControlStyleFilled.m
randallli: actually i missunderstood these never mind

randallli (approved):

PR #8670, opened 2019-10-28
[TextFields] Refactor snapshots in preparation for more types of textfields

andrewoverton:
TL;DR -- This PR results in less duplication than if I were to just rewrite the same snapshot tests for MDCBaseTextField, MDCFilledTextField, and MDCOutlinedTextField, but there is still more duplication than I would like for there to be because of a bug that makes inheritance based snapshot tests not workable internally.

Before bringing over MDCFilledTextField and MDCOutlinedTextField I wanted to have some way to try to reduce code duplication in the snapshot tests for the three textfields. The snapshot tests for MDCTextField make use of inheritance to achieve this. I eventually want to do something similar, but I want to wait until [b/133313258](http://b/133313258) is resolved before I do that, because I want for the snapshot tests to be able to run internally. Until then, I suggest we have something like this. There's a little less duplication than if we just wrote the same snapshot test "the old fashioned way" for each textfield class, and there is some shared infrastructure, like in `MDCTextControlSnapshotTestHelpers`. In the future, if/when other MDCTextControls are brought into the library, they can also use of things like this.

I want to stress that this is not supposed to be the final state of snapshot testing for the new textfields. It's just one step closer. I'm still figuring it out.

Related to #6942.

### DIFF
--- a/components/TextFields/tests/snapshot/MDCBaseTextFieldTestsSnapshotTestHelpers.h
+++ b/components/TextFields/tests/snapshot/MDCBaseTextFieldTestsSnapshotTestHelpers.h
@@ -19,16 +19,16 @@
 #import "../../src/ContainedInputView/MDCBaseTextField.h"
 
 @interface MDCBaseTextFieldTestsSnapshotTestHelpers : NSObject
-+ (void)configureTextFieldWithAssistiveLabelText:(MDCBaseTextField *)textField;
++ (void)configureTextFieldWithColoredAssistiveLabelText:(MDCBaseTextField *)textField;
 + (void)configureTextFieldWithText:(MDCBaseTextField *)textField;
 + (void)configureTextFieldWithLeadingView:(MDCBaseTextField *)textField;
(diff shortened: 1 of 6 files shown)


components/TextFields/tests/snapshot/MDCBaseTextFieldTestsSnapshotTestHelpers.h
```diff
@@ -0,0 +1,34 @@
...
+// limitations under the License.
+
+#import "MaterialSnapshot.h"
+
+#import <UIKit/UIKit.h>
+
+#import "../../src/ContainedInputView/MDCBaseTextField.h"
+
+@interface MDCBaseTextFieldTestsSnapshotTestHelpers : NSObject
++ (void)configureTextFieldWithAssistiveLabelText:(MDCBaseTextField *)textField;
```
bryanoltman: Because this will eventually be used as a base class for test cases, please document what these functions do.

bryanoltman (changes requested):

bryanoltman (approved):

